PR #3, opened 2016-08-26
Complete Wiki Transfer

oopsjpeg:
A complete transfer of every osu! wiki page, converted to Markdown format.
Needs to be cleaned and organized.


Ameobea:
:heart: MediaWiki

peppy:
The directory structure is a but confusing: each language has a folder except for EN? I feel like EN should also get a folder and all relative paths should be the same in each language's implementation.

oopsjpeg:
Yep, this was the type of sorting that was planned out. I still have to get around to actually sorting everything into langcode folders.

PR #11, opened 2016-08-31
Images - not sorted yet

### LINKED ISSUE #2: Full S3 wiki dump
This is for getting the images that are used on the wiki so articles can use it once the migration completes.

The current wiki S3 integration complicates things a bit due to mediawiki slowdowns and related stuff.

TPGPL:
Resolves #2 
?


peppy:
the fuck?

peppy:
all you did was dropped the export in? what's the point?

PR #14, opened 2016-09-02
Accuracy and osu!academy table fixes

MegaApplePi:
Fixes these tables:
- Accuracy
- osu!academy


Repflez:
Please change the remaining tables in your commits to markdown. We aren't using html at all in the wiki.

PR #30, opened 2016-09-05
Translate the rest of the Chat-Console article

HWSKY:
Various modifications were made here, adjusting the translation with proper formatting of the main file. Ff the master ou moderator user want a new table to user status, I'll easily make and require another commit and push.


Repflez:
Can you change the command list to use markdown?

HWSKY:
until Wednesday will be available for change

PR #54, opened 2016-10-12
Fixed errors and added an achievment

Trehorna123:
Edited to show that you can get an achievment with difficulty increasing mods and I added the achievment Hour Before The Dawn since it was missing


TPGPL:
That's ok, tho. You could add that achievement to **all** achievement articles

Repflez:
Agree with @TPGPL. It would be nice if you added it to the other achievement articles.

But for now, I'll merge it since it's a contribution.

PR #64, opened 2016-10-25
Installation

TPGPL:
- [x] Main Article
- [x] Languages


Repflez:
The articles could have the other stuff removed and only have the osu! instructions on it.
It's the _osu! wiki_, not the T-Aiko wiki or the osu!stream wiki (even if that one is an official derivative of osu!).

PR #65, opened 2016-10-27
Fixed some things in this English Article and added some things

sr229:
probably the most ugliest around here. Fixed some to Markdown

still WIP

unfortunately because MD is limited by design, I cannot unuse some HTML around here such as the font size

which is according to [this stack exchange topic](http://meta.stackexchange.com/questions/170480/is-it-possible-to-change-the-text-size-in-the-code-markdown). the text sizing is entirely up to the site.


sr229:
I want your feedback out of this guys. I will try do the same on HTML-infested articles but it depends on the severity and the limits of the Markdown language

MegaApplePi:
Use images for the examples, when necessary

sr229:
I think that can be done fairly quickly, will do that once I get home.

TPGPL:
I'm already fixing Installation. There is even a PR, so you can see how I have ported it.

TPGPL:
Also, add title to BBCode with `====` header and reduce headers' grade by 1 (like `###` into `####`)

sr229:
Updated to resolve Conflicts

sr229:
Closed for Inactivity

PR #67, opened 2016-11-01
Remove unnecessary whitespace

cl8n:
I noticed while editing some files that there are extremely large blocks of whitespace here and there that serve no purpose (likely artifacts from the initial conversion to markdown). Here's what I changed, taken from the commit message:
- Remove whitespace from lines without text
- Trim trailing whitespace to two spaces
- Remove newlines if there are more than one in a row
- Remove trailing whitespace on lists

None of these changes affect the display of files.


TPGPL:
Good job so far

Repflez:
uh what? was this outdated or something?

PR #70, opened 2016-12-08
Update: Multi (two table styles used [raw HTML, and GFM]).

RaikireHiuduo:
Change most tables to hardcoded HTML tables. Not yet refined. May need another check.


TPGPL:
Actually we **shouldn't** use any HTML.

Repflez:
Don't use HTML unless it's ***absolutely required.***

RaikireHiuduo:
rowspan/colspan attribute, mainly.

Listing in the GFM Table need to be done with \<ul\> tags (which might end up unreadable the longer the list is). It can be done if you want to enforce GFM Table only.
Reference: https://stackoverflow.com/questions/19950648/how-to-write-lists-inside-a-markdown-table

On unrelated note: Yes, I am reading; just have no idea why I am not allowed to post. New to this (Pardon the pixelated image):-
discord

RaikireHiuduo:
Push request.

Check Multi, and osu!academy (newly added) for changes.

PR #80, opened 2016-12-29
Legal (TOS and Copyright)

MegaApplePi:
- added `Legal/TOS` and `Legal/Copyright` folders
- added the `en.md` files
- copied-and-pasted the TOS and Copyright (DMCA) into the respective files
- fixed formatting to attempt to match original

### DIFF
--- a/wiki/Legal/Copyright/en.md
+++ b/wiki/Legal/Copyright/en.md
@@ -1,0 +1,64 @@
+# osu! Copyright Policy
+
+osu! has adopted the following policy concerning copyright infringement in accordance with the Digital Millennium Copyright Act, 17 U.S.C. 512.
+The contact information for osu!'s Designated Agent to Receive Notification of Claimed Infringement ("Designated Agent") is listed at the end of this policy, and is on file with the United States Copyright Office (http://www.copyright.gov/onlinesp/list/index.html).
+osu! takes copyright and other intellectual property rights very seriously.
+It is osu!'s policy to (1) expeditiously block access to or remove content that it believes in good faith may contain material that infringes the copyrights of third parties and (2) remove and discontinue service to repeat offenders.
+
+## Procedure for Reporting Copyright Infringements:
+
+If you believe that content residing on or accessible through the osu! website or service infringes your copyright, please send a notice of claimed copyright infringement containing the following information to the Designated Agent listed below (consult legal counsel or refer to 17 U.S.C. 512(c) to confirm these requirements): 
+
+1. A physical or electronic signature of a person authorized to act on behalf of the owner of the copyright that has been allegedly infringed; 
+
+2. Identification of the copyrighted works claimed to have been infringed, or if multiple copyrighted works at a single online site are covered by a single notification, a representative list of such works at that site; 
+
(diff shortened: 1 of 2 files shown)


cl8n:
While on the topic, we should probably do something with the other old /p/ pages as well, like art, challengeart, doyoureallywanttoaskpeppy, inactive, twitter, amazon, bestof20yy, etc. since they likely won't see a mention on the new site

MillhioreF:
As per discussion in the dev discord all completely static pages should be imported into the wiki format, and can be given special URLs if need be. Many of those pages aren't static and will have to be redesigned or removed, but some (twitter, askpeppy, maybe inactive?) can certainly be moved over.

peppy:
yep, we want to move as much of them to the new wiki system as possible (anything that doesn't require dynamic content implementations, basically).

wiki/Legal/Copyright/en.md
```diff
@@ -0,0 +1,64 @@
...
+
+Designated Agent to Receive Notification of Claimed Infringement: 
+
+**Email**: [pe@ppy.sh](mailto:pe@ppy.sh) (24 hour response guaranteed) 
+
+\- or -
+
+**Postal Address**: 
+
+<pre>
```
cl8n: use triple backticks (` ``` `) instead

wiki/Legal/TOS/en.md
```diff
@@ -0,0 +1,99 @@
...
+
+**9\. PRIVACY POLICIES**
+
+osu.ppy.sh finds your privacy most important, and we value your use of our services.
+osu.ppy.sh's services are free of any charge, no prior registration is needed for the use of these services.
+osu.ppy.sh will explicitly ask for information about you when such information is needed for use of the services provided by osu.ppy.sh.
+Any submission of any personal information is done only with your voluntary act (website), or automatically (game client software) where necessary to provide diagnostics and feedback.
+osu.ppy.sh keeps all private and personal information private, and does not, or will not share this information with any third parties except under the following circumstances :
+
+When such action is necessary to conform to legal requirements or comply with legal process.
```
cl8n: probably just use a list here

wiki/Legal/Copyright/en.md
MegaApplePi: Using triple backticks enables the syntax highlighter (idk about how it'll appear on osu!wiki though).

wiki/Legal/TOS/en.md
MegaApplePi: I am not sure if I should modify the tos any further then what was given.

wiki/Legal/Copyright/en.md
cl8n: if you don't specify a language it should have the same effect as `<pre>`

MegaApplePi:
The new lines is just an appearance wise thing while editing it in the text. Markdown considers them as a single line.

wiki/Legal/Copyright/en.md
MegaApplePi: Whoops, one of the markdown previewers I used has displayed it with syntax highlighting .-.

I'll a push a change.

cl8n:
That's a good point; I guess I just commented on it because it's different to most of the other articles.

This should be ready to merge now

PR #81, opened 2016-12-29
Move static pages to articles

cl8n:
Like #80


cl8n:
Decided to not add the inactive page, since it'd be kinda weird to see "your account is disabled" on the wiki if your account is in a normal state

PR #86, opened 2016-12-30
Handle /WEIRD

cl8n:
The language-specific translation notes aren't needed anymore because we're tracking with GitHub issues, the T-Aiko! article doesn't belong on the wiki and the feature requests article doesn't have any content.

Moved the skinning articles as well


cl8n:
Thanks for the catch!

PR #96, opened 2017-01-08
Correcting "Shortcut Key Reference" folder

TPGPL:
- Fixed English and other languages
- Fixed links
- Added Titles
- Added images
- Deleted categories
- Deleted HTML tags


TPGPL:
I'll have to uppercase some icons because they aren't working

PR #113, opened 2017-02-06
Move /Tournaments/ path out of /Announcements/

Loctav:
Along with some minor updates to the TWC 2017 article formatting, moving out the `/Tournaments/` path out of `/Announcements/` seem to be the best way to go, especially for the page header.


TPGPL:
We need to fix old paths for Tournaments folder (if there are any).

PR #153, opened 2017-02-24
Escape brackets

Loctav:
For testing purposes to see what makes it eat up the entire column's content.


Loctav:
Let's try something else.

PR #166, opened 2017-03-01
wiki/About/ko.md

CookieCoMa:
Translated.


wiki/About/ko.md
```diff
@@ -0,0 +1,6 @@
+정보
+=======
+
+osu! wiki는 Osu!를 돕기 위한 커뮤니티 중심의 프로젝트입니다. 사용자는 보다 효과적인 방식으로 게임을 이해합니다.
```
kj415j45: L3,L5:it should be **osu!** not **Osu!**

CookieCoMa:
Fixed, Thanks a lot

CookieCoMa:
That was stupid of me. Anyway thank you.

PR #173, opened 2017-03-07
fix Article Style

kj415j45:
fix Article Style in #167


kj415j45:
@RaikireHiuduo i will give up it  :p

PR #203, opened 2017-04-09
TWC2017: Grand Finals update

Loctav:


WalterToro:
All the osu links should be using osu.ppy.sh instead of new.ppy.sh as described [Here](https://github.com/ppy/osu-wiki/blob/master/wiki/Article_Style_Guide/en.md#links), i am wrong?

PR #206, opened 2017-04-10
Update es.md

apap04:
Fixed ¡ tiene un perfil! to ¡tiene un perfil!


MegaApplePi:
For future commits, please write more meaningful commit messages (use the message to describe the main change and the description to explain in-depth, if need be).

PR #217, opened 2017-04-22
Fixed How to get beatmaps on osu! ? to How to get beatmaps on osu!?

apap04:
Fixed accidental spacings. Common mistake.


sr229:
Nice blob

PR #225, opened 2017-04-29
added (or Beatmap Appreciation Team) in BAT

apap04:
Added this for users to know what BAT is.

### DIFF
--- a/wiki/Glossary/en.md
+++ b/wiki/Glossary/en.md
@@ -35,7 +35,7 @@ Bancho is the easily expandable server coded by peppy.
 Bancho replaces the need for a third-party IRC server in the management of all online content in osu!.
 Logging with an osu!account ingame will automatically attempt to connect to Bancho until done successfully.
 
-### BAT
+### BAT (or Beatmap Appreciation Team)
 
 **As of February 1, 2015, the BAT was renamed to _Beatmap Nomination Group_**
 


MegaApplePi:
The meaning of that term is in the definition.

I don't think there is a need for this and besides, this will only create more confusing section links.

wiki/Glossary/en.md
```diff
@@ -35,7 +35,7 @@ Bancho is the easily expandable server coded by peppy.
 Bancho replaces the need for a third-party IRC server in the management of all online content in osu!.
 Logging with an osu!account ingame will automatically attempt to connect to Bancho until done successfully.
 
-### BAT
+### BAT (or Beatmap Appreciation Team)
 
 **As of February 1, 2015, the BAT was renamed to _Beatmap Nomination Group_**
```
kj415j45: why not add them to this line(L40)

MegaApplePi: the meaning is placed under that line, but that is a notice to say that the team name was changed...

MegaApplePi: if really, the I guess the meaning can be moved to that line, but that line isn't meant to be used as the definition

kj415j45: if it's the first time that it appear in this article, it **can** be there(not **must**)

apap04: Close???

kj415j45: depends on you

i suggest close it
if you want to keep it, please also fix others like AFK, BPM, MAT, .....................

PR #251, opened 2017-05-15
Add participant group list OWC/2015

RaikireHiuduo:
See title.


sr229:
Can we validate this information @Loctav?

RaikireHiuduo:
@sr229 Not needed, check: https://osu.ppy.sh/wiki/Osu!_World_Cup_2015#Participants

PR #252, opened 2017-05-18
Initial commit of all tumblr news posts

peppy:


TPGPL:
would it be possible to sort them a bit?

peppy:
they are added this way intentionally for ease-of-procesing.

PR #256, opened 2017-05-21
redirects testing

MegaApplePi:
Testing for osu-web!
`redirects/pp` contains an intentional typo

### DIFF
--- a/wiki/redirect/archive/pl.md
+++ b/wiki/redirect/archive/pl.md
@@ -1,0 +1,1 @@
+/wiki/Announcements/Archive/?locale=pl
--- a/wiki/redirect/osu!academy/de.md
+++ b/wiki/redirect/osu!academy/de.md
@@ -1,0 +1,1 @@
+/wiki/Announcements/osu!academy/?locale=de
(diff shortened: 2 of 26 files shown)


wiki/redirect/archive/pl.md
```diff
@@ -0,0 +1 @@
+/wiki/Announcements/Archive/?locale=pl
```
TPGPL: I think redirects should be only english if they arent in foreign language

wiki/redirect/osu!academy/de.md
```diff
@@ -0,0 +1 @@
+/wiki/Announcements/osu!academy/?locale=de
```
TPGPL: Whats the purpose of this redirect

MegaApplePi: > Testing for osu-web!
For when I type `https://new.ppy.sh/help/wiki/osu!academy/?locale=de` instead of `https://new.ppy.sh/help/wiki/Announcements/osu!academy/?locale=de`

TPGPL: but there will be search bar for those? So it should find it anyways

MegaApplePi: alrighty, I will remove it but I want to first see if _that_ (see my comment above) works or not.
This is also why a couple of the redirects are in other languages as well I suppose.

Issue #177, opened 2017-03-11
[Draft] Main index page done with ducttapes


RaikireHiuduo:
Bone-structure based on the 63 folders in `/wiki`.

```Markdown
# osu!wiki

Welcome to the osu!wiki, a collaborative compendium of all _osu!_ things!

- [Welcome](/wiki/Welcome "Welcome") <-- I dunno
- [About](/wiki/About "About")
- [Legal](/wiki/Legal "Legal")

## osu! client

### Getting Started

- [Registration](/wiki/Registration "Registration")
- [Installation](/wiki/Installation "Installation")

### Help

- [Help Center](/wiki/Help_Center "Help Center")
- [Rules](/wiki/Rules "Rules")
- [Reporting Bad Behaviour](/wiki/Reporting_Bad_Behaviour "Reporting Bad Behaviour")
- [Glossary](/wiki/Glossary "Glossary")

### Available Game Modes

- [FAQ](/wiki/FAQ "FAQ")
- [Beginner](/wiki/Beginner "Beginner") <-- This look a lot like a general guide to me
- [Game Modes](/wiki/Game_Modes "Game Modes")
  - [External Ports](/wiki/Game_Modes/External_Ports "External Ports")
- [Multi](/wiki/Multi "Multi")

### Game Mechanics

#### Gameplay

- [Accuracy](/wiki/Accuracy "Accuracy")
- [Beatmaps](/wiki/Beatmaps "Beatmaps")
- [Game Modifiers](/wiki/Game_Modifiers "Game Modifiers")
- [Hit Objects](/wiki/Hit_Objects "Hit Objects")
- [Interface](/wiki/Interface "Interface")
- [Options](/wiki/Options "Options")
- [Play Styles](/wiki/Play_Styles "Play Styles")
- [Score](/wiki/Score "Score")
- [Shortcut Key Reference](/wiki/Shortcut_Key_Reference "Shortcut Key Reference")
- [Storyboards](/wiki/Storyboards "Storyboards")
- [Mascots](/wiki/Mascots "Mascots")
- [Replay](/wiki/Replay "Replay")

#### Online

- [BanchoBot](/wiki/BanchoBot "BanchoBot")
- [Chat Console](/wiki/Chat_Console "Chat Console")
- [Internet Relay Chat](/wiki/Internet_Relay_Chat "Internet Relay Chat")
- [Live streaming osu!](/wiki/Live_streaming_osu! "Live streaming osu!")
[...]

PR #264, opened 2017-05-25
Index page of osu-wiki and a minor osu!wiki history update

RaikireHiuduo:
See title.

Updated from #177.

Notes:-
1. Facebook link is not shown in GitHub for some reason.
2. `osu!news` and `osu!dev blog` still uses the old tumblr link. What is the new links for those two?


wiki/Index/en.md
```diff
@@ -0,0 +1,132 @@
...
+- [osu!store](/wiki/osu!store "osu!store")
+- [People](/wiki/People "People")
+
+### Follow
+
+- [Facebook](https://www.facebook.com/osugame "Facebook")
+- [Reddit](https://www.reddit.com/r/osugame/ "r/osugame at Reddit")
+- [Twitter](/wiki/Twitter "Twitter page")
+- [osu!news](http://osunews.tumblr.com/ "osu!news")
+- [osu!dev blog](http://pppy.tumblr.com/ "osu!dev blog")
```
TPGPL: Link typo

RaikireHiuduo: To where?

RaikireHiuduo:
@TPGPL So, do you want me to include all the subfolders (if available) too?

wiki/Index/en.md
TPGPL: pppy.tumblr.com/

TPGPL:
I think it would be slightly better, but let's ask other people

MegaApplePi:
@RaikireHiuduo Since this is technically an index page that indexes everything, I suppose it should contain links to everywhere, even if it makes the page appear long.

The links in there will always be in need of maintenance anyways since refactoring and new folders will constantly happen anyway (but redirects may help with half of the battle once they are added :p)

wiki/Index/en.md
```diff
@@ -0,0 +1,132 @@
+# osu!wiki
+
+Welcome to the osu!wiki, a collaborative compendium of all _osu!_ things!
+
+## Legal notice
```
MegaApplePi: Title case

wiki/Index/en.md
```diff
@@ -0,0 +1,132 @@
+# osu!wiki
```
MegaApplePi: /me talking (can ignore) _:c the title isn't is same as the folder name_

PR #325, opened 2017-06-06
Added osu!mania information to Game Modifiers

Shadowtail117:
Relax, Auto Pilot, Spun Out, and Auto didn't have exceptions for osu!mania. In addition, I added a disclaimer to xK, Co-Op, and Random that they were specific to osu!mania. Lastly, I revised the story about Flashlight being unranked for a while and clarified the 10K mod.


TPGPL:
not sure about coding the mod names tho, but nothing about it in ASG

PR #348, opened 2017-06-09
Rewrite section in ASG

kj415j45:


wiki/Article_Style_Guide/en.md
```diff
@@ -141,8 +141,14 @@ osu!mania (Español)
 # Skinning osu!catch (Français)
 ```
 
-You can find a list of native language names at [Language names in their own languages and scripts](http://www.omniglot.com/language/names.htm) or [List of language names](https://en.wikipedia.org/wiki/List_of_language_names).
-<!-- NOTE this may introduce more inconsistencies because some languages use more than one script to represent their language! -->
+You can find a list of native language names in [`osu-web`'s script file](https://github.com/ppy/osu-web/blob/master/app/Libraries/LocaleMeta.php), e.g:
```
MegaApplePi: remove `, e.g:` and replace with a `.`
new line: add `For example:`

kj415j45: list them all is too long, and more language may add in osu-web(

MegaApplePi: It would only be as long as this:
chrome_2017-06-09_10-08-01
and even so, that list would need to be updated anyways when they add more locales

kj415j45:
will continue by @MegaApplePi

PR #353, opened 2017-06-10
achievements: hotfix

MegaApplePi:
fix links from `Beatmap_Packs` (old) to `Beatmap_Packs_0916` (current)


MegaApplePi:
Raikire should have fixed it

MegaApplePi:
the linking in the index is OK

PR #380, opened 2017-06-18
CWC 2017: Semifinals update

Loctav:


wiki/Tournaments/CWC/2017/en.md
```diff
@@ -47,20 +47,40 @@ The osu!catch World Cup 2017 is run by various community members by distributing
 
 ------------------------------------------------------------------------
 
-## Match Schedule: Quarterfinals
+## Match Schedule: Semifinals
 
-### Sunday, 18. June 2017
+**Disclaimer: Just UTC times are guaranteed to be correct! No responsibility can be taken for the correctness of the local times!**
```
peppy: i can't understand what this means

peppy: if you aren't going to guarantee correctness just don't include local times?

PR #416, opened 2017-06-27
Fix titles for Rules/Short, Storyboard Scripting, AiBat

TPGPL:


wiki/Rules/Short/en.md
```diff
@@ -1,5 +1,5 @@
-Simplified osu! Rules
-========================
+Simplified
```
MegaApplePi: following the folder's name, it should be `Short` (otherwise change the folder's name)

PR #426, opened 2017-07-01
Fix timestamp

Loctav:
I hope we finally figured it.


TPGPL:
that's probably why the date of releasing the news post on the site was "one month ago" :eyes:

PR #435, opened 2017-07-04
shared: fix move mods icons to subfolder

MegaApplePi:
- use abbreviations instead of full names
- use lowres images
- downscale true/false icons further


TPGPL:
You should fix the paths to the images in some article?

MegaApplePi:
I don't know, I was going to do it in another pull request.

PR #444, opened 2017-07-08
guides: add Finding_Metadata

MegaApplePi:
quick-ish port from the forums


MegaApplePi:
will reopen after #450 is done

PR #446, opened 2017-07-09
mfc/*: use "Main page" instead of "Click here to ..."

MegaApplePi:
- use "Main page" instead of "Click here to ..."


peppy:
Isn't this whole "Announcements" thing just duplicating new posts?

MegaApplePi:
iirc, "Announcements" was originally "Announcements Archive" which only linked to older news posts for the purpose of quicker access. Since-- at the time-- <https://osu.ppy.sh/news/> only listed up to 20 news posts. (But I guess <http://osunews.tumblr.com/archive> also works too)

But now that I think about it... its pretty much <https://osu.ppy.sh/home/news>

MegaApplePi:
to be deleted

PR #360, opened 2017-06-12
[EN] Announcements/Contests

OniJAM:
Update beatmapping contests

PR #455, opened 2017-07-10
Contests

OniJAM:
Since #447 has deleted the Announcements folder, but unofficial contests which are official supported are still hard to be found in the news page, so unofficial contests should have a place to show just like unofficial tournaments do.
Files have been modified and merged from pr #354 and #360.

- move Contests out of the Announcements folder
- remove Announcements
- redirect the redirects


OniJAM:
@Repflez

PR #463, opened 2017-07-10
contests: hotdelete section Monthly Fanart Contest

MegaApplePi:
delete it? ¯\\\_(ツ)\_/¯


RaikireHiuduo:
Need context. Why?

MegaApplePi:
@RaikireHiuduo The MFC article was deleted while the Facebook page doesn't have all of the contests since they are located elsewhere in the osu! web.

Just to side track, I'm not sure if the two notes still applies.

PR #474, opened 2017-07-11
objects: merge beatmap_editor/sliders into Hit_Objects

MegaApplePi:


RaikireHiuduo:
Add new heading called `##Mapping Hit Objects` and put it there? You might also want to account for other hitobjects and osu!taiko mapping too.

Current is valid for osu! explanation only.

xtrem3x:
(in taiko) Spinners+Sliders works but Sliders+Hit Objets not ...

MegaApplePi:
@RaikireHiuduo I've began drafting the other game mode explanations, but this was intended to just get rid of the Sliders article

PR #484, opened 2017-07-13
Rules: Remove italics for osu! client

RaikireHiuduo:
osu! client is a term. Also, osu!client or osu! client?


peppy:
osu! client

PR #486, opened 2017-07-13
Fix broken images for Game_Modifiers pages

turbedi:
And delete an unused/duplicate image


Ephemeralis:
Is it not possible to replace the ```` tags with shorthand Markdown? I'm not sure what our styling rules are on this, or if this would even work, though I get the impression that it should.

RaikireHiuduo:
@Ephemeralis It is possible, but it might be an overkill and a waste of effort if the en.md was updated again (which it will need).

turbedi:
Closing for PR #511

PR #487, opened 2017-07-14
[osu!wiki : HOO]Minor fixes

sr229:
Just some minor fixes for the ``osu!wiki`` article (sorry if it's in a seperate PR)


peppy:
This kind of change seems pointless. Also this history page is kind of missing

PR #488, opened 2017-07-14
[OWH] osu!wiki History June Update

RaikireHiuduo:
Supersede #487 .

Updates **June** section of osu!wiki's history. No idea for July (this month).


peppy:
Keep in mind the backup is temporary and that link will break in the near future.

Also choose one name. We don't need "aka"s all over the place.

RaikireHiuduo:
Done.

PR #557, opened 2017-08-04
Changed the osu!catch embed video file (needs an update)

Raveille:


sr229:
You probably wanna edit that Commit PR message @Raveille ;joy:

PR #609, opened 2017-08-20
mods summary: update page

MegaApplePi:
Status: done
Description: updates for `Game_Modifiers/Summary`

## Changelog

- use game mode icons
- remove columns that explains certain things, if they want to read about it, they should just read the main article
- update game modifier icon links
- fix inverse mods section
- add note that the reverse of the inverse mods section is also true
- purge translations


RaikireHiuduo:
I would suggest to put a better generic explanation (as in, one or two sentences and less than 20 words) column before the next person said "DT/NC = 150% song speed" and "HT = 50% song speed" and they decide to pick the `/Summary` to read instead of the main page. The way it is done assumes you already know what those modifiers does.

Looks nice otherwise.

PR #613, opened 2017-08-22
[ZH] Translate Chat_Console/en.md to simplified Chinese

1004121460:


kj415j45:
this article was moves to our organization: https://github.com/osu-translate-zh/osu-wiki/tree/zh-Chat_Console

if we get it done, we'll re-pr to here

PR #673, opened 2017-09-07
Fix missing leading character from QAT gazette article URL link

Ephemeralis:
The rule of three prevails in all things.

Or the rule of tree, I should say.


TPGPL:
owo

PR #720, opened 2017-09-22
Un-implemented rule regarding two sets of one song in qualified after 7 days

Xinnoh:
https://osu.ppy.sh/forum/t/203558/


Ephemeralis:
Okorin/JBH will need to review this one and see if it's relevant to the current criteria or not. It's 3 years old, so hard to say.

sr229:
Requesting an update here.

PR #729, opened 2017-09-25
Add new glossary entries

osuShiro:
Adds the following entries to the Glossary:

- Accuracy
- NM
- Blanket
- BN
- QAT
- pp
- Loved
- Graveyard

Also edits the Modding entry to include "mod".

### DIFF
--- a/wiki/Glossary/en.md
+++ b/wiki/Glossary/en.md
@@ -8,6 +8,10 @@ Throughout the years, the osu!community has adopted its own set of slang that ma
 
 If the name is not obvious enough, the abandoned wasteland is a hidden sub-forum where the instructions to the moderators concerning it will: "Put stuff we do not really need in here." When a thread is or is said to be "wastelanded", it is moved there.
 
+### Accuracy
+
+A [game metric](/wiki/Accuracy) to measure the player's rhythmical accuracy. Sometimes shortened to _Acc_.
+
 ### AFK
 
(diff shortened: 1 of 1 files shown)


MegaApplePi:
I left out (or removed) some of those terms because by having them there, there would be two different locations to link to (the full articles and the glossary term). e.g. QAT redirects to the Quality Assurance Team article, but by having a term in the glossary, either readers will need to type out _Quality Assurance Team_ to see the article (and not the glossary) or the glossary term won't have a redirect link.

osuShiro:
That's a perfectly valid reason - I just thought they were missing. Tell me which ones to remove.

MegaApplePi:
they should be
- accuracy
- BN
- qat
- pp

loved and graveyard might be one too... iirc, they should be explained in the beatmaps page under the ranked statuses section... but I think don't remember if they have links to them or not

wiki/Glossary/en.md
```diff
@@ -8,6 +8,10 @@ Throughout the years, the osu!community has adopted its own set of slang that ma
 
 If the name is not obvious enough, the abandoned wasteland is a hidden sub-forum where the instructions to the moderators concerning it will: "Put stuff we do not really need in here." When a thread is or is said to be "wastelanded", it is moved there.
 
+### Accuracy
+
+A [game metric](/wiki/Accuracy) to measure the player's rhythmical accuracy. Sometimes shortened to _Acc_.
+
 ### AFK
```
Xinnoh: basic internet acronym, doesn't really fit with the rest of the osu specific terms, should remove

ThePooN: Maybe there should be a category for "internet acronyms" used in the game? I think it's fair to have them

PR #749, opened 2017-10-01
External Ports to Similar Games

RaikireHiuduo:
### Description

Move External Ports to Similar Games.

### Status

Done.

### Notes

- `fr.md` Game Modes index page is broken for the external ports redirection
   - So does `Index`.
- Touches FAQ, Similar Games, Game Modes.
- Purges are a mistake.
- Also updated that osu!stream packs that cost money are now all costed and locked at 0.99USD.


RaikireHiuduo:
_nothing really matters to meeeeeeeeeeeeeeeeeeeeeeee_

What to do now? Delete `osu!stream` and fix `Index`?

sr229:
yes

MegaApplePi:
no, keep `osu!stream`

ppy wrote:
image

you could fix `Index` but that probably should be done in another pull request

sr229:
We should have a clear consensus on what to call this category, For instance, we won't call opsu! a "similar game" as it is a clone of osu! by nature.

sr229:
CC @RaikireHiuduo for updates

peppy:
I think "similar games" is fine. Another option would be "osu!-inspired games" i guess

RaikireHiuduo:
Redirection drama I am gonna leave it to @MegaApplePi .

Not in scope for this PR.

Also, done for now. I seriously forgot what I should be fixing here other than similar games title is fine.

PR #752, opened 2017-10-03
[RU] Accuracy

u624ka:
### Description

Clarify the description of hit accuracy
Add the relevant formulae

### Status

Finished; Awaiting Review

### Changelog

- Clarify the description of hit accuracy displayed on the profile (later referred to as "profile accuracy") calculation to explicitly refer to the weight of a score, rather than its pp value
- Specifically name it weighted for the same reason: there is no pp value in the calculation, only the weight
- Remove the explicit statement that only the top 100 scores affect the profile accuracy, since the change in weight is gradual and has no clear-cut boundary, and instead specify that each score affects the profile accuracy proportionally to its weight

- Add the step-by-step description of how exactly the profile accuracy is calculated
- Specify that the description applies to all game modes


wiki/Accuracy/ru.md
```diff
@@ -17,7 +17,7 @@
 
 Данный параметр убран с сайта, и увидеть его можно только на экране результатов, если прокрутить вниз.
 
--   [PP](/wiki/Performance_Points)-точность. Она считается почти так же, как общая, но в расчётах участвует ещё и количество PP, полученное за каждую карту: чем больше PP у результата, тем сильней он влияет на точность. Таким образом, как и при расчёте игрового ранка, на PP-точность влияют только 100 лучших рекордов; всё, что находится ниже, имеет вес, близкий к нулю.
+-   Взвешенная общая точность. Именно она отображается в профиле, и расчитывается по тому же принципу, что и [PP](/wiki/Performance_Points): влияние точности отдельно взятого результата прямо пропорционально его весу (указано на сайте как "weighted **X%**").
```
kj415j45: actually, you can use `/wiki/pp` :p

PR #760, opened 2017-10-08
Update community rule 1 to prevent wrangling weasels wresting worryingly wide

Ephemeralis:
I seriously cannot believe I had to write this.

### DIFF
--- a/wiki/Rules/en.md
+++ b/wiki/Rules/en.md
@@ -6,7 +6,7 @@ These rules are the basis from which we help keep the osu! community a fun and w
 Community Rules
 ---------------
 
-1.  **Each player may only have *ONE* account at any point of time.** This account is you. It is not anyone else - not your brother, your mother, your sister, your friend - it is YOU. Don't share your account with anyone else.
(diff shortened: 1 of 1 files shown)


wiki/Rules/en.md
```diff
@@ -6,7 +6,7 @@ These rules are the basis from which we help keep the osu! community a fun and w
 Community Rules
 ---------------
 
-1.  **Each player may only have *ONE* account at any point of time.** This account is you. It is not anyone else - not your brother, your mother, your sister, your friend - it is YOU. Don't share your account with anyone else.
+1.  **Each player may only have *ONE* account, ever.** The *first* account you make during account registration is the account you will have in *osu!* forever. This account is YOU. It is not anyone else - not your brother, your mother, your sister, your friend - it is YOU. Don't share your account with anyone else. You don't get to make a new one if you lose access to it. Keep it safe.
```
TPGPL: You don't get into You mustnt?

TPGPL: you will have in *osu!* forever. -> you will only use in *osu!*?

Ephemeralis: musn't implies it is a choice they can, but should not consider

'don't get' is conversational for 'you cannot do this. ever.'

Ephemeralis: this one is worded funny and I'll change it

PR #790, opened 2017-10-24
osu(ff): fix pixelLength type

### LINKED ISSUE #789: pixelLength is a double,not a float.
#https://github.com/ppy/osu-wiki/blob/master/wiki/osu!_File_Formats/Osu_(file_format)/en.md
The value pixelLength of a slider is double.

MegaApplePi:
## Description

see title (osu (file format): fix `pixelLength` type)

## Status

Done, anyone else here to double check? ¯\_(ツ)_/¯

## Changelog

~~Modernize from row 233, column 15 to row 233, column 20 to speak `Double`, no more flimsy `Float`, for `pixelLength`. Thus adding an extra column to row 233, totaling 194 columns from previous 193 columns from this point on at 9:30a (UTC-5) until further pull requesting.~~

~~tl;dr~~ changed `Float` to `Double` in line 233

## Related Issues

resolves #789


MegaApplePi:
#814 supersedes this pull request

PR #808, opened 2017-11-01
Remove locales from article headers

cl8n:
these aren't necessary anymore on the site

ready to go 🔥


cl8n:
it covers all first-level headers that are written with either `#` or `===`

I guess if there are `h1` tags they won't be counted, but those articles (if they exist) are in need of rewriting anyway

PR #810, opened 2017-11-01
Remove duplicate storyboard scripting content

cl8n:
all content in the english "miscellaneous" article can be found elsewhere. the japanese one has some differences compared to the existing articles so I've left it for now (might help translators in the future)

ready


RaikireHiuduo:
Define "elsewhere".

Is it in the main page?

cl8n:
In every other article under Storyboard Scripting, each header is its own article now

PR #812, opened 2017-11-01
Add top10 Halloween Fanart winners to newspost

Ephemeralis:
Includes stupid little blurbs and links to full size images.


Loctav:
We made an unheard deal here.

Issue #811, opened 2017-11-01
Expand the .osu format page


fmang:
This one: https://osu.ppy.sh/help/wiki/osu!_File_Formats/Osu_(file_format)

As mentionned in https://github.com/ppy/osu/issues/1443, it's perfectible.

I'm going to work on it at https://github.com/fmang/osu-wiki/blob/formats/osu/wiki/osu!_File_Formats/Osu_(file_format)/en.md and then make a pull request hopefully next week. This issue is for the record in case someone else would like to contribute.

osuShiro:
issued fixed in #814

osuShiro:
pending further fixes

PR #814, opened 2017-11-01
.osu file format: expand the hit objects section

fmang:
### Description

The current description of the hit objects in the article describing the .osu format is far from complete, and a source of frustration when writing a beatmap parser.

To start, I've expanded the [HitObjects] section, which the the subject of this pull request. I'd plan to expand the other sections in later pull requests.

### Status

Help wanted.

I have prefixed some paragraphs with *Unconfirmed* or *TODO* when I was missing a piece of information.

### Related Issues

See #811.

### DIFF
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -201,9 +201,9 @@ This section is made of CSV lines. The first 5 fields are common to every hit ob
 
 To map these coordinates for a standard 640x480 screen, you need to add 64 pixels to *x* and 48 pixels to *y* to respect a uniform padding. Without the padding, an object at (0, 0) will be cut on the top left for the screen.
 
-For some hit objects, like spinners or notes in osu!taiko, the position is completely irrelevant.
+For some hit objects, like spinners, the position is completely irrelevant.
 
-For osu!mania, only *x* is relevant.
+For osu!mania, only *x* is relevant. See the osu!mania hold note section below.
 
 #### Time
 
@@ -369,12 +369,14 @@ A hold note unique to osu!mania.
 
 **Example**: `329,192,16504,128,0,16620:0:0:0:0:`
 
-*x* (Integer) determines which column a note will fall on. The value does not have to be precise. *y* is ignored.
+*x* (Integer) determines which column a note will fall on. *y* is ignored.
 
-TODO: Range of *x*?
+The number of column is defined by the *CircleSize* property in the Difficulty section. Columns are indexed from 0.
 
-TODO: Define *x* more precisely.
+The column for a note is computed with `x / column width` with `column width = 512 / number of columns`. The resulting value is floored, then clamped between 0 and (#column - 1) for safety.
 
-TODO: Default value for *y*?
+To avoid rounding errors, *x* is best picked in the middle of the range for its column. For example, in 4K mode, the first column ranges from 0 to 128, so the safest value for x is 64.
+
+For the 7K + 1 mode, the column index is `1 + x / (512 / 7)`, leaving the column 0 for the specials.
 
 *endTime* (Integer) is the time when the key should be released, in milliseconds from the beginning of the song.
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -191,31 +191,23 @@ Hit Objects
 
 ### Common structure
 
-This section is made of CSV lines. The first 5 fields are common to every hit
-objects, and an optional last *addition* fields.
+This section is made of CSV lines. The first 5 fields are common to every hit objects, and an optional last *addition* fields.
 
 **Syntax**: `x,y,time,type,hitSound...,addition`
 
 #### Position
 
-*x* and *y* are integers representing the position of the center of the hit
-object. *x* ranges from 0 to 512 pixels, inclusive, and *y* ranges from 0 to
-384 pixels, inclusive. The origin, (0, 0) is at the top left of the screen.
+*x* and *y* are integers representing the position of the center of the hit object. *x* ranges from 0 to 512 pixels, inclusive, and *y* ranges from 0 to 384 pixels, inclusive. The origin, (0, 0) is at the top left of the screen.
 
-To map these coordinates for a standard 640x480 screen, you need to add 64
-pixels to *x* and 48 pixels to *y* to respect a uniform padding. Without the
-padding, an object at (0, 0) will be cut on the top left for the screen.
+To map these coordinates for a standard 640x480 screen, you need to add 64 pixels to *x* and 48 pixels to *y* to respect a uniform padding. Without the padding, an object at (0, 0) will be cut on the top left for the screen.
 
-Hit objects whose position is irrelevant, like spinners in osu! or any note in
-osu!taiko, have the special position (256,192), which is the center of the
-screen.
+Hit objects whose position is irrelevant, like spinners in osu! or any note in osu!taiko, have the special position (256,192), which is the center of the screen.
 
 TODO: What about osu!mania?
 
 #### Time
 
-*time* is an integral number of milliseconds from the beginning of the song,
-and specifies when the hit begins.
+*time* is an integral number of milliseconds from the beginning of the song, and specifies when the hit begins.
 
 #### Type
 
@@ -225,16 +217,12 @@ and specifies when the hit begins.
 - Bit 1 (2): slider.
 - Bit 2 (4): new combo.
 - Bit 3 (8): spinner.
-- Bits 4-6 (16, 32, 64) form a 3-bit number (0-7) that chooses how many combo
-  colors to skip. Note that 0 means 1 color is skipped anyway by the new combo
-  field. A skip value of 1 therefore skips 2 colors.
+- Bits 4-6 (16, 32, 64) form a 3-bit number (0-7) that chooses how many combo colors to skip. Note that 0 means 1 color is skipped anyway by the new combo field. A skip value of 1 therefore skips 2 colors.
 - Bit 7 (128) is for an osu!mania hold note.
 
-Circles, sliders, spinners, and hold notes can be OR'd with new combos and the
-combo skip value, but not with each other.
+Circles, sliders, spinners, and hold notes can be OR'd with new combos and the combo skip value, but not with each other.
 
-Unconfirmed: a non-zero combo skip value is ignored when the new combo bit is
-not set.
+Unconfirmed: a non-zero combo skip value is ignored when the new combo bit is not set.
 
 Examples:
 
@@ -244,8 +232,7 @@ Examples:
 
 #### Hit sounds
 
-*hitSound* (Integer) is a bitmap of hit sounds to play when the hit object is
-successfully hit.
+*hitSound* (Integer) is a bitmap of hit sounds to play when the hit object is successfully hit.
 
 - Unconfirmed: Bit 0 (1): hitnormal.
 - Bit 1 (2): hitwhistle.
@@ -260,17 +247,13 @@ Unconfirmed: Multiple sounds will overlap if multiple bits are set.
 
 #### Addition
 
-The *addition* field is optional and define extra parameters related to the hit
-sound samples. It defaults to `0:0:0:0:`.
+The *addition* field is optional and define extra parameters related to the hit sound samples. It defaults to `0:0:0:0:`.
 
-Unconfirmed: This field is deprecated. You may omit it entirely, or write
-`0:0:0:0:` instead.
+Unconfirmed: This field is deprecated. You may omit it entirely, or write `0:0:0:0:` instead.
 
 **Syntax**: `sampleSet:additions:customIndex:sampleVolume:filename`
 
-*sampleSet* (Integer) changes the sample set of the normal hit sound, and
-*additions* (Integer) changes the sample set for the other hit sounds (whistle,
-finish, clap). The values for these are:
+*sampleSet* (Integer) changes the sample set of the normal hit sound, and *additions* (Integer) changes the sample set for the other hit sounds (whistle, finish, clap). The values for these are:
 
 - 0: Auto. In that case, you need to use the sample set information from the timing point.
 - 1: Normal.
@@ -279,16 +262,13 @@ finish, clap). The values for these are:
 
 *customIndex* (Integer) is the custom sample set index, e.g. 3 in `soft-hitnormal3.wav`.
 
-Unconfirmed: The special index 1 doesn't appear in the filename. For example:
-`normal-hitfinish.wav`.
+Unconfirmed: The special index 1 doesn't appear in the filename. For example `normal-hitfinish.wav`.
 
-Unconfirmed: The special index 0 means you need to get the sample index from
-the timing point.
+Unconfirmed: The special index 0 means you need to get the sample index from the timing point.
 
 *sampleVolume* (Integer) is the volume of the sample, and ranges from 0-100 (percent).
 
-*filename* (String) names an audio file in the folder to play instead of sounds
-from sample sets, relative to the beatmap's directory.
+*filename* (String) names an audio file in the folder to play instead of sounds from sample sets, relative to the beatmap's directory.
 
 Unconfirmed: A comma in the *filename* may break everything.
 
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -189,6 +189,111 @@ Combo# (Integer List) is a list of three numbers, each from 0 - 255 defining an 
 Hit Objects
 -----------
 
+### Common structure
+
+This section is made of CSV lines. The first 5 fields are common to every hit
+objects, and an optional last *addition* fields.
+
+**Syntax**: `x,y,time,type,hitSound...,addition`
+
+#### Position
+
+*x* and *y* are integers representing the position of the center of the hit
+object. *x* ranges from 0 to 512 pixels, inclusive, and *y* ranges from 0 to
+384 pixels, inclusive. The origin, (0, 0) is at the top left of the screen.
+
+To map these coordinates for a standard 640x480 screen, you need to add 64
+pixels to *x* and 48 pixels to *y* to respect a uniform padding. Without the
+padding, an object at (0, 0) will be cut on the top left for the screen.
+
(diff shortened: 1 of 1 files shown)


wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+**Syntax**: `x,y,time,type,hitSound...,addition`
 
-A hit circle is a single hit in all osu! game modes.
+#### Position
+
+*x* and *y* are integers representing the position of the center of the hit object. *x* ranges from 0 to 512 pixels, inclusive, and *y* ranges from 0 to 384 pixels, inclusive. The origin, (0, 0) is at the top left of the screen.
+
+To map these coordinates for a standard 640x480 screen, you need to add 64 pixels to *x* and 48 pixels to *y* to respect a uniform padding. Without the padding, an object at (0, 0) will be cut on the top left for the screen.
+
+Hit objects whose position is irrelevant, like spinners in osu! or any note in osu!taiko, have the special position (256,192), which is the center of the screen. For osu!mania, only *x* is relevant.
```
cl8n: maybe note that the position doesn't actually matter either. that value is just the default set by the osu! editor

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+
+#### Type
+
+*type* is a bitmap specifying the object type and attributes.
+
+- Bit 0 (1): circle.
+- Bit 1 (2): slider.
+- Bit 2 (4): new combo.
+- Bit 3 (8): spinner.
+- Bits 4-6 (16, 32, 64) form a 3-bit number (0-7) that chooses how many combo colors to skip. Note that 0 means 1 color is skipped anyway by the new combo field. A skip value of 1 therefore skips 2 colors.
```
cl8n: the note about skips isn't really correct; advancing to the next combo color isn't the same thing as "skipping" a color

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+
+Circles, sliders, spinners, and hold notes can be OR'd with new combos and the combo skip value, but not with each other.
+
+Unconfirmed: a non-zero combo skip value is ignored when the new combo bit is not set.
+
+Examples:
+
+- `1`: circle.
+- `5 = 1 + 4`: circle starting a new combo.
+- `22 = 16 + 4 + 2`: slider starting a new combo, skipping 2 colors.
```
cl8n: order numbers from least to greatest, for consistency

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+- Unconfirmed: Bit 0 (1): normal.
+- Bit 1 (2): whistle.
+- Bit 2 (4): finish.
+- Bit 3 (8): clap.
+
+Unconfirmed: The special value 0 is the same as 1 (hitnormal).
+
+Unconfirmed: If bit 0 is not set, the normal sound is not played.
+
+Unconfirmed: Multiple sounds will overlap if multiple bits are set.
```
cl8n: confirmed 👍

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+
+- *sample set* is normal, soft, or drum.
+- *sound* is normal, whistle, finish, or clap.
+- *index* is the custom index. It is omitted when equal to 0 or 1.
+
+#### Addition
+
+The *addition* field is optional and define extra parameters related to the hit sound samples. It defaults to `0:0:0:0:`.
+
+Unconfirmed: This field is deprecated. You may omit it entirely, or write `0:0:0:0:` instead.
```
cl8n: it's not deprecated at all, this is used in almost every ranked map

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
 
-*x (Integer)* ranges from 0 to 512 (inclusive) and *y (Integer)* ranges from 0 to 384 (inclusive).
+**Syntax**: `sampleSet:additions:customIndex:sampleVolume:filename`
```
cl8n: maybe change `additions` to `additionSet` to be clear about what it actually means. it's also confusing to have two things named "addition" when they aren't the same thing

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+- 0: Auto. In that case, you need to use the sample set information from the timing point.
+- 1: Normal.
+- 2: Soft.
+- 3: Drum.
 
-Bit 0 (1) = circle, bit 1 (2) = slider, bit 2 (4) = new combo, bit 3 (8) = spinner. Bits 4-6 (16, 32, 64) form a 3-bit number (0-7) that chooses how many combo colors to skip. Bit 7 (128) is for an osu!mania hold note. Circles, sliders, and spinners can be OR'd with new combos and the combo skip value, but not with each other.
+*customIndex* (Integer) is the custom sample set index, e.g. 3 in `soft-hitnormal3.wav`.
 
-`1` - circle, `5` - circle starting a new combo, `22` - slider starting a new combo, skipping 2 colors.
+Unconfirmed: The special index 1 doesn't appear in the filename. For example `normal-hitfinish.wav`.
```
cl8n: confirmed 👍

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
 
-*hitSound (Integer)* is a bitmap of hitsounds to play on top of the normal hitsound:
+Unconfirmed: The special index 0 means you need to get the sample index from the timing point.
```
cl8n: confimed 👍

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+A **perfect** circle slider is defined by three points. In that order: start, pass-through, and end. *x* and *y* define the start point, and *curvePoints* defines the pass-through and end point. For example `P|250:200|200:15`.
+
+A perfect circle slider could be represented as a center point, a radius, and two angles for convenience. See this source code for the conversion algorithm: [CircularArcApproximator.cs](https://github.com/ppy/osu/blob/master/osu.Game/Rulesets/Objects/CircularArcApproximator.cs).
+
+A **Bézier** slider is made of one or many Bézier curves, sharing common ends. The degree of each curve is arbitrary. The first control points is defined with *x* and *y*, and the other ones by *curvePoints*.
+
+To identify the separation between two curves, the intersection point is repeated. Consider the sequence ABCDDEFFG. You would get the 3 Bézier curves: ABCD (cubic), DEF (quadratic) , FG (linear).
 
-`424,96,66,2,0,B|380:120|332:96|332:96|304:124,1,130,2|0,0:0|0:0,0:0:0:0:`
+Note that this representation is ambiguous, because ABBBC could actually represent ABBBC, AB+BBC, AB+B+BC, or AB+BBC.
```
cl8n: not sure if this note is important; you'd never see three in a row anyway

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
 
-*x*, *y*, *time*, and *type* behave the same as described in Hit Circle Syntax.
+Example: `476,340,6419,2,0,B|437:336|422:309|422:309|384:309|359:337|359:337|328:308|300:304|300:304|272:352|237:383|176:356|159:287|224:256,1,420,6|2,0:0|0:0,0:0:0:0:`
 
-*hitSound* applies to the body of the slider, but only whistle sounds will play during the slider body.
+The first Bézier curve in the segment is quadratic with the following points: (476, 340), (437, 336), (422, 309).
 
-*sliderType* will be `L` (linear), `P` (perfect), `B` (Bezier), or `C` (Catmull). A slider created in the editor with only a start and end point will be a linear slider. A slider with only its start, end, and one grey point will be a perfect circle slider. All others will be Bezier. Catmull sliders are deprecated.
+Note that the multiple curves in the same Bézier slider need not have the same length, so you *will* need to approximate their lengths, otherwise the slider ball would speed up or slow down brutally in the middle of a slider.
```
cl8n: what do you mean here? this sounds like you're talking about an external project

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
 
-*repeat (Integer)* is the number of times a player will go over the slider. A value of 1 will not repeat, 2 will repeat once, 3 twice, and so on.
+*repeat* (Integer) is the number of times a player will go over the slider. A value of 1 will not repeat, 2 will repeat once, 3 twice, and so on.
+
+#### Duration
 
 *pixelLength (Float)* is the length of the slider along the path of the described curve. If the length is greater than that of the described curve, the slider will continue in a straight line.
 
-*edgeHitsounds (hitSound|...)* is a `|`-separated list of hitSounds to apply to the circles of the slider. The values are the same as those for Hit Circle hitSounds.
+TODO: What's the unit? Is it really pixels?
```
cl8n: osu!pixels, yeah. it's in the same units as the 512 x 384 virtual screen

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+
+TODO: What happens if a Bézier or perfect circle slider's *pixelLength* is too big? Would you still get a straight line at the end?
```
cl8n: perfect circle will extend the curve. bezier makes a straight line

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Is it only about the last point or the above points too?

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: So far I've only downloaded beatmap packs, and never found a single beatmap using that feature. I'll check out some ranked beatmaps to extract some examples.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: This is an edge case, and I did find a beatmap with ABB. If we were consistent, it's would become AB+B, which is the same as AB. However, it might as well be a quadratic curve ABB.

My point was that this edge case means the editor must ensure you won't see 2 identical control points in a row for a given curve. The cubic curve ABBC is perfectly reasonable, so I'll focus on it rather than use ABBBC which is more convoluted. What do you think?

wiki/osu!_File_Formats/Osu_(file_format)/en.md
cl8n: most hitobjects will be all zero, because they can simply inherit that information from timing points, but there should be at least a few in nearly every map that's properly hitsounded. [random example](https://osu.ppy.sh/b/1195081)

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Haha, I actually am. I've had that problem writing [my port](https://github.com/fmang/oshu).

I'll drop that paragraph, because it's not related to the format but is more a gotcha for port makers.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
cl8n: oh, I just meant the point about three in a row. beatmaps will never use more than two duplicate points for the reasons you described

I think you can just remove this line and the rest of the explanation is fine. It's true that this implementation does not do justice to all possible bezier curves, but it's not worth mentioning here because it's beyond the scope of osu! (and mapping with sliders). If the reader already knows what bezier curves are, they will understand it anyway, just as you have

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: By the way, is it safe to assume the editor will always write the correct length?

wiki/osu!_File_Formats/Osu_(file_format)/en.md
MegaApplePi: If you are going to refer to this document for making such a port, then I'd assume that others would as well. You could rephrase it to so it can target them, `Note for third party project devs, [...]` or similar. ¯\\\_(ツ)\_/¯

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Aw, this is complicated.

What if the pixelLength is shorter than the curve? Is the curve shrinked?

To summarize, the parser shoud 1. parse the path, 2. compute its actual length, 3. extend the path until its length matches pixelLength. Right?

Also, why would anyone make the actual length and pixelLength differ? For a perfect circle, it's just about moving the end point, and for a Bézier curve it's about adding an explicit straight curve. I guess the answer is *historical reasons* here.

cl8n: for your parser, that sounds about right. but keep in mind that most sliders will be *shorter* than their underlying curve (when the length is shorter, it follows the same shaped path, but only as much as it can)

I think you're missing the point of this feature. without it, it'd be nearly impossible to create on-time sliders--- the editor is designed so that the mapper can build a curved path for a slider to sit on. the slider won't always fill the entire curve (in fact, it almost never will), but it'll fill as much as it needs to to snap in time with the currently selected beat divisor

so, there are three reasons that the curve length and pixelLength differ:
- the curve length is almost never the correct length to snap in time with the music. and that's the most important part of sliders
- placing points so that they *exactly* match to both the curve and the music is impossible for most sliders because it'd require the user to place points on a non-integral position
- if they didn't differ, why would it even be written here? you can calculate the length of any curve given just it's points

just try it in the editor yourself if you don't understand how it works. it's like this because it's most intuitive for the mapper, not because of historical reasons

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Reformulated.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Fixed.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Fixed.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: You're right. I think I'll rename the whole addition field to *extras* for even more clarity.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Got it! Thank you for the detailed explanation.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
cl8n: just line 244 here. not sure about the others

fmang:
Only the notes about hit sounds remain unconfirmed. I'll dig the source code of ppy/osu a bit or seek help on Discord for answers.

Thank you very much clayton for your help so far!

fmang:
Should be okay now, thanks to smoogipoo and Damnae.

It's not perfect, but I hope it's a good improvement over the current version.

More to come in future pull requests.

MegaApplePi:
@fmang can you verify or do you know if #789 is correct, if so, can you change it?

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -176,90 +179,206 @@ Example of an inherited Timing Point:
...
+#### Repeat
 
-*curvePoints (x:y|...)* is a series of `|`-separated coordinates describing the control points of the slider. Red points appear twice. NOTE: curvePoints is separated from sliderType with a `|`, not a comma.
+*repeat* (Integer) is the number of times a player will go over the slider. A value of 1 will not repeat, 2 will repeat once, 3 twice, and so on.
 
-*repeat (Integer)* is the number of times a player will go over the slider. A value of 1 will not repeat, 2 will repeat once, 3 twice, and so on.
+#### Length and duration
 
-*pixelLength (Float)* is the length of the slider along the path of the described curve. If the length is greater than that of the described curve, the slider will continue in a straight line.
+*pixelLength* (Float) is the length of the slider along the path of the described curve. It is specified in osu!pixels, i.e. relative to the 512×384 virtual screen.
```
cl8n: yeah, it's a `Double` here

fmang: I believe there's the *format*, and the *parser implementation*.

For a format specification, what matters is the notation. Here a *floating number* is implicitly something like `[0-9]*(.[0-9]*)?`. I could write a thousand decimals of pi, it would not matter and it would *work* because we're talking about a textual representation, not a binary representation. Conversely, the value 0.1 cannot be expressed exactly as a Double, technically.

The notion of *Double* is only relevant to the C# implementation, and doesn't even matter in most cases because 32-bit is already super precise for a number which is most of the time integral. High-level languages like Python or JavaScript don't even have the word double, because their floats are already 64-bit.

Even from the point of view of a mapper with no programming experience, the word *double* is going to be confusing I think.

cl8n: by that logic I'd just write "Decimal" then

fmang: Sold!

PR #828, opened 2017-11-08
OWC2017: Update rosters

Loctav:
### Description

Just one more update to go (3 teams left pending)


Ephemeralis:
IF YOU WANNA JAM

PR #840, opened 2017-11-14
Remove perfect sliderbody overlapping rule

Okorin:
### Description

taken from https://osu.ppy.sh/community/forums/topics/634624
reasons for this that i can think of additionally are
- maps can be made intuitive around the concept by introducing the concept instead of doing it randomly
- snaking sliders kind of defeat the point of this

when trying to move it to guidelines instead the following was the best i could do:

-   **When perfectly overlapping two slider bodies, the first slider should be faded out fully before the second one is fully faded in.** While beatmaps can be built around this, doing so without properly introducing it can cause reading hazards.

this doesnt sound convincing at all and just seems random among all other guidelines
=> opted for removal

### Status

someone merge or something help

### Changelog

removed `When perfectly overlapping two slider bodies, the first slider must be fully faded out before the second slider is fully faded in.`

### DIFF
--- a/wiki/Ranking_Criteria/osu!/en.md
+++ b/wiki/Ranking_Criteria/osu!/en.md
@@ -25,6 +25,7 @@ Refer to [this thread](https://osu.ppy.sh/forum/t/178700) for alternative diffic
 
 ### Gameplay
 
+- **Avoid perfectly overlapping slider bodies in a way that causes reading issues.** Doing so can cause sliders to be misread as a circle due to obscuring the slider body.
 - **Jump:** Hit objects spaced further apart from each other in comparison to the average spacing for such patterns. Usually snapped to 1/2 beats.
 - **Stream:** Consecutive circles grouped together. Usually snapped to ¼ beats.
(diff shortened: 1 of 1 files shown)


TheMoMan:
Guideline suggestion:

- **Avoid perfectly overlapping slider bodies in a way that causes reading issues.** Doing so can cause sliders to be misread as a circle due to obscuring the slider body.

Clearer, has reasoning, and doesn't have that "don't do this unless you know what you're doing" vibe.

wiki/Ranking_Criteria/osu!/en.md
```diff
@@ -25,6 +25,7 @@ Refer to [this thread](https://osu.ppy.sh/forum/t/178700) for alternative diffic
 
 ### Gameplay
 
+- **Avoid perfectly overlapping slider bodies in a way that causes reading issues.** Doing so can cause sliders to be misread as a circle due to obscuring the slider body.
```
TheMoMan: Probably should be under guidelines and not the glossary lol

PR #861, opened 2017-11-25
Beatmap_Nominators: Move from BNG to BN

### LINKED ISSUE #713: Move Beatmap_Nomination_Group to Beatmap_Nominators
Just a reminder for the next rewrite to move the `Beatmap_Nomination_Group` articles to `Beatmap_Nominators` (and stop referring the BNG as the BNG and use BN instead).
Ref: <https://github.com/ppy/osu-wiki/pull/561#discussion_r132090032>

MegaApplePi:
## Description

#713 

## Status

Completed; a quick review would be nice ✨ 

## Related

resolves #713


peppy:
Do we usually just nuke all translations in these rename cases?

MegaApplePi:
No, I don't think we do. I wasn't sure what to do with `es.md` and `tr.md` since both articles needed some quick translation fixes regarding the use of BNG and Beatmap Nomination Group. Should I put them back?

PR #872, opened 2017-11-26
redirect: update/fix and rearrange keys

### LINKED ISSUE #887: Remove case-only redirects
We now have case-insensitive url support, so any case-sensitive redirects can now be removed.

MegaApplePi:
### Description

update some redirects

### Status

~~Done~~
~~WIP~~
Done

Resolves #887.


wiki/redirect.yaml
```diff
@@ -497,102 +503,102 @@
...
-"guides/osu!mania_mapping_guide":                             "Guides/osu!mania_Mapping_Guide"
-"guides/recording_osu!":                                      "Guides/Recording_osu!"
-"guides/searching_and_downloading_beatmaps":                  "Guides/Searching_and_Downloading_Beatmaps"
-"guides/setting_the_offset_on_the_correct_beat":              "Guides/Setting_the_Offset_on_the_Correct_Beat"
-"guides/starting_a_beatmap_project":                          "Guides/Starting_a_Beatmap_Project"
-"guides/tablet_purchase":                                     "Guides/Tablet_Purchase"
-"guides/tips_and_tricks_on_skill_improvement":                "Guides/Tips_and_Tricks_on_Skill_Improvement"
-"guides/videos_from_youTube":                                 "Guides/Videos_From_YouTube"
 "adding_custom_hit_sounds":                                   "Guides/Adding_Custom_Hit_Sounds"
+"guides/adding_custom_hit_sounds":                            "Guides/Adding_Custom_Hit_Sounds"
```
peppy: are these redirects actually needed where the only difference is case?

MegaApplePi: Yes, since if someone were to type `guides/Adding_Custom_Hit_Sounds` (with a lowercase G), `guides/adding_custom_hit_sounds` (with all lowercase), or any in letter case (except for the ones used by the folder) it would 404.

I had thought of a possible fix for this that would remove most of these, but it may cause some unintended things to happen (mainly, the subtitle would be using the wrong case since it appears to be using the path). That being said, I was going to ask about using lowercase letters for all folder names (so osu!web could strtolower the path).

peppy: Tracking at https://github.com/ppy/osu-web/issues/1817

MegaApplePi:
should I just remove the whitespace entirely and leave it at that?

peppy:
it can't change every commit, put it that way.

MegaApplePi:
will redo soon:tm:

PR #873, opened 2017-11-26
Divide MWC2014 into 4K and 7K

TPGPL:
### Description

Divided the MWC2014 article into 4K and 7K and added redirects for those articles.

### Status

Done/Review


peppy:
why closed?

WalterToro:
Because MWC2014 was one tournament with two brackets, so, its wrong to split the article.

peppy:
I'd be great to leave a comment in this case, rather than just closing.

PR #894, opened 2017-12-03
.osu: Expand the events, timing points, and colours sections

fmang:
### Description

I'm back with more sections.

The main work here is around the [TimingPoints] section, which was missing a few key elements like the default custom index, and wasn't clear about inherited timing points.

I have also expanded the [Events] and [Colours] section without much knowledge because they were almost empty in the first place.

### Status

Pending review.

### Related Issues

This is the second pull request for issue #811.

### DIFF
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -165,11 +165,23 @@ See [Storyboard Scripting](/wiki/Storyboard_Scripting)
 Timing Points
 -------------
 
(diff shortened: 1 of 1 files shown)


wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -160,16 +160,42 @@ SliderTickRate (Decimal) specifies how often slider ticks appear. Default value
...
+
+### Breaks
+
+**Syntax**: `2,start,end`
+
+The *start* and *end* fields are both an integral number of milliseconds from the beginning of the song defining the start and end point of the break period, respectively.
+
+### Storyboards
+
+Storyboards are now defined in a separate *.osb* file.  See [Storyboard Scripting](/wiki/Storyboard_Scripting).
```
peppy: Might be worth adding a note that you can still add local storyboards for difficulty specific elements (in conjunction with an .osb or without).

fmang:
Updated. Thank you for the feedback.

peppy:
nice work on this one, looking good!

PR #928, opened 2017-12-17
[Help wanted] .osu: Expand the Difficulty section

fmang:
### Description

The difficulty section of the beatmaps is by far the most obscure one, because the property names are ambiguous, and the units completely unspecified.

> CircleSize (Decimal) specifies the size of hit object circles.

My feel: https://www.youtube.com/watch?v=OYt1kqDNlMY

I think that all the properties deserve a proper conversion formula to common units: seconds for time, and osu!pixels for lengths.

The OverallDifficulty property might require a table to show the 50, 100 and 150 time windows, depending on the mode.

### Status

Help wanted.

Search for lines starting with TODO.

### Related Issues

This is 3rd merge request for https://github.com/ppy/osu-wiki/issues/811

### DIFF
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -133,29 +133,41 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
 Difficulty
 ----------
 
-HPDrainRate (Decimal) specifies the HP Drain difficulty.
+### Stars
 
-`HPDrainRate:5`
+All the difficulty properties defined in this subsection are expressed in *stars*, where 1 star means very easy, and 9 stars means very hard. The number of stars is always a decimal number.
 
-CircleSize (Decimal) specifies the size of hit object circles.
+```
+HPDrainRate:4
+CircleSize:3.3
+OverallDifficulty:4
+ApproachRate:4.8
+```
 
-`CircleSize:4`
+*HPDrainRate* specifies how fast the health bar decreases.
 
-OverallDifficulty (Decimal) specifies the amount of time allowed to click a hit object on time.
+TODO: The health bar decreases continuously, but also when the player misses a note. Which of the two this settings affect? How much exactly?
 
-`OverallDifficulty:6`
+*OverallDifficulty* is the most visible parameter. It defines the number of stars of a beatmap appears with in the overview. It affects the time window a hit object remains clickable.
 
-ApproachRate (Decimal) specifies the amount of time taken for the approach circle and hit object to appear.
+TODO: Doesn't it affect anything else? How is the time window defined?
 
-`ApproachRate:7`
+*CircleSize* defines the size of the hit circles and sliders in the osu!standard mode. The resulting circle radius in osu!pixels is defined by the formula `54.4 - 4.48 * CircleSize`. In osu!mania mode, *CircleSize* is the number of columns.
 
-SliderMultiplier (Decimal) specifies a multiplier for the slider velocity. Default value is 1.4 .
+*ApproachRate* defines how long hit objects appear before they should be hit.
 
-`SliderMultiplier:1.3`
+TODO: How to convert the approach rate to seconds?
 
-SliderTickRate (Decimal) specifies how often slider ticks appear. Default value is 1.
+### Sliders
 
-`SliderTickRate:1`
+```
+SliderMultiplier:1.3
+SliderTickRate:1
+```
+
+SliderMultiplier (Decimal) specifies a multiplier for the slider velocity. The reference velocity at slider multiplier = 1 is 100 osu!pixels per beat. A slider multiplier of 2 would yield a velocity of 200 osu!pixels per beat. The default slider multiplier is 1.4 when the property is omitted.
+
+TODO: SliderTickRate (Decimal) is the number of ticks per beat. It defaults to 1 tick per beat.
 
 Events
 ------
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -135,7 +135,7 @@ Difficulty
 
 ### Stars
 
-All the difficulty properties defined in this subsection are expressed in *stars*, where 1 star means very easy, and 9 stars means very hard. The number of stars is always a decimal number.
+All the difficulty properties defined in this subsection are expressed in *stars*, where 0 star means very easy, and 10 stars means very hard.
 
 ```
 HPDrainRate:4
@@ -148,15 +148,40 @@ ApproachRate:4.8
 
 TODO: The health bar decreases continuously, but also when the player misses a note. Which of the two this settings affect? How much exactly?
 
-*OverallDifficulty* is the most visible parameter. It defines the number of stars of a beatmap appears with in the overview. It affects the time window a hit object remains clickable.
+*CircleSize* (CS) defines the size of the hit circles and sliders in the osu!standard mode. The resulting circle radius in osu!pixels is defined by the formula `54.4 - 4.48 * CircleSize`. Ranked beatmaps must have a CircleSize value betwenn 2 and 7, inclusive. In osu!mania mode, *CircleSize* is the number of columns.
 
-TODO: Doesn't it affect anything else? How is the time window defined?
+### Overall Difficulty
 
-*CircleSize* defines the size of the hit circles and sliders in the osu!standard mode. The resulting circle radius in osu!pixels is defined by the formula `54.4 - 4.48 * CircleSize`. In osu!mania mode, *CircleSize* is the number of columns.
+*OverallDifficulty* (OD) is the harshness of the hit window and the difficulty of spinners.
 
-*ApproachRate* defines how long hit objects appear before they should be hit.
+- When OD < 5: `spins_per_second = 5 - 2 * (5 - OD) / 5`
+- When OD > 5: `spins_per_second = 5 + 2.5 * (OD - 5) / 5`
+- When OD = 5: `spins_per_second = 5`
 
-TODO: How to convert the approach rate to seconds?
+TODO: How is the time window defined?
+
+### Approach Rate
+
+*ApproachRate* (AR) defines when hit objects start to fade in relative to when they should be hit.
+
+```
+                                        X = perfect hit
+                p r e e m p t           ↓
+ ├───────────────────────┬──────────────┤
+ 0%      fade_in           100% opacity
+```
+
+The circle starts fading in at `X - preempt` with:
+
+- When AR < 5: `preempt = 1200ms + 600ms * (5 - AR) / 5`
+- When AR > 5: `preempt = 1200ms - 750ms * (AR - 5) / 5`
+- When AR = 5: `preempt = 1200ms`
+
+The amount of time it takes for the hit object to completely fade in is also reliant on the approach rate:
+
+- When AR < 5: `fade_in = 800ms + 400ms * (5 - AR) / 5`
+- When AR > 5: `fade_in = 800ms - 500ms * (AR - 5) / 5`
+- When AR = 5: `fade_in = 800ms`
 
 ### Sliders
 
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -150,7 +150,7 @@ ApproachRate:4.8
 
 The resulting circle radius in osu!pixels is defined by the formula `32 * (1 - 0.7 * (CircleSize - 5) / 5)`, alternatively written `54.4 - 4.48 * CircleSize`.
 
-Ranked beatmaps must have a CircleSize value betwenn 2 and 7, inclusive.
+Ranked beatmaps must have a CircleSize value between 2 and 7, inclusive.
 
 In osu!mania mode, *CircleSize* is the number of columns.
 
@@ -166,9 +166,9 @@ The width of the hit window is defined by the following table:
 
 | Score | Hit Window |
 | --- | --- |
-| 50 | `150ms + 50ms * (5 - AR) / 5` |
-| 100 | `100ms + 40ms * (5 - AR) / 5` |
-| 300 | `50ms + 30ms * (5 - AR) / 5` |
+| 50 | `150ms + 50ms * (5 - OD) / 5` |
+| 100 | `100ms + 40ms * (5 - OD) / 5` |
+| 300 | `50ms + 30ms * (5 - OD) / 5` |
 
 ### Approach Rate
 
--- a/wiki/osu!_File_Formats/Osu_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Osu_(file_format)/en.md
@@ -155,10 +155,16 @@ TODO: The health bar decreases continuously, but also when the player misses a n
 *OverallDifficulty* (OD) is the harshness of the hit window and the difficulty of spinners.
 
 - When OD < 5: `spins_per_second = 5 - 2 * (5 - OD) / 5`
-- When OD > 5: `spins_per_second = 5 + 2.5 * (OD - 5) / 5`
 - When OD = 5: `spins_per_second = 5`
+- When OD > 5: `spins_per_second = 5 + 2.5 * (OD - 5) / 5`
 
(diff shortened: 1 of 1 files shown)


wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,41 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
+TODO: Doesn't it affect anything else? How is the time window defined?
 
-`ApproachRate:7`
+*CircleSize* defines the size of the hit circles and sliders in the osu!standard mode. The resulting circle radius in osu!pixels is defined by the formula `54.4 - 4.48 * CircleSize`. In osu!mania mode, *CircleSize* is the number of columns.
 
-SliderMultiplier (Decimal) specifies a multiplier for the slider velocity. Default value is 1.4 .
+*ApproachRate* defines how long hit objects appear before they should be hit.
 
-`SliderMultiplier:1.3`
+TODO: How to convert the approach rate to seconds?
```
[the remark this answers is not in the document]
MegaApplePi: on second thought, it seems like you want a formula...  ¯\_(ツ)_/¯.. never mind about this

fmang:
That's right, I'm looking for the formula. However, your table is quite relevant as it will help confirm theory.

I found the relevant source file in ppy/osu: `osu.Game.Rulesets.Osu/Scoring/OsuPerformanceCalculator.cs`

I'll dig into that and extract what I can.

FreezyLemon:
Some things:
- The file you linked there is not what you're looking for. That one is for PP calculations.
- I know that the old beatmap listing shows the difficulty values as stars, but I'm not sure if they should be called that. It can easily be confused with star rating. (also the new site doesn't use stars anymore)
  - the values generally go from 0 to 10 and not 1 to 9
- That term "decimal number" really bothers me for some reason. Do you mean a number base 10? A number that is represented with a comma? (integers don't contain commas, are they still valid?)
  - technically the maps can have any value within limits. the game (stable and lazer) is able to handle values of e.g. 4.928174, as long as it is within bounds. Not sure if you want to mention that, since it is unrankable and requires file editing.
- `OverallDifficulty` is definitely not the Star Rating that appears in the overview or game. It's a setting that applies to the size of hit windows and spinner difficulty (the amount of spins required to pass a spinner)
- You forgot to mention that CS is capped. In the current editor, you can only get values in [2, 7]. You can ofc file edit to change these, but it's unrankable. should probably at least mention the cap
- The explanation for AR is kinda... meh. You might want to say something like "AR defines when hit objects start to fade in relative to when they should be hit" or something like that, that would be more precise. 

#### On to the TODOs (everything based on the osu! gamemode):

##### AR ([source](https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu/Objects/OsuHitObject.cs#L78))
For a better image, let's say we have a hitcircle positioned at time X. This X is the "middle" of the hit window, so hitting the circle when you are at X is a perfect hit.

The circle starts fading in at `X - preempt` with preempt being split into two formulas:
AR < 5: `preempt = 1200 + 600 * (5 - AR) / 5`
AR > 5: `preempt = 1200 - 750 * (AR - 5) / 5`
AR == 5 is 1200. (all values in milliseconds)

The amount of time it takes for the hitcircle to completely fade in is also reliant on AR:
AR < 5: `fade_in = 800 + 400 * (5 - AR) / 5`
AR > 5: `fade_in = 800 - 500 * (AR - 5) / 5`
AR == 5 is 800. (again, everything in milliseconds)

##### OD ([source](https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu/Objects/Spinner.cs#L26))
`spins_required = duration_in_seconds * spins_per_second`
again, split up into two formulas:
OD < 5: `spins_per_second = 5 - 2 * (5 - OD) / 5`
OD > 5: `spins_per_second = 5 + 2.5 * (OD - 5) / 5`
OD == 5 is 5. 

Can't tell you about the hit window timing, since it's not implemented in lazer yet (or at least I can't find it).

##### CS ([source](https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu/Objects/OsuHitObject.cs#L81))
Since the lazer window is easily resizable, it also has a different approach to Circle Size. Instead of having some kind of fixed formula that calculates exact pixels, lazer scales up and down based on CS values:
`scale = (1 - 0.7 * (CS - 5) / 5) / 2`
This is linear scaling that does not split into two formulas like the other values.

HP is similar to OD and hit window timing, I don't think it's implemented yet.

--
If you have any other questions or need more info for other game modes, feel free to ping me or mention me on discord and I'll see what I can find.

fmang:
Thank you @FreezyLemon for the review.

I have integrated your formulas with hardly any edit since your explication is clear.

For hit windows, one of your links point to the answer, doesn'it it? [Here](https://github.com/ppy/osu/blob/098cfa71074364563f13b4ab8362a2bc5e0133a5/osu.Game.Rulesets.Osu/Objects/OsuHitObject.cs#L81-L83). Applying the same logic as the approach rate, I'll fill the section.

Since OD is not the number of stars in the listing, how are these stars computed then? Is it a mix of everything?

For the CS, there is an [OBJECT_RADIUS](https://github.com/ppy/osu/blob/098cfa71074364563f13b4ab8362a2bc5e0133a5/osu.Game.Rulesets.Osu/Objects/OsuHitObject.cs#L16), multiplied with the *Scale*, to compute the final [Radius](https://github.com/ppy/osu/blob/098cfa71074364563f13b4ab8362a2bc5e0133a5/osu.Game.Rulesets.Osu/Objects/OsuHitObject.cs#L40). The formula I wrote is a substituted and normalized variant of the *Radius* property. A scale of 0.5 doesn't mean anything without that reference of 64 osu!pixels.

FreezyLemon:
You're welcome :)

I like what you did with AR, that's a pretty good visualisation.
The logic for the hit window times got added after I made that comment (the links always refer to the current version of the code).

One small thing regarding CS: Yes, I know the scale does not mean anything without an initial size. But that's not really the issue here. Isn't the Wiki supposed to give info for the current osu! client? Because that one still uses the old way of directly calculating pixel sizes. You should at least include that information, and maybe add the lazer thing as a footnote. Just my two cents on it though.

The star rating... let's say it's somewhat complicated, and it's also very different for each and every game mode. If you want to look into it yourself, take a look in `osu.Game.Rulesets.<Mode>` and look for the `<Mode>DifficultyCalculator`. I can't guarantee anything though :P. 

I'll give you a TL;DR: for osu! standard. If you want to know it for the other game modes, ask :D .

### Star Difficulty in osu! "standard"

There are two important "skills" that it considers: Aim and Speed. These skills can calculate something called "strain" for every hitobject. These strain values basically represent how difficult a hitobject is for that specific skill. ("high Aim strain" meaning this was a hitobject that was difficult to aim at)

#### Aim

Aim calculates its strain value for an object based on the distance it is apart from the previous object and the time between the two objects (this one is kinda simple).

#### Speed

Speed tries to figure out what kind of "object composition" you're hitting at the moment:
"is this a stream?", "are these single circles (e.g. a jump)?".
It then decides on a constant to use (single objects > streams > closely stacked > directly stacked) and divides that by the time between this and the last object. (so optimal values would be fast maps (low delta time) and single objects (high constant value))

#### Putting these together for the whole map
The map is split up into sections of 400ms, and then for each section, the highest value of "strain" is calculated. It then creates a [weighted sum](https://en.wikipedia.org/wiki/Weight_function#Discrete_weights) (w = 0.9) for every skill, starting at the highest strain values. This means that technically (**and very simplified**), difficulty peaks award a higher strain value here than a continuously difficult map of the same caliber.

It then adds both these weighted sums plus half the difference between them together, and there you have it. Keep in mind that this is only the star rating and does not include mods (note: DT just means 1.5x speed here), I tried to keep it... _simple_

Fun fact: the "aim" and "speed" strain values are also used for (at least the local) PP calculations, though that is _way_ more complicated.

fmang:
There. I think I'm done. It's far from perfect but gotta move on. Thank you everyone.

@FreezyLemon Nice explanation for the stars. I don't think it's relevant to integrate it in the .osu article though. Is there a related wiki article on this?

For the CS, the radius constant is hard-coded clear as day in osu!lazer, so I'll leave it in the formula.

@TPGPL Could you merge this if this looks good to you?

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
 Difficulty
 ----------
 
-HPDrainRate (Decimal) specifies the HP Drain difficulty.
+The following 4 difficulty properties are expressed in *stars*, where 0 star means very easy, and 10 stars means very hard.
```
MegaApplePi: - `The following 4 difficulty` -> `The following four difficulty`
- `where 0 star means very easy, and 10 stars means very hard.` -> `where 0 stars is very easy while 10 stars is very hard.`

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
 
-`HPDrainRate:5`
+```
+HPDrainRate:4
+CircleSize:3.3
+OverallDifficulty:4
+ApproachRate:4.8
+```
+
+*HPDrainRate* specifies how fast the health bar decreases. At the time of this writing, the near future definition of this property remains unsettled.
```
MegaApplePi: - put `HPDrainRate` in code instead of italics-> `` `HPDrainRate` `` (and for the others too)
- `At the time of this writing, the near future definition of this property remains unsettled.` -> `The near future definition of this property is unknown.` (or just remove the first part)

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
+*CircleSize* (CS) defines the size of the hit circles and sliders in the osu!standard mode.
 
-CircleSize (Decimal) specifies the size of hit object circles.
+The resulting circle radius in osu!pixels is defined by the formula `32 * (1 - 0.7 * (CircleSize - 5) / 5)`, alternatively written `54.4 - 4.48 * CircleSize`.
 
-`CircleSize:4`
+Ranked beatmaps must have a CircleSize value between 2 and 7, inclusive.
 
-OverallDifficulty (Decimal) specifies the amount of time allowed to click a hit object on time.
+In osu!mania mode, *CircleSize* is the number of columns.
```
TPGPL: In osu!mania, *CircleSize** is replaced with the Key Amount.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
 
-`OverallDifficulty:6`
+### Overall Difficulty
 
-ApproachRate (Decimal) specifies the amount of time taken for the approach circle and hit object to appear.
+*OverallDifficulty* (OD) is the harshness of the hit window and the difficulty of spinners.
```
TPGPL: harshness into size/stretch/width? not sure

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
-`SliderMultiplier:1.3`
+| Score | Hit Window |
+| --- | --- |
+| 50 | `150ms + 50ms * (5 - OD) / 5` |
+| 100 | `100ms + 40ms * (5 - OD) / 5` |
+| 300 | `50ms + 30ms * (5 - OD) / 5` |
+
+### Approach Rate
+
+*ApproachRate* (AR) defines when hit objects start to fade in relative to when they should be hit.
```
TPGPL: relative -> relatively

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
+
+### Sliders
+
+```
+SliderMultiplier:1.3
+SliderTickRate:1
+```
 
-SliderTickRate (Decimal) specifies how often slider ticks appear. Default value is 1.
+SliderMultiplier (Decimal) specifies a multiplier for the slider velocity. The reference velocity at slider multiplier = 1 is 100 osu!pixels per beat. A slider multiplier of 2 would yield a velocity of 200 osu!pixels per beat. The default slider multiplier is 1.4 when the property is omitted.
```
TPGPL: specifies a multiplier for the slider velocity -> specifies the multiplier of the slider velocity
delete reference

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
 
-`SliderTickRate:1`
+SliderTickRate (Decimal) is the number of ticks per beat. It defaults to 1 tick per beat.
```
TPGPL: It defaults to 1 tick per beat -> The default value is 1 tick per beat.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: Abusing `code` makes the document harder to read in my humble opinion.

When the article is complete, we could review its global shape for that kind of things. Right now I'm staying consistent with the other sections.

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -133,29 +133,76 @@ BeatmapSetID (Integer) is the ID of the beatmap set.
...
+```
+
+*HPDrainRate* specifies how fast the health bar decreases. At the time of this writing, the near future definition of this property remains unsettled.
+
+### Circle Size
+
+*CircleSize* (CS) defines the size of the hit circles and sliders in the osu!standard mode.
 
-CircleSize (Decimal) specifies the size of hit object circles.
+The resulting circle radius in osu!pixels is defined by the formula `32 * (1 - 0.7 * (CircleSize - 5) / 5)`, alternatively written `54.4 - 4.48 * CircleSize`.
```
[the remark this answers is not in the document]
fmang: I don't like adding *the following*, it adds no meaning.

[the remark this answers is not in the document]
fmang: I'm surprised, but you're more involved than I am.

The spelling *osu!pixel* is used on other articles, including the other sections here, and https://osu.ppy.sh/help/wiki/Ranking_Criteria

Should I update the other sections here too?

wiki/osu!_File_Formats/Osu_(file_format)/en.md
MegaApplePi: the ASC says that keywords must be in code, but yeah, I guess it could wait since it is still WIP (I forgot that you're doing this in sections)

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: What to you mean? The property is called `KeyAmount`, or it's still `CircleSize`, but the value is interpreted as the *key amount* instead?

wiki/osu!_File_Formats/Osu_(file_format)/en.md
fmang: I've used *width* in the table that follows it, but I like harshness. It might not be the most accurate word, but it's easy to picture as a human being :)

wiki/osu!_File_Formats/Osu_(file_format)/en.md
```diff
@@ -142,15 +142,15 @@ OverallDifficulty:4
 ApproachRate:4.8
 ```
 
-*HPDrainRate* specifies how fast the health bar decreases. At the time of this writing, the near future definition of this property remains unsettled.
+*HPDrainRate* specifies how fast the health decreases. However, the definition of this property may be changed in the near future.
```
MegaApplePi: Shouldn't this be under a section called `### HP Drain Rate` and the abbreviation for HPDrainRate be *HP*?

PR #980, opened 2018-01-18
Installation: update the page

MegaApplePi:
and throw everything else out


MegaApplePi:
will merge `/wiki/Installation` and `/wiki/Registration` together

PR #984, opened 2018-01-21
add banner and mentorship program opening post

Okorin:
### Description

adds a newspost and banner about mentorship signups being a thing, i forgot we had newspost for this in the past so we're kind of behind schedule

### Status

done, needs review


news/2018-01-26-community-mentorship-program-spring-2018-signups-now-open.md
```diff
@@ -0,0 +1,26 @@
+---
+layout: post
+title: "Community Mentorship Program Spring 2018 Signups Now Open"
+date: 2018-01-26 12:00:00 +0000
+---
+Have you ever opened up the editor and just looked at everything it offers in confusion? Do you have no idea where to even begin beatmapping? Worry no further, the Community Mentorship Program might be the place for you!
+
+![](/wiki/shared/news/2018-01-26-community-mentorship-program-spring-2018-signups-now-open/banner.png)
+
+If you've ever opened up the editor and tried to make a beatmap you've probably noticed that everything can feel overwhelming. If this feeling rings a bell, the Community Mentorship Program might be the place for you!
```
Naxesss: Might be a good idea to change that last part since it's already in the first paragraph. Something like
"If you've ever opened up the editor and tried to make a beatmap you've probably noticed that everything can feel overwhelming. There are so many things to learn, and the most difficult part is often the beginning, especially if you're learning everything on your own."
would probably be better, so you don't read the same thing twice.

PR #951, opened 2017-12-31
Contests Page Update - WIP

Xinnoh:
### Description

Contest page is really out of date

### Status

WIP

### Changelog

- Filled in Monthly Mapping Contests for 13/14/15
- Added Aspire Stage 2/3
- Added Osu!idol 2017

### Checklist 

- [ ] Monthly Mapping Contests
- [x] Aspire
- [ ] Osu!idol
- [ ] Pending Cup
- [ ] Newspaper Cup
- [ ] Other Community 

### Related Issues 

Could use some suggestions on other things to add

PR #998, opened 2018-01-27
Update Contests

Xinnoh:
### Description

Replacing #951 

### Status

Need to find other contests not listed here

### Changelog

- Filled in Monthly Mapping Contests for 13/14/15
- Added Aspire Stage 2/3
- Added Osu!idol 2017
- Updated Pending and Newspaper Cup
- Finished

### Checklist 

- [x] Monthly Mapping Contests
- [x] Aspire
- [x] Osu!idol
- [x] Pending Cup
- [x] Newspaper Cup


Xinnoh:
Should be done. There are more new contests that have popped up since this was last updated, but I will be busy for a while so I won't have time to do it. Good enough to update existing ones for now.

PR #1010, opened 2018-01-31
History of osu!: additions to April in 2013 and add January section in 2018

MegaApplePi:
- add peppy's osu! from 2007-09 release forum post (2013)
- add January (2018)
  - ~~I wish I took a photo of the previous leaderboards to show the before and after looks :(~~


Ephemeralis:
2013 needs to be split up into multiple paragraphs, or at least two. Massive block of text atm.

PR #1011, opened 2018-01-31
Update people tables

cl8n:
Small formatting fixes and moving people to their correct groups


TPGPL:
could you add `outdated` mark to the outdated translations and change the group link in the translations? 👀

WalterToro:
[Stymlice](https://osu.ppy.sh/users/5122436) (Russian moderator) and [Tommay](https://osu.ppy.sh/users/3132818) (French moderator) should be added to the GMT list
Atrue, Pittigbassje and Sylphi should be moved into alumni.

Ephemeralis:
The BN part of this PR looks to conflict with #1016, which is also significantly more complete and up to date.

cl8n:
yep, I'll let 1016 cover the BN stuff. beat me to it /shrug

otherwise this should be ready to go after addressing the notes in OP

Ephemeralis:
spboxer is chat & mod and Mili only speaks English afaik with a passing amount of Norwegian? Not sure there.

PR #1041, opened 2018-02-13
Sitemap (I forgot that these are what they're called)

MegaApplePi:
- updates and replaces `/wiki/Index`
- sorted alphabetically by directory
- the links use full path names relative to the root directory (so clicking on them in GitHub should lead you to a directory listing on GitHub)


TPGPL:
Maybe we could add a system to auto-update this article basing on directories. Like we would use it once a month or something

sr229:
@TPGPL the most viable way is coding a GitHub bot to make edits based on repository status, this can be done really quick but our problem would be hosting

wiki/Sitemap/en.md
```diff
@@ -0,0 +1,270 @@
+# Sitemap
+
```
TPGPL: Add something like "Complete list of osu!wiki articles. Updated every month."

wiki/redirect.yaml
```diff
@@ -1028,6 +1026,9 @@
 
 "skr":                         "Shortcut_Key_Reference"
 
+"main_page": "Sitemap"
+"index": "Sitemap"
+
```
TPGPL: all_articles, main_page

MegaApplePi:
Those are spaces. It was done that way to make it a but easier for me to read in raw text.

TPGPL:
Okay

wiki/Sitemap/en.md
sr229: I don't think it would be every month, once the bot is live it would be updated every 15 minutes

MegaApplePi: sitemap is and should already be self-explanatory... a leading paragraph for this isn't needed imo

wiki/Sitemap/en.md
```diff
@@ -0,0 +1,270 @@
+# Sitemap
```
TPGPL: Add some tags? :thinking:

PR #1057, opened 2018-02-20
[ZH/TOS]Quick format fixes

axmmisaka:
Got some little format error - offline syntax checker did not catch them.


axmmisaka:
Don't know, but the original version has those.

TPGPL:
i think you should delete that whitespace, as it's not needed 🤷‍♂️

axmmisaka:
See if this is Okay

TPGPL:
the numbers were okay, the whitespace like shown below isn't; check the diff

axmmisaka:
Okay - I didn't get you in the first comment.
And yes - I will delete them in the next commit, but actually this is done on purpose, since they can be interpreted as 3 paragraphs and they will sqeeze together as in English version if there's no newline.

PR #1084, opened 2018-03-02
Beatmap Nominator Rules: Recent changes

Naxesss:
Added the italic part in "**Do not qualify any beatmap if you are not yet a full member of the Beatmap Nominators, *unless it was nominated by someone who is***." Beatmaps can now reach qualified as long as a full member of the Beatmap Nominators placed one of the nominations, rather than just the second.

First and second nominations are essentially the same thing, carrying the same weight and responsibility, so in which order the nominations are done should logically not matter, although with the old wording it did. It was also possible to evade the rule, if a full nominator had placed the first nomination, by self-popping and then renominating, which isn't necessary anymore.

*This change is in effect immediately.*


Xinnoh:
should add changelog at the bottom

PR #1117, opened 2018-03-07
Forced revert article

osuShiro:
after discussion, etc

### DIFF
--- a/wiki/Help_Center/en.md
+++ b/wiki/Help_Center/en.md
@@ -92,6 +92,34 @@ Lastly, we are aware that some providers do not accept automated emails from us.
 - `sbcglobal.net`
 - `t-online.de`
 
+### Name Changes
+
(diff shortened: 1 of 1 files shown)


wiki/Help_Center/en.md
```diff
@@ -120,6 +120,14 @@ We only grant free reverts for a user's direct previous username (the one shown
 
 We only grant this one time per account, so please consider carefully before requesting!
 
+#### My username has been reverted back!
+
+**We only force username reverts for inappropriate usernames.**
+
+If you find your recent namechange has been undone, it likely is because the username you changed to was abusive or inappropriate. We will not perform the change again, even if we requested, but if you purchased it on the store, we'll offer a free change for you.
```
TPGPL: it likely is because -> it is likely that?

PR #1176, opened 2018-03-26
SOFT3: Ro16 Update

WalterToro:
### Description

Round of 16 update and small fixes.

### Status

Finished


wiki/Tournaments/SOFT/3/en.md
```diff
@@ -85,71 +90,116 @@ Springtime Osu!mania Free-for-all Tournament 3
 
 ## Mappools
```
TPGPL: Mappool?

PR #1090, opened 2018-03-03
Centre not center

Joehuu:
### Description

Changed "center" to "centre" according to the ASC.

### Status

- [X] Finished

PR #1185, opened 2018-03-29
Rename help centre page

Joehuu:
### Description

Supersedes #1090 ~with one commit.~ And fixed formatting according to the ASC.

### Status

- [X] Finished


wiki/redirect.yaml
```diff
@@ -465,8 +465,9 @@
 "tips_and_tricks_on_skill_improvement":                       "Guides/Tips_and_Tricks_on_Skill_Improvement"
 "videos_from_youTube":                                        "Guides/Videos_From_YouTube"
 
-"help_center":                 "Help_Center"
-"help":                        "Help_Center"
+"help_center":                 "Help_Centre"
+"help_centre":                 "Help_Centre"
```
MegaApplePi: wait, did I ask you to put this one here?

Joehuu: yes
> add the other varient too (so other links and American English people can go to the right location)

MegaApplePi: I'm sure that was me talking about keeping `help_center` (not `help_centre`). Anyways, remove it, case sensitive redirects aren't needed since osu-web handles it better.

PR #1218, opened 2018-04-06
SOFT3: Round of 16 Update

WalterToro:
### Description

Round of 16 Update, also applied Shoegazer comments and changes.

### Status

Finished


wiki/Tournaments/SOFT/3/en.md
```diff
@@ -174,9 +195,35 @@ Springtime Osu!mania Free-for-all Tournament 3
 
 ## Match Results
```
TPGPL: for the stages:  "through this link" -> "here"

PR #1220, opened 2018-04-07
redirect: add tournament redirects and other stuff

MegaApplePi:
add some missing tournament redirects and other stuff


wiki/redirect.yaml
```diff
@@ -772,6 +767,8 @@
 "achievements/beatmap_packs_0916": "Medals/Beatmap_Packs_0916"
 "beatmap_packs_0916":              "Medals/Beatmap_Packs_0916"
 
+"mods": "Mod"
```
TPGPL: prob shouldnt give a redirect for article which doesnt exiat yet

PR #1221, opened 2018-04-07
Add the disambig pages

MegaApplePi:
There were only three on the MediaWiki :shrug:

## Description

Don't hardcode ambiguous redirect names to a specific article.

## Status

Open for discussion if there is any

### DIFF
--- a/wiki/Disambiguation/EZ/en.md
+++ b/wiki/Disambiguation/EZ/en.md
@@ -7,3 +7,5 @@
 -   The difficulty of a beatmap:
     -   ![](/wiki/shared/diff/easy-m.png) EZ, for osu!mania.
     -   ![](/wiki/shared/diff/easy-s.png) [Easy](/wiki/Difficulties/Easy), for osu!standard.
+    -   ![](/wiki/shared/diff/easy-t.png) [Kantan](/wiki/Difficulties/Kantan), for osu!taiko.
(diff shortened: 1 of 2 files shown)


wiki/Disambiguation/EZ/en.md
```diff
@@ -0,0 +1,9 @@
+# EZ
+
+**EZ** may also refer to:
+
+-   The [Easy game modifier](/wiki/Game_Modifiers/#easy).
+-   The version name of a beatmap.
+-   The difficulty of a beatmap:
+    -   ![](/wiki/shared/diff/easy-m.png) EZ, for osu!mania.
```
TPGPL: add the rest of the game modes too

wiki/Disambiguation/Mod/en.md
```diff
@@ -0,0 +1,7 @@
+# Mod
+
+**Mod** may also refer to:
+
+-   A moderator, such as a [QAT](/wiki/QAT) or [GMT](/wiki/GMT) member.
+-   A person who looks over a beatmap gives feedback.
```
TPGPL: and gives feedback

wiki/Disambiguation/EZ/en.md
MegaApplePi: I don't think when someone writes EZ, they think taiko or catch, but alright.

PR #1235, opened 2018-04-10
TWC2018 Ro16

juankristal:
Added mappool for Ro16.
Added schedule.
Added results of Group Stage.


HappyStick:
Here's the RO16 bracket

bracket

PR #1303, opened 2018-05-01
camellia themed spotlights

Shadowthegod:
### Description

The article for Camellia Themed Spotlights

### Status

- [ ] WIP
- [ ] Pending Review
- [ ] Finished


news/2018-05-05-camellia-themed-spotlights-2018.md
```diff
@@ -0,0 +1,52 @@
+---
+layout: post
+title: "Camellia Themed Spotlights 2018"
+date: 2018-05-05 14:00:00 +0000
+---
+
+Greetings everyone! This time Camellia makes his appearance in Themed Spotlights! Come and check out our selection of excelent beatmaps!
```
TPGPL: excellent

news/2018-05-05-camellia-themed-spotlights-2018.md
```diff
@@ -0,0 +1,52 @@
...
+
+![](/wiki/shared/news/2018-05-05-camellia-themed-spotlights-2018/spotlight-header.jpg)
+
+We just released the Camellia Themed Charts and can be found [here](https://osu.ppy.sh/p/chart?ch=CACH17). Their respective packs are available at the [Beatmap Pack Listing](https://osu.ppy.sh/p/packlist?t=r) for easier access.
```
TPGPL: We have just released the Camellia Themed Charts which can be found

news/2018-05-05-camellia-themed-spotlights-2018.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Past Winners
+
+Congratulations to the winners of the Camellia Themed Spotlights. They receive two month of supporter status for their efforts!
```
TPGPL: Past Winners -> Winners?

They will receive

TPGPL: Camellia Themed Spotlights -> Anime Themed Spotlights

Ephemeralis:
after some talk internally, we're not sure we can push any sort of themed spotlight out as semi-official content supported via a newspost, due to the fact that we own precisely 0 of the tracks involved in the showcase and don't want to push the impression across that we do

Xinnoh:
So spotlights like these can still be done if not accompanied by a newspost?
If so, why not give a mentiont to these charts into the monthly spotlights, something like
"By the way there's also the Camellia spotlights if you play mania and have nothing to do"
It's not really as official and still gives a mention to people that are interested in that content

Ephemeralis:
it's essentially still official promotion even if referred to that way and i have absolutely no idea where we stand on the legality of that

Shadowthegod:
I find it weird that while the Monthly can be announced the Themed one is not. Both are essentially the same thing with the only difference that the second builds around a certain theme. Technically due to the copyrights, we should not announce the Monthly Spotlights either.

PR #1322, opened 2018-05-08
TWC 2018 Finals Update

juankristal:
ignore the fact this is on a cwc branch

Updated results and schedule for Grand Finals of the TWC2018


Joehuu:
Not on the PR, but change:
`[![][flag_AR] juankristal](https://osu.ppy.sh/users/443656)` -> `![][flag_AR] [juankristal](https://osu.ppy.sh/users/443656)` for every user. This is a new rule on the updated asc.

PR #1336, opened 2018-05-08
[ES] About / Add space between osu! and wiki

KaiMinCL:
- [ ] WIP
- [ ] Pending Review
- [x] Finished


Joehuu:
Can you close this? You missed an instance on that file you changed. Don't worry I'll change all the instances from all the wiki files with this instance.

Issue #1338, opened 2018-05-09
Correct pages from updated ASC


Joehuu:
~Opened this, so people don't waste time. I'll be doing this.~

~Doing the originally planned idea. A commit for every minor change or fix. **Edit: Might be per PR.**
(i.e. See the first referenced PR that I closed. It'll have that format.)~

~Tagging outdated for articles will come last.~

Joehuu:
Can't actually do this until the ASC gets merged and officially applied. Also I'm kind of busy with other things.

PR #1349, opened 2018-05-13
Accuracy: add mode icons and tag pl and pt-br as outdated

Joehuu:
Edit: pl is not updated under the osu!taiko heading while pt-br is really outdated everywhere.

### Status

- [X] Finished

### Additional Notes 

Part of #1338.

### DIFF
--- a/wiki/Accuracy/pl.md
+++ b/wiki/Accuracy/pl.md
@@ -1,10 +1,13 @@
+---
+outdated: true
+---
 # Precyzja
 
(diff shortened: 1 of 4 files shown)


wiki/Accuracy/pl.md
```diff
@@ -1,10 +1,13 @@
+---
+outdated: true
```
TPGPL: remove the tag, the article is up-to-date, but it uses different format

PR #1417, opened 2018-05-31
Remove huge piece of example from chat commands table

FreezyLemon:
### Description

I removed a big part of the `/np` command example in the "chat commands" table. Apparently, this used to look different in the old wiki, but in its current state, it's basically unreadable. If someone wants this info back, it should probably be moved to its own paragraph below the table. Not sure it's needed though.

### Status

- [ ] WIP
- [ ] Pending Review
- [x] Finished


Joehuu:
It is not that much different actually. Markdown has its limitations on a lot of stuff(in this case, new lines).

oldwiki

PR #1425, opened 2018-06-02
CWC 2018 Groups Update

juankristal:
Added schedule and map pool.


wiki/Tournaments/CWC/2018/en.md
```diff
@@ -96,7 +96,93 @@ The osu!catch World Cup 2018 is run by various community members by distributing
 
 ## Mappools
 
-_Mappools will be revealed one week in advance of each stage_
+- NoMod
```
WalterToro: You should note that this is the Group stage mappool

PR #1433, opened 2018-06-05
Remove an unnecessary bracket in Privacy Policy

TheWildTree:


Joehuu:
Those are parentheses. Anyway someone please add the block tag, so @peppy can rightfully merge it.

Joehuu:
Also, someone needs to change the `Last Updated` date. Just letting them know before merging this.

peppy:
I don't think this warrants a change to the date. That's for real changes.

PR #1461, opened 2018-06-15
Remove "Article of the month" section on Welcome page

### LINKED ISSUE #1448: Article of the month needs to be updated every month
The last updated one was on April. I suggest we remove it completely if it doesn't get attention.

Off topic, but when the wiki has its main page. It should be moved to a page other than `Welcome`. *(if it ever stays)*

Joehuu:
Closes #1448.

I came to a conclusion to myself that this isn't working. First off, the English page isn't getting monthly updates. That's bad enough, and then these translations. They have to updated monthly which is not happening.

*Note: This PR is purely at removing the AotM, nothing more. After this gets merge, I will make another one regarding to the ASC.*


Ephemeralis:
it is a little pointless to have things like this if they aren't done properly or the maintainers for them aren't interested in keeping them going, so i guess we could let them go

PR #1462, opened 2018-06-15
[THOM CUP] Round of 16 Update

WalterToro:
### Description

Multiple updates for THOM CUP, Round of 16 and Group Stage update, staff update and rules update

### Status

Finished


WalterToro:
Whoops, should be fixed now, blame Atom (?

PR #1502, opened 2018-07-02
Update .db file format

DorJoogl:
Update article about the .db file format by adding "loved" and "unsubmitted" rank status identifiers. Fixed link at bottom of the article targeting the .osr file format article.

Found out about the missing identifiers while writing an API that wraps around the database file formats in the osu directory.

Identifier for /b/545076
See above, identifier for [/b/545076](https://osu.ppy.sh/beatmapsets/545076/) (loved).

Identifier for local beatmap
See above, identifier for a local, unsubmitted beatmap.

There used to be a link to the .osu file format as well but it wasn't evident how it correlated with the section it was mentioned in. Can add again if desired. Otherwise ready for review.

### DIFF
--- a/wiki/osu!_File_Formats/Db_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Db_(file_format)/en.md
@@ -68,7 +68,7 @@ Some data types specific to osu!.db are defined below.
 | String | Audio file name |
 | String | MD5 hash of the beatmap |
 | String | Name of the .osu file corresponding to this beatmap |
-| Byte | Ranked status (4 = ranked, 5 = approved, 2 = pending/graveyard) |
+| Byte | Ranked status (4 = ranked, 5 = approved, 2 = pending/graveyard, 7 = loved, 1 = unsubmitted) |
 | Short | Number of hitcircles |
 | Short | Number of sliders (note: this will be present in every mode) |
 | Short | Number of spinners (note: this will be present in every mode) |
--- a/wiki/osu!_File_Formats/Db_(file_format)/en.md
+++ b/wiki/osu!_File_Formats/Db_(file_format)/en.md
@@ -179,6 +179,4 @@ This database contains the scores achieved locally.
 | Int | Should always be 0xffffffff (-1). |
 | Long | Online Score ID |
 
-Apart from the online score ID, the individual score format is the same as the replay format. [Osr (file format)][Osr Link]. This explains the empty string and -1 int.
-[Osr Link]: /wiki/osu!_File_Formats/Osr_(file_format)
-[Osu Link]: /wiki/osu!_File_Formats/Osu_(file_format)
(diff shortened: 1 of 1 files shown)


DorJoogl:
After digging around some more, I found out that the status 6 is used to mark qualified maps. See screenshot below for an example on [/b/770246](https://osu.ppy.sh/beatmapsets/770246).

Qualified map status

There's also a status 0 that seems to be an undetermined/unknown status. When importing mapsets, the status for every map is set to 0. When a map is focused in the song selection, the game client seems to check the map status online and update the field in the cache file. See below an example of an unsubmitted map from an imported mapset before (status 0) and after (status 1) being clicked on in the song selection.

Status 0 before

Status 0 after

Now this also seems to imply that there must be a status 3, deprecated or not. In that case all statuses from 0 to 7 would be covered. I haven't found a map with a status 3 yet, let alone any mean of reproducing it. I'd add the missing statuses as soon as such a status has been confirmed.

DorJoogl:
That's the issue. I've found no reliable means of verifying if it's correct apart from running my wrapper over my locally stored beatmap cache. The closest official source I found is [this file from the osu repo](https://github.com/ppy/osu/blob/632befb2101d0e886901c1b407189eabca1540cc/osu.Game/Beatmaps/BeatmapSetOnlineStatus.cs) but it seems to only internally represent the statuses - not how they're saved in the cache. The code also distinguishes graveyarded and pending maps which the status identifier in the beatmap cache doesn't. 

All the added statuses are not verified. They're just my best guesses from reverse engineering the .db file format. Any help in verifying is greatly appreciated.

wiki/osu!_File_Formats/Db_(file_format)/en.md
```diff
@@ -179,6 +179,4 @@ This database contains the scores achieved locally.
 | Int | Should always be 0xffffffff (-1). |
 | Long | Online Score ID |
 
-Apart from the online score ID, the individual score format is the same as the replay format. [Osr (file format)][Osr Link]. This explains the empty string and -1 int.
-[Osr Link]: /wiki/osu!_File_Formats/Osr_(file_format)
-[Osu Link]: /wiki/osu!_File_Formats/Osu_(file_format)
+Apart from the online score ID, the individual score format is the same as the replay format. This explains the empty string and -1 int. See [this](/wiki/osu!_File_Formats/Osr_(file_format)).
```
TPGPL: `See [this](/wiki/osu!_File_Formats/Osr_(file_format))` -> `For more information, see [.osr (file format)](/wiki/osu!_File_Formats/Osr_(file_format)).`

cl8n:
2 can be pending, wip or graveyard. 3 is unused and the rest is correct

DorJoogl:
Sorry, was out and about. But I fixed it now.

PR #1527, opened 2018-07-16
Add Project Loved: Week of July 15th

cl8n:


cl8n:
should be ready to go

PR #1528, opened 2018-07-16
[LMS Shift] End of Tournament & Conclusion

fajar13k:
### Description
- End of Taiko Last Man Standing: Shift Cup
- Podium and Grand Finals results are added


wiki/Tournaments/LMS/Shift/1/en.md
```diff
@@ -28,10 +28,8 @@ The **Taiko LMS: Shift Cup** is a 3 on 3, 1v1 knock-out style osu!taiko tourname
 | Placing | Prize(s) |
 | --- | :-- |
 | ![Gold Crown](/wiki/shared/GCrown.png "1st place") | Special profile badge |
```
WalterToro: Looking that the 2nd and 3rd places will get nothing, i think it would be better to write a sentence or two saying that the winner of the tournament will get the a profile badge (you can also put that badge here), better than doing a table and filling the prizes with "none"

PR #1573, opened 2018-08-05
Removes the misc link and mark pt/ru as outdated

Okorin:
the link to /wiki/Storyboard_Scripting/Miscellaneous is taken out in this

pt/ru are 2 liners and thus marked as outdated
the link to misc is ded in other languages or contains duplicate info according to #810 so linking to it is just confusing

---


Okorin:
redoing

PR #1588, opened 2018-08-14
[ZH] Legal/Terms

axmmisaka:
added original text, and finished amending. It seems that the main difference is between 'osu.ppy.sh' and 'osu', **1. foobar** and ## foobar.
There's only minor changes, so I didn't push to osu-translate-zh and ask for amends.


wiki/Legal/Terms/zh.md
```diff
@@ -0,0 +1,111 @@
+# osu! 服务条款
+
+## 声明
+**这是一份 osu! 版权政策的非正式、非官方授权的中文翻译。它不是由 osu! 官方或 ppy（Dean Herbert）发表，不具有任何法律效力。只有英文原版文本具有此等效力。**
```
cl8n: remove the disclaimer

wiki/Legal/Terms/zh.md
```diff
@@ -0,0 +1,111 @@
+# osu! 服务条款
```
cl8n: add this to the beginning
```
---
legal: true
---
```

PR #1661, opened 2018-09-03
Add a link to the Storyboard Scripting Shorthand in the Commands article

yugecin:
The shorthand page was not linked anywhere (except for the sitemap), there was only a vague reference in the fade command description (which is how I found out about this page in the first place).

---


TPGPL:
mark the japanese translation as outdated by adding this at the beginning of the article:
```
---
outdated: true
---
```

PR #1675, opened 2018-09-10
Storyboard fix and indexes

EmilySunpy:
Removed misinformation as floating point numbers are allowed in move commands.
Fixed order of which origin is mentioned as it did not match the indexes before.
Added numbers for indexes for layers and origin in lists.
(Useful for those who make storyboard creation tools)

---


TPGPL:
could you give a source for those informations, please?

wiki/Storyboard_Scripting/Commands/en.md
```diff
@@ -98,7 +98,7 @@ See the [shorthand](/wiki/Storyboard_Scripting/Shorthand) section for an explana
 
 | Affects | Value definition | Default value |
 | ------- | ---------------- | ------------- |
-| The location of the object in the play area. | An (x,y) position, as specified above. Decimals are not allowed. | An (x,y) position, as specified above. Decimals are not allowed. |
+| The location of the object in the play area. | An (x,y) position, as specified above. Decimals are accepted. | An (x,y) position, as specified above. Decimals are accepted. |
```
cl8n: for default value, it'd be the initial sprite/animation position

wiki/Storyboard_Scripting/Cheat_Sheet/en.md
```diff
@@ -41,24 +41,25 @@ Z-order (back to front) is determined by the order the files appear in the .osu
...
--   TopLeft
--   TopCentre
--   TopRight
--   CentreLeft
--   Centre
--   CentreRight
--   BottomLeft
--   BottomCentre
--   BottomRight
+-   0 - TopLeft
```
cl8n: I would just add a note here that 6 can also be TopLeft, but you shouldn't use it. Custom is not used for storyboarding

cl8n:
@TPGPL it's correct, a lot of storyboarders use numbers instead of full names now because it saves filesize when you have tons of effects

yugecin:
can confirm they're correct

you can see the values in the osu source, see https://github.com/ppy/osu/blob/5e5a74293e2710690328dda0086ec0fe9617ac24/osu.Game/Beatmaps/Formats/LegacyDecoder.cs#L150

wiki/Storyboard_Scripting/Commands/en.md
yugecin: see also #1660 I guess

wiki/Storyboard_Scripting/Cheat_Sheet/en.md
cl8n: doesn't have to be confusing, just note that it does the same thing as topleft. if a storyboarder did use it for some reason, other people might want to know what 6/Custom means

yugecin: okay, readded them

Wieku:
Also can confirm it's correct, without looking at lazer's code my storyboard parser in Go wouldn't work properly.

wiki/Storyboard_Scripting/Cheat_Sheet/en.md
```diff
@@ -41,24 +41,25 @@ Z-order (back to front) is determined by the order the files appear in the .osu
...
+-   1 - Centre
+-   2 - CentreLeft
+-   3 - TopRight
+-   4 - BottomCentre
+-   5 - TopCentre
+-   6 - Custom (same effect as TopLeft but shouldn't be used)
```
TPGPL: comma before but
shouldn't -> should not

EmilySunpy: This was fixed over a month ago... maybe you should stop blocking? >.>

PR #1677, opened 2018-09-10
Add an URL fraction to a link pointing to the Layers section in the Storyboard Scripting General Rules article

yugecin:
The link was just pointing to the page, which might confuse users. By adding the fragment to the URL, the page will automatically jump to the correct spot.

---


TPGPL:
i don't think it's necessary to add this to be honest

PR #1687, opened 2018-09-13
Fixing dead link

RockRoller01:
adressing this https://osu.ppy.sh/forum/t/802897/start=0


cl8n:
if it's only referenced from within the wiki, just change the links in the articles instead of adding redirects

RockRoller01:
I don't know if it is used anywhere else, thats why I added a redirect isntead of doing that

TPGPL:
superseded by #1689

PR #1724, opened 2018-10-02
TSC Grand Finals & Conclusion update

fajar13k:
- Add Finals match results
- Add Grand Finals match schedules (time not included)
- Add Grand Finals mappools
- Add Grand Finals match results
- Add Podium

---


fajar13k:
Label as `blocked` please.
Will update alongside the conclusion after 7th of October.

PR #1749, opened 2018-10-14
Ranking Criteria Updates

pishifat:
add exception to song content rule discussed here: https://osu.ppy.sh/community/forums/topics/811164
adjust wording to .osu editing rule because someone said it was misleading
revert a change from #1747 
change to set contribution rule discussed here: https://osu.ppy.sh/community/forums/topics/804448
change to object snapping rule discussed here: https://osu.ppy.sh/community/forums/topics/798129
add exception to rule about mapping the last 20% of an mp3 (was part of earlier rc, but deleted when it shouldn't have been)

---


Okorin:
just 3 hours of debate about a few words later we determined that this way of saying it does in fact not alter the meaning, it just needed a whole lot of brains to figure out it doesnt (thx naxess)

PR #1753, opened 2018-10-15
taiko RC fix aimod

Raidencio:
removed that line since AImod now doesn't screw with taiko diffs

if the first commit gives conflict i guess i'll die


Okorin:
change is generally fine, can go live when stable is actually changed

PR #1758, opened 2018-10-18
Fix remaining line endings

Joehuu:


sr229:
What line endings are precisely we're using here for this file?

Joehuu:
Idk if the problem is even line endings. Some weird thing with these files is that when you save it, the diff does this.

`Ctrl+S` any of these files and it'll show the diff replacing all lines. (I use VS Code.)

sr229:
Doesn't answer my question however, Are we using CRLF or LF initially?

Joehuu:
CRLF, since most files have that.

sr229:
I inquired @Repflez regarding normalizing line endings, I think we should enforce CRLF as our defualt line endings from now on since it was our line endings since the start.

PR #1764, opened 2018-10-19
Add index pages in folders with no articles

### LINKED ISSUE #1658: Make index pages for every folder
Enforcing https://osu.ppy.sh/help/wiki/Article_Styling_Criteria#index-articles. Even if they just link to articles below it, better than 404

related to discussion in https://github.com/ppy/osu-web/pull/3595

cl8n:
closes #1658 

turns out that Legal and Tournaments/\*\* didn't have index pages. some minor re-organizing done to eliminate useless index pages too

---


Joehuu:
We need to remove the redirects that go to the tournament index, or it will conflict with the new index articles.
E.g.
~`"tournaments/owc":             "Tournaments/#-osu!-world-cup"`~

sr229:
Last change is since Oct 23, updates?

cl8n:
the only further changes would be adding new tournaments since October 23. last this was active, I don't believe that @Joehuu, @TPGPL and I really came to a conclusion about whether this should be merged, or if we should work on an osu-web solution instead

PR #1752, opened 2018-10-15
Categorize Issues

sr229:
This will focus on having a more contributor-friendly issue making
Issues will now have a dialogue that will allow the issue maker to select to further narrow down their issue.

image

PR #1766, opened 2018-10-19
Add issue template

cl8n:
to supersede #1752 

will hopefully direct people to the right place if they have issues with something else in osu!

---


Joehuu:
I'm all for this. Don't want to press another button, and news shouldn't have any issues to begin with.

PR #1795, opened 2018-10-29
Help Centre Update

### LINKED ISSUE #1793: Dead link in master/Help_Centre
Hello, 
While checking the translation of [Help_Centre](https://github.com/ppy/osu-wiki/blob/master/wiki/Help_Centre/en.md), we found a dead link in line 609, as it says:
> On most Wacom tablets, this is done via the `Mapping` tab of the `Wacom Tablet Properties` application. An example may be viewed [here](https://w.ppy.sh/d/dc/Tabarea.png).     

The link of the picture is dead now. Hope you can fix it!

mikser572:
Added a new image to the folder and fixed the link in the article. Ready for reviews.

- Closes #1793 

---


TPGPL:
this still won't work, should be uploaded to puush instead i think

mikser572:
Changed it back

PR #1818, opened 2018-11-04
Update beatmap status list in the Interface article

mikser572:
Added "loved" to searching by ranking status. Marked all the translations as outdated.

---


TPGPL:
better add it to every translation instead of marking as outdated, doesnt require language knowledge i think

mikser572:
I mean, I can do it, but they're gonna still be outdated. Images don't really work and translations contain almost half of the lines of the original article.

TPGPL:
then let's leave them outdated 🤷‍♂️

Joehuu:
Can do it here too. https://github.com/ppy/osu-wiki/pull/1818/files#diff-6d0672fc5036826c3f110250d3cb17c4R128

PR #1825, opened 2018-11-06
Update osu!taiko BN listing

Nardoxyribonucleic:


Raidencio:
try to have commit names more descriptive (update bn listing for example) for next time!

Xinnoh:
uh why wasn't the whole bng updated

PR #1827, opened 2018-11-07
Update the rest of the beatmap nomintors

Xinnoh:
need to add languages, will start spam pm


cl8n:
should update or outdate the indonesian article as well

Xinnoh:
need to redo this since i did some dumb stuff on my end

PR #1840, opened 2018-11-16
[FR] Legal/Terms, osu!supporter

Krystool:
I have translated the French Page of osu!supporter to be up to date !


TPGPL:
reopen after you fully translate Legal/Terms

PR #1842, opened 2018-11-17
Don't recommend using cutting edge

smoogipoo:
CE has been in flux as of the last couple of tournaments, as more and more stuff has been moved to lazer and designs have been adjusted accordingly.

Telling people to use CE will result in their client displaying a design that doesn't fit with the latest provided [template](https://osu.ppy.sh/help/wiki/osu!tourney/Skinning/).

Stable has been stable for a long period for use in tournaments, thus far.


peppy:
For what it’s worth I’m going to roll back CE to how it was today. Not sure this needs to change after doing so.

peppy:
Has since been reverted, should be good now.

PR #1859, opened 2018-11-26
Updated OWC 2018 Bracket Image

VINXIS:
Updated the bracket image since no one else had. Taken from [reddit](https://www.reddit.com/r/osugame/comments/a0dyxl/i_updated_the_2018_tourney_poster_updated_all_the/)

---


WalterToro:
Actually denied since is not the right format and the PR with the Semifinals update and the updated bracket should be today

PR #1867, opened 2018-12-01
Staff Log: November and December 2018 update

MegaApplePi:


wiki/People/Staff_Log/2018/en.md
```diff
@@ -317,3 +317,21 @@ All users listed below were added on 2018-07-27.
 - [Arzenvald](/users/3027421) (2018-10-22)
 - [Yuii-](/users/2935923) (2018-10-23)
 - [Yukiyo](/users/4541873) (2018-10-28)
+
+## November
+
+-   Moved [Aloda](/users/1190127) to the osu! Alumni (2018-11-07)
```
Joehuu: I think the format for lists is one space after the dash.
```suggestion
- Moved [Aloda](/users/1190127) to the osu! Alumni (2018-11-07)
```

MegaApplePi: Why was this changed in #1821?

Joehuu: I changed the spacing because the format was widespread. Also, I use a extension that checks for format and it recommends this.

MegaApplePi: The spacing from before was how a parser I used gave it to me. I did prefer the extra spacing since it helped me read the lists in raw text better while editing (especially if the list was long and/or had indentation).

Using an extension that recommends using one format over another, I think, isn't a good way to justify this change. Since I could have another extension that recommends spacing it out the way I had it.

wiki/People/Staff_Log/2018/en.md
```diff
@@ -317,3 +317,21 @@ All users listed below were added on 2018-07-27.
...
+## November
+
+- Moved [Aloda](/users/1190127) to the osu! Alumni (2018-11-07)
+- Moved [Pereira006](/users/537344) to the Global Moderation Team (2018-11-09)
+- Added [Gottagof4st](/users/6122935) to the Quality Assurance Team (2018-11-10)
+- Added [Kuron-kun](/users/2697284) to the Quality Assurance Team (2018-11-10)
+- Added [Noffy](/users/1541323) to the Quality Assurance Team (2018-11-10)
+- Added [Yuii-](/users/2935923) to the Quality Assurance Team (2018-11-10)
+- Added [Regraz](/users/3076909) to the Quality Assurance Team (2018-11-11)
+- Moved [\-kevincela-](/users/266596) to the osu! Alumni (2018-11-17)
```
Joehuu: This escape is not needed. Other than that I approve.
```suggestion
- Moved [-kevincela-](/users/266596) to the osu! Alumni (2018-11-17)
```

PR #1871, opened 2018-12-02
[KO] OWC 2018

chocoliti:
translate korean owc 2018 article


TPGPL:
reopen after you update the article

PR #1882, opened 2018-12-06
[KO] Accuracy

chocoliti:
korean translate of accuracy article


TPGPL:
reopen after you apply changes

peppy:
I'm not sure we should be closing contributions like this. Translating all but images s still better than nothing, no?

PR #1888, opened 2018-12-10
Fixed broken link in the Rules article

mikser572:
The "Beatmap Submission System" is no longer in the Glossary so I changed the link to "/wiki/Submission" in every language.

---


Joehuu:
It exists. https://osu.ppy.sh/help/wiki/Glossary/#bss

There is a parsing issue when it redirects from `/wiki/Beatmap_Submission_System` to `/Glossary/#bss`. See ppy/osu-web#3988.

cl8n:
@Joehuu , `/#` should instead be `#` following https://github.com/ppy/osu-wiki/pull/1926 right?

PR #1891, opened 2018-12-11
OMMFF - RO32 osu! Division update 

fajar13k:
- Add TimmyAkmed as Taiko commentator
- Add Tigzick as Referee
- Move Head Referee below Management

---


fajar13k:
The French translator said it suppose to be like that btw, So i just follow their instruction.
It's not like Round of 16, Round of 8. Maybe it's like Battle to 16, Battle to 8.

PR #1900, opened 2018-12-16
oPHF2018 End of Tournament Update

SurfChu85:
- finalized Prizes section
- added Tournament Winners section
- added Match Week 5 results
- Fixed Links


fajar13k:
Can you also add oPHF2018 to the tournaments' index page?

PR #1902, opened 2018-12-16
Update Osr Format description

SlowLogicBoy:
last Long Unknown = Online Score Id according to [LegacyScoreParser](https://github.com/ppy/osu/blob/master/osu.Game/Scoring/Legacy/LegacyScoreParser.cs#L83)


cl8n:
yeah, it's good as is 👍

PR #1914, opened 2018-12-22
Remove hyphen clarification on osu!taiko skinning files

### LINKED ISSUE #1892: Old info on Taiko skinning page
The hyphen doesn't seem to be anywhere, so it doesn't need to be left out
image

Joehuu:
Resolves #1892.

---

### DIFF
--- a/wiki/Skinning/osu!taiko/en.md
+++ b/wiki/Skinning/osu!taiko/en.md
@@ -19,7 +19,7 @@ You can override the osu!taiko playfield parts by creating a folder called `taik
 
 Notes:
 
-- Animation name: `pippidonclear{n}.png` (without the hyphen (`-`)).
+- Animation name: `pippidonclear{n}.png`.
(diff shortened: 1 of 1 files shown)


TPGPL:
actually some of the animations with the hyphen dividing frame name and skin item name so i don't know if this change is needed

Joehuu:
Reworded it, so it seems that the hyphen is used commonly to separate frame numbers.

Before, it was saying to omit something that is not there.

Okorin:
dont think you need to clarify that it's not needed if the file is given in the format `filename{n}.png` to begin with - in fact clarifying makes this more confusing to read https://osu.ppy.sh/help/wiki/Ranking_Criteria/Skin_Set_List#osu!taiko doesnt clarify any of that either

Joehuu:
Awaiting another opinion. Reverted back to removing them.

PR #1926, opened 2018-12-25
Remove end slashes from redirect links

Joehuu:


sr229:
On a side note, would this break in osu-web side? CC from anyone in osu-web if this would break anything or fix?

Joehuu:
Don't think it'll mess up with anything. The other links without anchor links do not have end slashes and it works.

cl8n:
it won't, all is good 👍

PR #1896, opened 2018-12-12
Ranking Criteria clarification

pishifat:
adjusting ranking criteria to clarify meaning and match recent client update: https://osu.ppy.sh/community/forums/topics/815518

---

PR #1928, opened 2018-12-27
Ranking Criteria: Unsnapped object clarification

Naxesss:
Seems pishifat forgot to include this from https://osu.ppy.sh/community/forums/posts/6905070, as right now the rule is less clear than it was before that was amended.


Naxesss:
duplicate of https://github.com/ppy/osu-wiki/pull/1896

PR #1930, opened 2018-12-28
BN Rules: Clarify probation hybrid interaction

Naxesss:
Since it's only one nominator looking at a their respective mode difficulties of a hybrid set instead of 2, it would be very dangerous for someone new or facing issues to be allowed to, alone, be held responsible for there not being any issues. As such each mode needs a full member involved in the process. Not very clear from the rule at the moment, so we're adding this onto it.


Xinnoh:
uhh the wording on this is a bit too strict
uncommon situation but lets say I'm a catch/mania full bn, and I nominate a catch/mania set. a probation bn from either mode should be allowed to qualify, but not under this wording.
something like "at least one full bn for each mode" or something similar would be more appropriate

PR #1937, opened 2019-01-01
Fix italic syntax

TsukiPoi:
_i_th -> *i*th


wiki/Performance_Points/en.md
```diff
@@ -87,7 +87,7 @@ This means that only your highest pp score will award the full amount of pp for
 
 To provide an example of this with a mathematical formula:
 
-Let _PP_ represent each score's individual pp value. _PP\[i\]_ denotes the _i_th score's pp value, sorted descending, where _i_ ranges from 1 to _n_, and _n_ is defined as the amount of scores you have.
+Let *PP* represent each score's individual pp value. *PP\[i\]* denotes the *i*th score's pp value, sorted descending, where *i* ranges from 1 to *n*, and *n* is defined as the amount of scores you have.
```
sr229: What's the point of this? They're equally valid Markdown.

TsukiPoi: > What's the point of this? They're equally valid Markdown.
image

sr229: This is the fault of the Markdown engine, and not the syntax itself, I can confirm from GitHub's Markdown engine the following syntax renders fine.

TsukiPoi: Underscore parsing requires space before and after ```i```. For many Markdown, not just osu-web.
e.g. _i_and*i*   ```_i_and*i*```
I should not change all underscore to asterisk. The only problem is on ```i```.

sr229: @TsukiPoi that's not true, it works with either for GitHub
screenshot from 2019-01-05 01-47-55
screenshot from 2019-01-05 01-48-03

Joehuu: You have to use asterisks with italicizing part of a term. Italicizing a whole term works fine.
(e.g. `Markdownis_hard_` -> Markdownis_hard_ | `Markdownis*hard*` -> Markdownis*hard*)

https://www.markdownguide.org/basic-syntax/#italic

Issue #328, opened 2017-06-06
Flag reference links template


MegaApplePi:
Update: I have created a [GitHub page](https://megaapplepi.github.io/flag-wiki-osu/) to do the parsing for you.

---

Put these references at the bottom of your article, and remove the ones you don't use. If a user's flag doesn't exist, follow these rules for replacing the flag based on their country code:

- `BV` uses `flag_NO`
- `FX` uses `flag_FR`
- `UM` uses `flag_US`
- everything else uses `flag___`

```markdown
[flag___]: /wiki/shared/flag/__.gif
[flag_AD]: /wiki/shared/flag/AD.gif "Andorra"
[flag_AE]: /wiki/shared/flag/AE.gif "United Arab Emirates"
[flag_AF]: /wiki/shared/flag/AF.gif "Afghanistan"
[flag_AG]: /wiki/shared/flag/AG.gif "Antigua and Barbuda"
[flag_AI]: /wiki/shared/flag/AI.gif "Anguilla"
[flag_AL]: /wiki/shared/flag/AL.gif "Albania"
[flag_AM]: /wiki/shared/flag/AM.gif "Armenia"
[flag_AN]: /wiki/shared/flag/AN.gif "Netherlands Antilles"
[flag_AO]: /wiki/shared/flag/AO.gif "Angola"
[flag_AR]: /wiki/shared/flag/AR.gif "Argentina"
[flag_AS]: /wiki/shared/flag/AS.gif "American Samoa"
[flag_AT]: /wiki/shared/flag/AT.gif "Austria"
[flag_AU]: /wiki/shared/flag/AU.gif "Australia"
[flag_AW]: /wiki/shared/flag/AW.gif "Aruba"
[flag_AX]: /wiki/shared/flag/AX.gif "Aland Islands"
[flag_AZ]: /wiki/shared/flag/AZ.gif "Azerbaijan"
[flag_BA]: /wiki/shared/flag/BA.gif "Bosnia and Herzegovina"
[flag_BB]: /wiki/shared/flag/BB.gif "Barbados"
[flag_BD]: /wiki/shared/flag/BD.gif "Bangladesh"
[flag_BE]: /wiki/shared/flag/BE.gif "Belgium"
[flag_BF]: /wiki/shared/flag/BF.gif "Burkina Faso"
[flag_BG]: /wiki/shared/flag/BG.gif "Bulgaria"
[flag_BH]: /wiki/shared/flag/BH.gif "Bahrain"
[...]

PR #1958, opened 2019-01-11
Remove multi-flag icons

### LINKED ISSUE #1948: Multi-flag icons should be removed
As countries gather, we need to make a new whole flag by editing which is hard. Also this instance... we can't divide a flag into three. Best to just add flags side by side. Need an opinion, and will make a PR nuking the images and changing their references.

cl8n:
closes #1948


cl8n:
@MegaApplePi could you remove them from #328 ?

MegaApplePi:
ok

PR #1983, opened 2019-01-20
readme: Add ppy and MillhioreF to news posts section

MegaApplePi:


cl8n:
keep as is

MegaApplePi:
Hmm, alright then. I thought that there would be some issue if both Shiro and Ephemeral were out.

PR #1984, opened 2019-01-20
oSEAT2: Grand Finals

fajar13k:
- Add Finals match results
- Add Grand Finals match schedules
- Add description to Finals mappools

---

### DIFF
--- a/wiki/Tournaments/oSEAT/2/en.md
+++ b/wiki/Tournaments/oSEAT/2/en.md
@@ -97,37 +97,30 @@ The osu! South East Tournament 2 is run by various community members.
 
 -----------------------
 
-## Match Schedules: Finals
+## Match Schedules: Grand Finals
 
-### Friday, 18 January 2019
-
-| Match Code | Player A |  |  | Player B | Match Time (UTC+8) |
-| :-: | --: | :-: | :-: | :-- | :-: |
-| MB | Crezz | ![][flag_ID] | ![][flag_SG] | GSBlank | **22:00** |
-
-### Saturday, 19 January 2019
+### Saturday, 26 January 2019
 
 | Match Code | Player A |  |  | Player B | Match Time (UTC+8) |
 | :-: | --: | :-: | :-: | :-- | :-: |
-| MA | Raindrop | ![][flag_SG] | ![][flag_SG] | SeeL | **22:00** |
+| LB | Rexeez | ![][flag_ID] | ![][flag_MY] | GSBlank | **21:00** |
 
-### Sunday, 20 January 2019
+### Sunday, 27 January 2019
 
 | Match Code | Player A |  |  | Player B | Match Time (UTC+8) |  |
 | :-: | --: | :-: | :-: | :-- | :-: | :-: |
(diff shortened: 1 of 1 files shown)


wiki/Tournaments/oSEAT/2/en.md
```diff
@@ -97,38 +97,31 @@ The osu! South East Tournament 2 is run by various community members.
...
-¹ Encounter dependent on winner of Match MA and Match MB
+¹ Encounter dependent on winner of Match LB
 
 -----------------------
 
 ## Mappools
 
 ### Finals
 
+ **This mappool will be played in Finals and Grand Finals.**
```
cl8n: I don't think you should bold here since it seems reserved for the download links

just a style thing, either way works probably

wiki/Tournaments/oSEAT/2/en.md
```diff
@@ -97,38 +97,31 @@ The osu! South East Tournament 2 is run by various community members.
...
+### Sunday, 27 January 2019
 
 | Match Code | Player A |  |  | Player B | Match Time (UTC+8) |  |
 | :-: | --: | :-: | :-: | :-- | :-: | :-: |
-| LA | Rexeez | ![][flag_ID] | ![][flag_ID] | Skydiver | **13:30** |  |
-| NA-1 | Raindrop | ![][flag_SG] | ![][flag_MY] | Crezz | **13:30** | ¹ |
-| NA-2 | Raindrop | ![][flag_SG] | ![][flag_MY] | GSBlank | **22:00** | ¹ |
-| NA-3 | SeeL | ![][flag_SG] | ![][flag_MY] | Crezz | **13:30** | ¹ |
-| NA-4 | SeeL | ![][flag_SG] | ![][flag_MY] | GSBlank | **22:00** | ¹ |
+| LC-1 | Skydiver | ![][flag_ID] | ![][flag_ID] | Rexeez | **21:00** | ¹ |
```
cl8n: these footnotes should probably go after the match code, rather than in their own cell

fajar13k: That would be a disaster for the readers and to look at.

cl8n: "LC-1 ¹"?

if you mean that the superscript could be confused with the match number, adding a space like above seems to fix the issue

cl8n: also you don't need to make it a numeral, it could just be an asterisk I think

fajar13k: > 
> also you don't need to make it a numeral, it could just be an asterisk I think

Don't, last week i have 16 superscript.  And you don't want 16 asterisks filling a cell.

cl8n: well in this case you've only got one :wink:

fajar13k: but consistency matters dad, even though match schedules only have to be filled when the tournament's still running.

fajar13k: Alongside the asterisks, will apply the suggestion

PR #2046, opened 2019-02-13
Catch RC combo colour rule fix

Xinnoh:
When adding rules about combo colours, I assumed kiai would always be used but it doesn't have to be. This fixes my oversight.


JBHyperion:
Seems pretty logical to clarify wording here yeah

PR #2053, opened 2019-02-16
Update Beatmap Nominators List

celerih:
Still waiting on a few ppl to give me their Additional Languages, but this adds the new wave of beatmap nominators, as well as removing the resignations and removals of the past month or so

PR #2055, opened 2019-02-18
Revert "Update Beatmap Nominators List"

Krystool:
Reverts ppy/osu-wiki#2053, too many extra spaces


cl8n:
that's... not how it works

please close this if you aren't going to add anything else

Krystool:
Should be reversed because it can broke the osu wiki page

cl8n:
every language's page look good to me, can you screenshot?

PR #2062, opened 2019-02-20
Remove code styling from headers

cl8n:


wiki/osu!_Program_Files/zh.md
```diff
@@ -16,32 +16,32 @@ osu! 目录文件
 文件夹
 ------
 
-### `_MACOSX`
+### _MACOSX
```
Joehuu: Don't know why there's an underscore. Update on `en.md` too. https://en.wikipedia.org/wiki/MacOS
```suggestion
### macOS
```

cl8n: It's styled this way because that's how the directory is named, each header lists a directory

Joehuu: Can you apply the backslash, edited the suggested.

PR #2064, opened 2019-02-20
Standardize spelling of "Hit object"

cl8n:


cl8n:
it's because some files were missing the final line ending, I added them back in for the changed files. and I edited outdated articles anyway because why not /shrug

PR #2031, opened 2019-02-07
Misc. formatting fixes

cl8n:
don't review this yet, I'll be redoing commits as necessary

- [x] HitObject -> Hit object (changing caps where necessary)
- [x] Remove whitespace from tables
- [ ] Standardize image filenames (only hyphens for separating words) + ASC
- [x] Remove code styling from headers
- [ ] Match article titles to their folder names (except cases where it makes sense to have different) - waiting on #1849 
- [ ] Link users/beatmaps/etc by ID instead of name
- [x] Remove tabs
- [x] Remove trailing (single) spaces
- [ ] Remove other kinds of whitespace not in ASCII

---

PR #2072, opened 2019-02-21
Cleanup MediaWiki leavings in the osu!wiki

TPGPL:


cl8n:
I'm going to apply all of the changes listed in #2031 throughout a series of PRs so you don't need to change things if it's on that list already

that being said can we just merge this one with the current changes asap, and split further ones into more specific PRs? we are going to run into a ton of conflicts otherwise

PR #2079, opened 2019-02-22
Cleanup MediaWiki leavings in the Modding article

TPGPL:


TPGPL:
because they need to be translated so they can work

cl8n:
this makes it a little difficult for translators though, like when they go through these articles again I really doubt they'll realise that the glossary links need to be updated

cl8n:
maybe add a comment before the affected lines?

Joehuu:
`Glossary` does not even have any translations, so it's fine.

PR #2086, opened 2019-02-26
otst! 2019: Group Stage

fajar13k:
- Add GS Mappools
- Add GS Match schedules
- Add Groups (for GS) on participants
- Update participants & staff

---


WalterToro:
Change group stage schedule from March 2 to 11

PR #2094, opened 2019-03-01
Add missing "s" in privacy policy

Phippe:
- change "... a user register an ..." to  "... a user registers an ..."
---


cl8n:
remove the part of your comment below the `---` before merging

peppy:
@cl8n technically which is more correct?

cl8n:
this change

Joehuu:
Can you add the `---` in the OP and merge this @peppy ?

PR #2110, opened 2019-03-06
Rename Expert+ diff name in catch

Shadowthegod:
Changing the Expert+ difficulty name, which is recently changed, to Overdose+. The reason is that Deluge was never meant to serve that purpose, it was just the name for a second Overdose, not officially even at that. Thus the rare case usage. 

---


Shadowthegod:
Should be done, if I did it right :eyes:

PR #2115, opened 2019-03-07
New Featured Artist: Imperial Circus Dead Decadence

Ephemeralis:
👀 🙊 

---

Imperial Circus Dead Decadence joins the fray as the latest Featured Artist!


Raidencio:
OH MY GOD

Nivalyx:
i think alongside linking the forum thread it would be better to link directly to the new MG site (http://mappersguild.com) as well though so that the readers could be redirected easier to the system

on the other side i'm so positive and happy about this news though btw :)

peppy:
I think the forum link is fine for this one.

PR #2141, opened 2019-03-26
[SK] About

FajsiEx:


wiki/About/sk.md
```diff
@@ -0,0 +1,5 @@
+# O osu!
```
TPGPL: more like `About osu!wiki`

Joehuu: It's `osu! wiki` with a space.
```suggestion
# O osu! wiki
```

FajsiEx: Fixed

PR #2018, opened 2019-02-05
Compendium rework

RockRoller01:

PR #2144, opened 2019-03-28
Delete the Skinning Compendium

TPGPL:
Closes #2018 

I think there's no actual reason to keep this list on the wiki, as it will be always outdated and will need constant updates.

---


cl8n:
the "See also" list of Skinning should link to the forum post instead of this now removed page

PR #2158, opened 2019-04-05
Update Shortcut Key Reference to include "H" keybind

Cyuubi:
Looks like it was never added, so I added it. :P


wiki/Shortcut_Key_Reference/en.md
```diff
@@ -143,6 +143,7 @@ Note: Not all of these work in the [multi](/wiki/multi) mode's song select scree
 - `Space`: Skip opening cut-scene
 - `Tab`: Toggle scoreboard
 - `Shift` + `Tab`: Toggle in-game interface
+- `H`: Toggle scoreboard and in-game interface at once (and hide "REPLAY MODE" text if you are viewing a replay)
```
cl8n:
```suggestion
- `H`: Toggle scoreboard and replay interface (only available when watching a replay or spectator stream)
```
seems to hide different components than "in-game interface"

Cyuubi:
Fixed.

PR #2159, opened 2019-04-07
GO 3: Semifinals

fajar13k:
- Update participants
- Update match results
- Update mappools
- Update event timeline
- Staff changes

---

### DIFF
--- a/wiki/Tournaments/GO/3/en.md
+++ b/wiki/Tournaments/GO/3/en.md
@@ -54,7 +54,7 @@ The Game\_Osu Tourney 3 is run by various community members.
 | Developers | ![][flag_FR] [Loazo](https://osu.ppy.sh/users/3925586), ![][flag_SE] [AntoN](https://osu.ppy.sh/users/2538562), ![][flag_FR] [0xy](https://osu.ppy.sh/users/9239673) |
 | Wiki | ![][flag_ID] [Naruk](https://osu.ppy.sh/users/7100002) |
 
-*Special thanks to kymutsojason, ![][flag_FR] [FayeurS 5](https://osu.ppy.sh/users/6177263), ![][flag_GB] [Bubbleman](https://osu.ppy.sh/users/5182050), ![][flag_FR] [Mimir](https://osu.ppy.sh/users/7382734), and ![][flag_FR] [Nozhomi](https://osu.ppy.sh/users/2716981) who served their role but left early.*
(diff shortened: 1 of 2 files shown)


wiki/Tournaments/GO/3/en.md
```diff
@@ -73,13 +75,184 @@ The #Game\_Osu Tourney 3 is run by various community members.
 
 ## Participants
 
-*The participants will be revealed after the registration phase closes.*
+| BWS Rank |  | Player |
```
cl8n: is there a wiki page about BWS? this would make no sense to most people

fajar13k: Not an official article, but there's a link in the ruleset that explains it.
Or should i import the short definition to the Participants section?

cl8n: yeah just a short description is enough. maybe we can give it its own article sometime and have that description link to it in the future

cl8n:
```markdown
*Note: Badge-weighted seeding (BWS) is used in this tournament to more accurately seed players and enforce rank restrictions. It works by taking into account the amount of recorded tournament wins (via badges) that a player has in addition to their rank. See the Game\_Osu website linked above for more details.*
```
?

cl8n:
where does the number sign in `#Game_Osu` come from? everywhere I'm looking does not have the number sign

fajar13k:
Original name is `#Game_Osu Tourney Winter/Spring` and **is the 3rd iteration** of the tournament.
So rather than using #GO TWS, the host and I agree to use #GO 3 instead.
For other examples, i've done this in Asian Taiko Showdown article.

wiki/Tournaments/GO/3/en.md
```diff
@@ -139,9 +139,11 @@ The #Game\_Osu Tourney 3 is run by various community members.
 | #569.62 | ![][flag_AU] | [GranDSenpai](https://osu.ppy.sh/users/3997580)
 | #700 | ![][flag_US] | [Tj7223](https://osu.ppy.sh/users/10210668) |
 | #737 | ![][flag_MY] | [Rampax](https://osu.ppy.sh/users/3995630) |
-| #766 | ![][flag_ME] | [Andros](https://osu.ppy.sh/users/9072586) |
+| #766 | ![][flag_MX] | [Andros](https://osu.ppy.sh/users/9072586) |
 | #972.71 | ![][flag_NL] | [GladiOol](https://osu.ppy.sh/users/23326) |
 
+*Note: Badge-weighted seeding (BWS) is used in this tournament to more accurately seed players and enforce rank restrictions. It works by taking into account the amount of recorded tournament wins (via badges) that a player has in addition to their rank. See the #Game\_Osu website linked above for more details.*
```
cl8n: should go above the table i think

cl8n:
GO 3 is fine for the acronym, but the title doesn't seem correct for this tournament, the only reference I could find using `#Game_Osu` at all was their YouTube channel which doesn't seem very active or meaningful

maybe the first iteration was called #Game_Osu, but this one seems to be Game_Osu

PR #2160, opened 2019-04-07
BN Rules: Disqualifications

TheMoMan:
As a result of [restructuring](https://osu.ppy.sh/community/forums/topics/862002?start=7005591) the QAT and BN, the Beatmap Nominators will be given access to the disqualification button. These additions to the rules will outline what is expected of a BN when disqualifying a beatmap. 

Changes should be applied immediately after BNs get the disqualify button.


wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -27,6 +27,15 @@ Violations of those rules may be reported to a member of the [Quality Assurance
...
 + **Do not partake or participate in paid modding, ranking OR beatmapping commissions.** There should never be any question as to the legitimacy or motives of members of any "official" osu! team (BN, QAT or otherwise) involved in the ranking process.
++ **Disqualifications should primarily be done for beatmaps with unrankable issues, or if the mapset owner requests it.** 
++ **Qualified beatmaps found with unrankable issues must first be posted to the [Report a Beatmap](https://osu.ppy.sh/community/forums/topics/447428) thread.** If the issue has already been reported by someone else, then you are free to disqualify the beatmap. Beatmaps should not be disqualified without at least two people confirming the issue.
++ **Disqualification requests by the mapset owner can be applied immediately.** 
++ **For more subjective issues or concerns, waiting for a mapper’s reply is preferable if the map still stays in qualified section for more than 24 hours.** In case there is very little time left and the mapper gives no response, the map may be disqualified for further discussion.
```
Naxesss:
```suggestion
+ **For non-unrankable issues or concerns, waiting for a mapper reply is preferable if the map still has more than 24 hours until rank, otherwise it may be disqualified for further discussion.**
```
"subjective" is subjective in itself, better to draw clearer lines. Also better to clearly state what "too little time" is to prevent inconsistencies. From what I understood the whole thing can also be shortened down since the former and latter sentences basically say the same things.

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -27,6 +27,15 @@ Violations of those rules may be reported to a member of the [Quality Assurance
...
++ **Disqualification posts must contain a short reason for the disqualification, as well as links to any related issues posted in the beatmap discussions.**
```
Naxesss:
```suggestion
+ **Disqualification posts must contain a short reason for the disqualification, as well as links to any related issues posted in the beatmap discussions.** This includes requested disqualifications; simply writing that it was requested is not enough.
```
avoids "disqualified upon mapper's request" type reasons, which say nothing about why it was requested

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -27,6 +27,15 @@ Violations of those rules may be reported to a member of the [Quality Assurance
...
++ **Disqualification requests by the mapset owner can be applied immediately.** 
++ **For more subjective issues or concerns, waiting for a mapper’s reply is preferable if the map still stays in qualified section for more than 24 hours.** In case there is very little time left and the mapper gives no response, the map may be disqualified for further discussion.
++ **Disqualification posts must contain a short reason for the disqualification, as well as links to any related issues posted in the beatmap discussions.** 
++ **Ensure that you are looking at the most up to date version of the beatmap before disqualifying.** 
++ **Beatmaps qualified in modding v1 that need to be disqualified should be brought to the attention of the NAT.** 
++ **Issues that can be resolved via the map management panel must still be disqualified.** In the case that this is not possible, contact a member of the NAT.
```
Naxesss:
```suggestion
+ **Issues that can be mitigated through online offsets, tags or similar must still be disqualified.** In the case that this is not possible, contact a member of the NAT. Online offsets and tags only apply for the website, so players will still have the issues in the client unless fixed before ranked.
```
explains why this is the case, as well as replaces "map management panel" with what we would actually change in it, since I doubt nominators will know about the panel better than what it enables.

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -27,6 +27,15 @@ Violations of those rules may be reported to a member of the [Quality Assurance
...
++ **Beatmap Nominators in probation aren’t able to disqualify a map.** They may however participate in such discussions if desired.
```
Naxesss:
```suggestion
+ **Only full members may disqualify beatmaps, regardless of reason.** Probation members may however still participate in such discussions and report beatmaps like everyone else.
```
rewording, avoids the contraction and statement about being unable, even though they are.

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -70,3 +79,15 @@ _Change date: 12. April 2018, **in effect immediately**_
...
```
Naxesss: remember to update this as well, should stuff above be changed

Naxesss:
Should probably also look into the vetoing rule before merging, since it's also affected by the ability to dq.

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -70,3 +79,15 @@ _Change date: 12. April 2018, **in effect immediately**_
...
+
+_Change date: 7. April 2018, **in effect immediately**_
```
Kagetsu24:
```suggestion
_Change date: 7. April 2019, **in effect immediately**_
```

wiki/People/Beatmap_Nominators/Rules/en.md
Naxesss:
```suggestion
+ **For non-unrankable issues or concerns, waiting for a mapper reply is preferable if the map still has more than an estimated 24 hours until rank, otherwise it may be disqualified for further discussion.** This does not apply to posts made less than 48 hours before the estimated rank date of the beatmap set, to give mappers ample time to reply and decide whether to make changes
```
discussed this part a bit further and decided that we'll probably need a measure to ensure people don't post right before the map gets ranked to force a dq.

cl8n: the second revision is really confusing to read. if I'm understanding this correctly...
for posts made before 48 hours remaining: unrankable issue = dq, non-unrankable = wait, but only until the 24-hour mark at which it is also OK to dq
for posts made after 48 hours remaining: unrankable issue = dq, non-unrankable = ...dq? I think?

the second sentence is totally useless in that case, it should just be removed and then the 24-hour rule will cover this case

wiki/People/Beatmap_Nominators/Rules/en.md
cl8n: well we could jsut fix this on the site too

wiki/People/Beatmap_Nominators/Rules/en.md
Naxesss: It's rather
**unrankable at any point in time** = dq
**non-unrankable** posted when there's **>48h left** = wait until 24h left and then dq if no reply (so there's a 24 hour time period in which the mapper can reply)
**non-unrankable** posted when there's **<48h left** = can be ignored

if there's <48h left it's not a reasonable amount of time in which the mapper could respond to it, so dqing for those cases is way too easily abusable (e.g. just wait until there's little time left to post an issue to force a dq, then repeat multiple times means you're basically stalling a set for things the mapper may not even want to apply but has literally no way to convey that before a dq may happen).

cl8n: two problems then
- what if there is an important (but not unrankable) issue noticed after 48h left? just send it to ranked anyway?
- the second sentence is wrong. if "this does not apply", then the dq is valid, but you are saying it is not valid

cl8n: fwiw I don't think stalling by a single DQ is even a problem, I don't see why this would count as "abuse" if the BN thinks they have a good reason to DQ

Naxesss: - call for a bn to veto should they agree
- "does not apply" refers to "otherwise it may be disqualified for further discussion"

bns are given the benefit of the doubt that they're correct so if they feel the need to dq they can do so at any point in time assuming they have a reason to do so

cl8n: is there any difference between a veto and a DQ if the beatmap is already qualified? .-. now I'm more confused

TheMoMan: It should be so that:
72 hours before rank, suggestion posted -> 23 hours before rank, no response, DQ
47 hours before rank, suggestion posted -> No DQ

To make it clearer, I think:

```suggestion
+ **For non-unrankable issues or concerns, the mapper should be given time until 24 hours before estimated rank time to respond.** If the mapper fails to respond, the beatmap may be disqualified for further discussion.  This does not apply to posts made less than 48 hours before the estimated rank date of the beatmap set, to give mappers ample time to reply and decide whether to make changes.
```

Naxesss: vetoes ensure that no other nominator renominates until the vetoer thinks the issue is resolved or it's otherwise lifted (it's basically just a tool to make sure people actually discuss things and don't just try to escape back to ranked without the other person/mediators agreeing to it)

cl8n: so basically this rule is like
you can disqualify, but only if \<these conditions\> are met, and \<those conditions\> don't apply when \<this other condition\> is met, and also if you don't happen to be disqualifying for \<this other rule mentioned later about vetos\>

there's got to be a better way to write all of that. reading this alone doesn't even allude to the veto thing at all and the 48hour thing is unclear in which part of the sentence it is saying no longer applies

cl8n: this would change the rule and not just the wording, but I think it'd make more sense to just say if the mapper doesn't have enough time to respond, the map can be dq'd *only via* veto. if the mapper *does* respond and agree with the concern, a normal dq can be done

Naxesss: Is the thing Mo suggested clearer? Not sure why you're talking about defining vetoes in the rule, this rule doesn't have anything to do with vetoes, unless I'm misunderstanding

cl8n: the rule seems to be overridden entirely by the fact that a BN can just veto-disqualify instead of normally disqualifying, so it's confusing when veto is not mentioned at all

cl8n: your x -> y chart leaves out the case
no reply from mapper due to not having any time to respond -> dq only possible with veto

Naxesss: It's not overridden, it's just overlapping with it in terms of when you can dq. In this specific rule we're talking about dqing for the mapper not replying to some suggestion in accordance to the code of conduct, and this can be to allow someone that isn't bn to have their points not be ignored. Put this in contrast to dqing for a veto (which is another rule), in which case the bn is held responsible for discussing and trying to reach an agreement, and can be done at any time completely unrelated to this.

Naxesss: Also yes I could have included that in the x y chart. It's less of mapper declines -> veto and more of just straight up vetoing whenever, it's just that the chart was put into context of this rule, which assumes that the start point is that some post has been made which needs a response.

cl8n: another unrelated issue with this rule is that it doesn't specify what the mapper is supposed to be replying to; you'd have to read previous rules to know that this is about disqualification-worthy concerns. so maybe this should go as a sub-bullet one of the previous rules?

what you and Mo wrote is correct, technically, I'm just saying that it's hella confusing if I were a BN looking to disqualify a map around or during that 48-hour range. I shouldn't have to read multiple different (and pretty unrelated) rules and then figure out based on all these conditions whether or not i can dq, there should just be a single rule that answers the question, "when am I allowed to DQ a map"

Shadowthegod: As a BN technically you should already know these rules or at least have some kind of vague knowledge about them. Just my two cents. Especially because due to these changes, the BNG becomes the well-known Quality Assurance Team (in other games, listed in the credits). I wouldn't be surprised if this knowledge would be required to be a BN in the future, to be honest. 

But as I do not see inside anymore, nor actually want to to be fair, these are just my thoughts.

cl8n: this article is about rules for BNs, it's here so that it can be used as a simple reference

Shadowthegod: Yes, but as a BN you must know these rules and not search for it when you are in need. If you as a BN do not know what can you do or not and when then you should not be a BN at all. Or maybe I just have higher expectation towards the content management group.

Shadowthegod: I mean, at least for me the very first version was understandable enough to know what to do and when. Note that I am not an expert english speaker.

cl8n: I think the original is fine too. or the first revision

Naxesss: Shadow is correct. As a nominator you should already know all of the rules before you start doing anything bn related. Nominators are also actively updated on any changes to them, so once you know them by heart or at least feel familiar around them, the need to read through all of them again is pretty much non-existent. As such we can expect that nominators know about all other rules and terms involved, or if not, will.

cl8n: here's my suggestion then, has the same rule as Mo's, just reworded with the changes i mentioned
- makes it clear which rule this contrasts with (line 30)
- bolded text now includes the defining reason of this type of dq, being that the mapper eventually agrees with the BN
- that last sentence is no longer ambiguous with its reference
```suggestion
+ **Disqualifications may be issued proactively for non-unrankable issues, so long as the mapper agrees with the reasons for doing so.** A post must be made on the beatmap discussion prior to 48 hours before the estimated rank date to give the mapper ample time to make a decision and reply. The beatmap set may also be disqualified if the mapper gives no response by 24 hours before the estimated rank date, to allow more time for discussion.
```

TheMoMan: I'd say that works for me.

TheMoMan: We decided on my rewording since that's the clearest without repeating what is said in other rules. Feel free to reopen discussion if there's still concerns.

cl8n: still doesn't address any of the bulleted things above but it's mostly nitpicky grammar anyway, this is fine for now. I'll likely make another pass over RC and all the related articles sometime so I may mention you later if I edit this part

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -70,3 +73,25 @@ _Change date: 12. April 2018, **in effect immediately**_
...
 
 _Change date: 12. Dec 2018, **in effect immediately**_
+
++ **Disqualifications should primarily be done for beatmaps with unrankable issues, or if the mapset owner requests it.** 
++ **Qualified beatmaps found with unrankable issues must first be posted to the [Report a Beatmap](https://osu.ppy.sh/community/forums/topics/447428) thread.** If the issue has already been reported by someone else, then you are free to disqualify the beatmap. Beatmaps should not be disqualified without at least two people confirming the issue.
++ **Disqualification requests by the mapset owner can be applied immediately.** 
++ **For non-unrankable issues or concerns, the mapper should be given time until 24 hours before estimated rank time to respond.** If the mapper fails to respond, the beatmap may be disqualified for further discussion.  This does not apply to posts made less than 48 hours before the estimated rank date of the beatmap set, to give mappers ample time to reply and decide whether to make changes.
```
Joehuu:
```suggestion
+ **For non-unrankable issues or concerns, the mapper should be given time until 24 hours before estimated rank time to respond.** If the mapper fails to respond, the beatmap may be disqualified for further discussion. This does not apply to posts made less than 48 hours before the estimated rank date of the beatmap set, to give mappers ample time to reply and decide whether to make changes.
```

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -2,15 +2,15 @@
...
+These rules are the result of discussion within the [Nominator Administration Team](/wiki/People/Quality_Assurance_Team) and have taken into account feedback from the [Beatmap Nominators](/wiki/People/Beatmap_Nominators) upon their proposal.
 
 Beatmap Nominators possess the tools to nominate beatmaps and shape the world of osu! by promoting new content towards being included in the official listings, for this reason they have to abide by the following rules at all times:
 
 ## Rules
 
 Rules are exactly that: **rules.** Violating any of these may get the respective Beatmap Nominator removed permanently from the Beatmap Nominators.
 
-Violations of those rules may be reported to a member of the [Quality Assurance Team](/wiki/People/Quality_Assurance_Team) directly.
+Violations of those rules may be reported to a member of the [Nominator Administration Team](/wiki/People/Quality_Assurance_Team) directly.
```
Loctav: Change to Nomination Assessment Team

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -36,16 +45,10 @@ Violations of those rules may be reported to a member of the [Quality Assurance
...
-+ **You must not place any icons besides the star, bubble, bubble pop and heart.** Whilst you are able to set other icons, they are either deprecated and discouraged from use (flame, circle, drum, apple, piano) or reserved for use by the Quality Assurance Team (nuke). If you accidentally set an unwanted icon, contact a QAT immediately or overwrite the icon in case you wanted to post a bubble or heart.
++ **You must not place any icons besides the star, bubble, bubble pop and heart.** Whilst you are able to set other icons, they are either deprecated and discouraged from use (flame, circle, drum, apple, piano) or reserved for use by the Nominator Administration Team (nuke). If you accidentally set an unwanted icon, contact a NAT immediately or overwrite the icon in case you wanted to post a bubble or heart.
```
Loctav: Chnage to Nomination Assessment Team

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -2,15 +2,15 @@
 
 The rules listed here affect what Beatmap Nominators can and cannot do when nominating beatmaps as well as set the tone for the general conduct expected from them.
 
-These rules are the result of discussion within the [Quality Assurance Team](/wiki/People/Quality_Assurance_Team) and have taken into account feedback from the [Beatmap Nominators](/wiki/People/Beatmap_Nominators) upon their proposal.
+These rules are the result of discussion within the [Nominator Administration Team](/wiki/People/Quality_Assurance_Team) and have taken into account feedback from the [Beatmap Nominators](/wiki/People/Beatmap_Nominators) upon their proposal.
```
Loctav: Change to Nomination Assessment Team

PR #2164, opened 2019-04-11
Update BN listing

Xinnoh:
Not updating the QAT page because the entire thing need an overhaul.
I think that NATs should be included on the BN list in future updates because outside of managing the team, their role is identical to the BNG in regards to mapping.

Removals:
Maridius, Krfawy, Namki, alienflybot, Ascendance (moved to qat)
Added:
JBHyperion, Raiden, Gabe, Hobbes2, Regraz

alienflybot and Kuron-kun are currently BNs but they've already resigned.


cl8n:
should fix the other groups too if users moved

TheMoMan:
Add: Noffy
Remove: WWWWWW, wonjae, DTM9, Haruki, Yales
Add language: ailv: Chinese

WalterToro:
Superseeded by #2265

PR #2192, opened 2019-04-23
Update multi match history image

Joehuu:
Moving away from old format, while also doing small changes.

---


Joehuu:
Blocked until the match history page is updated with `Multi`. Just wanting it to be official.

Joehuu:
Can be unblocked/merged.

PR #2202, opened 2019-04-27
Remove Official Songs wiki page

Joehuu:
Article content (songs) has since been rewarded. Program renamed/broadened to `Featured Artists`???

Article is outdated (2015), and content is temporary (meaning it shouldn't be an article).

---


cl8n:
if there's not an article about Featured Artists yet open an issue probably

seems good to remove this

PR #2102, opened 2019-03-03
Update difficulty thresholds

Poyo-SSB:
Documents ppy/osu-web#4181.

---

Updates the listed difficulty thresholds.

PR #2225, opened 2019-05-05
Reword osu!mania difficulties drain time restriction

### LINKED ISSUE #2109: Mania ranking critera rephrase because of new difficulty icons
So with the new icon changes a part of the ranking criteria that affects mania has become very confusing. Speceficly here "...lower than 3:30, the lowest difficulty of each included game mode cannot be harder than a Normal. Because osu!mania does not have a difficulty-specific Ranking Criteria yet, an osu!mania beatmapset's Normal difficulty is defined as a difficulty below 2.00 stars." The reason this is confusing is because mania's normal icon is no longer below 2 stars, and with the new icons that is an easy.
My suggestion for a quick fix would be to rephrase the whole sentence to "...lower than 3:30, the lowest difficulty of each included game mode (except osu!mania) cannot be harder than a Normal. Because osu!mania does not have a difficulty-specific Ranking Criteria yet, an osu!mania beatmapset's lowest difficulty must be an easy difficulty."

Joehuu:
Fixes #2109.

---

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -105,7 +105,7 @@ For the record, these include the recent changes made [from the discussion forum
     - **...any two or more osu!taiko, osu!mania, or osu!catch difficulties must be arranged in a reasonable spread.**
 - **Every difficulty of a beatmapset must have a minimum drain time of 30 seconds.**
 - **If the drain time of each difficulty is...**
-    - **...lower than 3:30, the lowest difficulty of each included game mode cannot be harder than a Normal.** Because osu!mania does not have a difficulty-specific Ranking Criteria yet, an osu!mania beatmapset's Normal difficulty is defined as a difficulty below 2.00 stars. For hybrid beatmapsets that include osu! difficulties, the additional modes' lowest difficulties cannot be harder than a Hard.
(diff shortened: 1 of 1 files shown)


cl8n:
this doesn't help because "Easy" is left undefined (there is no osu!mania-specific RC)

I think the issue is invalid anyway because "2.00 stars" isn't meant to refer to a difficulty icon, it's just an arbitrary star rating that was chosen. RC difficulties don't have much to do with the icons on the website

Joehuu:
I thought it was defined in https://osu.ppy.sh/help/wiki/Difficulties#star-rating with the general star rating ranges for every mode.

The issue author probably knew that `Easy` now fits into that arbitrary range. /shrug

cl8n:
that article seems really misleading, I'm not sure why it references RC at all

the characteristics of each difficulty should be defined under the "Difficulty-specific" header of each game mode's RC article (which I believe they are for all except osu!mania)

Joehuu:
https://github.com/ppy/osu-wiki/pull/2102#discussion_r261884894

cl8n:
that was a step in the right direction but I think that sentence can be removed altogether

Joehuu:
Guess I'll close then.

PR #2249, opened 2019-05-12
Staff Log: May and June 2019 update

MegaApplePi:


wiki/People/Staff_Log/2019/en.md
```diff
@@ -93,3 +93,72 @@ All users listed below were added on 2019-02-16.
 - [wonjae](https://osu.ppy.sh/users/5032045) - resignation (2019-04-25)
 - [DMT9 Nowa](https://osu.ppy.sh/users/5428909) - resignation (2019-04-27)
 - [\_WWWWWW_WWWWWW_](https://osu.ppy.sh/users/1981781) - unsatisfactory behavior (2019-04-30)
+
+## May
+
+- **Quality Assurance Team becomes the Nomination Assessment Team** (2019-05-06)
+- Moved [-Mo-](https://osu.ppy.sh/users/2202163) to the Nomination Assessment Team (2019-05-06)
```
cl8n: instead of listing the new NAT, it would be more helpful to list any QAT who were not moved to NAT because they ended up in different groups (gmt/bn/alumni)

the previous line already implies that all QAT became NAT

WalterToro: Agree with @cl8n on this one, better to let know the ones that were moved to nat or gmt (and or) bn

MegaApplePi: @cl8n I don't understand on what you mean. Were there other changes to some users than the ones listed here?

This list of changes were based off of [this post](https://osu.ppy.sh/community/forums/topics/179514?n=62).

WalterToro: Wait no, now I remember those changes were done before (see April 2019 log).
So the remainder members at that point were moved from QAT to NAT.
_whoooops_

cl8n: that addresses the first part of my comment but not the second, I'll just make another PR to show what I meant then lol

PR #2252, opened 2019-05-12
Add iDOL@NSTER tournament

cl8n:


cl8n:
depends on #2253 now

cl8n:
i put event details near the top bc it hasn't happened yet and that's important info for anyone wanting to attend

cl8n:
and for the images-- idk, I think they're kinda cute so I just used them instead of text (that's what the website opts to use, too). after the tournament ends I'll probably remove the schedule section entirely cuz these LANs never end up sticking to a concrete schedule anyway. as long as the final version includes the day of the event I think that's good enough

wiki/Tournaments/Zotcade/2019/en.md
```diff
@@ -0,0 +1,294 @@
...
+## Ruleset
+
+![iDOL@NSTER Bracket Format](img/format.png)
+
+Any disregard for these rules will be met with punishment, ranging from a warning to immediate disqualification. It will be up to the discretion of the staff to punish based on the severity of the offence. **These rules and regulations are subject to change during the event.**
+
+### Lobby and Matches
+
+1. Missing your match counts as a loss; no rematches will be made.
+1. Score V2 will be used as the win condition.
```
Joehuu: We increment according to ASC. Section was deleted on recent ASC PR for some reason :(.

cl8n: is that just an example or a rule (also why, doesn't that make it harder to update?)

Joehuu: > The Article Styling Criteria (ASC) serves as the osu! wiki's enforced styling standards to keep **consistency** in **clarity**, **formatting**, and **layout** in all articles.

It should be considered a rule or preferred to fulfill ASC's goal. Anyway, it has been used in most articles and doesn't confuse new people about this behavior. It also keeps track of numbers when it really matters (i.e. referencing a step, but having to go to parser instead to get said number).

wiki/Tournaments/Zotcade/2019/en.md
```diff
@@ -0,0 +1,294 @@
...
+| :-: | :-- |
+| ![Gold Crown](/wiki/shared/GCrown.png "1st place") | Profile badge, HyperX Cloud Alpha Headset |
+| ![Silver Crown](/wiki/shared/SCrown.png "2nd place") | HyperX Alloy Core RGB Keyboard |
+| ![Bronze Crown](/wiki/shared/BCrown.png "3rd place") | HyperX Pulsefire FPS Mouse |
+
+![iDOL@NSTER Profile Badge](img/badge.png)
+
+## Organisation
+
+The iDOL@NSTER is run by various osu! community members at UCI or visiting for the event.
```
TPGPL:
```suggestion
The iDOL@NSTER is run by various osu! community members at UCI or visiting the event.
```

cl8n: what's wrong with the for? you're changing the meaning of the sentence with this suggestion

PR #2265, opened 2019-05-22
Update Beatmap Nominator Listing

celerih:
Hasn't been updated in like 2 months and we got some new additions recently so did some changes. Not done to other languages since the BN changes have to be translated and not just update listing


cl8n:
not in alphabetical order

PR #2274, opened 2019-05-26
Strip BOM

cl8n:
for some reason there's a BOM in these few files, just annoying because dos2unix and sed remove them automatically

---


sr229:
Let me guess, UTF-8 with BOM?

cl8n:
BOMs aren't required (or encouraged) for UTF-8 so 💥

PR #2276, opened 2019-05-26
Amend DQ post rule

TheMoMan:
The first line in a DQ post is what gets posted into the disqualification reason below the nomination status, and so it should ideally contain the summary of why a disqualification has taken place. This is for the convinience of new readers coming into beatmap discussion.

Posting just the link, for example "DQ to fix 123456" should also be discouraged as it does not properly explain why the disqualification has taken place.


Mordredosu:
beyond unnecessary, we really don't need to remove every fun aspect from being bn and turn it into lifeless machine work

people aren't dumb enough to completely ignore everything after the first line, if they can't read anything but that they shouldn't be using a pc in the first place; as long as the reason is clearly visible this should not be considered an issue at all, here's an example that's as easy to read as possible https://i.imgur.com/WvY6qn5.png 

99% sure the nat will still click on the link and check the dq reason no matter what's written in the first line, so this seems very not needed


as for the 2nd part, posting "DQ to fix 123456" does say why the dq has taken place?... clicking another link won't kill you and you'd do it anyways to check if what's written is actually true

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -31,7 +31,7 @@ Violations of those rules may be reported to a member of the [Nomination Assessm
...
 + **Disqualification requests by the mapset owner can be applied immediately.** This includes yourself, in case you want to self-dq.
 + **For non-unrankable issues or concerns, the mapper should be given time until 24 hours before estimated rank time to respond.** If the mapper fails to respond, the beatmap may be disqualified for further discussion. This does not apply to posts made less than 48 hours before the estimated rank date of the beatmap set, to give mappers ample time to reply and decide whether to make changes.
-+ **Disqualification posts must contain a short reason for the disqualification, as well as links to any related issues posted in the beatmap discussions.** This includes requested disqualifications; simply writing that it was requested is not enough.
++ **The first line in the disqualification post must contain a short reason for the disqualification.** This includes requested disqualifications; simply writing that it was requested is not enough. You must also link any related issues posted in the beatmap discussions unless there are too many to post, however, avoid posting only links in the first line of the post.
```
cl8n:
```suggestion
+ **The first line in the disqualification post must contain a short reason for the disqualification.** This includes requested disqualifications; simply writing that it was requested is not enough. You must also link any related issues posted in the beatmap discussions unless there are too many to post. Avoid writing only links in the first line of the post.
```
two subordinations in a single sentence :blobsweats:

pimpG:
i like this change because it will also get rid of childish nomination resets, there are much better ways to have fun with bn work than mocking other mappers and nominators for the mistakes they made.

this should be applied for bubble pops/first nomination resets too.

cl8n:
oh woops I didnt see Mordredosu's comment before

um... I'm not sure how this change makes anything less "fun", u can still type however you normally would, just make sure to have a real reason in the first line because the first line is used for preview on the event pages and nomination status

TheMoMan:
I mean you could still write jokes into the body of the post yeah. Main reason is so people get a quick reference to the reset, kinda comparable to git commits.

Also how would you suggest rewording to incorporate "bubble pop" resets?

pimpG:
Me?

Would just replace the words dq and disqualification for reset and nomination reset from your post

The wording of the actual reset post can be anything as long as it meets the requirements stated.

Issue #2275, opened 2019-05-26
Misc. formatting issues


cl8n:
adding onto #2031

- [ ] Strip BOM (#2274)
- [x] HitObject -> Hit object (changing caps where necessary) (#2064)
- [ ] Remove whitespace from tables
- [ ] Standardize image filenames (only hyphens for separating words) + ASC
- [x] Remove code styling from headers (#2062)
- [ ] Match article titles to their folder names (except cases where it makes sense to have different)
- [ ] Link users/beatmaps/etc by ID instead of name
- [x] Put osu.ppy.sh in front of osu! links (#2076)
- [ ] Remove tabs
- [x] Remove trailing (single) spaces (#2067)
- [x] Remove other kinds of whitespace not in ASCII (#2063)
- [ ] Add pipes to tables without pipes on the ends
- [ ] Fix old MediaWiki image entries
- [ ] Limit word boundary whitespace to a single space in normal paragraphs
- [ ] Make all headers use hashes instead of underscoring lines
- [ ] Don't use horizontal rules before headers
- [ ] Remove slashes from anchors
- [ ] Remove trailing slashes from links
- [ ] Remove useless newlines
- [ ] Handle news.ppy.sh links
- [ ] Handle w.ppy.sh images
- [ ] Convert /beatmaps/ links into /beatmapsets/
- [ ] Other MediaWiki things (expand this later)
- [ ] Handle special variables (?m=0, etc) on /beatmap(set)s and /users
- [ ] Point /p/ links to new site
- [ ] Point /forum/<module> links to new site

MegaApplePi:
> Single-space lists

So, we going with single spaces for lists? 

~~nts I wrote a thing explaining why I thought we should use 4 characters. Only to cut myself off, realising that it is some kind of hypothetical that would only occur if we are in need of long and heavily indented lists. /shrug~~

cl8n:
oh that is sort of misleading how I wrote it sorry. i dont mean the indenting, I mean the amount of blank lines between bullets in a list (nearly all lists should have no blank lines, but some do for seemingly no reason)

TicClick:
@cl8n this is something you have the most context for -- wanna tick some boxes/add short explanations for what needs to be done, so that others could pick some tasks up?

cl8n:
[...]

PR #2278, opened 2019-05-27
Add formatting tool

cl8n:
**don't use this yet**, I'll unmark as draft when I can verify that it works in all cases. see known issues below. it's also just a huge mess lol

adds a script to check and fix formatting errors. I don't expect to ever merge this, but it could be helpful for regular contributors to use. it should probably be written as a CI check at some point

two approaches can be used here: either parse the markdown and rebuild it, which should give 100% valid results every time but lack any detail on what errors were fixed, or search and fix errors individually (current), which is helpful for contributors but prone to introducing errors or keeping more around

current rules:
1. remove extra/unneeded whitespace from the beginnings and ends of files
2. use atx headers instead of setext
3. check for tabs and other disallowed whitespace (*doesn't actually change any files*, misleading, I know)

known issues:
- line numbers aren't accurate sometimes
- post-yaml newline addition isn't logged as an issue (it should be a rule, too)
- code blocks are not exempt from formatting rules

usage:
run all rules - `node bin/format.js`
run some rules - `node bin/format.js --rule 1,2`


cl8n:
the plan is to hit all of #2275 with this thing eventually /shrug

sr229:
We could probably rename it as `wikifmt` instead.

peppy:
Have you checked for available CI tools with github integration? last i checked there were a few for markdown.

cl8n:
I'm working on extending markdownlint with rules specific to osu-wiki, and that works nicely with CI tools

bin/format.js
```diff
@@ -0,0 +1,198 @@
...
+    changedFilesTotal = changedFilesTotal.concat(changedFiles);
+
+    console.log(`\x1b[30m\x1b[43mRULE ${i}\x1b[0m${indent}Made \x1b[33m${changeCount.toString().padStart(3)}\x1b[0m changes to \x1b[33m${changedFiles.length.toString().padStart(3)}\x1b[0m articles`);
+};
+
+const changedFilesCount = changedFilesTotal.filter((f, i, files) => files.indexOf(f) === i).length;
+
+console.log(`\x1b[30m\x1b[42mDONE!\x1b[0m    Made \x1b[32m${changeCountTotal.toString().padStart(3)}\x1b[0m changes to \x1b[32m${changedFilesCount.toString().padStart(3)} / ${articles.length}\x1b[0m articles`);
+
+function rule_1_boundaryWhitesapce(content, meta, log) {
```
sr229: document what this function does, because definition name does not exactly state what the function is supposed to do.

bin/format.js
```diff
@@ -0,0 +1,198 @@
...
+        content = content + '\n';
+    } else if (trailingWhitespaceMatch[0] !== '\n') {
+        log(lineAtIndex(original, trailingWhitespaceMatch.index), 'Whitespace at end of file');
+        content = content.substring(0, content.lastIndexOf(trailingWhitespaceMatch[0])) + '\n';
+    }
+
+    return [content, meta];
+}
+
+function rule_2_atxHeaders(content, meta, log) {
```
sr229: document what this function does as well.

bin/format.js
```diff
@@ -0,0 +1,198 @@
...
+            return match;
+
+        log(lineAtIndex(content, index), 'Setext header used (atx is preferred)');
+        return (l1 !== undefined ? '# ' : '## ') + header.trim();
+    });
+
+    return [content, meta];
+}
+
+function rule_3_noTabs(content, meta, log) {
```
sr229: document this line as well

bin/format.js
```diff
@@ -0,0 +1,198 @@
...
+
+    const filteredContent = content.replace(/Ｆａｓｔ　＆　Ｃｈｉｌｌ 早く/g, '');
+    const nonAsciiPattern = /[^\S\n\r\t ]+/g;
+    while (match = nonAsciiPattern.exec(filteredContent))
+        log(lineAtIndex(filteredContent, match.index), 'Non-ASCII whitespace is not allowed');
+
+    return [content, meta];
+}
+
+function rule_4_smartCharacters(content, meta, log) {
```
sr229: document this as well.

bin/format.js
```diff
@@ -0,0 +1,198 @@
...
+let changeCountTotal = 0;
+let changedFilesTotal = [];
+
+for (let i = 0; i < rules.length; i++) {
+    if (changes[i] === undefined)
+        continue;
+
+    const changeCount = Object.values(changes[i]).reduce((sum, c) => sum + c, 0);
+    const changedFiles = Object.entries(changes[i]).filter(c => c[1] > 0).map(c => c[0]);
+    const indent = ' '.repeat(4 - i.toString().length);
```
sr229: This variable is only declared once, just put it at the declaration it represents.

cl8n:
next version of this will be documented properly @sr229 
i dont think anyone is using this so i will close, i'll post here with the next project sometime

PR #2294, opened 2019-06-01
BN Rules: Minor fixes

Naxesss:
Stating that a veto is a veto was considered common sense, but it probably shouldn't be, so we're adding that. Also changed around wording of another rule which had a spelling issue.

edit: also standardizes "beatmapset" for consistency with the ranking criteria


wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -47,7 +47,7 @@ Violations of those rules may be reported to a member of the [Nomination Assessm
...
-+ **Do not veto for unrankables.** Unrankables in themsleves mean that no other nominator should nominate and that the mapper has no other choice but to fix it, in the same way an upheld veto works. Because of this, disqualifying for unrankables often requires a second person to confirm the issue.
++ **You can veto a nominated/qualified beatmap set by posting the issues you think need to be addressed before reaching Ranked status in the form of problems in the respective modding discussion.** This will reset all nominations the beatmap set has received and thereby stop it from proceeding to Ranked. Other nominators will be unable to nominate it again until the veto is lifted. The veto post must clearly state that it is a veto and include a clear problem and reasoning for stopping the ranking process of the beatmap set. You must participate in the resulting discussion and attempt to reach a mutual agreement.
```
cl8n:
```suggestion
+ **You can veto a nominated/qualified beatmapset by posting the issues you think need to be addressed before reaching Ranked status in the form of problems in the respective modding discussion.** This will reset all nominations the beatmapset has received and thereby stop it from proceeding to Ranked. Other nominators will be unable to nominate it again until the veto is lifted. The veto post must clearly state that it is a veto and include a clear problem and reasoning for stopping the ranking process of the beatmapset. You must participate in the resulting discussion and attempt to reach a mutual agreement.
```
beatmap set => beatmapset as decided in RC a while ago

Naxesss:
@cl8n Vetoes don't differ much from regular disqualifications apart from the signal that the user can argue against it. Having a separate label might work if users learn to recognize it and/or some text appears above/below it saying what a veto means, but considering that nominators currently need to make "suggestions" only to say "[Problem Stamp]" on the post when reporting it, I'd rather leave that for later when things are more fleshed out.

The veto system in general is likely to keep evolving as we go, but since it's all managed by NAT who currently have little influence on what is being worked on development-side, anything implemented may end up being detrimental if we have no means to change it when needed. So I'd rather refrain from having that engraved in stone before we know for sure that this works well.

cl8n:
there's an issue open for that too! if you want to have influence in development, the best way is to let the developers know what is needed =)

Naxesss:
fwiw there are issues open for basically everything we need, some dating back to 2017

Joehuu:
why was the space added back on `beatmapsets`?

cl8n:
mrw checked backwards

Naxesss:
oh it's supposed to be the other way around lmao, sec

PR #2309, opened 2019-06-17
Rules update: disallow BSS content from restricted users

Ephemeralis:
This change brings beatmap content in line with the rest of the established 'read only' practices regarding account restriction as unobtrusively as possible.

---
Content provided by restricted users is no longer allowed in any BSS submissions. Sets uploaded before restriction may remain, but cannot be Ranked, Loved, or altered in any way.

### DIFF
--- a/wiki/Rules/en.md
+++ b/wiki/Rules/en.md
@@ -42,6 +42,7 @@ A set of more detailed rules for beatmap construction exists, known as the [Rank
 1. **Make sure you have permission to use any content involved in your beatmap.** This includes songs, videos, hit sounds, graphics, and any other content that isn't your own creation. If you are unsure of where to find content that is free to use, we have a growing library of tracks available at the [Featured Artist](https://osu.ppy.sh/beatmaps/artists/) listing, all 100% licensed for use in osu!.
 2.  **Stay on topic in Beatmap discussions.** These threads/discussions are about the beatmap in question, and nothing else. If you have an issue with something not directly pertaining to the beatmap, post it in the appropriate forum.
 3.  **Do not plagiarise or attempt to steal the work of others.** Do not also upload or use other people's work without their explicit permission (including, but not limited to, skins and guest difficulties).
+4.  **Beatmaps must not contain any content attributed to or contributed by any restricted users.** This includes graphics, audio, beatmap data and storyboard sequencing data. Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way, and cannot be Ranked or Loved until their creator returns to the community in good standing.
 
 ## Off-Topic Rules
 
--- a/wiki/Rules/en.md
+++ b/wiki/Rules/en.md
@@ -42,7 +42,7 @@ A set of more detailed rules for beatmap construction exists, known as the [Rank
 1. **Make sure you have permission to use any content involved in your beatmap.** This includes songs, videos, hit sounds, graphics, and any other content that isn't your own creation. If you are unsure of where to find content that is free to use, we have a growing library of tracks available at the [Featured Artist](https://osu.ppy.sh/beatmaps/artists/) listing, all 100% licensed for use in osu!.
 2.  **Stay on topic in Beatmap discussions.** These threads/discussions are about the beatmap in question, and nothing else. If you have an issue with something not directly pertaining to the beatmap, post it in the appropriate forum.
 3.  **Do not plagiarise or attempt to steal the work of others.** Do not also upload or use other people's work without their explicit permission (including, but not limited to, skins and guest difficulties).
-4.  **Beatmaps must not contain any content attributed to or contributed by any restricted users.** This includes graphics, audio, beatmap data and storyboard sequencing data. Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way, and cannot be Ranked or Loved until their creator returns to the community in good standing.
+4.  **Beatmaps must not contain any content attributed to or contributed by any restricted users.** This includes graphics, audio, beatmap data, and storyboard sequencing data. Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way, and cannot be Ranked or Loved until their creator returns to the community in good standing.
 
 ## Off-Topic Rules
 


wiki/Rules/en.md
```diff
@@ -42,6 +42,7 @@ A set of more detailed rules for beatmap construction exists, known as the [Rank
...
 2.  **Stay on topic in Beatmap discussions.** These threads/discussions are about the beatmap in question, and nothing else. If you have an issue with something not directly pertaining to the beatmap, post it in the appropriate forum.
 3.  **Do not plagiarise or attempt to steal the work of others.** Do not also upload or use other people's work without their explicit permission (including, but not limited to, skins and guest difficulties).
+4.  **Beatmaps must not contain any content attributed to or contributed by any restricted users.** This includes graphics, audio, beatmap data and storyboard sequencing data. Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way, and cannot be Ranked or Loved until their creator returns to the community in good standing.
```
Joehuu:
```suggestion
4.  **Beatmaps must not contain any content attributed to or contributed by any restricted users.** This includes graphics, audio, beatmap data, and storyboard sequencing data. Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way, and cannot be Ranked or Loved until their creator returns to the community in good standing.
```

RushFTK:
links about discussion on osu! bbs?

Ephemeralis:
We generally don't discuss anything involving restrictions or restriction-related policy. There's largely never been a need to.

RushFTK:
this rule is aim to fix some leaks on rules about  restriction-related policy. at least a public discuss is needed  in case leaks are still exist.

Ephemeralis:
Okay.

Restrictions are designed to apply a 'read only' state onto users affected by them. It hides their scores, it prevents them from interacting with the community at large, it generally makes them a 'shadow' until they successfully appeal and place themselves as ready to interact with the community again.

This generally works fine, except when mappers are restricted. There is currently no direct ruling on whether a restricted mapper can create guest difficulty for other beatmapsets which are being handled by non-restricted users. In relative abundance, some restricted mappers have continued to proliferate their works via other users placing them on the BSS for them. This is tantamount in mechanics to a non-restricted user participating in chat in lieu of a restricted user, which would be met extremely poorly under the approach we currently use.

Anecdotally, many mappers consider restriction a less serious offense for them in light of this. Given a cooperating friend, they may continue to contribute in direct defiance of the punishment placed upon them. 

Some have even used their restriction as an in-circle method of garnering views and attention for their maps. Given the way that current RC credit rules work, restricted users must be credited for their work as would be appropriate. This can give the impression to other users that said user is not actually under any sort of administrative sanction - ergo, it makes it seem as if they are not being punished. Which in actuality, they really aren't, given that a significant amount of the social interaction involved in mapping often happens outside of the game platform itself.

Allowing this 'loophole' to continue to be abused widely dis-incentivizes mappers to abide by the community rules if they are already inclined to be oppositional or defiant. They may continue to contribute via users without penalty beyond what restriction classically grants. It essentially provides them with a way to bypass the most important punitive measure of restriction for their particular social context.

The proposed rule addition cuts off that loophole - on paper. In practice, it will likely force restricted users to forego appropriate credit for their works and contribute via 'ghostmapping', at risk to both the restricted party and the beatmapset owner hosting the set. This creates considerable incentive for the restricted person to appropriately pursue appeal and restore their account standing, as is the **entire point of restriction in the first place.**

tldr; this rule change makes restriction appropriately threatening for mappers, where as at the moment, it currently isn't that big of a deal if you're suitably socially connected. it also brings behavior in line with other restriction penalties as per the concept we've been pushing for the past three and a half years.

wiki/Rules/en.md
```diff
@@ -42,6 +42,7 @@ A set of more detailed rules for beatmap construction exists, known as the [Rank
...
 2.  **Stay on topic in Beatmap discussions.** These threads/discussions are about the beatmap in question, and nothing else. If you have an issue with something not directly pertaining to the beatmap, post it in the appropriate forum.
 3.  **Do not plagiarise or attempt to steal the work of others.** Do not also upload or use other people's work without their explicit permission (including, but not limited to, skins and guest difficulties).
+4.  **Beatmaps must not contain any content attributed to or contributed by any restricted users.** This includes graphics, audio, beatmap data, and storyboard sequencing data. Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way, and cannot be Ranked or Loved until their creator returns to the community in good standing.
```
dorothy3242: Okay, this current wording *sucks*.

> Existing sets uploaded before a user's restriction will be allowed to remain as is, but **may not be updated or altered in any way**, and cannot be Ranked or Loved until their creator returns to the community in good standing.

Disallowing any kind of updates means that mapsets with guest difficulties or etc made by a restricted user can not be updated to remove the restricted user's content and move on to rank. 

in a simplified scenario:
person A uploads a map with a gd by person Q
person Q gets restricted.
person A can't update their map, because it has content by a restricted user, and so their set **may not be updated or altered in any way**. This means they can't update the set to remove the difficulty by person Q and then rank their own map. The mapset becomes stuck.
 
This would be punishing mappers that aren't restricted, force-stopping their mapsets because of something that is not in their control. The only way around that flaw would be to *break the rule*, potentially putting what were innocent mappers in trouble for no good reason.


Suggestion:
> Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way unless it is to remove the restricted user's content, and cannot be Ranked or Loved until their creator returns to the community in good standing.

Ephemeralis: The 'existing sets' clause is supposed to refer to the restricted user's own beatmaps. Any normal standing community member could simply remove the restricted user's GD from their set and proceed with ranked like normal.

I'll think of a way of rewording that to be more clear.

peppy:
How I would have worded this:

```
We have noticed that a number of mappers who have been placed in restricted status have been ignoring their restriction, turning to submitting guest difficulties as a way of interacting with the community. While it's great to see these users contributing positively to the game, we do hope that they follow correct process to restore their account status before doing so.

To ensure this happens, we are introducing a rule effectively disallowing "guest difficulty" submissions from restricted users. This will be applied to all new difficulties submitted (existing ones can still be ranked) from this day forward.
```

On further discussion @Ephemeralis has mentioned that this was not triggered by any particular user or case, but is an effort to patch a loophole in the system.

Still not sure I agree with this one (as it will create more tension, work for the team, and has easily-apparent loopholes). Feel free to continue discussion.

RushFTK:
thanks for taking so much time to explain that , Ephemeralis. Maybe i could understand why stuff want fix that, its resonable.
In my opinion, a user is restricted because he/she did a harmful things for osu! community and disallowed he/she do future more. so makes them a 'shadow' themself is not only a punishment for user self, but also a protect for osu! community.   
Before add this rule, i think the punishment is already enough that people cant do things as usual. if they really cant do anything ,it imply "pushing them away from osu! instead of make them realize their mistakes." if him/her map is aim for rank, or in other words, their map have enough quaility for rank. we can assumed their gds is a contributed for osu! , just make osu! more playable, instead of harmful osu! community. map host may also a grund to make sure if their gd is "not harmful." if a inappropriate gd uploded, the host should be punishment.  
so im fear add this rule will cut-off these player last way to connect with osu! community, push them away for osu! some of restricted player is bad and need cleared out, but somes are not, they just need a moderate punishment to let them realized fault. and its a little pity that them own fault let osu! lost some beautiful mapset , which can let player enjoyable.  

maybe let them cant edit their gd is ok, but if their gd is done and already in ranking process, neednt edited (patterns). let them ranked is acceptable.

> Some have even used their restriction as an in-circle method of garnering views and attention for their maps.   
im really confused why a restriction can be a way for garnering views and attention, instead of making lots of pp map and keep them alive. in fact a gd's name usually be ignored unless he/she is already famous, in latter case, restriction to garnering views is a more ridiculous things -- saying something junk word on twitter or make a inappropriate beatmap submission is a more useful and less harmful for they self.

Poliwrath:
There's a (one-sided) discussion about this happening on Twitter currently, and one reply was interesting:

>also, its not stopping anything, people will just upload without credit/ghostmap 

(https://twitter.com/sammishyy/status/1140632912842371072)

How are you going to combat this?

Ephemeralis:
> 
> 
> There's a (one-sided) discussion about this happening on Twitter currently, and one reply was interesting:
> 
> > also, its not stopping anything, people will just upload without credit/ghostmap
> 
> (https://twitter.com/sammishyy/status/1140632912842371072)
> 
> How are you going to combat this?

We can't.

It is my hope that the loss of being able to directly credit work to themselves plus making it potentially risky for set owners will encourage restricted mappers to seek appeal rather than attempting more clumsy ways to circumvent the issue at hand.

There's no need for a lot of these long-term restricted mappers to remain outside of the game. I feel as if some choose to do so as appealing is either too protracted or too difficult. I want people to be re-engaging officially and returning to the community, not existing in a fugue state where people are overlooking what is arguably a deliberate circumvention of the penalties that restriction imposes.

This is also being accompanied by a full review of most appeals made in the past few years in order to make sure we're not the ones responsible for the delay in attempting to return. Sometimes we are.

Ephemeralis:
We've decided that it is "for the best" that we don't consider pursuing this rule addition for the time being.

Yes, that is tacit endorsement to continue the current way of circumventing things.

I would prefer vastly if any long-banned mappers that would've been affected by this rule to shoot me an email at ephemeral@ppy.sh detailing their history as much as possible, ideally from an email associated with their account. I'll dedicate some time to resolving these cases as much as I can in lieu of them.

Kyuunex:
Hi, just wanna add my opinions.

The way I see restricted mappers creating guest difficulties is, the only way they can contribute to the game while restricted, as a way to redeem themselves (so they can brag they ranked something in their appeal email, prove their worth, or something).

I do completely understand your point with "read-only" users, I just wanted to add my opinion. In fact, I know one talented mapper who ranked a guest difficulty while being restricted. (he's unrestricted now tho)

Poliwrath:
# of GDs / maps ranked during restriction seems irrelevant though?

Shadowthegod:
Not sure if my opinion matters but while I agree that this should be the process going forward I disagree with this part: 

**Existing sets uploaded before a user's restriction will be allowed to remain as is, but may not be updated or altered in any way, and cannot be Ranked or Loved until their creator returns to the community in good standing.**

Why these maps needs to be stopped especially if we are talking about good quality ones meaning that they could go for ranked? In my opinion the set owner should just take over the difficulty, in worst case even re-name it, and proceed (they can do this already anyway so this rule is actually useless in this case). 

I understand the intentions behind this but you are trying to stop the unstoppable. These users, especially if they are well know, will be able to make their creations Ranked or even Loved behind the staff. In the long run maybe this will change but even then, this rule should only affect their future contributions and not the ones before their restriction.

insomnyawolf:
From my point of view, readonly state is not the best that you can do.
If the user is famous:
there will be other ways to contact them outide osu and make drama and shit (Reddit, Twitter, Youtube, DTwitch, Discord).
Ellse:
For the unknown users, noone will notice that they have  got banned, is just making them forcedly "offline" so they can't talk with any friend they have made if they can't contact them out of the game.
About Bss
I kinda agree with @Shadowthegod , but i dont see the point why something that may have good quality should be filtered away, isn't  that just damaging the end users?

I personally think that readonly ban state should be reworked and this rule should be done over the new "banned" status.
--Sorry for my bad english.

PR #2316, opened 2019-06-21
translations for osu!taiko difficulties

CtrlAltWaverick:
added translations of the osu!taiko difficulty names


TPGPL:
Thanks for your contributions, but I'm afraid we don't need this change in the article, sorry for trouble!

PR #2322, opened 2019-06-25
ASC: Updates

MegaApplePi:


wiki/Article_Styling_Criteria/en.md
```diff
@@ -48,39 +48,43 @@ Listed below are the properly-supported locales for the osu! wiki. The file name
...
 #### Tagging articles
 
-*Note: You should only need to include words that are not found in the article.*
-
-Tagging articles will help the website's search engine query articles better. It is unlikely that you will need to do this, but the markup is as follows:
+Tagging articles will help the website's search engine query articles better. Tags should use lowercase letters. It may be unlikely that this is needed for most articles; however, the markup is as follows:
 
 ```md
 ---
 tags:
```
sr229: Nit: I think since this is a front matter with a YAML syntax, we should have it as `yml` instead.

wiki/Article_Styling_Criteria/en.md
```diff
@@ -108,55 +114,51 @@ Folder names must not contain URL reserved characters. Generally, all folder nam
...
 
-Parts of the GFM (GitHub Flavored Markdown) markup are supported in the osu! wiki. GFM has a small amount of markup that anyone can learn in about 3 minutes. However, do not expect a lot out of GFM as it is intended to be used as a bare minimum markup language.
+Some parts of the GitHub Flavored Markdown markup are supported in the osu! wiki.
 
 ### HTML
 
 <!-- ppy ruled HTML is not allowed. from:ppy#0001 during:2017-06-26 in:osu-wiki markdown -->
 
-**HTML use is prohibited!** Resturcture the formatting if you need to use HTML.
+HTML must not be used, with exception (see section below). The structure of the article must be redone if HTML is used.
```
sr229: I think we should link the ASC section as-is.
```suggestion
using HTML is prohibited, with an exception for comments (see [comments](#comments)). The structure of the article must be redone if HTML is used.
```

wiki/Article_Styling_Criteria/en.md
MegaApplePi: Down below on line 87, I included a title to show that a space is required. Changing it to `yaml` will make that look like a comment instead. (keeping it as-is)

wiki/Article_Styling_Criteria/en.md
```diff
@@ -405,9 +430,9 @@ Whenever linking to a beatmap, use this format as the link text:
 {artist} - {title} ({creator}) [{difficuty_name}]
 ```
 
-When linking to a beatmap, make sure that the link actually links to that difficulty. All beatmap difficulty URLs looks like this: `https://osu.ppy.sh/beatmaps/{BeatmapID}`.
+The link to a beatmap must actually link to that difficulty. All beatmap difficulty URLs looks like this: `https://osu.ppy.sh/beatmaps/{BeatmapID}`. The other URL format may be used instead.
```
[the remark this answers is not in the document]
MegaApplePi: Should beatmap difficulty URLs be formatted like that? If so, `URLs should be` -> `URLs must be`.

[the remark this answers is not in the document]
Joehuu: Edited suggestion.

Joehuu:
I thought the line endings were CRLF.

MegaApplePi:
No, line endings of checked in files must be `LF`. Most discussions about this happened through Discord (ref: `in:#osu-wiki from: MegaApple_Pi#9995 line ending`).

MegaApplePi:
@Joehuu I forgot that git users on Windows will see CRLF, assuming default installation. I'm going to go with ``Articles must be checked in using the `LF` (line feed) end of line sequence.``.

Joehuu:
Doesn't the `.gitattributes` automate line endings, so we don't have to worry about some local setting?

MegaApplePi:
Yeah, that file will automate/normalize the line endings if you are using git. If you upload via GitHub, this process is skipped (that is, GitHub will commit the line endings as-is).

wiki/Article_Styling_Criteria/en.md
```diff
@@ -108,55 +114,59 @@ Folder names must not contain URL reserved characters. Generally, all folder nam
...
-HTML comments may be used for annotating text, but limit usage.
+HTML comments may be used for annotating text.
 
 ### Editing
 
-#### Escaping
+#### End of line sequence
 
-*Notice: The title of an article must not be escaped, as titles are parsed as plain text, not Markdown.*
+<!-- Parsing will break if it reads a file using CRLF. -->
```
cl8n: I don't think this is true, if you're referring to parsing on the website
not rly a problem though just a comment

MegaApplePi: "Parsing" is broad, I could have worded that better. Probably something like: `<!-- YAML will display in the article on the website if the file is using CRLF -->`?

Joehuu:
`Note: If you do not want to change your current settings for the osu! wiki, you can install a separate installation of osu! in another directory.` -> ``` Note: If you do not want to change your current settings for the osu! wiki, you can move your `osu!.<ComputerUser>.cfg` out of the osu! folder and move it back later. ```

Something along the lines.

Downloading multiple installations results in broken uninstall locations, beatmap files (`.osz`) opening in new installation, etc.

wiki/Article_Styling_Criteria/en.md
```diff
@@ -108,55 +114,59 @@ Folder names must not contain URL reserved characters. Generally, all folder nam
...
+Redirects must be updated to have the ambiguous keyword(s) redirect to the disambiguation article.
 
 ## Markdown
 
 All osu! wiki articles must adhere to the following rules.
 
 ### GitHub Flavored Markdown
 
-Parts of the GFM (GitHub Flavored Markdown) markup are supported in the osu! wiki. GFM has a small amount of markup that anyone can learn in about 3 minutes. However, do not expect a lot out of GFM as it is intended to be used as a bare minimum markup language.
+Some parts of the GitHub Flavored Markdown markup are supported in the osu! wiki.
```
sr229: Let's make it a bit more formal-sounding.
```suggestion
A subset of the GitHub Flavored Markdown syntax are supported in the osu! wiki.
```

sr229: You might as also link this [article](https://help.github.com/en/articles/about-writing-and-formatting-on-github) about GFM on "GitHub Flavored Markdown" as well to give contributors an idea what GFM is.

MegaApplePi: I forgot why I wrote that here. Does the website support any markup from GFM, other than the regular Markdown? A subset (or "some parts") of GFM seems vague or doesn't sound right if it is just Markdown.

I'll just remove the section instead, if there is no other discussion.

MegaApplePi: Also, I think that article should be linked in the `osu! wiki contribution guide` instead.

Joehuu: https://github.github.com/gfm/

Tables are not supported in regular markdown, maybe that's why.

sr229: Markdown tables are part of the Markdown syntax spec @Joehuu

sr229: > Also, I think that article should be linked in the `osu! wiki contribution guide` instead.

good wiki practice is that this should also be linked in the criteria page as well. It's better you link GFM's definition since one reference on one article is not enough.

Joehuu: GFM as an extension, yes. Commonmark or regular markdown, no.

cl8n: they aren't part of commonmark, which is the spec that osu! uses. osu! is also using other extensions and options that aren't standard though, so that's why these things are possible

anyway i think this note about GFM is a little pointless since any info it potentially gives is still meaningless (osu! has nothing to do with GFM, even if they happen to be similar). there's plenty of info about markdown in this article already

sr229: Yeah I think we should remove this altogether if that's the case.

PR #2340, opened 2019-07-04
Add links to active packs torrenting

glukki:
Old topic has no activity, and has a reference to a new one.
New topic is active, and there're seeders.


sr229:
CC @TPGPL to review anything else here that I may missed.

PR #2344, opened 2019-07-07
Rule Change: Allowing more than 6 Notes at once

Feerum:
As discussed at https://osu.ppy.sh/community/forums/topics/915024 

---

Rule Change for allowing to press more than 6 notes at once from Expert difficulties.

### DIFF
--- a/wiki/Ranking_Criteria/osu!mania/en.md
+++ b/wiki/Ranking_Criteria/osu!mania/en.md
@@ -54,7 +54,7 @@ All rules are exactly that: *rules*. They are *not* guidelines and may *not* be 
 - **The `Auto` mod must be able to achieve the full score (1,000,000) in all difficulties.**
 - **Two notes cannot be placed in a single column at the same timestamp.**
 - **No column can be left empty.** If you want to use fewer keys, change the number of keys in the Song Setup.
-- **There must not be more than 6 notes pressed at any given moment.** Most USB-connected keyboards cannot handle 7 simultaneous inputs at a time. Note that this rule does not apply to ends of long notes, as they are released, not pressed.
+- **There should not be more than 6 notes pressed at any given moment in Insane or lower difficulties.** Using more than 6 notes must also follow a reasonable spread to the next lower difficulty. This rule does not apply to ends of long notes, as they are released, not pressed. 
 - **Beatmaps can only use 4 to 9 keys.** Anything else is not supported for the Ranked section. 
 - **Each key mode in a beatmap set requires its own spread.** For example, a 4 key beatmap would require its own Normal/Hard/Insane in order to create a proper spread, independent of any other key modes present in the beatmap set.
 - **If multiple key modes are present in a single beatmap set, the key mode must be denoted in the difficulty name.**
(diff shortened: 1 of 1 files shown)


wiki/Ranking_Criteria/osu!mania/en.md
```diff
@@ -54,7 +54,7 @@ All rules are exactly that: *rules*. They are *not* guidelines and may *not* be
 - **The `Auto` mod must be able to achieve the full score (1,000,000) in all difficulties.**
 - **Two notes cannot be placed in a single column at the same timestamp.**
 - **No column can be left empty.** If you want to use fewer keys, change the number of keys in the Song Setup.
-- **There must not be more than 6 notes pressed at any given moment.** Most USB-connected keyboards cannot handle 7 simultaneous inputs at a time. Note that this rule does not apply to ends of long notes, as they are released, not pressed.
+- **There should not be more than 6 notes pressed at any given moment in Insane or lower difficulties.** Using more than 6 notes must also follow a reasonable spread to the next lower difficulty. This rule does not apply to ends of long notes, as they are released, not pressed.
```
Hydria: "should" should still be "must" as it remains as a rule for insane and lower diffs

Ephemeralis:
despite the connotations this poses towards improving top-end osu!mania beatmapping, something about the idea of having a soft hardware requirement for progression in a gamemode feels fundamentally at odds with the whole free-to-win aspect of osu! and is probably a major factor as to why the rule is how it is in the first place

have any stats been pulled on what the average KRO is on standard and low-end hardware?

peppy:
@Ephemeralis discussion is in the linked forum thread.

RushFTK:
so player should give up use USB Keyboard before they become Expert

Feerum:
@peppy Setting up a poll with a sample beatmap is not a problem! Can be done really quick. Seems also to be a good idea. The first idea of the Rule was it to have this allowed from a certain difficulty. 
So we can setup an Expert or higher Difficulty 7K Beamap with intentional 7-Noted chords and see what the people think/see if they can play it.

Additional i heard from several players that such pattern got already used in maps for the 7K osu!mania World Cup and the participants didn't had problems playing them.

wiki/Ranking_Criteria/osu!mania/en.md
cl8n: or move to Guidelines if that wasn't the intention^

also remove the trailing space

Feerum:
Little update: Poll created and can be found here: https://osu.ppy.sh/community/forums/topics/936996

wiki/Ranking_Criteria/osu!mania/en.md
Feerum: Applied. Sorry that it took so long. I am too scared to break something unintentionally lmao

Feerum:
@peppy The poll ended and the result can be found here. https://osu.ppy.sh/community/forums/topics/936996

Would like to hear from you if it would be okay to continue with this Rule change now or if you need anything more. According to the Poll, almost 90% of the people were able to play the beatmap which had added 7-Noted chords.

PR #2361, opened 2019-07-15
BN/NAT updates

cl8n:
closes #2180 

- move People/QAT to People/NAT and update article accordingly
- move People/QAT/Blog to QAT_Blog
- fix references to Quality Assurance Team and links going to Quality_Assurance_Team
- fix references to QAT and links going to QAT
- make glossary entry for QAT and point historical links towards that instead of NAT
- redirect old QAT pages to glossary, redirect QAT blog to top-level page
- add some short NAT redirects
- outdate a lot of things that were previously unmarked by mistake (did not update qat/nat references in these articles)

PR #2366, opened 2019-07-16
Move and update "Changing the Title Text"

### LINKED ISSUE #1223: "Changing Title Text" is incomplete and partially incorrect
https://osu.ppy.sh/help/wiki/Guides/Changing_the_Title_Text

among other things:
- `wait` isn't listed as an option at all
- the defaults aren't all correct
- `time` seems to increment by some amount for each line independent of your setting for it
- `hold` doesn't include the transitions (or at least not the fade-out...)

the format of this whole article is a little strange and probably needs to be re-written anyway

only an osu!dev will be able to document some of these things, because the only way to test them in-game is to bug a QAT about it and have them update maps on your behalf. would be helpful if a dev can provide the relevant code so we can document it correctly

cl8n:
closes #1223
opted to make an article about title text, instead of a guide describing how to change it. though most of the article is still one section dedicated to that

---


cl8n:
depends on #2361 for a link to NAT

Joehuu:
Image doesn't comply with ASC. See https://osu.ppy.sh/help/wiki/Article_Styling_Criteria#images under `Screenshots of gameplay` for settings (resolution, fps counter, etc.).

PR #2369, opened 2019-07-17
Clarification to BN Vetoes in BN Rules

ascendancelol:
allows BNs to veto mapsets of modes they are not qualified for when the veto is related to audio, backgrounds, or other content unrelated to the mapset's game mode


JBHyperion:
Makes logical sense that general issues such as those described above can be raised / addressed by all Nominators / NATs, considering these typically don't require detailed knowledge of specific game mode(s).

PR #2373, opened 2019-07-22
fix grammar issue

aspriddell:
fixes a grammar issue with personal pronouns


peppy:
Not sure about this one, arguable. The original was meant to read a bit weird.

cl8n:
either works, I kinda like the way original version sounds

peppy:
I also have to say I prefer the original version. Let's leave it for now.

PR #2375, opened 2019-07-23
Fix wrong suggested skin sizes of osu!'s hit circles

### LINKED ISSUE #2370: Skinning wiki suggested sizes are incorrect
https://github.com/ppy/osu-wiki/blob/master/wiki/Skinning/osu!/en.md

Hitcircle template on page is 128 but listed as 118. Many other cases of this.

Joehuu:
Fixes #2370.

---


peppy:
Thanks!

PR #2377, opened 2019-07-24
BN Rules: Remove the need to state that a veto is a veto

Naxesss:
ok so after observing how vetoes are done and people responding to them, along with making one, I don't think stating that a veto is a veto is necessary and more harms the process than benefits it

for context I changed this rule a while ago here https://github.com/ppy/osu-wiki/pull/2294 and have since changed my mind; the change was mainly to make sure people didn't mistake vetoes for unrankables, but in reality many issues people have accepted as unrankables aren't and need to be vetoed for, like snapping mistakes and spread issues

if they believe it could be part of the ranking criteria (mistaking a veto for an unrankable), it probably makes sense to them regardless and so highlighting that it can be argued against or mediated is unnecessary

look at it this way:
- if the mapper **agrees** with the reasoning, it doesn't matter
- if the mapper **disagrees** with the reasoning, you'd want to hear their thoughts on it
    - if it's **unrankable**, perhaps the rule needs to be changed and something was overlooked?
    - if it's **a veto**, you'd want to discuss the issue and hear their side before making it seem like their only choice is mediating it (which is what stating that it's a veto generally does)
- if a nominator wants to nominate a popped/dqed map, they'd know whether they can based on if the pop/dq is properly resolved, which they'd have to check anyway

**tl;dr:** stating that it's a veto only makes the process worse if the mapper disagrees, which is the only case it's there for, since it makes the vetoer seem unwilling to hear the mapper's thoughts before fully deciding what to do


UndeadCapulet:
fully support this, i've been protesting this since day 1 ^^' just makes mappers overly upset for no real benefit

PR #2394, opened 2019-08-02
Fix format of skinning sounds wiki article

Joehuu:


Joehuu:
The beatmap skinnable tables were removed as it was too repetitive and somewhat also existed on interface sounds showing always false. Only gameplay ones are, so adding a sentence on that section will simplify everything.

sr229:
> The beatmap skinnable tables were removed as it was too repetitive and somewhat also existed on interface sounds showing always false. Only gameplay ones are, so adding a sentence on that section will simplify everything.

Makes sense, I do agree there's no need for the tables when i read it again. Good catch.

wiki/Skinning/Sounds/en.md
```diff
@@ -1,17 +1,12 @@
-[true]: /wiki/shared/true.png
-[false]: /wiki/shared/false.png
-
 # Sounds
 
 `.wav`, `.mp3`, and/or `.ogg` are valid formats for sounds.
 
-## Main Menu
+## Interface
```
Joehuu: This is a test.
```suggestion
## Interfaces
```

PR #2407, opened 2019-08-05
Staff Log: July 2019

MegaApplePi:


MegaApplePi:
@cl8n Not WIP but I want to prevent merging because I am waiting for the next post on the forum thread. I am doing this on the assumption that the next month's (August) post concludes/finalizes the previous (July) month. But yeah, from what @Joehuu said, it would be great if they moved from there to here.

---

Or maybe have osu-web automatically log the changes and have a specific page to view it on. (random idea)

WalterToro:
> @cl8n Not WIP but I want to prevent merging because I am waiting for the next post on the forum thread. I am doing this on the assumption that the next month's (August) post concludes/finalizes the previous (July) month. But yeah, from what @Joehuu said, it would be great if they moved from there to here.

You can go with this by the end of each month, even if BNs takes most of the activity and regularly happens, we shouldnt go with the assumption that every month has changes.

> Or maybe have osu-web automatically log the changes and have a specific page to view it on. (random idea)

I like this idea, is something that is done on Aiess so I can see this happening on web, still that will remove the reasons why someone is removed of a group (that could be optional).
Still, this is something to propose on osu-web, and not here.

PR #2413, opened 2019-08-10
Enforce in-game screenshot feature on ASC

Joehuu:
For consistency reasons. Also I feel people will use third party apps if this is not clearly stated.


MegaApplePi:
1d8c5f882dccee5f53d9a3d316512836d5eb9311?

Joehuu:
It defaults to `Enabled` with a clean cfg with no yellow indicator. After restart, it does show. Probably a bug and doesn't happen on lazer, so I'll probably revert.

Issue #2247, opened 2019-05-11
Implementation of grouping news posts


Joehuu:
Might be a way to shorten the [news directory](https://github.com/ppy/osu-wiki/tree/master/news) significantly as not many diverse news posts have been made yet.

The news directory will have folders for the type of news:
```
/news
	/osu!weekly
		/2017-06-07-osu-weekly-106.md
	/Project_Loved
		/2019-05-06-project-loved-week-of-may-5th.md
```
There can be links to previous and next iterations like osu!weekly had done, but it is done automatically.

cl8n:
I feel like having the order of these posts implicitly defined by their file name isn't the best option... but for all the news posts currently I think it'd work. so idk . I will try this in a few weeks

sr229:
Categorization may help us in the near future as well, so I think this directory layout may be a good start for a cleaner news layout.

cl8n:
will add an `order` attribute to each post to help with sorting. it could also be a doubly-linked list but that's probably kinda annoying for news post writers since they have to update the last post when they make a new one

cl8n:
ok I think I'm dumb I can just use the date for sorting. I can't imagine a case where the "next" post would be posted before the current one 🤔

cl8n:
something to note, osu!weekly news posts have tables linking to previous/next posts within the articles that should be removed when this is done

peppy:
The osu-web design has categorisation done via tags. Adding support will require some degree of implementation at osu-web.

TicClick:
update: after https://github.com/ppy/osu-web/pull/9280 is merged, news posts will be grouped by years -- https://github.com/ppy/osu-wiki/pull/8027

Joehuu:
Closing as the original issue was resolved by https://github.com/ppy/osu-wiki/pull/8027. Having the directories by year was way better than series as posts from a series can realistically reach 1000+.

I guess the series grouping can be done via tags, will track that in the search implementation: https://github.com/ppy/osu-web/issues/2305.

PR #2432, opened 2019-08-15
Remove QAT Blog wiki article

Joehuu:
This article just lists the news articles with a bit of background and nothing else.

When https://github.com/ppy/osu-web/issues/2305 and https://github.com/ppy/osu-wiki/issues/2247 are implemented, lists like this totally shouldn't exist anymore.


cl8n:
hmm... I don't think I agree with this one, the history is important enough to warrant some mention

while news posts may be "grouped" on the site in the sense that you can go to previous/next posts, there's no design (yet) for news categories, so you still won't be able to view a listing of all QAT Blog posts there

maybe at some point the list can be removed and the content moved to History of osu!, but for now I think all of this is relevant

peppy:
Also disagree with throwing this away. It provides a good summary of the QAT.

If anything, the page could be renamed at some point?

Joehuu:
Alright, closing. Will just leave this as is if there are no suggestions.

PR #2452, opened 2019-08-25
[RU] Visual Settings

vp2woon:


cl8n:
a little strange that the screenshot is in english but I guess that's a problem for another day since all interface screenshots are english

PR #2499, opened 2019-09-07
Update for NAT / BN / Staff Log pages

dorothy3242:
Updates Staff Log, BN page, and NAT page accordingly for the new NAT addition.


cl8n:
can u also add rhytoly to probation nominators

PR #2501, opened 2019-09-07
Fix links to osu!academy article

Joehuu:


cl8n:
should add a redirect if this page was moved at some point

PR #2508, opened 2019-09-08
Update groups lists and standardize formatting

### LINKED ISSUE #2492: Update user group lists
I think alumni and GMT are the only outdated ones left. need to get these up to speed so that people who manage staff log can update them regularly

cl8n:
closes #2492


cl8n:
is the logic there that cells need spaces on either side of content (and content is 0 chars ) ? not sure how useful that is but alright

Joehuu:
You can add content in the empty cells easily without pressing space.

cl8n:
that's got to be the laziest thing I've ever heard but I love it

PR #2525, opened 2019-09-11
Revise Registration

TPGPL:
- fixes formatting of up-to-date translations
- deletes unusable translations

---


cl8n:
changes like e14c658e059f87c79156e8eb6e69f677f4cc1e91 probably aren't useful if we're going to handle this automatically at some point anyway

PR #2440, opened 2019-08-19
Update main page to work with new wiki stuff

LiquidPL:
ppy/osu-web#4809 implements new main page stuff, and the source pages need to be updated for it.

In short the main page is just in Markdown instead of YAML.

Closes #1209. Resolves #1652.

PR #2533, opened 2019-09-13
[FR] Main_Page

DoryanR:


cl8n:
this will have to be changed when the new main page is finished in #2440 . also it looks like a lot of the text is still in English here, is that intentional?

DoryanR:
I leave words in English because I think their translation would make it bad

PR #2550, opened 2019-09-14
Revise osu!supporter folder

TPGPL:


cl8n:
request changes (if necessary) to #2549 instead of changing fr here

TPGPL:
will just revert it to normal state

PR #2576, opened 2019-09-17
[FR] osu!taiko

almaloste:


almaloste:
I copied it from english version yes.

cl8n:
this article is a nightmare in EN too so I just fixed very obvious formatting problems and left the rest as is

PR #2595, opened 2019-09-20
Delete broken Options translations

TPGPL:
- they are broken
- they are full of html and broken images
- outdated

---


cl8n:
ES/ZH can be deleted but I'm not sure about the others. there's a lot of content that is still up-to-date there, and pretty much the entire article can be easily converted to markdown (the few parts that can't are just colspan tables). we have an archive of images too

TPGPL:
will make a new PR

PR #2599, opened 2019-09-23
Add osu!mapping

### LINKED ISSUE #2588: pishifat's mapping videos deserve a wiki page
pishifat's videos are famous in mapping due to the number of topics they cover, and function much the same way as osu!academy mapping related videos did once upon a time. We could make a wiki page about these in a similar format as the osu!academy's listing, making the videos even more accessible to new mappers :)

cl8n:
closes #2588


wiki/osu!mapping/en.md
```diff
@@ -0,0 +1,162 @@
...
+
+<!-- TODO: this isn't really concrete history -->
+The series focused on this tutorial-style content at first, but started branching out a few months in to include discussions about the state of mapping and modding, personal/opinionated commentaries, and interviews. There have also been some research projects presented in the form of osu!mapping videos, most notably [a study of mappers' influences](https://www.youtube.com/watch?v=TigwDc5iOcw).
+
+## Episodes
```
MegaApplePi: Do we want to list out the episodes, not link to a playlist?

cl8n: it's not going to be updated very often so I think this case may be okay

also there's not currently a playlist of these videos (besides Uploads, but that includes his other videos too that don't have anything to do with osu)

cl8n:
na I think it's fine for now. maybe I'll write more the next time I have to update this article

PR #2605, opened 2019-09-25
Update en.md

IcicleFall:
`| Int | Unknown Int, always seems to be 4 |` is the first instance of `| Int | Size in bytes of the beatmap entry |` in the `### Beatmap Information` section directly underneath it.


cl8n:
turns out this was something else and nobody got around to documenting it previously. will be fixed in #2607

PR #2635, opened 2019-10-04
Update song select filter list

bdach:
Add info about the two new song select filters: `artist=` and `creator=` (added [in stable in 20190924.2](https://osu.ppy.sh/home/changelog/stable40/20190924.2) and [in lazer in 2019.925.0](https://osu.ppy.sh/home/changelog/lazer/2019.925.0)) to the *Interface* article.

Additionally rephrased the introductory sentence for the list, since I couldn't parse it correctly as written. I added the new ones first because it seemed more natural looking at how the song select screen is structured (artist/creator is generally closest to the top, AR/CS/other metadata is lower), but I'm not going to insist on that if there are other ideas.


cl8n:
im fixinf formatting

PR #2649, opened 2019-10-06
[PT-BR] Add Main Page

pimpG:
I preferred not to translate some things because the Brazilian community is already familiar with them.


cl8n:
this is going to be outdated after #2625 is finished, so let's wait until that is done

MegaApplePi:
@pimpG Can you update this again to match the current one?

PR #2654, opened 2019-10-06
Update en.md

pimpG:
removed duplicate: Support Team... also known as the "Support Team"


TPGPL:
i think it's correct as `account support team` is also known as `support team`

pimpG:
support team is a separate team so i don't think this article should support the misleading information.

cl8n:
it's not misleading, they're both referred to as "support team" in various parts of the game. this part of the article clarifies that we're talking about the tech support team, not the account support team, despite them having the same name

PR #2655, opened 2019-10-06
Remove Off-Topic forum rules from Rules

### LINKED ISSUE #2644: Remove Off-Topic forum rules from Rules
feels really out of place with the serious tone of Rules article, and they're pinned to the forum already

cl8n:
resolves #2644

rationale being they're already pinned to the forum and I don't see why this forum's rules are listed here but no other forums' are. marked some translations outdated along the way (they were outdated prior to this PR)


cl8n:
skipped pt-br because there's a different PR open where we can change this

TicClick:
I can almost hear Ephemeral going on about this in the background

Ephemeralis:
as dead as Off-Topic is these days, it was once an important part of the community's shitposting history and I would be loathe to lose that history by scrubbing any mention from it of the rules

cl8n:
I don't think it's "losing history" when this content is already duplicated from somewhere else where it belongs 

would be nice if OT were mentioned on the wiki somewhere, but I don't think this is the place to do it. if there's history to write about, it could be its own article?

cl8n:
@Ephemeralis beep

cl8n:
@Ephemeralis :b:lease

TicClick:
@Ephemeralis come out to play-ay

Ephemeralis:
i'm supremely reluctant to delete history. what's achieved by removing these lines exactly?

cl8n:
https://github.com/ppy/osu-wiki/pull/2655#issuecomment-539066342 for ur first sentence and https://github.com/ppy/osu-wiki/pull/2655#issue-325080906 for ur second

TicClick:
> i'm supremely reluctant to delete history. what's achieved by removing these lines exactly?

Because, despite having a historical value, it's still _rules_.

There is a couple of articles and pages that have to be official, or at least appear to be suchs. While one may tell the readers to [go and screw themselves](https://osu.ppy.sh/home/news/2019-11-07-new-featured-artist-ricky-montgomery) in an official announcement (although this is also often frowned upon, at least by yours untruly), they wouldn't (hopefully) put crude jokes on legal matter pages and/or Ranking Criteria, would they?

Ephemeralis:
those rules are still in effect on the forum to this day afaik, even if they are (somewhat redundantly) supposedly repeated in forum posts within it

also all of them are salient rules and not at all jokes, save for the anime girl inclusion in birthday threads (which is tradition more than anything)

cl8n:
they are still in effect and not all of them are jokes. that doesn't address any of the reasons for removing them

this isn't the place for forum-specific stuff, you've got pinned threads for that (and you're already using one). you don't see any other forums' rules here cuz it makes no sense

Ephemeralis:
move rule #3 from off-topic rules to to forum-wide rules in that case, for starters, since it is a forum-wide rule

i'll concede on everything else because nobody else seems to give a fuck, but i'd still like to lodge that i don't really approve of demolishing tradition for the sake of 'completeness', remember that we're not a faceless corporate enterprise, and it's ok for "official" documentation to sometimes be a little bit quirky

cl8n:
did u not see #2804 or the part where I repeated many times that equivalent stuff exists pinned to the top of the forum (where more people who care about it will see it) :neutral_face: 

I don't think anyone but u is interpreting this like "demolishing tradition" or "[becoming a] faceless corporate enterprise". would appreciate if u dont put words in my mouth cuz I am strongly against both of those things

Ephemeralis:
yes but pinned to the top off an offtopic forum vs encoded in the official rules documentation for a given community are two extremely different things in terms and reach and scope

off-topic was a much bigger thing when these rules were written and even though it has since faded, its official omission from the major rules document signals a departure of a tradition that has persisted for over 6 years

for reasons that boil down to "it doesn't belong here" and "it's in a document that should *appear* official", pretty much. it's hardly worth arguing over, but it is what it is.

PR #2656, opened 2019-10-07
Additional Main Page style fixes

LiquidPL:
As requested by wiki contributors when we were initially merging the main page.

Requires:

- [ ] ppy/osu-web#5073


sr229:
We should be blocking this since #2625 is still out there unmerged and it would be nice to be fixed post-merge instead of pre-merge.

cl8n:
as long as this goes in around the same time as web part being deployed it doesn't matter

wiki/Main_Page/en.md
```diff
@@ -3,9 +3,11 @@ layout: "main_page"
 ---
 
 <div class="wiki-main-page__blurb">
+
```
cl8n: the other markdown-in-html parts use 3 newlines at the start, is that needed here?

LiquidPL: Nah, you could just remove all those newlines.

LiquidPL: i mean those under the blurb ofc

cl8n: I'm asking about something like line 13-16 cuz it has 3 newlines there but you only use 2 here

LiquidPL: How I understand it is that it didn't matter here since it was separating two html elements. You would have to ensure there's enough newlines here if there were any markdown elements between them.

PR #2672, opened 2019-10-15
Remove "use osu! account to contribute" note on Main Page

### LINKED ISSUE #2648: Remove line "You can use your osu! account to start contributing!" since this is not possible anymore
https://github.com/ppy/osu-wiki/blob/master/wiki/Main_Page/en.md

using our own osu! account to edit the wiki is not possible anymore.
currently we can only contribute by creating a github account and we are not forced to use our osu! user names in github or anything like that so keeping "You can use your osu! account to start contributing!"  in that article makes no sense

LiquidPL:
closes #2648


cl8n:
for now can we just remove the second paragraph and call it good, we're going to remove that at some point anyway

the styling fixes will be helpful later but we don't need em for this PR

cl8n:
flipped PR lol

PR #2679, opened 2019-10-17
Reorganize People articles

### LINKED ISSUE #2220: People shouldn't exist
Not sure why https://osu.ppy.sh/help/wiki/People/ exists when it's just The_team minus some.

Just redirect to The_team and call it a day?

Also consider moving inactive below user groups

cl8n:
the hierarchy makes a little more sense this way and removes the need for duplicate indices. resolves #2220


cl8n:
conflict should be very straightforward to resolve (all I did was rename those files), I'll leave it for merge time in case if other staff log / group updates happen between now and then

PR #2680, opened 2019-10-17
Add .editorconfig

cl8n:
mostly for the `trim_trailing_whitespace = false` in markdown, a lot of people have their editor trim by default and forget that this actually does matter for markdown

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false


MegaApplePi:
How would this work exactly? Does the person editing need to have something else installed for this to work?

cl8n:
an editorconfig extension if their editor doesn't support this out of the box

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
MegaApplePi:
```
end_of_line = lf
```

?

cl8n: no reason to standardize the endings locally as long as git checks them back in as LF

just annoying for windows users

MegaApplePi: I was looking more towards people who upload files via GitHub. As it sounded like GitHub implemented EditorConfig natively but I couldn't find any documentation about this.

cl8n:
there's no time like the present 🎉

Issue #2634, opened 2019-10-03
RFC: Enforce mandatory parity with English articles for localisations


sr229:
Prior to GH-2626, there has been conflict about content parity with some localized articles within this PR. So I'm proposing in the ASC to add a "Mandatory content parity" clause to prevent localization teams and contributors to deviate from the base article (English articles are the *de facto* base articles).

Reasons for this is because we want to enforce better consistency in translations, at the sacrifice of lesser flexibility.

cl8n:
it shouldn't "sacrifice flexibility", just let translators know that if they need to change the info presented in the article they should do one of
- ask someone first if they aren't sure
- change en.md as well
- leave a comment on the PR about it so someone else can review and/or update en.md

IMO any changes in tone/style/language are okay, even moving around sentences/paragraphs where necessary, as long as it's the translator's best effort and they think it works well. the reality is we don't have the resources to get more strict than that anyway (except for maybe French/Russian rn, but that's still up to each set of reviewers and not rly enforceable)

sr229:
I think as long as we can mandate it, the reviewers should be aware and notified of the new changes.

peppy:
It should be as close as possible, agree.

sr229:
Putting this as my TODO since I want this RFC to be in the constitution to prevent further inconsistencies, the less time we get this sorted the better since the wiki is a large project by its own.

PR #2687, opened 2019-10-19
Add Content Parity clause to ASC

sr229:
Resolves GH-2634


sr229:
Whoops, I accidentally clicked Review instead of commit, my bad 😅

PR #2688, opened 2019-10-20
Added importaint information missing from score documentation

columna1:
Missing important information regarding additional mod values in score records. Not having this information could lead to errors when reading or writing to files. osu! gives an error if this information is missing from scores.db.


cl8n:
good catch, thank you

PR #2690, opened 2019-10-20
Ranking Criteria Update

dorothy3242:
https://osu.ppy.sh/community/forums/topics/949456
Additionally changes "ranked" for the consistent song title guideline to "has leaderboard" (which is Ranked and Loved)

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -146,8 +146,8 @@ All forms of Artist/Title standardisation apply to both the `Romanised` and `Uni
 
 - **Logos should not be used as references for capitalisation of titles or artists.** Because logos are often stylised, apply standard capitalisation unless other textual metadata supports it.
 - **Tracks created by artists belonging to doujin circles should list the circle's name as the main artist.** The exception to this is when the artist(s) of a given track is well-known enough by their own name. In this case, the specific artist name(s) may be used instead.
-- **If the same song exists in the `Has Leaderboard` section already, the metadata should be followed unless it breaks other rules in the Ranking Criteria or the Official Sources state something completely different.**
-- **Artist names and field formatting should be consistent between different songs from the same person or group as other maps in the `Has Leaderboard` section.** This is unless the person or group intentionally uses a different alias for different song or album releases.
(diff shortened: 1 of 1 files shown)


cl8n:
think I mentioned before but "has leaderboard " can be potentially confusing because of the qualified section 

so I think it's better to write out "ranked or loved"

wiki/Ranking_Criteria/en.md
```diff
@@ -146,7 +146,8 @@ All forms of Artist/Title standardisation apply to both the `Romanised` and `Uni
 
 - **Logos should not be used as references for capitalisation of titles or artists.** Because logos are often stylised, apply standard capitalisation unless other textual metadata supports it.
 - **Tracks created by artists belonging to doujin circles should list the circle's name as the main artist.** The exception to this is when the artist(s) of a given track is well-known enough by their own name. In this case, the specific artist name(s) may be used instead.
-- **If the same song exists in the Ranked section already, the metadata should be followed unless it breaks other rules in the Ranking Criteria or the Official Sources state something completely different.**
+- **If the same song exists in the Ranked or Loved sections already, the metadata should be followed unless it breaks other rules in the Ranking Criteria or the Official Sources state something completely different.**
```
cl8n: i dont understand why Official Sources is in title case but it was like this already so /shrug

dorothy3242: maybe because it's a glossary term, but not even the glossary capitalizes it like that and we don't do that for other glossary terms, oop

wiki/Ranking_Criteria/en.md
```diff
@@ -146,7 +146,8 @@ All forms of Artist/Title standardisation apply to both the `Romanised` and `Uni
...
+- **Artist names should be consistent between different songs from the same person or group in the Ranked or Loved sections.** This is unless the person or group intentionally uses a different alias for different song or album releases.
```
cl8n:
```suggestion
- **Artist names should be consistent between different songs from the same person or group in the Ranked or Loved sections.** This does not apply if the person or group intentionally uses a different alias for different song or album releases.
```

cl8n: more direct

PR #2691, opened 2019-10-20
Remove sitemap content to make place for the autogenerated one

LiquidPL:
Requires:
- [ ] ppy/osu-web#5152

After the web-side PR is merged, the sitemaps will be generated automatically from the pages stored on the website.

Also it would probably be a good idea to add placeholder pages for each language here, so that it loads the respective sitemaps for them, instead of English.


LiquidPL:
The idea with this was so the wiki editors can add some text at the beginning of the page if they wish so, but if you don't feel the need to do so then sure.

cl8n:
seems fine to me, though I feel like this might fit better as a template- sort of thing rather than a page layout. like the ability to write `<Sitemap/>`

not worth looking into for this change, just a thought

LiquidPL:
Yeah I don't think it'd be worth implementing wiki-level templates right now.

cl8n:
web doesnt need any article now

PR #2693, opened 2019-10-21
Make rules sound stricter

trillyon:


peppy:
not sure this is necessary, nor sounds better

trillyon:
ight imma close this

PR #2702, opened 2019-10-24
Remove Difficulties/ subpages

### LINKED ISSUE #2579: Difficulties/ pages are probably not useful
at least in their current state. they're basically mapping guides and the same info can be found in RC (but it's more updated/accurate there)

cl8n:
resolves #2579 (if ppl agree with my latest comment there)


wiki/Disambiguation/en.md
```diff
@@ -5,6 +5,6 @@ Disambiguation in the osu! wiki is the process of resolving conflicts that arise
 ## List of disambiguation articles
 
 - [Bancho](/wiki/Bancho)
-- [Easy](/wiki/Easy)
+- Easy
```
MegaApplePi: The link here should not be removed. It is pointing to a redirect, but should really link to the page underneath it.

cl8n: goop

is this the only one that's referring to disambig (did you check all of them) ?

MegaApplePi: Just checked again but yeah, it looks like this is the only one that actually intends to link to the disambiguation page while all of the others were linking to it by mistake.

PR #2703, opened 2019-10-24
Add a new place where BBCode is used

trillyon:
Added a new place where BBCode is used.


wiki/BBCode/en.md
```diff
@@ -153,7 +153,7 @@ _Not to be confused with [Spoilerbox](#spoilerbox)._
 
 **Hides a paragraph** inside a nameable box from the reader's view. The reader, however, can still read the paragraph by clicking on the box to toggle the hidden paragraph.
 
-These are commonly used to hide large images or a large amount of images.
+These are commonly used to hide large images or a large amount of images, usually in the [Skinning forum](https://osu.ppy.sh/community/forums/15).
```
cl8n: doesn't seem necessary to mention this forum. boxes are used everywhere

trillyon: It's ok if you delete. Just something I kinda wanted to mention.

PR #2705, opened 2019-10-24
Add "omit first bar line" to .osu file format

pavlukivan:


cl8n:
I'll fix it in that PR^ and commit under your name, since we will have to fix the conflict anyway if I merge this first

thanks!

PR #2712, opened 2019-10-26
[DE] osu!stream

xCoast:
I had to do a commit before submitting because my fork and branch were still holding on to my [past mistakes](https://github.com/ppy/osu-wiki/pull/2693).
You are welcome to translate this.


trillyon:
Also, there are other commits in there that I don't need. Please delete the osu!supporter changes if you can.
Thanks

cl8n:
pls reset/rebase this and make a single commit with the osu!stream article, you can do `git push -f` when you're done to overwrite the history in this PR

also please test your changes before committing, there is a lot of broken markdown because you added extra spaces

xCoast:
Ok. Will resub

PR #2722, opened 2019-10-27
Reference Song Content Rules and Visual Content Considerations in Rules

cl8n:
follow-up to #2719 suggested by @squirrelpascals


Joehuu:
Are we not using `outdated_since` tags here?

cl8n:
i can't tell if the articles are more outdated than that, would rather have less info than be wrong

cl8n:
on second thought idk why I marked this as a rule change, it's just listing things that were mistakenly not on this page before

PR #2723, opened 2019-10-27
Don't use pp redirect link in Beatmaps page

Circleen:
Somebody forgot to make loved and qualified the same link (pp was a blank redirect).


Circleen:
Worthless pull request. Continue on with your day

MegaApplePi:
I think this pull request was fine. That link should probably match the other one in the loved section.

Circleen:
It shouldn't be a redirect.

Circleen:
:)

PR #2726, opened 2019-10-27
Add Outdated Tags to Registration Localizations

Circleen:
Now that we have then new registration page done, the localization's are outdated.


Circleen:
I’m finished with the outdated tags

cl8n:
looks like there's no ref to update here, all we get to work with is a patch

PR #2734, opened 2019-10-28
Add New Item to Miscellaneous

Pelbix:
Some players don't use tablets, and tablets aren't required to play osu!.


Pelbix:
Conflicts were found. Closing.

PR #2730, opened 2019-10-27
Hotfix for Main Page

MegaApplePi:
Hot and temporary fix for the main page. This pull request is not the rewrite; for discussion on rewriting this page, please see #2606.

PR #2735, opened 2019-10-28
Fix osu!lazer hyperlink in main page

phosphene47:
Fix `osu!lazer` hyperlink which is linking to the wiki repository rather than lazer.


Joehuu:
Will be fixed in https://github.com/ppy/osu-wiki/pull/2730.

PR #2752, opened 2019-10-31
[KO] Fix markup in Ranking Criteria/osu!mania

Dubstek1:
The command was added where it should have been.


cl8n:
looks like only the first sentence of each is supposed to be bold

PR #2759, opened 2019-11-02
Add HD Size to Skin Elements Lists

### LINKED ISSUE #2756: Skinning Elements Page does not have HD Sizes
most osu! players have an HD resolution, and most skinners use both HD and SD files.

circleen:
Fixes #2756. Most players use HD resolution.

- [X] Interface
- [X] osu!standard
- [x] osu!taiko
- [x] osu!catch
- [x] osu!mania

### DIFF
--- a/wiki/Skinning/Interface/en.md
+++ b/wiki/Skinning/Interface/en.md
@@ -1123,9 +1123,9 @@ Notes:
 
 ![](img/scorebar-kidanger2.png)
 
-| Versions | Animatable | Beatmap Skinnable | Blend Mode | Origin | Suggested SD Size |
-|:-:|:-:|:-:|:-:|:-:|:-:|
-| All | ![No][false] | ![Yes][true] | Normal | Centre | - |
+| Versions | Animatable | Beatmap Skinnable | Blend Mode | Origin | Suggested SD Size | Suggested HD Size |
+|:-:|:-:|:-:|:-:|:-:|:-:|:-:|
+| All | ![No][false] | ![Yes][true] | Normal | Centre | - | - |
 
 Notes:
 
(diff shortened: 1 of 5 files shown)


circleen:
Interface is finished!

cl8n:
if every one is twice in each dimension, I feel like it'd be better to mention that instead of listing out redundant information for every element (at a glance the only one you changed that doesn't follow this rule is the main menu bg?)

circleen:
The main menu background was supposed to be 1920x1080, according to that skin element list spreadsheet

cl8n:
I don't know what "that spreadsheet" is; unless if it's an official source you should verify these yourself

also the commit you titled "cl8n suggestions" has nothing to do with what I suggested ಠ_ಠ

circleen:
What I ment by 'cl8n suggestions' is the double file size on the menu-background element. Also [this was the spreadsheet I was talking about.](https://docs.google.com/spreadsheets/d/1bhnV-CQRMy3Z0npQd9XSoTdkYxz0ew5e648S00qkJZ8/edit#gid=2074725196)

circleen:
The thing I don't get is how I'm at Size XXL yet I have 2 things done on the list

cl8n:
read my comment again, that's not what I suggested

cl8n:
- size tags are based on lines changed https://github.com/apps/pull-request-size
- spreadsheet seems to be missing `arrow-generic` so idk if any other parts of it are outdated
- apparently menu-background doesn't support 2x sizing (https://osu.ppy.sh/community/forums/posts/6723524), should check if this post is true

circleen:
- yes ik
- hmm...
- oh. never realized.

wiki/Skinning/Interface/en.md
```diff
@@ -31,16 +31,17 @@ Notes:
     - If the background is transparent, the background color will be black.
 - The player can drag and drop an image to overwrite the skin's skinned image
   - **This will delete the skin's image in the process!**
+- Background images can only be 1366x768 and will be upscaled to the player's resolution.
```
cl8n: this isn't true, you have to read more carefully and/or test using your own game client before committing these changes

the forum post says (again, I'm not sure if forum post is true so you should check) that the game doesn't use the @ 2x variant of this, so the skinner should use the highest quality image available for the "SD" version

circleen: The game does not use the 2x varient. Just tested.

circleen:
Ready for review!

wiki/Skinning/Interface/en.md
cl8n: ... so can you fix this part? not sure why you marked as resolved

circleen: The note is gone now. Should I leave the HD Size as a - or add the 2x resolution?

cl8n:
per https://github.com/ppy/osu-wiki/pull/2759#issuecomment-549015408 this doesn't seem useful

PR #2767, opened 2019-11-03
Fix Spelling Errors on Help Centre

circleen:


peppy:
Most of these aren't really errors but changes to style.

cl8n:
you don't need to block these if they don't change anything beyond language/style. I think the only articles we block for that are legal

PR #2769, opened 2019-11-04
Outdate storyboard list

Coppertine:
The storyboard list should either be updated or deleted.


WalterToro:
Would mark for deletion as User Generated Content shouldnt be on the wiki

cl8n:
deleted @ #2770

Issue #2621, opened 2019-09-28
Tournament template needs some updates


cl8n:
didn't check exactly what needs to change yet but I keep having to comment the same stuff on tournament PRs because some parts of the template are dumb

cl8n:
this PR linking here^ has a lot of good examples for what could change. should probably apply backwards to recent world cups

PR #2772, opened 2019-11-04
Update tournament match results template

### LINKED ISSUE #2271: Tournament match results tables are unlabeled/confusing
for most tournaments
- dates are written in the table header for some reason (are dates even needed if they're in the schedule already?)
- table isn't labeled
- nothing describes match links, currently you have to hover over a random "# 1" to tell that it's a match link at all
- double elim brackets aren't specified

cl8n:
part of #2621, closes #2271


cl8n:
i'll do the other sections on different branch

PR #2784, opened 2019-11-06
Fix oversized images in old fanart contest

peppy:


cl8n:
how did you find/notice this lol

peppy:
I get exceptions bubbling up to my alert stream if they occur more than five times in succession (which was caused in this case by someone visiting the page and triggering image loads for every image).

If I don't acknowledge within 5 minutes the world basically explodes around me.

Issue #2681, opened 2019-10-17
Mark outdated, incomplete, unnecessary, etc. English articles


cl8n:
it's well-known to people familiar with the wiki that half of it is in need of updates, but we still haven't properly marked which articles those are. at some point we should make a full pass with the `outdated` tag and open more specific follow-up issues that we can milestone

https://github.com/cl8n/osu-wiki-log-outdated should be updated to ignore outdated articles in the "missing translations" section too

sr229:
I'm making a educated guess about 30% of the translations might be outdated since the English versions have more churn vs these translations.

cl8n:
high priority because translators need to know if articles are worth their time to translate

cl8n:
for reference there is now a `needs_cleanup` attribute available that displays a message like `outdated` does:

> This page does not meet the standards of the osu! wiki and needs to be cleaned up or rewritten. If you are able to help out, please consider updating the article!

Issue #2486, opened 2019-09-07
List of outdated articles and missing translations


cl8n:
select your language on **[the osu-wiki-status website](https://clayton.cc/osu-wiki-status/en)**, it tells you articles marked with `outdated` or `needs_cleanup`, and missing translations

some English articles *shouldn't* be translated, and need to be updated in English first:

- Articles marked with `outdated` (this is just for en.md, if the translation has `outdated` that's okay)
- Articles marked with `needs_cleanup`
- Articles that seem to be written poorly or contain misleading info. Message in `#osu-wiki` if you aren't sure about this one

the "Missing articles" sections on osu-wiki-status also omit contests, tournaments, and staff log. you can translate them if you want but they're very low priority. staff log is also set to be deleted, as it'll be replaced with an equivalent feature on the osu! website. there may be some other articles omitted by me, I'm not keeping this issue up-to-date.

Akylzhan:
@cl8n Can I translate Missing Translations to Russian?

cl8n:
go for it, just try to avoid articles that haven't been updated in a long time because we might revise the English versions soon

GalaxyOsu:
@cl8n I would like to contribute to the wiki, editing images, changing some info and translating pages

TPGPL:
@GalaxyOsu You are free to start contributing whenever you want. Here's the [contribution guide](https://osu.ppy.sh/help/wiki/osu%21_wiki_Contribution_Guide) if you need more information on how to contribute.

DoryanR:
@cl8n I was wondering if it was possible to have a specific category for articles that have the `no_native_review` tag, as this would be useful to know which ones have not had any reviews, and to look at them for potential mistakes.

cl8n:
added

Soetch:
I'll translate the french missing translations.

cl8n:
some outdated info here and we've since added links to this website elsewhere, going to close and unpin this.

PR #2792, opened 2019-11-07
[FR] Videos From Youtube

UltayJr:


cl8n:
this probably wasn't worth your time to translate because the original isn't written well / isn't very useful

I added some guidelines in https://github.com/ppy/osu-wiki/issues/2486#issue-490619773 to help decide if an article is worth translating. also when #2681 is done that should be helpful for you

cl8n:
I'll just merge this now and updates can be made later after the English one changes

PR #2793, opened 2019-11-07
Update Videos from YouTube downloader and formatting

UltayJr:
Using another link, because the past one is broken, i used another, if you have better ones !?


cl8n:
for now it's better than a dead link.

PR #2809, opened 2019-11-12
Clean up wording in Installation

circleen:
/wiki/Solo_mode doesn't exist. Who put that here


cl8n:
solo mode was probably a page in the past (it must have predated `osu-wiki` because there's nothing in git history about it)

PR #2812, opened 2019-11-12
Update transaction completed text on osu!supporter

circleen:
It was wrong


circleen:
that was quick

circleen:
world record for quickest pr ever merged

PR #2827, opened 2019-11-17
[ES] Main_Page

Makamizake:
Translated version of Main_Page en.md.


wiki/Main_Page/es.md
```diff
@@ -0,0 +1,62 @@
+<!-- Translated to Latin American Spanish, there it might be some differences with Castilian. -->
```
MegaApplePi: I think you need to move these comments underneath the meta. Can you move them down to line 6?

PR #2854, opened 2019-11-27
Fix spelling mistake on Projects

circl3n:


peppy:
Was this fixed elsewhere?

cl8n:
it's spelled the same way on the forum post, probably why OP closed this

PR #2788, opened 2019-11-07
Rewrite Beatmap ranking procedure

pishifat:
also deletes duplicate outdated guide and hopefully fixes all relevant links

fixes part of #2615

PR #2858, opened 2019-11-27
[KO] Update Beatmap Ranking Procedure

Dubstek1:


cl8n:
#2788 will be changing almost all of this article. some of this translation might be relevant still but I want to see where that one goes first (because it was significant enough of a change to warrant deleting the translations)

PR #2874, opened 2019-12-01
Skinning/Interface: Add note about menu-background.jpg sizing

### LINKED ISSUE #2823: Mention menu-background's (lack of) @2x behavior
and probably explain what @ 2x does in more places (each skin elements page... idk)

MegaApplePi:


wiki/Skinning/Interface/en.md
```diff
@@ -15,12 +15,14 @@ Interface skinning elements are used in multiple game modes or parts of the clie
 
 | Versions | Animatable | Beatmap Skinnable | Blend Mode | Origin | Suggested SD Size |
 |:-:|:-:|:-:|:-:|:-:|:-:|
-| All | ![No][false] | ![No][false] | Normal | Centre | 1366x768 |
+| All | ![No][false] | ![No][false] | Normal | Centre | 1366x768 (see notes) |
```
cl8n: probably use a higher recommended size

I was going to say "use the same size osu! defaults use", but it seems like the default bgs *do* support @ 2x... so I guess list the 2x-but-not-available-as-2x size,  2732x1536

cl8n:
I don't have supporter &can't test so I hope this is correct ¯\\\_(ツ)\_/¯

PR #2880, opened 2019-12-02
Add markdownlint config

cl8n:
a very late start on #2275


cl8n:
this is for the node implementation of markdownlint. it matches ASC with the MD rules available and I'll be adding more rules in a separate project to cover everything possible

PR #2883, opened 2019-12-02
Update MP7K for Semifinals

benzerka:


cl8n:
check if your editor has a file encoding option for "UTF8", instead of "UTF8 with BOM". we only support the former here

PR #2885, opened 2019-12-02
Clean up timing exceptions in Ranking Criteria

Naxesss:
The BN tests shows that a many applicants fail on questions related to unsupported beat snap divisors tied with BPM changes.

This aims to clear up the confusion by making the exceptions clearer and more readable.


wiki/Ranking_Criteria/en.md
```diff
@@ -180,11 +180,15 @@ This category contains explicit allowance statements of concepts and rules that
...
 - **Beatmaps with Single-BPM and Multi-BPM timing must be perfectly timed.** This means BPM and offset are exactly synchronized with the song.
 - **Uninherited timing points must be the same in every difficulty of a beatmapset.** Each point must have the same BPM and offset in each difficulty.
-- **There must not be extra uninherited timing points in any difficulty.** These can affect main-menu pulsing, the Nightcore mod, and cause timing to shift due to millisecond rounding errors. Adding uninherited timing points at the starts of musical phrases to match a song with the Nightcore mod's beat is acceptable.
+- **There must not be extra uninherited timing points in any difficulty.** These can affect main-menu pulsing, the Nightcore mod, and cause timing to shift due to millisecond rounding errors. Exceptions to this rule:
```
cl8n: the two things you list as exceptions aren't exceptions, they're exactly what the rule is meant to cover. the wording of "extra uninherited timing points" makes this pretty clear imo, because those "exceptions" aren't extra timing points, they're required to make the map work well

revert and then tack on the bit about 1/5 1/7 to the last sentence, that should clarify it enough if the reader doesn't make it to line L187

Naxesss: changed to allowance instead. kept the points for readability and consistency with the other change (allows for less text and easier processing/look-up)

Coppertine:
Just to say, if you are placing down a timing point that is not on a full beat, you can enable the "Omit First Bar Line" option to remove the crash sound in the nightcore mod.

PR #2886, opened 2019-12-02
Clean up and clarify RC and Scaling BPM

pishifat:
fixes https://osu.ppy.sh/community/forums/posts/7284928 and some more


cl8n:
I guess catch should be the same too?

pishifat:
all should be the same

cl8n:
i know they dont all have rhythm rules but it's not wrong to refer to a set of 0 things

anyway catch does have both rhythm guidelines and rules but no mention of em so lets do that

PR #2887, opened 2019-12-03
Add exception to exclusive Qualified songs BN rule

pishifat:
discussion: https://osu.ppy.sh/community/forums/topics/979133

### DIFF
--- a/wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
+++ b/wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
@@ -22,9 +22,8 @@ Beatmap Nominators possess the tools to nominate beatmaps and shape the world of
 
 ### Qualification
 
-- **To maintain a variety of songs in the Ranked section, there are two requirements:**
-  - Different beatmaps of similar versions of a song cannot be qualified in the same mode at the same time. Similar versions of a song include cuts, extensions, and simple covers.
-  - Only 3 significantly different versions of a song may be qualified in the same mode at the same time. Significantly different versions include remixes, compilations, and mashups.
+- **Different beatmaps of similar versions of a song cannot be qualified in the same mode at the same time.** Similar versions of a song include cuts, extensions, and simple covers.
+- **Only 3 significantly different versions of a song may be qualified in the same mode at the same time.** Significantly different versions include remixes, compilations, and mashups.
 - **Do not qualify beatmaps of modes which you are not a full member of, unless they were nominated by someone who is.** This ensures that a full nominator is always part of the process for each mode.
 - **Do not qualify a beatmap directly without any communication with the mapper.** The mapper might have done self-changes after the last nomination, so nominating a beatmapset without communicating with the mapper can lead to pointless disqualifications.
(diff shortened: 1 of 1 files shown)


wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
```diff
@@ -22,7 +22,7 @@ Beatmap Nominators possess the tools to nominate beatmaps and shape the world of
 
 ### Qualification
 
-- **Do not qualify beatmapsets whose song already has a Qualified beatmapset, unless no game modes of both sets overlap—this includes remixed versions**. For example an osu!taiko and osu! beatmapset of the same song can exist in the Qualified section, whilst an osu!taiko and osu!/osu!taiko hybrid set cannot. This ensures variety among Ranked beatmaps within their respective game modes.
+- **Only one beatmap of a song may be qualified per game mode at a time. This includes different length cuts and simple covers of a song, but excludes remixes, up to three different versions of which may be qualified per game mode at a time.** This ensures variety among Ranked beatmaps within their respective game modes.
```
cl8n: - should use similar language as the other points, speaking directly to the bn. so first sentence could be "Do not qualify beatmapsets with songs already used in Qualified beatmaps of the same mode." ?
- "includes" and "excludes" could be interpreted to mean the opposite of your intention, I think that's why the original was confusing. "Cuts and covers do not count as different songs when applying this rule, but remixes do" or somethign similar that sounds better for second sentence 
- possibly remove the remixes thing altogether per The Dm
- can drop the "within their respective game modes" from last sentence cuz its obvious and just puts morewords in already wordy rule

pishifat: reworded

wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
```diff
@@ -22,7 +22,9 @@ Beatmap Nominators possess the tools to nominate beatmaps and shape the world of
 
 ### Qualification
 
-- **Do not qualify beatmapsets whose song already has a Qualified beatmapset, unless no game modes of both sets overlap—this includes remixed versions**. For example an osu!taiko and osu! beatmapset of the same song can exist in the Qualified section, whilst an osu!taiko and osu!/osu!taiko hybrid set cannot. This ensures variety among Ranked beatmaps within their respective game modes.
+- **To maintain a variety of songs in the Ranked section, there are two requirements:**
+    - Different beatmaps of similar versions of a song cannot be qualified in the same mode at the same time. Similar versions of a song include cuts, extensions and simple covers.
+    - Following the above, only 3 significantly different versions of a song may be qualified in the same mode at the same time. Significantly different versions would be remixes, compilations and mashups.
```
dorothy3242:
```suggestion
    - Only 3 significantly different versions of a song may be qualified in the same mode at the same time. Significantly different versions would be remixes, compilations and mashups.
```

cl8n:
the bold is on the wrong part, it doesn't say anything by itself. bold first part of each bullet instead

cl8n:
and probably just remove the first line too

cl8n:
merge master into this branch if you're going to use it again

PR #2891, opened 2019-12-03
Make osu!mania tournament names consistent

cl8n:


Joehuu:
Should we merge `MP4K` and `MP7K` to `MP` and have subfolders `3_4K` and `2019_7K`?

Also the order of the naming (year / installment number and mode) needs normalizing. See https://github.com/ppy/osu-wiki/tree/master/wiki/Tournaments/MWC and https://github.com/ppy/osu-wiki/tree/master/wiki/Tournaments/oPHN. I would go with MWC.

cl8n:
sure

PR #2892, opened 2019-12-03
Add Winter 2019 Fanart Contest Voting newspost

Ephemeralis:
Contest is live and pending announcement by this post.


cl8n:
would it be hard to make the osu cookie beat to the song in the video? I think that'd make it nicer to look at

Issue #2606, opened 2019-09-25
Update Welcome and Main Page


cl8n:
the main page is live with the new format, and most of Welcome can be replaced by it. the main page just needs a small introduction to osu! and descriptions for each major category, then we can redirect welcome --> main_page

peppy:
Here's what the welcome page on the old site used to include. Note that a lot of these are outdated, but the general feel of them still stands. It should be aimed at new players to help them jump into playing the game

https://gist.github.com/peppy/a387c8fc05da0188fd5d57e757298c01

sr229:
How do we go around re-implementing this? Sounds like web have to do some parts as well.

cl8n:
if it's just text at the beginning and within each box, it looks like it'll work as-is. whoever closes this issue should be testing (or asking someone to test for them) just in case, though.

sr229:
I'll put it at my todo for Monday, I can port what was in the snapshot HTML peppy sent.

sr229:
My strategy is to condense `welcome.md` into something similar to @peppy's gist. This should remove `welcome.md`. Might need to redirect to index.

PR #2897, opened 2019-12-04
[ZH] Add Main Page

cdwcgt:
Add Chinese translation of Main_Page
Status: Ready for being reviewed.

### DIFF
--- a/wiki/Main_Page/zh.md
+++ b/wiki/Main_Page/zh.md
@@ -1,0 +1,61 @@
+---
+layout: "main_page"
+---
+
+<!-- Do not add any empty lines inside this div. -->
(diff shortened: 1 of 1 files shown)


cdwcgt:
I saw this(This page does not meet the standards of the osu! wiki and needs to be cleaned up or rewritten. If you are able to help out, please consider updating the article!) in https://osu.ppy.sh/help/wiki/Main_Page
Maybe this translation need to change?

cl8n:
the translation is fine for now, that note means the English version needs to be updated first (#2606)

cdwcgt:
ok.I know it.

cdwcgt:
I update the translate of "Support Team"
I change it to "制作团队"

PR #2900, opened 2019-12-04
Fix formatting inconsistency across Registration translations

cl8n:


cl8n:
that doesn't sound useful

Issue #2615, opened 2019-09-27
beatmapping pages rework


dorothy3242:
this is a thing that's happening slowly and surely

- [ ] People/Beatmap Nominator pages
- [ ] Beatmap Discussion
- [ ] Beatmap Editor
- [ ] Beatmapping
- [ ] Difficulties
- [ ] Relevant pages in Guides
- [ ] Modding

All need to be reviewed, combined where possible (some pages may seem like duplicates of each other) and generally rewritten to be made up to date. Lots of outdated and conflicting information is currently floating around these pages.

Maybe adding a history page for good to know old info when that's all done too.

cl8n:
I'm gonna assign this to October but we'll probably need some help to complete it

cl8n:
milestone never gets done on this lol

Pennek:
im currently rewriting the Beatmap Discussion page.

cl8n:
I have a thing for Difficulties WIP thanks mr venix for reminding me to post. involves tying together "Difficulty Naming" from RC and some other jazz

Pennek:
For stuff like Hitsounding it might be worth cross-referencing usermade guides, since there are quite a lot of those not directly related to wiki

TicClick:
> Difficulties (clayton is working on this)
is @cl8n still working on this?

PR #2912, opened 2019-12-07
Add "outdated" or "needs_cleanup" tags to articles

### LINKED ISSUE #2681: Mark outdated, incomplete, unnecessary, etc. English articles

cl8n:
skipped:
- BanchoBot, added tag in #2909
- some things related to beatmapping because #2615 is tracking
- Ranking Criteria
- Tournaments
- Storyboard / scripting
- File formats
- Mapping Techniques
- Guides

whole categories may need to be reorganized so i didnt bother with tags

closes #2681

### DIFF
--- a/wiki/Help_Centre/en.md
+++ b/wiki/Help_Centre/en.md
@@ -1,3 +1,7 @@
+---
+needs_cleanup: true
+---
+
 # Help Centre
 
 Having trouble with something? We're here to help! Check out some solutions to common issues in the sidebar to the left of your screen.
--- a/wiki/Registration/en.md
+++ b/wiki/Registration/en.md
@@ -1,3 +1,7 @@
+---
+needs_cleanup: true
+---
+
 # Registration
 
 *Warning: Having more than one osu! user account at any time is an infringement against the [osu! rules](/wiki/Rules)!*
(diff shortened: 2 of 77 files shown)


wiki/Help_Centre/en.md
```diff
@@ -1,3 +1,7 @@
+---
+needs_cleanup: true
```
TPGPL: not sure if needed here though

wiki/Registration/en.md
```diff
@@ -1,3 +1,7 @@
+---
+needs_cleanup: true
```
TPGPL: does it though?

wiki/Help_Centre/en.md
cl8n: i thought this one was strange because it has so many lines for no reason
some of the language is weird later on it just needs small changes

wiki/Registration/en.md
cl8n: low priority if anything so ill remove

PR #2925, opened 2019-12-13
Remove hitsound additions requirement for difficult osu!mania maps in RC

cl8n:


Coppertine:
Is there a reason to include the exception to mania maps? Players can instead turn their hitsound volume to 0%

cl8n:
https://osu.ppy.sh/community/forums/topics/996091 OP has info and links

super tl;dr is most of the mania community doesn't care for or actually dislikes hitadditions on difficult charts, mappers don't wanna map them, all around nobody benefits and it's just demotivating for charters to rank harder maps (which mania desperately needs)

discuss on thread instead of here please

PR #2926, opened 2019-12-13
Remove space from "hit sound"

cl8n:
nobody says "hit sound" or "key sound"


cl8n:
should have mentioned in pr description there is one unchanged "hit sound" in RC because it's in the other PR i have open. i think that one's going to be open for at least another few weeks though so u can fix it here and i solve conflict later

PR #2944, opened 2019-12-18
Add information about [audio] tag on BBCode page

KatouMegumi-osu:
Since this was misisng there, I have decided to add it. Though, I think this tag was mainly intended for FA news pages, it still works in forums.


cl8n:
formatting is really awkward, but so is the rest of this article and it's already marked as outdated so w/e

PR #2954, opened 2019-12-20
[JA] Replace 譜面 with ビートマップ

ethernetcat:
Replace words that differ from Japanese translations uploaded to osu-web with correct words.

*Replaced word*
`譜面` -> `ビートマップ`


cl8n:
seems appropriate, in English we don't really think of the meaning of "beatmap" when we say it, it's more like the osu! equivalent of "level"

ethernetcat:
In Japanese, "譜面" refers to  music sheet or notes written on music sheet.
I don't know how it interprets it in English. but I think `ビートマップ` is the best choice for the proper noun of osu!.

I don't know if the Japanese osu! community considers this appropriate.
If anyone considers it inappropriate, it needs discussion in the community.

cl8n:
ビートマップ is good 👍

PR #2955, opened 2019-12-20
Remove ranking criteria rule

pishifat:
recent client update made this rule obsolete (taiko now has drum samples)


pishifat:
yes

cl8n:
bad idea to do this one so quickly, the pitches are swapped from the usual expectation of taiko samples and this probably won't play well if you use them

PR #2960, opened 2019-12-22
Conclude OWC 2019

WalterToro:
We are done!


cl8n:
is omgforz spelled with the wrong casing on some internal sheet, I remember having to fix it once before

PR #2962, opened 2019-12-23
Update AiMod checks

Coppertine:
Set as outdated due to other functions that might have changed to the AIMod that I have not noticed.

### DIFF
--- a/wiki/Beatmap_Editor/AiMod/en.md
+++ b/wiki/Beatmap_Editor/AiMod/en.md
@@ -146,7 +146,7 @@ AiMod typically completes its beatmap analysis in under a minute and subsequentl
 | Unicode artist conflicts with {0} diff. | | |
 | Unicode title conflicts with {0} diff. | | |
 | This mapset cannot have an Insane diff without a Hard diff. | | |
-| Beatmap exceeds allowed upload size ({0}kb allowed) | The .osz file has found to be larger than the allowed limit set depending on the song. | |
+| Beatmap exceeds allowed upload size ({0}kb allowed) | The zipped `.osz` file is too large | |
 | This mapset is missing diff: {0} | | |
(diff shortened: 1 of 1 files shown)


wiki/Beatmap_Editor/AiMod/en.md
```diff
@@ -163,9 +145,8 @@ AiMod typically completes its beatmap analysis in under a minute and subsequentl
 | Title conflicts with {0} diff. | | |
 | Unicode artist conflicts with {0} diff. | | |
 | Unicode title conflicts with {0} diff. | | |
-| This mapset cannot have an Insane diff. without a Hard diff. | | |
-| Filesize of the mapset exceeds 10MB without video. | | |
-| Filesize of the mapset exceeds 24MB with video. | | |
+| This mapset cannot have an Insane diff without a Hard diff. | | |
+| Beatmap exceeds allowed upload size ({0}kb allowed) | The .osz file has found to be larger than the allowed limit set depending on the song. | |
```
cl8n: probably remove this explanation considering it's super straightforward and none of the others have one rn
```suggestion
| Beatmap exceeds allowed upload size ({0}kb allowed) | | |
```

Coppertine: Not exactly, the upload that is made is for the zipped file, the upload guidelines looks at the zipped size and not the folder size when unpackaged

Coppertine: The mapper who would see this error would at first think it's the folder size

cl8n:
```suggestion
| Beatmap exceeds allowed upload size ({0}kb allowed) | The zipped `.osz` file is too large | |
```

PR #2964, opened 2019-12-23
[JA] Add Difficulties

ethernetcat:
Add Japanese translation of `Difficulties`.
The difficulty name, which is unique word of osu!, is not translated.

---

### DIFF
--- a/wiki/Difficulties/ja.md
+++ b/wiki/Difficulties/ja.md
@@ -1,0 +1,71 @@
+---
+tags:
+  - difficulty
+  - difficulty name
+  - SR
+  - star rating
+  - stars
+---
+
+# 難しさ
+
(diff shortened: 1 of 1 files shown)


wiki/Difficulties/ja.md
```diff
@@ -0,0 +1,71 @@
+---
+tags:
+  - difficulty
+  - difficulty name
+  - SR
+  - star rating
+  - stars
```
cl8n: these should be written in japanese. they don't have to be translated from en but they should help people search for this article on https://osu.ppy.sh/home/search?mode=wiki_page

ethernetcat: Okay. Translate them.

ethernetcat:
Thank you for the review!
Translated and committed the tag.

PR #2965, opened 2019-12-23
Update tags and content parity ASC rules

### LINKED ISSUE #2833: Add ASC rule about translating tags
they're usually copied from English but I think they're supposed to be translated. need to double-check if that actually does anything

cl8n:
resolves #2833

### DIFF
--- a/wiki/Article_Styling_Criteria/en.md
+++ b/wiki/Article_Styling_Criteria/en.md
@@ -54,13 +54,18 @@ If your language is not listed above, you can either append `?locale={langcode}`
 
 ### Content parity
 
-*Note: This does not apply to articles originally written in a language other than English.*
-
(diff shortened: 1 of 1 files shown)


wiki/Article_Styling_Criteria/en.md
```diff
@@ -54,13 +54,18 @@ If your language is not listed above, you can either append `?locale={langcode}`
...
-*Note: This does not apply to articles originally written in a language other than English.*
-
 Translations are subject to strict content parity with their English article, in the sense that they must have the same message, regardless of grammar and syntax. Any changes to the translations' meanings must be accompanied by equivalent changes to the English article.
 
+There are some cases where the content is allowed to differ:
+
+- Articles originally written in a language other than English (in this case, English should act as the translation)
+- Explanations of English words that are common terms in the osu! community
+- External links
+- Tags
```
MegaApplePi:
```suggestion
- [Tags](#tagging-articles)
```

MegaApplePi: Never mind, level 4 gets no id iirc

PR #2977, opened 2019-12-27
Remove filler words from Installation

Suliper:


wiki/Installation/en.md
```diff
@@ -20,7 +20,7 @@ This page will roughly tell you how to get osu! working on your Windows device.
 3. While the installer is running, there is a ten second period to allow you to select the installation location.
    - If you want to change it, click on the text underneath the word, `Hi!` and select the installation location.
    - osu! will be installed to that location when the timer is done.
-4. When the installer is completed, osu! will start up. osu!direct will be used to download some starter beatmaps for you.
+4. When the installer is completed, osu! will start up. osu!direct will be used to download some starter beatmaps for you, even if you don't have supporter.
```
cl8n: this part is unnecessary, if the reader doesn't know what supporter is they'd be going out of their way to figure that out for no reason
```suggestion
4. When the installer is completed, osu! will start up. osu!direct will be used to download some starter beatmaps for you.
```

PR #2985, opened 2020-01-01
Update osu!catch RC difficulty setting guidelines

pishifat:
as per https://osu.ppy.sh/community/forums/topics/1002512


pishifat:
we good

PR #3001, opened 2020-01-05
Fix preview in MBC news post

pishifat:
oh no


cl8n:
@peppy was I supposed to be allowed to merge this ? not sure how the codeowners thing works

peppy:
i don't think so, will look into it

PR #3002, opened 2020-01-05
Remove Nardo, Mao and Stack from NAT list

pimpG:


cl8n:
add or update their alumni entries too

this is usually done at the same time as updating staff log, fwiw (if you plan to do more group updates beyond this)

pimpG:
sure

PR #3008, opened 2020-01-05
[ID] Update osu!academy

febrimuhnas:
add needs_cleanup tag

### DIFF
--- a/wiki/osu!academy/id.md
+++ b/wiki/osu!academy/id.md
@@ -1,17 +1,21 @@
+---
+needs_cleanup: true
+---
+
 # osu!academy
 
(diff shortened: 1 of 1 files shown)


wiki/osu!academy/id.md
```diff
@@ -1,17 +1,21 @@
+---
+needs_cleanup: true
```
MegaApplePi: @cl8n do translations also get this too?

cl8n: content/meaning needs work => add to en
the translation itself needs work, but en is fine => add here

MegaApplePi: the translation copied en, but en needs work => add to en only?

cl8n: ya

PR #3009, opened 2020-01-06
Update CGTS 2019 for Global phase

fajar13k:
- Add Qualifiers and Semifinals mappool (Global phase)
- Add Grand Finals match results (Regional phase)
- Add participants for Global phase (WIP)


cl8n:
the beginning of the article should have some info explaining how the tournament works cuz it's not clear how regional and global are related until you get to rules (which aren't expected to be read unless you're a participant)

cl8n:
also if there's a challonge for the global phase don't forget to add that. otherwise additions look good

PR #3028, opened 2020-01-10
[FR] Update Beatmap Editor/Kiai_Time

DoryanR:
This Pull Request contains the most recent version of the Kiai_Time page.

### DIFF
--- a/wiki/Beatmap_Editor/Kiai_Time/fr.md
+++ b/wiki/Beatmap_Editor/Kiai_Time/fr.md
@@ -1,21 +1,20 @@
 # Kiai Time
 
-![Etoiles jaillissant des côtés de l'écran pour montrer que le Kiai Time a commencé.](img/Kiai_Time_test.jpg "Etoiles jaillissant des côtés de l'écran pour montrer que le Kiai Time a commencé.")
+![Stars shower the screen indicating Kiai Time has started.](img/Kiai_Time_test.jpg "Les étoiles illuminent l'écran indiquant que le temps Kiai a commencé.")
 
-![Kiai Time indiqué dans l'éditeur de beatmaps par un indicateur en haut à gauche de l'écran.](img/Kiai_Time_edit.jpg "Kiai Time indiqué dans l'éditeur de beatmaps par un indicateur en haut à gauche de l'écran.")
+![Kiai Time shows in the upper-left of the screen when active during edit mode.](img/Kiai_Time_edit.jpg "Le temps Kiai s'affiche dans le coin supérieur gauche de l'écran lorsqu'il est actif en mode édition.")
 
-Le **Kiai Time** est défini par le [mappeur](/wiki/Beatmapping) sur une beatmap, entre deux [sections de timing](/wiki/Beatmap_Editor/Timing). Il est possible de reconnaître un Kiai Time grâce aux étoiles qui jaillissent des côtés de l'écran, aux flashs en rythme avec la musique et à la traînée d'étoiles qui suit le curseur (si l'option graphique correspondante est activée).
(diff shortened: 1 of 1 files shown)


cl8n:
this is how we handle translation PRs already, so you don't need to add the comment every time

the only exception is if nobody reviews it for a long time (two weeks is what we've been going with lately), we merge and add a tag that lets other reviewers search for these PRs later

cl8n:
ive been told most(/all?) of these PRs are machine translations, or otherwise just not that great, to the point where a reviewer needs to put in so much effort to review that they feel like it's not worth their time. if you want to help out, you're going to need to put more effort into these yourself, or be more proactive about asking other French speakers for specific advice

considering there's an active French translation group here, I think you should try to at least meet their standards. otherwise a lot of time will be wasted redoing translations of the same articles

DoryanR:
> ive been told most(/all?) of these PRs are machine translations, or otherwise just not that great, to the point where a reviewer needs to put in so much effort to review that they feel like it's not worth their time. if you want to help out, you're going to need to put more effort into these yourself, or be more proactive about asking other French speakers for specific advice
> 
> considering there's an active French translation group here, I think you should try to at least meet their standards. otherwise a lot of time will be wasted redoing translations of the same articles

It's true that my translations are done using a translator, I'm not going to hide it. Actually I like to translate content like this, I want to make content available in my language.

However I don't just translate them without proofreading them, I am conscientious in what I do, I do things right.

After concerning the French community, I had already been in contact with them in the past, but they weren't really grateful for the work I brought them, especially since I spent quite a bit of time there, so I decided to detach myself from them.

If it's a problem for me to translate this way, I can stop, but since I have the impression that not many French people are interested in translations, I tell myself that even if I do it this way, a good part of the work is done, and if you ask the other French people who have done translations, they'll probably tell you that they contain errors, but as I said, there's already a big part of the work that's done, so you just have to improve a few things.

cl8n:
starting from a machine translation might be fine, but the fact that people can *tell* that your work comes from a machine translation probably says something about how it reads. I can't say anything about the quality of your work myself so I'm just trusting the majority here... that probably seems unfair but idk what else I can do.

@Pachiru & the rest of the group, if you feel like there are too many problems to comment about, you can also leave more general reviews that cite just a few examples so Doryan has an idea of what to double-check himself. this would require some cooperation on Doryan's part too

fwiw, it's okay if translations aren't perfect---after all, I'm sure a French speaker (who can't speak English) would much prefer any translation compared to none. the absolute minimum quality I think we should all expect is that the PR is easier to review than it is to rewrite from scratch. also, you can make use of the `needs_cleanup` tag if you feel like the content is alright, but some of the language is awkward, and there aren't any reviewers with enough free time to suggest changes

DoryanR:
That's why I'm going to pay more attention to what I translate from now on, so there will be less things to check by the editors.

I'm also aware that not many French people have time to review my translations, but I'm thinking that there's someone who will be able to take care of it one day or another, I'm not saying I want my translations to be accepted quickly (even if at the time it was the case, but I didn't realize that people didn't have time), even if it's what I'd like, I can wait, I'd only make changes if for example the English version has been updated.

I will be more careful in what I do, there will be fewer problems concerning me, I don't want to discourage future French editors, that's not my intention.

wiki/Beatmap_Editor/Kiai_Time/fr.md
```diff
@@ -1,21 +1,20 @@
...
+2.  **Soyez cohérent** ; c'est ennuyeux quand vous avez 3 sections de Kiai à différents endroits sur la même [difficulty](/wiki/Difficulties) et ils ont tous l'air différents. De plus, c'est bien que toutes les difficultés d'un même [mapper](/wiki/Glossary) ont exactement les mêmes temps Kiai.
+3.  **Ne pas utiliser de blanc pur [combo colour](/wiki/Glossary)** (255,255,255) ; cela rend le Kiai aveuglant et fait mal aux yeux.
+4.  Il est normal qu'une section Kiai soit légèrement plus difficile que le reste du jeu de cartes, car le temps Kiai est aussi la **partie la plus forte de la carte**.
+5.  **[Guest difficulties](/wiki/Glossary) peuvent avoir des kiai différents. Ceci est dû au fait que tout le monde ne pense pas que le même endroit est le "cool" de la carte des temps ; cependant, le Kiai doit toujours être cohérent.
+6.  Colorationhaxing pour faire Kiai ont un ensemble différent de [combo colours](/wiki/Glossary) est une belle touche.
```
almaloste:
```suggestion
6.  Un mélange de couleur pour faire Kiai avec un ensemble différent de [couleur de combos](/wiki/Glossary) est une bonne idée.
```

PR #3029, opened 2020-01-11
Update tournaments article

fajar13k:
- Add missing official tournaments (OWC 2019, MWC 4K 2019)
- Add TBDs to on-going tournaments
- Rename BTT Season 4 -> BTT 4
- Update flag calling in Korean translation


cl8n:
not sure what we should do with this article going forward, I get the feeling it's not very useful

fajar13k:
¯\_(ツ)_/¯
If there's nothing planned so far, i might close the PR instead.

cl8n:
ok revisiting this now I remember the situation. we wanted to use this article as the index for all tournaments (which would be nice since we list info more immediate info like dates/winners here, and also iterations that didn't get wiki articles), but there was no way to redirect each series index (e.g. Tournaments/OWC) back to this page in a way that the site supported

so basically it's not very useful right now but could be in the future. or we could move this content into each series index. either way this PR is fine for now I'm just going to clean it upp

PR #3038, opened 2020-01-13
Fix syntax in BBCode article

TPGPL:


wiki/BBCode/fr.md
```diff
@@ -16,18 +16,18 @@ Quand il n'y a pas de texte surligné, cliquer sur le bouton fait apparaitre des
 
 Les balises peuvent être combinées pour un formatage plus riche. L'ordre et l'imbrication des balises **doivent** être respecté quand vous combinez des balises. Le non respect de l'ordre / imbrication peut casser le formatage de la publication.
 
-Par exemple (faire attention à l'ordre du code) :
+Par exemple (faire attention à l'ordre du code):
```
cl8n: this is correct as-is, same with all of the other spaces before colons in French

peppy:
If we're going to follow a standard, it's probably best to add some kind of CI?

TPGPL:
we could use https://github.com/marketplace/actions/markdownlint-cli for the repo and create some custom rules as well (set up example here: https://github.com/TPGPL/osu-wiki/pull/4/checks?check_run_id=387559963)

bdach:
I think if possible the markdown linter should be (a) runnable locally before push and (b) automatically fix syntactical errors by itself as much as possible (like `gofmt` or similar tools). But "just" a normal CI tool is definitely a good start.

cl8n:
you can run it locally, markdownlint is a popular tool with good editor support. we have it configured as best as possible for osu-wiki already (https://github.com/ppy/osu-wiki/blob/master/.markdownlint.json), but need to add more custom rules

PR #3040, opened 2020-01-13
Fix link to Getting Your Map Modded, replace bubble for nominations, …

pimpG:
…remove kudosu and star priority references (replaced by hype)


cl8n:
check open PRs before making a new one. most of this is in #2788

PR #3043, opened 2020-01-13
Disable MD040 rule in markdownlint

TPGPL:


cl8n:
any instances of ```text are not necessary then

PR #3044, opened 2020-01-13
Fix markdownlint warnings in A-lettered articles

TPGPL:


cl8n:
i dont think there's any point in doing this for outdated articles that have other unreported issues. the next updates to the articles will likely encompass the whole thing and we'll have to check it again anyway

PR #3046, opened 2020-01-15
Override multiple top-level headers lint rule for news posts

cl8n:
we allow them here because there's effectively an even-more-top-level header displayed in the header image above the post

### DIFF
--- a/news/.markdownlint.json
+++ b/news/.markdownlint.json
@@ -1,0 +1,4 @@
+{
+    "extends": "../.markdownlint.json",
+    "MD025": false
+}


MegaApplePi:
So in the main config file, rule `MD033` is `false`. HTML is allowed everywhere?

cl8n:
iirc a small subset of html is allowed in normal articles, and pretty much any html is allowed in news posts

at the very least we use it for embedding youtube videos and stuff so I just disabled the rule at the time

MegaApplePi:
Alright, I assume that HTML comments might get flagged.

PR #3062, opened 2020-01-15
[FR] Beatmap Discussion

DoryanR:
Translation of Beatmap Discussion page


cl8n:
per #2486 please avoid translating outdated articles, they're likely to be rewritten

PR #3077, opened 2020-01-19
Group osu! UCI tournaments

cl8n:
cuz there's a new one finishing soon 🎉


cl8n:
i left to go make dinner, remembered I didn't do redirects and thought you might comment on it, ran back here and committed the change, apparently just a few seconds before you sent this message

6th sense wiki editing

wiki/Sitemap/en.md
```diff
@@ -330,6 +330,8 @@ tags:
     -   [osu! Riverside](/wiki/Tournaments/osu!_Riverside)
         -   [MALE](/wiki/Tournaments/osu!_Riverside/MALE)
         -   [SWEAT](/wiki/Tournaments/osu!_Riverside/SWEAT)
+    -   [osu! UCI](/wiki/Tournaments/osu!_UCI)
+        -   [iDOL@NSTER](/wiki/Tournaments/osu!_UCI/iDOL@NSTER)
```
MegaApplePi: The ASC says to not use any percent encoding characters, that `@` is going to show up as `%40`. Did we want to ignore it and change it in the ASC later?

cl8n: exclamation marks are already allowed so I don't really see the point of this

MegaApplePi: Oh, I thought exclamation marks wasn't encoded, otherwise it wouldn't have been listed there. Alright, will make some changes to that section later.

PR #3078, opened 2020-01-19
[FR] Game Modifiers

DoryanR:
Translation of Game Modifiers page


cl8n:
in the future please avoid translating articles marked as outdated. see the list in https://github.com/ppy/osu-wiki/issues/2486 for things not to translate, I think I already linked this to you once...

PR #3083, opened 2020-01-19
[DA] Add Welcome

Pennek:


MegaApplePi:
@Pennek I made a mistake while trying to apply a mark onto the article you added. I think I brought your pull request back to where it was, but could you check to make sure everything there is where you left it?

Pennek:
can confirm, everything is as i left it

PR #3088, opened 2020-01-20
Updated FC logo details

arflyte:
Updated FC usage details. Bolded "Do not" to make sure they know the examples are bad examples. Will update examples sometime later.

### DIFF
--- a/wiki/Brand_identity_guidelines/en.md
+++ b/wiki/Brand_identity_guidelines/en.md
@@ -45,28 +45,28 @@ Since osu! is community-driven, the osu! cookie logo is designed to be simple an
 
 ![](img/restrictions-bad-single.png)
 
-- Do not change the cookie's aspect ratio.
-- Do not rotate the cookie. The cookie must align to the medium's orientation or reader's eye when reading.
-- Do not change any of the cookie's element into something else.
-- Do not remove any of the cookie's element.
-- Do not resize any of the cookie's element independently.
-- Do not modify any of the cookie's element
-- Do not rearrange any of the cookie's element.
+- **Do not** change the cookie's aspect ratio.
+- **Do not** rotate the cookie. The cookie must align to the medium's orientation or reader's eye when reading.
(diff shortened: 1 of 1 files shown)


wiki/Brand_identity_guidelines/en.md
```diff
@@ -45,28 +45,28 @@ Since osu! is community-driven, the osu! cookie logo is designed to be simple an
...
-- Do not even think about changing the element. Come on. That is not even our logo. That is just some text in a circle.
+- **Do not** apply any fancy (or really tacky) effects on the cookie.
+- **Do not** place any extra elements inside the cookie.
+- **Do not** apply outlines to the cookie. Use a different cookie colour.
+- **Do not** even think about changing the element. Come on. That is not even our logo. That is just some text in a circle.
 
 ### Full colour cookie usage restrictions
 
-Since osu! is community-driven, the osu! cookie logo is designed to be simple and versatile; it can adapt to many designs with ease. There is no strict colour restriction to the colour of the cookie. All the restrictions applied on the single colour cookie applies to the full colour cookie too.
+Since osu! is community-driven, the osu! cookie logo is designed to be simple and versatile; it can adapt to many designs with ease. Please use the logo as it is, without any modification. All the restrictions applied on the single colour cookie applies to the full colour cookie too.
```
Coppertine: This seems contradictory to the original statement. It's saying that there isn't a strict colour restriction to the cookie and you can use it to the cookie as is from the sentence before. But the commit now says that even thought the community can change and the logo can change, you can't modify the logo 🤔

peppy: I believe this is intentional. Note that there is a rule to not change the full-colour logo colour.

cl8n: I guess "it can adapt to many designs with ease" *could* be misinterpreted as "you can change this logo to match your design".... this new sentence clears it up, though

PR #3093, opened 2020-01-22
Lint BanchoBot article

TPGPL:


peppy:
Rather than doing this on a per article basis, please look at doing it all in one single PR, alongside adding CI-based linting. It's not really productive otherwise.

PR #3097, opened 2020-01-24
Add modding guide for osu!mania

Feerum:
So yeah, i wrote this guide in the past weeks. 
It shall show interested modders for osu!mania first steps and some tips for people who are already modding in osu!mania and help them to improve. 

Got already a lot of help by some people with grammar and re-writing stuff (Thanks Hydria), but i'm sure there's still some work which needs/could be done to improve it to make it fit into the osu! wiki


cl8n:
a lot of this info covers all modes and could be organised into Modding and Beatmap_Discussion. whole article may fit better as Modding/osu!mania considering Modding is already something like a "guide", and it'd be easy to find there

thnx for writing this, I'll review as soon as i can

cl8n:
@Pennek may want to look at this if you're working on Beatmap Discussions

Pennek:
heya from what i can see this is a great article but im just a bit uncertain in how to deal with articles like this, since it:

_1. contains elements from other articles, mainly /Modding and /Beatmap Discussions.
2. What is said in #part 1 should preferably be available in Beatmap Discussions for all modes to use and could be linked to from this article.
3. other stuff could also be moved to their own articles_

I just want to get this straight; do we want to keep content that covers this much stuff around? it would maybe be worth it as standalone articles that tie different topics together? But what's the current consensus around this?

im not sure what the best solution is, but it's definitely gonna be worth it to move relevant parts from this into /Beatmap_Discussions

Pennek:
Oh yeah, another thing: would it be okay for me to nab some of these directly into Beatmap Discussions?

cl8n:
> I just want to get this straight; do we want to keep content that covers this much stuff around? it would maybe be worth it as a standalone article that ties different topics together? But what's the current consensus around this?

no consensus, this doesn't happen often. I'm okay with the idea of having duplicate content as long as there's some purpose to it (in this case it seems useful because it's trying to introduce new topics to the reader). but at the same time, this definitely shouldn't be the *only* place where such info is available

> but it's definitely gonna be worth it to move relevant parts from this into /Beatmap_Discussions [...] would it be okay for me to nab some of these directly into Beatmap Discussions?

just keep in mind that Beatmap Discussions isn't supposed to be a guide. the writing style will probably differ a bit, but the content may be very similar

---

I feel like this guide is kinda lacking focus on mania overall, the majority of it could be applied to all modes... it's going to be hard to update this in the future if similar articles come along for other modes and I think we shoudl figure this out now before it's a pain

cl8n:
maybe we can handle game modes for this similar to how Wookieepedia handles Canon and Legends? when an article appears in both, they duplicate the article and only change the relevant details, then put a little navigation thing at the top so you can choose which "version" of the article you're looking at. it's a neat way to handle minor differences in otherwise similar articles

that would be a lot easier with template support but it's doable now too.

Pennek:
> when an article appears in both, they duplicate the article and only change the relevant details, then put a little navigation thing at the top so you can choose which "version" of the article you're looking at. it's a neat way to handle minor differences in otherwise similar articles

would that also work for articles like this one, where it has content from multiple articles?

cl8n:
it's addressing the game mode problem, not the Beatmap_Discussions/Modding stuff. for that it's probably fine to update these all separately and add some comments to the top of each reminding editors to try to keep their info consistent

Feerum:
Mhm.. i was well aware that several parts of this guide could be found around the wiki somewhere else while writing it. Something like Timing, AiMod or how the Beatmap Discussion works.
But i decided to still add everything into it because many articles are sadly outdated.
Another reason why i wanted to add is to simply have everything at one place so people who really want to learn modding have everything at one place to find.

What could be cut would be maybe Beatmap Discussion because this article is still somewhat up-to-date (Maybe here and there some changes but overall i think it works). But at least keeping the link to it would be useful.

Also overall guides would look in every mode the same, the only thing that changes is when it comes to pattern and spreads. And maaaaybe hitsounding. Everything else is in every mode same.

cl8n:
no updates to the splitting of this article, but the content seems helpful as-is so i'm merging. cleaned it up a bit and added notes to the top

PR #3102, opened 2020-01-25
Remove fillers on brand identity guidelines

Parapass:


wiki/Brand_identity_guidelines/en.md
```diff
@@ -53,14 +53,14 @@ Since osu! is community-driven, the osu! cookie logo is designed to be simple an
...
-- **Do not** apply any fancy (or really tacky) effects on the cookie.
+- **Do not** apply any fancy or tacky effects on the cookie.
 - **Do not** place any extra elements inside the cookie.
 - **Do not** apply outlines to the cookie. Use a different cookie colour.
 - **Do not** even think about changing the element. Come on. That is not even our logo. That is just some text in a circle.
 
 ### Full colour cookie usage restrictions
 
-Since osu! is community-driven, the osu! cookie logo is designed to be simple and versatile; it can adapt to many designs with ease. Please use the logo as it is, without any modification. All the restrictions applied on the single colour cookie applies to the full colour cookie too.
+Since osu! is community-driven, the osu! cookie logo is designed to be simple and versatile. It can adapt to many designs with ease. Please use the logo as it is, without any modification. All the restrictions applied on the single colour cookie applies to the full colour cookie too.
```
Coppertine: semicolons have the special use case that you can use the word "It" with the object still being said what "It" is from the previous part of the sentence (in this cace, the osu! cookie logo). When you make a new sentence instead, the "It" doesn't have any meaning since there wasn't any object that it's refferring to.

Aitten: I reverted that part.

cl8n: revert was a good choice, but this explanation is wrong... you can use pronouns across sentences and the meaning would have been very clear in this case

PR #3125, opened 2020-01-30
Move images to an img directory

DoryanR:
I put an "img" folder in the place where there were loose images.

I didn't touch the "shared" folder because as it's related to a lot of articles I don't want to do anything stupid.


cl8n:
in your editor disable any settings that trim whitespace from the end of lines, at least for markdown files. two spaces at the end of lines can be significant and you accidentally removed a lot of those here

thnx for finding all of these, I made a note to look further into broken references soon

Issue #3128, opened 2020-01-31
Beatmap Editor Rework


Pennek:
As part of #2615 i am splitting the Editor rework out into its own issue as it is a heavy topic with lots of articles to be rewritten/changed.

Tasks:

- [ ] AiMod - overall well-written but text is dated and has some fluff. needs update with mention of flaws. mentions of alternative tools like Modding Assistant would also be helpful.
- [ ] Beat Snap Divisor - overall well-written but could use some formatting changes to cut the length down. needs mention of unsupported beat snap divisors. could possibly merge with [this](https://osu.ppy.sh/community/forums/topics/935026) user created guide
- [ ] Compose - needs a rewrite. images as well. mode-specific sections for this article would be great as its a huge mess right now. potential users reading these are here for one mode anyways. This would be a perfect place to implement https://github.com/ppy/osu-wiki/pull/3097#issuecomment-579857343_
- [ ] Design - i know little about storyboarding. mention of other tools is one thing that should be mentioned here at least
- [ ] Distance Snap - should this even be its own article? mapping community has moved on from using this as the denominator for emphasis. could implement into an "editor tools"-article
- [ ] Kiai Time - overall well-written. needs a bit of cleanup and linking to ranking criteria for some parts. could be merged into the 'Timing' article
- [ ] Menu - maybe not worth having since this is more like an 'editor shortcuts' article, which is a duplicate of everything in wiki/Shortcut_key_reference
- [ ] SB Load - again, i know little about storyboarding. is this relevant?
- [ ] Song Setup - needs update to fit the Ranking criteria. a lot of this needs to reference the metadata ranking criteria section. rewriting this to be a simple explanation of the song setup screen, instead of what is allowed and what isnt might be the way to go
- [ ] Timelines - I can see nothing wrong here. looks good
- [ ] Timing - could use some cleanup but otherwise good

Coppertine:
>  SB Load - again, i know little about storyboarding. is this relevant?
No, it isn't SB Load is to the GPU load, but most performance in Storyboard is just cpu.

cl8n:
even for "gpu load" it's a pretty irrelevant metric these days. it definitely doesn't need an entire article for an explanation

Pennek:
ahh, so that's what pulled me down in my bn app, lol!

Pennek:
what do you suggest we do with that article?

cl8n:
both sound good to me

Coppertine:
[...]

PR #3129, opened 2020-01-31
Rewrite Beatmap editor index

Pennek:
Closes part of #3128

Needs further reorganizing once more tasks in #3128 are completed.


wiki/Beatmap_Editor/en.md
```diff
@@ -1,13 +1,35 @@
...
+The **beatmap editor** is osu!'s in-game graphical interface for creating [beatmaps](/wiki/Beatmaps). This is where all [beatmapping](/wiki/Beatmapping) takes place, and all content for the game is created. Creating content for osu! is, and always will be, free. So feel free to jump right in and start beatmapping!
 
-- [Compose](Compose): The place where a majority of beatmapping happens. This is where you place circles, sliders, and spinners.
-- [Design](Design): The editor for where Storyboards are made; however, most mappers only use the "Background Image/Video" button here.
-- [Timing](Timing): The second place of where all new beatmaps begin. This controls the BPM (Beats Per Minute), Timing Points, hitsounds, and hitsound volume.
-- [Song Setup](Song_Setup): The first place of where all new beatmaps begin. This sets the values regarding the title, artist, version name, difficulty, combo colours, and more.
+The beatmap editor is separated into four main sections which can be seen on the top-left corner of the editor. The categories are:
 
-Other components of the beatmap editor include:
+- Compose
```
cl8n: this format of list + sections makes sense for longer documents of this type, but here I think it might be better to just have the paragraph for each part and get rid of the list/sections, since it's so short. or maybe those paragraphs could be the list entries themselves. it doesn't seem useful to have this mini-index when everything fits on a typical screen

wiki/Beatmap_Editor/en.md
```diff
@@ -1,13 +1,35 @@
 # Beatmap Editor
 
-The **beatmap editor** is osu!'s in-game graphical interface for creating [beatmaps](/wiki/Beatmaps). It is separated into four parts:
+The **beatmap editor** is osu!'s in-game graphical interface for creating [beatmaps](/wiki/Beatmaps). This is where all [beatmapping](/wiki/Beatmapping) takes place, and all content for the game is created. Creating content for osu! is, and always will be, free. So feel free to jump right in and start beatmapping!
```
cl8n: second to last sentence doesn't belong here, and the last sentence is too imperative (this article isn't meant to encourage people to do anything). as long as u can hit Beatmaps and Beatmapping in this first part and give some very minimal context for each that's probably good. something like the intro of https://osu.ppy.sh/help/wiki/Modding

also "all content for the game" isnt very explanatory and kinda misleading

wiki/Beatmap_Editor/en.md
```diff
@@ -1,13 +1,35 @@
...
+- Compose
+- Design
+- Timing
+- Song Setup
+
+For Beatmap Editor Shortcuts, see [Shortcut key reference](/wiki/Shortcut_key_reference).
+
+## Compose
+
+Compose is the section where a majority of beatmapping happens. Most mappers spend more time in this category than all of the other categories combined, since this is where gameplay elements like circles, sliders, and spinners are placed. To learn more about this section, see the [compose article](/wiki/Beatmap_Editor/Compose).
```
cl8n: instead of the repeated "to learn more, see this article", you could link the article to the first word. also, since the link goes to something under Beatmap_Editor, you don't need the `/wiki/Beatmap_Editor/` prefix

sometimes we use the "*Main article:*" note at the beginning of sections too, but for something so short its prob not worth doing that

cl8n: could link to Hit Objects since that's what it's all about

wiki/Beatmap_Editor/en.md
Pennek: im unsure how you want me to format this, but I'll try deleting the list and keep them as paragraphs instead

Issue #3143, opened 2020-02-02
Reorganizing and cross-connecting articles in the wiki


Loctav:
Hey!

We have debated in the recent days that the decreased interest in the osu!wiki might find its origin in its rather unorganized shape and presentation. While the wiki itself certainly lacks core functions (such as proper image embedding, etc.), I believe that we could improve the wiki by a tenfold when we treat it less like a stockpile of copy-pasted ancient forum posts.

A wiki thrives and lives from having each article be intertwined with each other. It lives from the idea that you can go to an article you find or look up and you can follow down the rabbit hole of information by clicking every important term and be sent to a new article that tells you more about it. 

Right now, majority of the articles are orphans or the information is scattered across various articles while making it incredibly hard to find specific information (because it is buried within miles long guides or articles or because it doesn't exist)

The aim here is to reorganize all present information and add missing information by following the simple logic of most wikis that use articles to define individual terms and explain more on-top of it. (e.g. you don't find a guide "How to visit Japan" on Wikipedia. You have an article about "Japan")

There are many articles present already that do exactly that (such as the mascot article or some hit object articles), but they lack integration into other articles via cross-linking. 

Ideally, every obscure osu! word (e.g. Hitsound, Hit Circle, Beatmap, osu!lazer, Kiai Time) present in any article should be clickable on its first mention and should lead to an article or a sub header of a larger article that defines this word and explains more in-depth what's the matter with it.

For that purpose, I have created a google doc that lists a whole lot of terms that require stubs to be created, barebone articles that can be linked to when found on existent ones. They (almost) all should be separate articles for each and contain a basic definition of each. Later on, we can add more explanations, depth and visual appeal to each of them. For some terms, we can consider grouping them up into bigger articles (e.g. Sliderhead doesn't need an own article. We only need Slider). 

While building these stubs, we need to cross link each stub by referring to each other from the very beginning. 

You can find my aforementioned list of terms here: https://docs.google.com/spreadsheets/d/1sYphJvwxAU0av_dQVwdqYgRC3UqW6cv-TZSl2pX34oE/edit#gid=0

Loctav:
The list is nowhere near complete. It's a whole bunch of stuff I came up with by just browsing existing articles such the RC and the website itself. We can expand the list as we go, but we need a starting point somewhere.

Endaris:
[...]

PR #3144, opened 2020-02-02
Add Overlap stub

Pennek:
Still need input on how tags should look


wiki/Mapping_Techniques/Overlap/en.md
```diff
@@ -0,0 +1,9 @@
+---
+stub: true
+---
+
+# Overlap
+
+![Rotational_Symmetry.png](/img/overlap.jpg "An example of a hitcircle overlapping a sliderhead.")
```
MegaApplePi:
```suggestion
![Rotational_Symmetry.png](img/overlap.jpg "An example of a hitcircle overlapping a sliderhead.")
```

the beginning `/` in the link basically means to start with the root folder, to use current directory remove it or use `./`

MegaApplePi:
Temporarily blocking for now. I want to see the general game plan and other specifics in #3143 first before jumping the gun on this.


You also mentioned about something needing input on how tags should look. Do you mean that `stub: true` part?

Pennek:
no problem for the blocking. these take little effort.

yes the stub. I want that cleared up before pushing these forward, since we dont have a conses on how to add them for now

wiki/Mapping_Techniques/Overlap/en.md
```diff
@@ -0,0 +1,9 @@
+---
+stub: true
+---
+
+# Overlap
+
+![overlap.png](./img/overlap.jpg "An example of a hitcircle overlapping a sliderhead.")
```
MegaApplePi: The alt text should be a description of the image. If you don't have one, it should be left blank.

Pennek: what more would you add to this? the curvature of the slider? it simply is a hitcircle overlapping a sliderhead

MegaApplePi: Just to be sure we are talking about the same thing,
```
![overlap.png](./img/overlap.jpg "An example of a hitcircle overlapping a sliderhead.")
  ^ This is the alt text          ^ This is the title/hover text, also used as the caption on web
```

Alt text only displays when the image doesn't load, but it is also used for accessibility purposes too. If you don't have an image description for the alt text, then don't put anything in there.

Pennek: that clears it up. so leave it blank?

MegaApplePi: Yeah, that'll be fine.

cl8n:
for now it's probably best to do what Loctav suggested, add a note at the bottom saying this is a stub. I don't think the "stub: true" thing is ever going to turn into a web feature cuz this is starting to get pretty redundant with all of these `marker: true` tags that have similar effects

cl8n:
@MegaApplePi this would be 1. from https://github.com/ppy/osu-wiki/issues/3143#issuecomment-581197907 , and even if that were never to be completed i dont think adding these would hurt anything

all that comes to mind for me is that we may need to revisit how indices and Sitemap display this kind of stuff.

MegaApplePi:
Here are some notes on what was discussed through discord regarding the stub mark:

image

---

For now, we'll mark stubs with an html comment to help identify them in the future. If you think a stub can be further expanded, you could also include the note that clayton wrote at the bottom.

PR #3145, opened 2020-02-02
Add Stack stub

Pennek:
depends on #3143


cl8n:
should mention that stacks aren't (usually) made by the mapper, the game stacks objects automatically whenever the mapper puts objects very close together

PR #3169, opened 2020-02-05
[FR] Update Organisations/osu! UCI

DoryanR:
Translation of osu! UCI page


wiki/Organisations/osu!_UCI/fr.md
```diff
@@ -1,36 +1,39 @@
 ---
-outdated: true
 tags:
-  - Californie
```
cl8n: the tags should be things that people using the French locale would search for--- so you probably want French translations of common nouns here

Issue #3163, opened 2020-02-04
Add Article Stubs (Modding)


Loctav:
See the main issue https://github.com/ppy/osu-wiki/issues/3143
See the project: https://github.com/ppy/osu-wiki/projects/3

Here you find all article stubs that need to be created. Some articles might already exist. Some articles can be deemed unnecessary as singular article and be summed up into a bigger article. 

Many of these singular topics are included as subset in some articles already or information is scattered around multiple articles. These articles might need to be split up into multiple ones and information must be pulled together into one common article then (e.g. https://github.com/ppy/osu-wiki/tree/master/wiki/Glossary)

Each article should only contain information of one topic. 

You can claim articles beforehand. Just comment here.

- [x] Modding v1 #6260
- [x] Beatmap Discussion (aka Modding v2) #3126
- [x] Time Stamp #3348
---
- [x] Ranked
- [x] Qualified
- [x] ~~Disqualified~~ (covered in Beatmap ranking procedure)
- [x] Unranked
- [x] ~~Nominated~~ (covered in Beatmap ranking procedure)
- [x] ~~Nuked~~ (covered in Modding/Nuke)
- [x] Loved
- [x] Approved
---
- [x] Proto Bubble #3173
- [x] Bubble #6239
- [x] Bubble Pop #6239
- [x] Veto Mediation
- [x] Veto
- [ ] Nomination
---
- [x] Kudosu #3307
- [x] Hype #3399
---
- [x] Mapping Assistance Team #3173
- [x] MAT Leaders #3173
- [x] Beatmap Approval Team / Beatmap Appreciation Team #3173
- [x] BAT Leaders #3173
---
- [x] NAT meetings #3461

venix12:
i can probably do most of these

JBHyperion:
With some research I could probably look into the various "Teams" stubs, though they were a little before I really got involved in the scene.

Everything else should be doable for me also, though isn't this pretty much a deconstructed Glossary in its current format? 🤔 Guess this is up to interpretation with:

> Some articles can be deemed unnecessary as singular article and be summed up into a bigger article.

Pennek:
> isn't this pretty much a deconstructed Glossary in its current format?

as i've understood it that's pretty much the point for now. after we've built the foundation of stubs we can begin to link them together and go into more detail. eg. the 'slider'-article could go way more indepth about what a slider is, anchors, notepad creation, external programs, etc.

it's like constructing the wiki rabbit hole

[...]

PR #3173, opened 2020-02-05
Add defunct modding usergroups & pubble stubs

JBHyperion:
Part of #3163 - Adds stub wiki articles for the MAT, BAT, MAT Leaders, BAT Managers and the proto-bubble icon

Looking for help with history of the BAT stub article in particular


cl8n:
some formatting things to keep in mind

- metadata has to be surrounded by `---` to do anything, for example
  ```markdown
  ---
  tags:
    - MAT
  ---

  # Mapping Assistance Team
  ```
- titles don't need linguistic articles("the"/"a"/"an") at the start, so here `The Beatmap Appreciation Team` should be `Beatmap Appreciation Team`
- sections don't need to be qualified by the title again, here `History of the BAT` should be `History`
- an article not clearly missing any info for what it's trying to be isn't a stub
- first mentions of acronyms or alternate names should be italicised to show that you may refer to them again like that
- should probably refer to acronyms in further usages instead of the full name (repeating something like "Quality Assurance Team" five times gets tiring to read)

wiki/Modding/BAT/en.md
```diff
@@ -0,0 +1,26 @@
...
+
+Existing members of the Beatmap Appreciation Team at this time were given the option of remaining in the BAT to focus on beatmap qualification (giving up their moderation powers and responsibilities in the process), or transitioning to the Quality Assurance Team, who retained these moderation powers and responsibilities.
+
+From the 31st of January onwards, the group was renamed the "Beatmap Appreciators", (and shortly thereafter the "Beatmap Nomination Group"). At this time, the rules on admission of new members to the BAT were changed to allow entry via application only on a monthly basis.
```
dorothy3242:
```suggestion
From the 31st of January onwards, the group was renamed the "Beatmap Appreciators", and shortly thereafter the "Beatmap Nomination Group". At this time, the rules on admission of new members to the BAT were changed to allow entry via application only on a monthly basis.
```

if you were to remove the part currently in parenthesis you'd get `From the 31st of January onwards, the group was renamed the "Beatmap Appreciators", .` which makes no sense. I don't think the parenthesis are really needed around the second half of the sentence either, so best to just remove.

JBHyperion: Fixed with clayton's suggestions

JBHyperion:
@cl8n Cleaned up everything from your review regarding style and metadata tagging. Regarding

> an article not clearly missing any info for what it's trying to be isn't a stub

I can't guarantee these articles aren't missing information, since everything written here took place before I started modding and I can only go based on old hearsay and buried forum posts. If someone with better knowledge can confirm they contain everything they need to contain, then the stub tag can be removed.

peppy:
pubble lives on!

wiki/Modding/MAT_Leaders/en.md
```diff
@@ -0,0 +1,19 @@
...
+  - Mapping Assistance Team Leaders
+---
+
+# MAT Leaders
+
+**MAT Leaders** oversaw the [Mapping Assistance Team](/wiki/Modding/Mapping_Assistance_Team) (*MAT*). They were responsible for managing and mentoring team members, as well as raising awareness of important announcements and shaping the future goals and direction of the team. There was only one MAT leader at a time.
+
+## History
+
+![][flag_AU] [Ephemeral](https://osu.ppy.sh/users/102335) was chosen as the inaugural MAT Leader at the group's formation on the 29th of December, 2009.
```
cl8n, line 12: either this is the incorrect date, or it should say "shortly after the group's formation", because the group was formed on the 27th

JBHyperion: Correct, this should be the 27th

Issue #3151, opened 2020-02-03
Add Article Stubs (Beatmaps)


Loctav:
See the main issue https://github.com/ppy/osu-wiki/issues/3143
See the project: https://github.com/ppy/osu-wiki/projects/3

Here you find all article stubs that need to be created. Some articles might already exist. Some articles can be deemed unnecessary as singular article and be summed up into a bigger article. 

Many of these singular topics are included as subset in some articles already or information is scattered around multiple articles. These articles might need to be split up into multiple ones and information must be pulled together into one common article then (e.g. https://github.com/ppy/osu-wiki/tree/master/wiki/Glossary)

Each article should only contain information of one topic. 

You can claim articles beforehand. Just comment here.

- [x] Beatmap
- [x] ~~Beatmap Download - Has a [guide in Chinese](https://osu.ppy.sh/wiki/zh/Guides/Searching_and_Downloading_Beatmaps) that needs to be translated/adapted to English~~ https://github.com/ppy/osu-wiki/issues/8128
---
- [x] **Beatmap submission**
- [x] Beatmap Set #3405
- [x] Beatmap Spread https://github.com/ppy/osu-wiki/pull/9036
---
- [x] .osu file
---
- [x] **Difficulty**
- [x] Difficulty name
---
- [x] **Beatmap creator**
- [x] Guest difficulty
- [x] Guest mapper (covered in *guest difficulty*)
- [x] Collaborative difficulty #3404
---
- [x] **Beatmap creation** https://osu.ppy.sh/wiki/en/Beatmapping
- [x] Stack leniency https://github.com/ppy/osu-wiki/pull/9028
- [x] Beat Snap https://osu.ppy.sh/wiki/en/Client/Beatmap_editor/Beat_Snap
- [x] Grid Snap https://osu.ppy.sh/wiki/en/Grid_snapping
- [x] Time-distance equality #3488
- [x] Distance Snap https://osu.ppy.sh/wiki/en/Client/Beatmap_editor/Distance_snap
- [x] Overlap #3144
[...]

PR #3174, opened 2020-02-05
Add Health articles

Pennek:
resolves part of #3151


wiki/Beatmapping/Health_Drain/en.md
```diff
@@ -0,0 +1,7 @@
+# Health drain
+
+Health drain refers to the natural drain of the health bar that occurs when playing a [beatmap](/wiki/Beatmaps) while disregarding the influence of [hit objects](/wiki/Hit_Objects). This drain is dependent on the HP difficulty setting, which is set by the mapper of a beatmap. This value ranges from 0 to 10, where 10 is the highest and most punishing setting and 0 is the most forgiving. The HP difficulty setting can be changed alongside other difficulty settings in the [Beatmap Editor](/wiki/Beatmap_Editor). Doing this will change a Ranked beatmap's status to unranked and will erase all local scores on it.
+
+osu!standard and osu!catch have health drain at a constant rate throughout the course of a beatmap, with the exception of breaks. As for osu!taiko and osu!mania, health drain is disabled, so only the hit objects will affect the health bar.
```
JBHyperion: > As for osu!taiko and osu!mania, health drain is disabled, so only the hit objects will affect the health bar.
Kinda makes it sound like HP doesn't matter, might be better worded as "In osu!taiko and osu!mania, health drain occurs when an object is not successfully hit."

wiki/Beatmapping/Health_Drain/en.md
```diff
@@ -0,0 +1,7 @@
+# Health drain
+
+Health drain refers to the natural drain of the health bar that occurs when playing a [beatmap](/wiki/Beatmaps) while disregarding the influence of [hit objects](/wiki/Hit_Objects). This drain is dependent on the HP difficulty setting, which is set by the mapper of a beatmap. This value ranges from 0 to 10, where 10 is the highest and most punishing setting and 0 is the most forgiving. The HP difficulty setting can be changed alongside other difficulty settings in the [Beatmap Editor](/wiki/Beatmap_Editor). Doing this will change a Ranked beatmap's status to unranked and will erase all local scores on it.
```
JBHyperion: Para is pretty long, would consider splitting at "The HP difficulty setting can be changed...", since here you're discussing how to change the HP drain, rather than what HP drain _actually_ is.

wiki/Beatmapping/Health_Drain/en.md
```diff
@@ -0,0 +1,7 @@
+# Health drain
```
JBHyperion: Could preface this with an "HP" tag, e.g.

\---
tags:
  \- HP
\---

wiki/Beatmapping/Health_Drain/en.md
Pennek: yea, still need to change some things regarding misses and geki/katu health gain. Still unsure if i should split it up more and put the full HP description under the health bar article. Or do the opposite and link to hpdrain from UI health bar article once that exists

Pennek: probably going to end up doing the latter of those two.

cl8n: imo latter option is better, this article seems more about the mechanics and the UI article would be about how it's displayed& skinned

wiki/Beatmapping/Health_Drain/en.md
```diff
@@ -1,7 +1,99 @@
...
+
+## Health drain
+
+A common system across game modes is known as health drain, which refers to the natural drain of the health bar that occurs when playing a beatmap. This value is also known as *HP drain*, and is the one mostly referred to when talking about health. Health drain is dependent on the HP difficulty setting, which is set by the mapper of a beatmap. HP values range from 0 to 10, where 10 is the highest and most punishing setting and 0 the most forgiving. On HP10 the health drains rapidly and a single miss may be enough to fail a player, whereas on HP0 multiple combos can be missed before failing a player.
```
JBHyperion: Probably don't need to discuss the difference between HP0 and HP10 in both Health and Health drain sections. I would remove

> Health drain is dependent on the HP difficulty setting, which is set by the mapper of a beatmap. HP values range from 0 to 10, where 10 is the highest and most punishing setting and 0 the most forgiving.

from here and keep the rest.

wiki/Beatmapping/Health_Drain/en.md
```diff
@@ -1,7 +1,99 @@
...
+
+## osu!standard
+
+In osu!standard health will decrease at a steady rate (passive health drain), but can be replenished by tapping notes at the right time or spinning spinners. Score bursts will appear after tapping hit circles, changing score and health values depending on the accuracy of the hit. A perfectly timed hit (300) will recover health more than a badly timed hit (50). The player also receives larger boosts if he achieves a [katu](#katu) or a [geki](#geki) at the end of a combo. A miss will drain health instead. If spinners are cleared early, continuing to spin will gain additional health back. For an overview of health recovery and health loss see below:
```
JBHyperion: he achieves -> they achieve (in osu! and osu!mania sections)

wiki/Beatmapping/Health_Drain/en.md
```diff
@@ -1,7 +1,99 @@
...
+
+Health loss:
+
+- Passive health drain.
+- Missing a hit circle.
+- Tapping a slider head early.
+- Missing a slider tick, or slider reverse arrow.
+- Failed to achieve at least a 50 on a spinner.
+
+Health reserved:
```
JBHyperion: I would probably just make a new sentence here (or better yet at the start) mentioning that health drain won't take effect before the start of the beatmap or during breaks in any scenario, since I think this applies to all modes?

wiki/Beatmapping/Health_Drain/en.md
```diff
@@ -1,7 +1,99 @@
...
+- Constant life drain.
+- Not collecting any fruits or juice drops.
+
+Health reserved:
+
+- None.
+
+## osu!mania
+
+In osu!mania passive health drain is disabled, so only hit objects will affect the health bar. Score bursts will appear after hitting notes, changing score and health values depending on the accuracy of the hit. A perfectly timed hit (300) will recover health more than a badly timed hit (50). The player also receives larger boosts if he achieves a [katu](#katu) or a [geki](#geki) at the end of a combo. A miss will drain health instead. Health bar regeneration occurs slowly while holding hold notes. For an overview of health recovery and health loss see below:
```
JBHyperion: Does a rainbow 300 restore more health than a regular 300? I honestly don't know but it might be worth mentioning this if it's the case.

wiki/Beatmapping/Health_Drain/en.md
Pennek: This serves more as an overview of ALL things that apply to a game mode, so that everything that affects a game mode is easily accessible. This leaves no room for confusion.

but i agree it's definitely worth mentioning in the health drain section

wiki/Beatmapping/Health_Drain/en.md
Pennek: no idea, cant seem to find any info indicating this on the wiki either

Pennek: had confirmation that rainbow 300s do indeed gain more hp back

wiki/Beatmapping/Health/en.md
```diff
@@ -0,0 +1,101 @@
...
+- HP Drain
+- Passive Drain
+- Drain
+---
+
+# Health
+
+Health or more commonly known as *HP*, is the value that decides if a player has failed or passed a [beatmap](/wiki/Beatmaps). Health can be seen in the health bar which updates as a beatmap progresses. Health is largely determined by the HP difficulty setting, which is set by the mapper of a beatmap.
+
+The HP difficulty settting controls how heavily a player is penalized for missing notes and how much health is gained back by accurately hitting [hit objects](/wiki/Hit_Objects). HP values range from HP0 to HP10, where HP10 is the highest and most punishing setting and HP0 the most forgiving. On HP10 the health drains rapidly and a single miss may be enough to fail a player, whereas on HP0 multiple combos can be missed before failing a player. The HP difficulty setting can be changed alongside other difficulty settings in the [Beatmap Editor](/wiki/Beatmap_Editor). Doing this will change a Ranked beatmap's status to unranked and will erase all local scores on it.
```
MegaApplePi:
```suggestion
The HP difficulty settting controls how heavily a player is penalized for missing notes and how much health is gained back by accurately hitting [hit objects](/wiki/Hit_Objects). HP values range from HP0 to HP10, where HP10 is the highest and most punishing setting while HP0 is the most forgiving. On HP10, health drains rapidly and a single miss may be enough to fail the player. On HP0, multiple combos can be missed before the player fails. The HP difficulty setting can be changed alongside other difficulty settings in the [Beatmap Editor](/wiki/Beatmap_Editor). Doing this will change a Ranked beatmap's status to unranked and will erase all local scores on it.
```

MegaApplePi: > Doing this will change a Ranked beatmap's status to unranked and will erase all local scores on it.

Wouldn't changing anything in the beatmap and saving it locally erase all local scores for that beatmap? If so, it seems strange to me for why this article says this. Maybe put it in the beatmap editor article instead or make sure that all articles relating to the beatmap editor/song settings also has this too?

wiki/Beatmapping/Health/en.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Health drain
```
MegaApplePi: I think this section should be its own article/stub. It seems to be talking more about the gameplay mechanics of health, so I don't think it fits well in this article.

wiki/Beatmapping/Health/en.md
Pennek: agree, ill remove for now and make a reminder to put it in the beatmap editor article when i get to that

wiki/Beatmapping/Health/en.md
Pennek: I've thought about it for a bit and come to the conclusion that multiple sections here should be split up into their own thing to keep articles focused. I've gone ahead and tried to do that with my next commit

wiki/Beatmapping/Health/en.md
```diff
@@ -0,0 +1,101 @@
...
+- Health drain.
+- Missing fruits or juice drops.
+
+Health reserved:
+
+- Break time.
+
+## osu!mania
+
+In osu!mania only hit objects will affect the health bar, as [health drain](/wiki/Beatmapping/Health_drain) is disabled.
```
mcendu: Note: health drain is not disabled in lazer.

Pennek: It makes no sense to start detailing lazer when it is unfinished. will require too much upkeep as lazer is changed

cl8n: i'd also expect game mechanics on ranked maps to eventually match osu stable

PR #3175, opened 2020-02-05
Add Circle size article

Pennek:
Resolves part of #3151


cl8n:
`git merge master` should fast-forward by default to avoid making the extra commits at the start. how did you make these?

Pennek:
no idea, i used the `push to` command inside VSCode to push to origin

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
+---
+tags:
+- CS
```
MegaApplePi:
```suggestion
  - CS
```

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
...
+---
+
+# Circle Size
```
MegaApplePi:
```suggestion
# Circle size
```

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+Circle Size, commonly abbreviated as *CS*, is a difficulty setting which determines the size of [hit objects](/wiki/Hit_Objects) on [beatmaps](/wiki/Beatmaps). The values range from 2 to 7; higher circle size values will make hit objects smaller. Using the highest value will make hit objects as big as the cursor. Most beatmaps use a value between 3 and 5, while values between 5 and 6 are sometimes used and values of 6 and above are almost never used. High circle size values are likely to be seen when using the [Hard Rock](/wiki/HR) mod as it increases circle size. Conversely, the [Easy](/wiki/HR) mod will lower circle size.
```
MegaApplePi:
```suggestion
Circle size, commonly abbreviated as *CS*, is a difficulty setting which determines the size of [hit objects](/wiki/Hit_Objects) on [beatmaps](/wiki/Beatmaps). The value for circle size ranges from 2 to 7; where a higher value will make hit objects smaller. Using the highest value will make hit objects as small as the cursor. Most beatmaps use a value between 3 and 5, while values between 5 and 6 are sometimes used. Values between 6 and 7 are almost never used. Circle size values higher than 7 or lower than 2 can be seen when the [Hard Rock](/wiki/HR) or [Easy](/wiki/EZ) mods are used respectively.
```

MegaApplePi: > Using the highest value will make hit objects as big as the cursor.

Would be incorrect if the cursor size is set other than 1X (or automatic cursor size was enabled)... not sure how technical we want to be. The suggestion I wrote for this line switches "big" with "small", but I just wanted to point this out, I guess.

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+It is possible to set circle size values outside normal bounds by manually editing the [.osu file](/wiki/osu!_File_Formats/Osu_(file_format)). For ranked beatmaps this is to be avoided since it is considered unrankable by the [ranking criteria](/wiki/Ranking_Criteria).
```
MegaApplePi:
```suggestion
It is possible to set the circle size value outside normal bounds by manually editing the [.osu file](/wiki/osu!_File_Formats/Osu_(file_format)). However, this is considered unrankable by the [ranking criteria](/wiki/Ranking_Criteria).
```

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+## osu!standard
+
+In osu!standard only the size of circles and sliders are adjusted through CS, while spinners remain unaffected. Mathematically, circle size is derived through the following formula:
```
MegaApplePi:
```suggestion
In osu!standard, only the size of hit circles and sliders are adjusted through CS, while spinners remain unaffected. Mathematically, circle size is derived through the following formula:
```

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+`r = 32 * (1 - 0.7 * (CS - 5) / 5)`, where `r` is the radius measured in osu!pixels. Alternatively, this can be written as:
+
+`r = 54.4 - 4.48 * CS`
+
+## osu!catch
+
+In osu!catch circle size determines the size of both the catcher and the fruit.
```
MegaApplePi:
```suggestion
In osu!catch, circle size determines the size of both the catcher and the fruit.
```

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+## osu!mania
+
+In osu!mania, circle size refers to the total number of columns on a beatmap.
+
+## osu!taiko
+
+The circle size setting is visible in osu!taiko beatmaps but does not do anything that affects gameplay.
```
MegaApplePi:
```suggestion
In osu!taiko, circle size does not affect gameplay.
```

wiki/Beatmapping/Circle_Size/en.md
Pennek: agree, this seems a bit incorrect. this was pulled straight from another article iirc. think I'll clean it up instead

wiki/Beatmapping/Circle_Size/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+# Circle size
+
+Circle size, commonly abbreviated as *CS*, is a difficulty setting which determines the size of [hit objects](/wiki/Hit_Objects) on [beatmaps](/wiki/Beatmaps). The value for circle size ranges from 2 to 7; where a higher value will make hit objects smaller. Most beatmaps use a value between 3 and 5, while values between 5 and 6 are sometimes used. Values between 6 and 7 are almost never used. Circle size values higher than 7 or lower than 2 can be seen when the [Hard Rock](/wiki/HR) or [Easy](/wiki/EZ) mods are used respectively.
+
+It is possible to set the circle size value outside normal bounds by manually editing the [.osu file](/wiki/osu!_File_Formats/Osu_(file_format)). However, this is considered unrankable by the [ranking criteria](/wiki/Ranking_Criteria).
+
+## osu!standard
+
+In osu!standard, only the size of hit circles and sliders are adjusted through CS, while spinners remain unaffected. Mathematically, circle size is derived through the following formula:
```
MegaApplePi:
```suggestion
In osu!standard, only the size of hit circles and sliders are adjusted through circle size, while spinners remain unaffected. Mathematically, circle size is derived through the following formula:
```

I just noticed that in the other game mode sections, they actually spell out circle size but this one only uses CS

wiki/Beatmapping/Circle_size/en.md
```diff
@@ -0,0 +1,33 @@
+---
+tags:
+  - CS
+---
+
+# Circle size
+
+**Circle size** (***CS***) is a difficulty setting which determines the size of [hit objects](/wiki/Hit_Objects) on [beatmaps](/wiki/Beatmaps). The value for circle size ranges from 2 to 7, where a higher value will make hit objects smaller. Most beatmaps use a value between 3 and 5, while values between 5 and 6 are sometimes used. Values between 6 and 7 are almost never used. Circle size values higher than 7 or lower than 2 can be seen when the [Hard Rock](/wiki/Game_Modifiers#hard-rock) or [Easy](/wiki/Game_Modifiers#easy) mods are used respectively.
```
cl8n: saying it ranges from 2 to 7 seems wrong, cuz 0-10 are all valid too. if you want to mention that the song setup panel recommends 2-7 i think u should make that more clear

cl8n: > Most beatmaps use a value between 3 and 5, while values between 5 and 6 are sometimes used. Values between 6 and 7 are almost never used.

seems pointless if there's no citation or explanation

wiki/Beatmapping/Circle_size/en.md
```diff
@@ -0,0 +1,33 @@
...
+
+It is possible to set the circle size value outside normal bounds by manually editing the [.osu file](/wiki/osu!_File_Formats/Osu_(file_format)). However, this is considered unrankable by the [ranking criteria](/wiki/Ranking_Criteria).
+
+## osu!standard
+
+In osu!standard, only the size of hit circles and sliders are adjusted through circle size, while spinners remain unaffected. Mathematically, circle size is derived through the following formula:
+
+`r = 32 * (1 - 0.7 * (CS - 5) / 5)`, where `r` is the radius measured in osu!pixels. Alternatively, this can be written as:
```
cl8n: should just use the second version of the formula

wiki/Beatmapping/Circle_size/en.md
```diff
@@ -0,0 +1,33 @@
...
+
+## osu!catch
+
+In osu!catch, circle size determines the size of the catcher and the fruit.
+
+## osu!mania
+
+In osu!mania, circle size refers to the total number of columns on a beatmap.
+
+## osu!taiko
```
cl8n: game mode sections are usually in order std->taiko->catch->mania

wiki/Beatmapping/Circle_size/en.md
```diff
@@ -0,0 +1,33 @@
...
+
+In osu!taiko, circle size does not affect gameplay.
+
+<!--TODO: Insert links to difficulty settings, ranked, osu!pixels, circles, sliders, spinners, catcher, fruit, columns, game modes -->
```
cl8n: could add some of these now ;)

wiki/Beatmapping/Circle_size/en.md
Pennek: agree

Pennek: yup deleted!

wiki/Beatmapping/Circle_size/en.md
Pennek: i dont think simplifying a formula like this and calling it a day is good when potential users may wonder why those exact values were used. I think i'll delete the simplification though.

cl8n: > when potential users may wonder why those exact values were used

the formula u kept also doesn't solve this issue either. to a reader it's way more complicated for no reason. if you want the constants to be explained further (this is already partially simplified, fwiw) pls add a todo comment instead

cl8n: poke

Pennek: will use the second version of the formula

PR #3199, opened 2020-02-09
Fix formatting in FAQ article

TPGPL:
posting this as a standalone PR because it's *big enough*

---


cl8n:
while working on these, could you add any new problems you come across to the list in #2275 ? doesn't have to be very organized or detailed, I'm using that issue as notes for new markdownlint rules to add

PR #3201, opened 2020-02-10
Update BTT5 for Round of 16

fajar13k:
- Add RO32 match results
- Add RO16 mappool
- Staff reorganisation


cl8n:
pls pull master into this branch if you're going to reuse it again, my aging hard drive doesn't like to checkout all these files ;__ ;

PR #3206, opened 2020-02-10
Use image figures and clean up formatting in Community Contributor news post

cl8n:
images parsed into figures will be centered by default, no need for the raw html


bdach:
> images parsed into figures will be centered by default, no need for the raw html

...as long as they're the single image in a paragraph, I think ([source for claim](https://github.com/ppy/osu-web/blob/437e30252922b983e791abdf4265f60f8091268e/app/Libraries/Markdown/OsuMarkdownProcessor.php#L223-L225))

cl8n:
yep, in this case they are alone in a paragraph

cl8n:
@peppy ^probably a project setting to make it work as intended

peppy:
Aha, I have enabled it now. Let's see how it goes.

PR #3215, opened 2020-02-12
[FR] Contests/Monthly Beatmapping Contest

DoryanR:
Translation of Monthly Beatmapping Contest page


DoryanR:
I'm delete my pr as they're not good, so don't be surprised.

PR #3231, opened 2020-02-18
Fix formatting in Game Modes folder

TPGPL:


Joehuu:
I have pending reviews to fix number of lines for all translations, but there are some lines that I can't review. Will add commits locally instead.

PR #3238, opened 2020-02-21
Remove all mention of "intersubjective" from the CoC

Ephemeralis:
Intersubjective is a silly word and conveys a mostly wrong impression of what the issues it refers to represent.

I've altered wording to refer to 'social issues' instead and expanded that this issue classification exists largely because of self-built communal expectations spanning across several years of creation.


wiki/Rules/Code_of_Conduct_for_Modding_and_Mapping/en.md
```diff
@@ -17,10 +17,10 @@ The **Code of Conduct for Modding and Mapping** is a set of rules and guidelines
...
+-   When making suggestions, try to differentiate between what is objectively bad or wrong, and what you simply dislike on a personal level. **All types of suggestion are fine as long as they are put forward accordingly**.
     -   **Objective Issues** - Prohibited by rulesets and must be fixed.
-    -   **Intersubjective Issues** - Not prohibited by rulesets, but are agreed upon as being "wrong" by a majority of people. Should be fixed in most cases unless there are opposing intersubjective arguments with equal size, or if the mapper can exhaustively explain why they feel it is acceptable
-    -   **Subjective Issues** - Something that you believe or think on a personal level. Might become intersubjective if many people agree with it.
+    -   **Social Issues** - Not prohibited by rulesets, but are agreed upon as being "wrong" by a majority of people, usually because they impinge upon or challenge established traditions or "unspoken" expectations as a whole. These should be fixed in most cases unless there is considerable, well-founded reasoning to the contrary, as many of these expectations have been naturally created across many years of collaborative creation.
```
JBHyperion: For me, the way this first is worded now swings a little too far in the other direction - that merely attempting to do something different is wrong.

I think you can strike a better balance by somehow clarifying that it's challenging the expected norms in a _disruptive_ way that we're trying to dissuade, rather than trying to change anything at all, under any circumstance.

cl8n: it's also missing the "unless there are opposing intersubjective arguments with equal size" part, I think that's still relevant to include because not every established tradition will be consistent across all of osu!'s mappers/modders

cl8n:
lgtm after resolving jbh

Naxesss:
Not sure about the term "social issues" because that makes it seem like we're talking about something completely different when people hear it, having heard of it prior and all, unlike "intersubjective".

Either way, I'm not even sure the entire "subjective"/"objective" wording is a good idea to begin with. In this case "Subjective" just means an opinion, "Intersubjective" just means a shared opinion, and "Objective" just means a rule (i.e. an established shared opinion). It's not even equivalent to the scientific expression because there's plenty of rules that are not objective. I think we should just stick the simpler terms rather than redefining things, for example look at this version:

- When making suggestions, try to differentiate between [Ranking Criteria](https://osu.ppy.sh/help/wiki/Ranking_Criteria) rules and your own opinions. For example, if you dislike something, but it is not prohibited by rules, do not post it as if it must be fixed.
    - **Opinion** - Something that you agree with.
    - **Shared opinion** - Something that multiple people agree with.
    - **Rule** - Something that multiple people have agreed with and has since been established and must be fixed.

See the similarity?  Rewritten like this you'll notice the subpoints don't even make sense to write out because they don't need to be explained.

From here we can continue by adding that *the reason behind* a rule or opinion is much more important than *the fact that it is* a rule or opinion (the latter something the current version and revision are hammering hard on). Consider the following:
- A mistake mappers and modders seem to do commonly is calling something "subjective" as if a valid reason to dismiss it, rather than saying why they disagree with it.
- It wouldn't make sense to apply something you have good reason not to, even if it is a rule; in these cases we should look into changing the rule rather than enforcing it blindly.

cl8n:
if you don't plan to update this further, give the word for someone to take it over or move to RC forum

Ephemeralis:
This is probably better served by discussion on the RC forums and a wholesale wording rework as described by Naxess.

cl8n:
there's a thread on rc forum now, I'm going to lock this so ppl dont reply to the wrong place

PR #3242, opened 2020-02-21
Avoid need for duplicate user tags

JBHyperion:
- [ ] Reach a consensus in this [thread](https://osu.ppy.sh/community/forums/topics/1025897)

https://osu.ppy.sh/beatmapsets/1101184 required a disqualification to add the username "10nya" to tags as the hitsounder of the beatmap, despite already being searchable via this username using the Artist field.

Adds a clause to the Technical Metadata Rules section to negate the need for duplicate username tags in cases such as this where the contributing user is already searchable and properly credited.

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -91,7 +91,7 @@ For the record, these include the recent changes made [from the discussion forum
 #### Technical
 
 - **Metadata must be consistent across all difficulties of a beatmapset.**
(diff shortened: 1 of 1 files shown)


JBHyperion:
I forgot this needs to be discussed in the RC subforum first, oops.

Don't push this anywhere until a consensus is reached [here](https://osu.ppy.sh/community/forums/topics/1025897) I guess

ascendancelol:
Considering this is a change to the RC, it should be brought up  and approved in the RC subforum before ever touching github

https://osu.ppy.sh/community/forums/topics/1025897

cl8n:
thread closed

PR #3243, opened 2020-02-21
Update staff log

Kuron-kun:


cl8n:
please update bn listing as well

cl8n:
i pmd everyone with "TODO" as their language so hopefully we'll get those written soon. a lot of em responded quickly

also, jsyk that last comment in Staff Log is meant to help the next editor find where to start looking through Aiess. the last event is written there along with the date that the editor last checked

PR #3245, opened 2020-02-22
Fix Xayah's user link

Coppertine:
a 5 was added to the end...;


cl8n:
I think the name was kept as the beatmapset one on purpose, w/e tho

cl8n:
@peppy news permission still isn't working

Joehuu:
Maybe add osu-wiki write access to the [new-approval team](https://github.com/orgs/ppy/teams/news-approval/repositories)?

https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
> When the code owner is a team, that team must have write permissions, even if all the individual members of the team already have write permissions directly, through organization membership, or through another team membership.

PR #3247, opened 2020-02-22
Fill in missing BN languages

cl8n:
waiting on a few more replies


cl8n:
assumed fruko and nardo based on previous entries as they were both in staff before

PR #3250, opened 2020-02-23
Fix grammar and spelling in interface skinning

LumpBloom7:


LumpBloom7:
Additional Question: Would it be preferred if location be written like `Top Left` instead of `TopLeft`?

MegaApplePi:
- We can go with `Top Left`.
- If you can, a single commit in this pull request would be great.

Joehuu:
Will wait for another review.

PR #3309, opened 2020-03-14
Fix formatting in I-lettered articles

TPGPL:


wiki/Installation/en.md
```diff
@@ -6,14 +6,14 @@ needs_cleanup: true
...
+This page will tell you how to get osu! working on your Windows device. The [osu!academy](/wiki/osu!academy) has a [video tutorial](https://youtube.com/watch?v=0V5GwzmMhpU) that explains how to install osu! on Windows.
 
 ## Minimum system requirements
 
-- .NET framework 4+ (the installer should prompt you to download it, otherwise, you can [find it here](https://www.microsoft.com/en-us/download/details.aspx?id=48130))
+- .NET framework 4+ (the installer should prompt you to download it, otherwise, you can [find it here](https://microsoft.com/en-us/download/details.aspx?id=48130))
 - A decent graphics card.
 
-## Installing osu!
+## Installing osu!ss
```
MegaApplePi:
```suggestion
## Installing osu!
```

peppy:
What is this? This should be automated and only done with CI in place. Applying this manually is a waste of review time.

wiki/Installation/pl.md
```diff
@@ -6,28 +6,28 @@ outdated: true
 
 *Zobacz także: [Instalacja/macOS](/wiki/Installation/macOS)*
 
-Ten artykuł przybliży ci proces instalacji osu! na urządzenia z systemem Windows. [Akademia osu!](/wiki/osu!academy) podjęła ten temat w swoim [poradniku wideo](https://www.youtube.com/watch?list=PLmWVQsxi34bMYwAawZtzuptfMmszUa_tl&v=cz522ZAs5aQ).
+Ten artykuł przybliży ci proces instalacji osu! na urządzenia z systemem Windows. [Akademia osu!](/wiki/osu!academy) podjęła ten temat w swoim [poradniku wideo](https://youtube.com/watch?v=0V5GwzmMhpU).
 
 ## Minimalne wymagania systemowe
 
-- .NET Framework 4+ (możesz znaleźć go [tutaj](https://www.microsoft.com/en-us/download/details.aspx?id=48130); instalator powininen automatycznie powiadomić cię o potrzebie zainstalowania tej platformy)
```
peppy: what is this? why are you removing www from urls?

TPGPL: because it's not needed in any way, so why should it be there?

peppy: oh my god.

peppy:
Please reopen applying changes to every page with an attached automation script (at very least) and proposed plan for CI. Any manual (non-formatting) changes should be PR'd separately from ones which can be automated.

As an aside, you should NOT remove `www.` from URLs because that's not how the internet works. Please read up on this if you think you know better.

TPGPL:
Since we have a direction to go now,  could we have this merged at least? It's still better than what we have currently and has many changes that can't be automated. It would be a waste to let it stay closed.

peppy:
Please open in a new PR with the following changes:

- No formatting fixes (leave that for an automated pass)
- Don't remove `www` for URLs
- Make your PRs in more digestable chunks. None of this "fix i articles". Either do it one per article, folder, language or any combination. The title should be something like "Fix outdated links in Beatmaps articles".

wiki/Installation/pl.md
cl8n: in this case the canonical url does include www so you should keep it, you can tell bc without www redirects to www. sometimes the page includes a `<link rel="canonical" href="...">` to let you know the preferred url

but likewise you *shouldn't* include www if the service says it's not canonical

PR #3313, opened 2020-03-15
Clean up the Interface article

TPGPL:
- cleaned up MediaWiki and HTML leftovers
- fixed broken links and unparsed syntax

I left formatting changes for CI to be fixed in the future


wiki/Interface/zh.md
```diff
@@ -19,9 +24,9 @@ outdated: true
 -   ［11］用户信息旁边的是地图总数，运行时间，以及电脑时显示间。
 -   ［12］用户面板，点击显示用户选项。
 
-## 用户选项
+### 用户选项
 
-<img src="useroptions1.png" title="User Options" alt="User Options" width="280" />
```
peppy: shouldn't the alt text still be in the markdown..? is this not regressing?

(applies to everywhere else)

TPGPL: I assumed it shouldn't be here as it's not translated. Can bring it back if thats not a good way

cl8n: keep title text, untranslated is better than nothing
remove alt text since it's duplicated from title text. block images on the site have their title text printed as a caption already so alt text is only useful when it differs

wiki/Interface/de.md
```diff
@@ -89,16 +105,17 @@ Tippen sie an einer beliebigen Stelle in der Songauswahl etwas ein und das Einge
...
-| Lokale Rangliste                |
-| Nationale Rangliste             |
-| Globale Rangliste               |
-| Globale Rangliste (Aktive Mods) |
-| Freundesrangliste               |
+**Ranking Tabelle:**
+
+| Name | Verwendung |
+| :-- | :-- |
+| Lokale Rangliste | Zeigt die lokale Rangliste an. Scores aus dem Multiplayer werden nicht gespeichert. Mit das Drücken der "F2" Taste lässt sich eine verpackte Version des Replays erstellen. Kann möglicherweise in bestimmten Fällen nicht funktionieren. |
```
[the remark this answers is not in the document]
TPGPL: https://web.archive.org/web/20160809182440/https://osu.ppy.sh/wiki/Interface

used web archive to check out old wiki

Issue #3310, opened 2020-03-14
Add CI and lint existing content


peppy:
There have been recent PRs attempting to fix markdown inconsistencies, mixed in with other manual changes that ends up being an absolute mess to review and just isn't how commits should be done.

I've added [codacy](https://app.codacy.com/gh/ppy/osu-wiki) which uses [remarkjs](https://github.com/remarkjs/remark-lint#using-remark-to-fix-your-markdown) for linting. It seems like a good starting point.

Next steps would be for contributors to check through the reported issues and help tweak the inspections / configuration file to match our expectation. For instance, we will need to disable the naked URL rule or see if we can train it to ignore the header content.

Once a configuration is decided on, it should be run on all articles in a single commit, then enforced via CI on pull requests (already enabled but *not* required for merging just yet).

Joehuu:
Will do.

TPGPL:
Are we using the `markdown-style-guide` preset for remark-lint?

Joehuu:
You can just edit the settings in codacy, I think.

bdach:
remarkjs/codacy has started to become a pain due to presets being broken. From my investigation it seems that the [overrides of the `present-lint-markdown-style-guide`](https://github.com/ppy/osu-wiki/blob/d8c1a7f3d0e8473d45aa32d3604e7f551bc13aab/.remarkrc.js#L13-L41) preset just plainly don't work at all, likely due to remarkjs/remark-lint#165.

For example, when I added a different line length limit of 40 to the [news subconfig](https://github.com/bdach/osu-wiki/blob/master/news/.remarkrc.js), codacy would start to randomly report 40 on some lines and 80 on others (80 being the limit from the preset itself). Locally where I installed rules with `npm install -g` (which is apparently what shouldn't be done) I got similar behaviour where I'd get *both* 40 and 80 reported on some lines. I figure codacy just randomly clips one of them.

[...]

PR #3318, opened 2020-03-16
Add Remark config

MegaApplePi:
part of #3310

After looking at Codacy's patterns page, some options don't even prompt what exactly we want to use (e.g. using ATX headings, not Setext (it might actually be confusing itself with the frontmatter/YAML stuff)). This should also enable the frontmatter plugin which should allow it to check for YAML and ignore it whatever's in it.

Things to do:

- [X] Check if codacy actually uses the YAML plugin
  - it looks like it does
- [X] Make some more adjustments to the config file ... the docs was giving me a headache.


MegaApplePi:
Some random things I found while looking through:

---

`emphasis-marker` might not work as intended. `strong-marker` has a similar issue as well.

```md
| [\_WWWWWW_WWWWWW_](https://osu.ppy.sh/users/1981781)   | Portuguese, Spanish, Japanese|
```

It's warning that this line should be using `*` for emphasis, maybe all of the underscores need to be escaped?

---

`list-item-spacing` might not be working as intended.

```md
- Item 1
  - Item 2
- Item 3
```

It is warning that there should be an empty space after item 2.

---

`no-missing-blank-lines` warns if an HTML comment don't have have a blank line in-between itself and something else.

---

The [NSC has a rule regarding filenames](https://osu.ppy.sh/help/wiki/News_Styling_Criteria#article-filenames), but it looks like some of them have been ignored. Not sure what to set `no-file-name-irregular-characters` to or if we should disable it. (setting it to `\\.a-zA-Z0-9-_!` for now)

---

`code-block-style` might get confused with HTML indenting with 4 spaces.

---

some `no-blockquote-without-marker` warnings may actually be lists/paragraphs missing an empty space before the blockquote (they are not supposed to be part of the blockquote)

---

`no-literal-urls` appears to give false positives. Not exactly sure what causes it, but I think other symbols inside the link syntax causes this or emphasizing.

peppy:
Thanks for this! Let's merge as-is and see how it adjusts inspections (we can make further changes post-merge).

PR #3319, opened 2020-03-16
[NL] Rules and Ranking_Criteria/osu!taiko update 

Joppe27:
Updated the dutch translations of the rules and taiko ranking criteria page to be on par with the current english versions, I also made some changes to make the translations more clear.


Joppe27:
I'm sorry, these commits don't have names! Will try to fix

PR #3322, opened 2020-03-16
Clean up Installation article

TPGPL:


wiki/Installation/de.md
```diff
@@ -2,15 +2,15 @@
 outdated: true
 ---
 
-# Installation
+# Installation (Deutsch)
```
cl8n: I guess the danger of changing this is that you don't know at a glance if this is actually a translation (loan word) or English. in this case it was already translated
```suggestion
# Installation
```
I think this is same for french, maybe more

wiki/Installation/en.md
```diff
@@ -39,7 +39,7 @@ There are two ways to add beatmaps, via the website or osu!direct (requires [osu
 4. Locate and open the beatmapset file, it has an `.osz` extension.
    - Alternatively, you can also place the beatmapset file into the `Songs` directory of the osu! installation.
 5. osu! will automatically handle the rest for you and load the beatmapset.
-   - If, for any reason, you do not see them, press `F5` in [Song Selection](/wiki/Song_Selection) to re-process your beatmaps.
+   - If, for any reason, you do not see them, press `F5` in [Song Selection](/wiki/Interface) to re-process your beatmaps.
```
cl8n:
```suggestion
   - If, for any reason, you do not see them, press `F5` in [Song Selection](/wiki/Interface#song-selection-screen) to re-process your beatmaps.
```
and wherever else it comes up

wiki/Installation/de.md
Walavouchey: you're correct, It's the same word in french and swedish as well

PR #3326, opened 2020-03-17
Fix remark-stringify config and add remarkignore

cl8n:
linting is unchanged, just removed the "remark-" prefix from things that didn't need it


cl8n:
not sure what to do about this

```diff
-*Also known as: **Flower Formation***
+\*Also known as: **Flower Formation\***
```

otherwise the only problems I'm seeing are with markdown that was clearly wrong to begin with

cl8n:
```diff
 | Option | Value type | Description |
-| :-- | :-- | :-- |
+| :----- | :--------- | :---------- |
```
no option to fix this without plugin

cl8n:
and empty lines between references > >

PR #3333, opened 2020-03-18
[SV] Add FAQ

Walavouchey:
Created Swedish translation of FAQ


Walavouchey:
The codacy warnings seem to also apply for the english file, since the formatting should be basically copied over... should they be fixed regardless?

Not sure if the osu! terms are translated/left untranslated properly. I looked at what the french file was doing and copied that. This should otherwise be ready for finishing touches

cl8n:
don't worry about codacy for now, we're still setting it up

TPGPL:
you need to translate images' alt texts (`<alt text>`) as well

Issue #3150, opened 2020-02-03
Add Article Stubs (Community)


Loctav:
See the main issue https://github.com/ppy/osu-wiki/issues/3143
See the project: https://github.com/ppy/osu-wiki/projects/3

Here you find all article stubs that need to be created. Some articles might already exist. Some articles can be deemed unnecessary as singular article and be summed up into a bigger article. 

Many of these singular topics are included as subset in some articles already or information is scattered around multiple articles. These articles might need to be split up into multiple ones and information must be pulled together into one common article then (e.g. https://github.com/ppy/osu-wiki/tree/master/wiki/Glossary)

Each article should only contain information of one topic. 

You can claim articles beforehand. Just comment here.

- [x] osu!api
- [x] osu!dev Discord #3400
- [x] osu!team
- [x] Global Moderation Team
- [x] ~~ChatOps~~ ↓
- [x] Language Surveillance Division https://github.com/ppy/osu-wiki/pull/6059
- [x] osu! Alumni
- [x] Live Streaming
- [x] User Titles
- [x] Community Contributors
---
- [x] **Chat** https://osu.ppy.sh/wiki/en/Client/Interface/Chat_console
- [x] Internet Relay Chat
---
- [x] **osu! forums** #6454
- [x] Forum Thread ^
- [x] Forum Post ^

Parapass:
Can I claim the Live Streaming stub?

Walavouchey:
chat is covered by [`Client/Chat_console`](https://osu.ppy.sh/wiki/en/Client/Interface/Chat_console)

Walavouchey:
closing because all tasks are completed

PR #3349, opened 2020-03-21
Add redirects and tags for The Team and Developers

cl8n:
checks off "osu!team" for #3150 and "osu!dev" if it were there. these are already articles but it's hard to get there using these terms

### DIFF
--- a/wiki/People/The_Team/en.md
+++ b/wiki/People/The_Team/en.md
@@ -1,8 +1,10 @@
 ---
 tags:
+  - osu! staff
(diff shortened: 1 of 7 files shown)


wiki/People/The_Team/en.md
```diff
@@ -1,6 +1,7 @@
 ---
 tags:
   - osu!team
+  - osu! team
   - staff
```
TPGPL: might add `team osu!` as well

PR #3352, opened 2020-03-22
Remove HTML entities from the wiki

TPGPL:
Used regex script: `&.{2,6};`. The only left ones are `&amp;`s in links but they shouldnt be deleted


cl8n:
&mdash is replaced with wrong character

PR #3353, opened 2020-03-22
Override some remark-lint rules for news posts

cl8n:
disallow html everywhere except news posts, and let news posts use top level headers

featuring a dumb hack for "extending" the config because there's no convenient way to load another remarkrc as a plugin... see [addModule](https://github.com/unifiedjs/unified-engine/blob/2f48feaf3483edce7321edb57eed1c67216009a8/lib/configuration.js#L210), it will ultimately call `require('../.remarkrc')`, node only recognizes json if the file ends in .json, and renaming it to `.remarkrc.json` isn't supported by remark.


peppy:
Weird it doesn't list any fixed issues. Should it not?

cl8n:
i'm not sure how codacy works, maybe it's only checking files that changed

remark-lint stops reporting all of the `multiple-toplevel` warnings under news/ after this change

peppy:
Looks good. Will merge and force full reanalysis.

PR #3355, opened 2020-03-22
[IT] Add Main_Page, Update Registration

Theighlin:
Updated italian registration page.


Theighlin:
Hi, first time editing a file here and I need help. How do I avoid [no-multiple-toplevel-headings]?. If I use ## it wouldn't match the others Main_Page.

peppy:
I believe the original pages are actually wrong (we only started applying linting recently). Matching them is fine for now, we will likely need to fix this in a separate pass.

MegaApplePi:
Ah shoot, missed the Registration article

PR #3356, opened 2020-03-22
Remove a redundant TWC rule about FM

### LINKED ISSUE #3268: twc2020 ruleset clarification
> 7. The FreeMod bracket will have FreeMod activated. Every individual player can pick Hidden, HardRock or both.
>    - When playing a FreeMod map, at least 1 player of each team must have at least one mod activated.
>    - **Players must select different mod combinations**. Both players using only Hidden, only HardRock or Hidden+HardRock is disallowed.
> 8. The tiebreaker will be played under FreeMod conditions.
>     - When playing the tiebreaker, **no one needs to have a mod activated**.

No mods is still a mod combination. That would mean at least one player still has to have a mod activated to avoid different mod combinations. That should be clarified - does the second rule get cancelled as well, or do two NoMods simply count as different mod combinations (Essentially "every individual player can choose whether they play it under NoMod or FreeMod conditions")?

pavlukivan:
resolves #3268


cl8n:
TWC ended but @WalterToro I'd use this wording in future tournaments

PR #3386, opened 2020-03-30
[ZH] Add flag-references translation

Kotoki1337:
meta/flag-references/zh.md


Kotoki1337:
ignore commits, i just gonna dumb.

PR #3400, opened 2020-04-03
Add osu!dev Discord server article

cl8n:


Ephemeralis:
Something about the difficulty category being in the list as one complete entry rubs me wrong. Might be worth using subheadings and lists to describe each category? (Dev, Difficulty, Community, Mapping) etc?

cl8n:
I thought each channel there was self explanatory cuz it's just a mode name added to the end. if it were expanded into subheadings I don't think I'd have any more to say about these (or any category really)

it was supposed to answer the question "where do I go to talk about \<this project\>?", so I flipping the table might make more sense?

peppy:
either that or just list all four channels in the left for that one

cl8n:
added more rows for it, that made room for links to the gamemodes so I like this now

PR #3409, opened 2020-04-04
Fix empty header in Mappers' Report news post

TPGPL:
also properly fixes a Codacy error


cl8n:
it's supposed to be a horizontal rule, with the link going to it

TPGPL:
I don't see a harm in making this a header, especially as it makes more sense linking to a header, not a horizontal rule.

cl8n:
I have no comment which one is "better", I just don't think we should change old news posts beyond what's actually broken

TPGPL:
i fixed it because codacy's automatic linter would change it into `# <a id="conclusion></a>` which is obviously wrong. Changing it to something like this doesn't make a difference for potential readers, but prevents breaking something by the lint

peppy:
I disagree with making "conclusion" visible since it wasn't intended to be visible.

PR #3414, opened 2020-04-05
Remove link reference style from the wiki

TPGPL:
Removes Raikire's reference style and uses inline links instead. Also removes manual image captions and changes `:-` into `:`

only used `\[.[^_\]]*\]: ` to find reference links and `:-[^:-]` to find `:-` occurences


peppy:
As always, could you please provide a script or list of regex or *something* to show how to reproduce these results? Manually checking these is absolutely bad.

PR #3417, opened 2020-04-06
add and remove dlls

penguin2233:
component list taken from fresh osu installation


peppy:
The two lines before this are also pretty dodgily wrong. The dlls listed there are not "Microsoft .NET" requirements. Not even sure why they are listed here (and why they "should be ignored" even if they are).

penguin2233:
would it be better to just list the files that you see without enabling hidden files/folders? the ss in this article doesn't show hidden files anyway

peppy:
I think it's fine as you have it.

PR #3423, opened 2020-04-07
Remove literal URLs from the wiki

TPGPL:
Properly fixes Codacy warnings, adds `mailto:` for literal mail links and removes most of remaining literal URLs on the wiki.


cl8n:
u can put angled brackets around links if it makes sense to display the whole thing, i did that in #3400

PR #3432, opened 2020-04-08
Escape quotation marks in image elements

TPGPL:
Possibly fixes broken remarklint errors?


cl8n:
just the title text seems problematic. quotes should be okay in alt

TPGPL:
lets see if it actually fixes it for now, then i will remove escaping in alt texts to check that

PR #3455, opened 2020-04-14
[EN] Update Beatmap_Nominators

Randelman4ik:
Remove Vulkin from "osu!taiko" BN's


cl8n:
we update group pages in batches with Staff Log so they're all at the same state, I'll do this soon since it's been a few weeks

PR #3460, opened 2020-04-15
[PT-BR] Add flag references and group info

Flantsune:
Created and translated: group-info/pt-br.yaml and flag-references/pt-br.md


cl8n:
for the group pages, i can update the user tables as long as the rest of the article is up to date. so they will need updates too to be able to use this

if you want to do that just put placeholders for the tables and i can fill them in

Flantsune:
Is it okay to leave the tables in English so they can be updated this way then? I'll get to translate the rest of the group pages tomorrow in a seperate PR

cl8n:
ya, for now the ones im updating are BN, NAT, GMT, i should probably add something similar for alumni but haven't gotten around to it yet

PR #3465, opened 2020-04-16
[PT-BR] Update The_Team and add sub articles

Flantsune:
Update The_Team and osu!_Alumni. Add the following articles: Global_Moderation_Team, Support_Team, Project_Loved_Team, Nomination_Assessment_Team, Developers, Beatmap_Nominators


Flantsune:
I forgot to mention that tables where the information is similar to meta/group-info were left in English intentionally in the hope those can all be updated with that instead. Any other that didn't seem like it would be translated with that were manually translated.

PR #3467, opened 2020-04-17
Update osu!tourney prize requirements

Trosk:
Updated requirements to reflect the current approval procedure


wiki/osu!tourney/Prizes/en.md
```diff
@@ -76,7 +78,7 @@ If your tournament meets the criteria listed above, and you have a design ready
 
 Please allow at **least** a week for a response. We will try our best to handle all requests within a week, but we occasionally have large spikes in workload that can make this impossible.
 
-If your request is approved, you may then submit a list of applicable recipients at the conclusion of your tournament, along with the finalized badge design.
+At the conclusion of your tournament you may submit a list of applicable recipients, along with all of the MP links of the winning team's matches. Your tournament will then be evaluated and based on professionality, quality and effort put into the tournament it may be granted a badge prize.
```
cl8n, line 81: should keep the "if your request is approved" language, otherwise it's not clear that this is a step *after* the previous email

as a process this doesn't make sense to me either...

- tournament host mails to see if it can have a badge as a prize
- you evaluate the tournament info and reply with approval or denial
- (after tournament ends) tournament host mails again with badge winners and their match info
- you evaluate the tournament again (?) and reply with approval or denial

if the tournament host can't be sure that their tournament will be badged or not until after it concludes, what's the point of the first email? given that first "approval" everyone would be under the impression that a badge will be awarded, but there's still a chance you will deny it the second time

Ephemeralis: my understanding of it is that the first email signals a request for suitability monitoring during the tournament, the organizers then conduct the tournament with said monitoring, then a decision is made and approval/denial is issued accordingly

cl8n: what "monitoring"? this new change clearly says the matches are evaluated after the tournament, not during

nwatti: The tournaments would be monitored while they are going on (checks for drama, professionalism, to see if no rules are broken, etc). Then after the tournament is finished, the winning team's matches will be reviewed to see if the matches were played and if the tournament was competitive enough.

nwatti:
Should add requirement for LAN tournaments to have MP links and/or VOD's in their post-tournament mail

wiki/osu!tourney/Prizes/en.md
cl8n: then let's write that all in, these changes don't mention it so far. and it's still not clearly written what comes in the first response email, and how the process continues from there ("if your request is approved" in my first comment)

I'm just not a fan of this uncertainty in prizing until after everything is concluded.  the winning team probably tried so hard in part because they wanted the reward, and it can be denied after the fact if tournament wasn't "competitive enough"... this seems like asking for some dramatic situation, if it didn't already happen

Ephemeralis: the way things work now is essentially a compromise in order to limit tournaments being sandbagged or deliberately abused for badges, which happened with abundance in the past

probably worth recording it somewhere though

nwatti: I think a big part of why a mail beforehand is required is so that we are aware of which tournaments are running, would be hard to monitor them if we didn't. Though I suppose that could be mentioned in the same mail the screening list is

nwatti (dismissed):

cl8n (approved):

PR #3472, opened 2020-04-21
Fix news' remark configuration

TPGPL:


cl8n:
unified-engine ends up doing `require()` on this, so it makes no difference

PR #3476, opened 2020-04-23
Remove "Outer Oni" from taiko diff names

Azuelle:
According to [this mod](https://osu.ppy.sh/beatmapsets/1137974/discussion/-/generalAll#/1518226) by Faputa

>Inner Oni -> 裏鬼
>Outer Oni -> 表鬼 -> Oni = easier than Inner Oni

And therefore the current wiki page is kinda misleading, as it marks Inner Oni as Expert and Outer Oni as Expertplus


Raidencio:
yes. I don't even know why Outer Oni was added to the diff name RC given that it's totally a custom diff name that hasn't been intersubjectively used much, unlike Hell Oni

cl8n:
it was going to be removed or clarified in the rewrite of difficulty pages months ago but that was never done. seems popular enough to warrant a mention in the 'common alternatives' section or w/e it will become

also, despite the language, outer oni usually is more difficult than inner oni.

xcyccg6:
It's a misleading diff name I would say.
Now we may correct this mistake, just like using album name as source

Raidencio:
it's absolutely not popular enough given that it mostly started as a custom diff name for one mapper (eiri-) and then maybe two more people used it once in their maps for some reason

it also makes no sense logically no matter how you look at it, because the concept is "the deeper you go, the harder it is", much like in TnT. so Outer Oni is def. not a common alternative, just wordplay for a custom top diff name.

cl8n:
"common alternatives" is probably not a great name for that section, it's meant to show what recurring names people use for the same level of difficulty as the default names

I'll edit to make it clear, that's why I self-assigned. this PR so far doesn't update every part of the wiki where outer oni is mentioned

xcyccg6:
I think you can put outer oni in "common alternatives“ and you may add some notes to this diffname like "Mapper eiri- first used this difficulty name and it is come from the standard mode. But it doesn't match the concept 'the deeper you go, the harder it is'. " and so on

Cynplytholowazy:
I would say it shouldn't even be put in the "common alternatives". There's only one people who uses it and a few others who used it in one of their maps, and many people agreed that it shouldn't be inside the wiki back then with the ranking criteria thread https://osu.ppy.sh/community/forums/topics/1002177

cl8n:
my bad for not get around to this soon, I'll edit further when I have time for wiki things

PR #3483, opened 2020-04-24
Update Interface

KevKjef:
updates the interface article, mainly to fix outdated parts.

Fixes outdated parts and the same spelling mistakes as in my previous upload:
Fixes:
Bubble icon: Change due to modding page instead of forum modding.
Delete beatmaps: Not 100% you can do this but I don't want to test delete all visible maps with 105k maps. Changed the part so that the 'joke' is now seen as something in the past.
Online Ranking:
Got outdated after updates to the screen you get. Updated pretty much the entire part. Also added a new screenshot for this section, and deleted the old one.
May require a different screenshot, as mine isn't the best.

Also fixed some grammar mistakes/misspellings.


TicClick:
@MegaApplePi do you remember why you tagged this as `blocked`?

wiki/Interface/en.md
```diff
@@ -273,7 +272,7 @@ Click the **[osu!cookie](/wiki/Glossary#cookie)** to **start playing the selecte
 
 This is the grade screen shown after you have successfully passed the beatmap. You can access your online results by scrolling down or pressing the obvious button.
```
peppy, line 273: let's rename "grade screen" to "results".

peppy (changes requested):

MegaApplePi:
@TicClick If I remember right, I was going to make some formatting changes, but I just did not get to it. I will remove the tag.

wiki/Interface/en.md
```diff
@@ -70,9 +70,9 @@ Access this screen by clicking your profile at the top left of the main menu. Yo
 
 ## Song Selection Screen
 
-![](img/Soloplay1.jpg "Song Selection")
+![](img/song-selection.jpg "Song Selection")
```
TicClick, line 73: @TPGPL at least to me, the difference between the bottom and the top screenshots wasn't immediately obvious without a caption like on `metadata-comparison.jpg`, mind adding one?

TicClick (changes requested):

wiki/Interface/en.md
TPGPL: i think i will just use a single image instead of a comparison, as its not needed at that exact context + a comparison is already made below

TicClick (approved):

PR #3493, opened 2020-04-25
[TR] Update The_Team

zeusminus:


zeusminus:
couldn't find a way to fix `[table-pipe-alignment]` issues on Codacy check. a review would be appreciated

PR #3498, opened 2020-04-25
Conclude TWC 2020

WalterToro:
long overdue for no reason


cl8n:
not sure why the stages table looks broken on the site, if that continues after this update I'll report on osu-web

PR #3507, opened 2020-04-29
add legitimate interest clause

osuShiro:
metadata remaining in the event of successful deletion


peppy:
@osuShiro i've combined it with the existing mention of this information in a hope to simplify things.

wiki/Legal/Privacy/en.md
```diff
@@ -88,7 +88,7 @@ In many cases, user submissions such as forum posts and beatmaps can be deleted
...
 
-As we have a strict one-account-per-user policy to maintain fair leaderboards, after account deletion you may not be able to return to our service. To enforce this, we will also maintain the unique identifying string as mentioned in "Information we collect".
+As we have a strict one-account-per-user policy to maintain fair leaderboards, after account deletion you may not be able to return to our service. To enforce this, we will also maintain the unique identifying string as mentioned in "Information we collect", even after account deletion. This is done so as a [legitimate interest](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/legitimate-interests/when-can-we-rely-on-legitimate-interests/) in order to enforce our rules. Rest assured, it is one-way hashed and cannot be used to identify you outside of osu!, nor the precise context of enforcing this rule.
```
Ephemeralis, line 91:
```suggestion
As we have a strict one-account-per-user policy to maintain fair leaderboards, after account deletion you may not be able to return to our service. To enforce this, we will also maintain the unique identifying string as mentioned in "Information we collect", even after account deletion. This is done so as a [legitimate interest](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/legitimate-interests/when-can-we-rely-on-legitimate-interests/) as outlined in the GDPR in order to enforce our rules. Rest assured, it is one-way hashed and cannot be used to identify you outside of osu!, or outside the precise context of enforcing this rule.
```

Ephemeralis: just specifies that it is a legitimate interest as outlined under the GDPR, and exchanges the 'nor' for something specifically outlining it as belonging to this particular rule since i thought the nor was a bit confusing

PR #3510, opened 2020-04-30
Add mention of next hitobject brightening in osu! mode

leduyquang753:
There is a small piece of information missing from the *osu!* gameplay page:

> The next hit circle or slider to be tapped/dragged is slightly brighter than the rest on the playfield.

The position of the line may not be the best, if so please move it to the most favorable position.


cl8n:
do u have a source for this, I can't tell in-game

leduyquang753:
> do u have a source for this, I can't tell in-game
A screenshot of the page itself:
image

Note how the 1st circle is brighter than the second and third.

Pennek:
That's happening because of fading. If you use lower ars more circles will be 100% faded in and the effect you're describing wont be accurate.

leduyquang753:
> That's happening because of fading. If you use lower ars more circles will be 100% faded in and the effect you're describing wont be accurate.

The highlight effect is just a part of the fading process anyway: fade until about 90% alpha, and when the hit object's time is almost reached, raise the rest to 100%. But that fact does not deny the existence of the highlight. When AR is high enough, I agree that the 0% ÷ 90% fade-in's stop time will collide with the 90% ÷ 100% fade-in, a sentence can be added for that.

You can try opening some easy maps, dim the background completely and see how the hit objects fade-in not fully and only highlight to 100% alpha when the hit time is near. The original game *osu!* based on – *Osu! Takadae! Ouendan* – also uses this highlighting effect. I'm not sure if you can also do this, but if possible, you may try contacting peppy directly.

peppy:
It is always the next circle which is faded. This will likely be kept with lazer, fwiw (although is not yet implemented).

Pennek:
Well this is news to me as this isn't detailed anywhere on wiki. We would need to amend this to the graph on /wiki/beatmapping/approach_rate and probably other places as well. @peppy do you have any specific values you'd want to share as to when exactly this happens?

peppy:
I'm not sure exact values matter too much. Probably fine to just add as-is?

PR #3517, opened 2020-05-02
Add audio compatibility mode rule to RC

TheMoMan:
Audio compatability mode enabled causes a -15ms offset compared to having it disabled. We should have everyone use a consistent setting for this, and since this setting uses a legacy engine we shouldn't be using it.

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -191,7 +191,7 @@ This category contains explicit allowance statements of concepts and rules that 
   - Remain unsnapped, as long as they align with the intended beat snap divisor.
   - Be snapped through a temporary change in BPM.
 - **An object which is wrongly snapped due to passing through or ending slightly before a new uninherited timing point must have its end snapped within the new timing section.** For spinners and osu!mania long notes, this can be achieved through dragging an object's tail in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the .osu file.
-- **Beatmaps must be timed with the audio compatability mode setting turned off.** This setting can be found in the game's audio options. Enabling this setting uses a legacy audio engine, and can cause discrepancies with timing. 
+- **Beatmaps must be timed with the audio compatibility mode setting turned off.** This setting can be found in the game's audio options. Enabling this setting uses a legacy audio engine, and can cause discrepancies with timing. 
 
 ### Guidelines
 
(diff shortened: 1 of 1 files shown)


wiki/Ranking_Criteria/en.md
```diff
@@ -191,6 +191,7 @@ This category contains explicit allowance statements of concepts and rules that
   - Remain unsnapped, as long as they align with the intended beat snap divisor.
   - Be snapped through a temporary change in BPM.
 - **An object which is wrongly snapped due to passing through or ending slightly before a new uninherited timing point must have its end snapped within the new timing section.** For spinners and osu!mania long notes, this can be achieved through dragging an object's tail in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the .osu file.
+- **Beatmaps must be timed with the audio compatability mode setting turned off.** This setting can be found in the game's audio options. Enabling this setting uses a legacy audio engine, and can cause discrepancies with timing.
```
STCapricorn, line 194:
```suggestion
- **Beatmaps must be timed with the audio compatibility mode setting turned off.** This setting can be found in the game's audio options. Enabling this setting uses a legacy audio engine, and can cause discrepancies with timing. 
```

cl8n:
seems out of place in RC to have a rule about what tool is used to make the map rather than what is in the map. what if I make the map by means other than the osu! editor? or with a huge universal offset? or my sound system has a delay way beyond 15ms otherwise? or I do need that audio compatibility option for osu! to work as intended (this option exists for a reason)?

the rule here is that maps should be properly timed, which exists at line 182. maybe add a note there if this is a common problem

TheMoMan:
I did think it might be out of place too, but the RC seems like the most visible place to have this information. Also while you could potentially create a map without the editor, the map must still be timed correctly with this setting disabled once it's loaded into the game.

I could amend this onto line 182 instead if you think that's more appropriate. Can understanding adding this feels very sudden but since this is causing inconsistencies in timing for ranked maps it probably needs to be.

Pennek:
Have to agree with clayton here. Ranking Criteria needs to make sense outside the context of the osu! editor. Amending on 182 would work

peppy:
This should NOT be in the ranking criteria. Game bugs should be fixed as such, not worked around.

Please report this on osu-https://github.com/ppy/osu-stable-issues/issues if you believe it is an issue. Please provide as much technical detail as possible (we did *extensive* testing before pushing out the legacy mode)

PR #3533, opened 2020-05-07
Add integration for Visual Studio Code

sr229:
This includes [`vscode-remark`](https://github.com/mrmlnc/vscode-remark) and a lint command to run a wiki-wide lint.

image

However, this does not include integration for other editors, hence the inclusion of the `lint:wiki` command as a way for users with other editors to lint locally on the fly.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+  "recommendations": [
+    "mrmlnc.vscode-remark"
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
     "name": "osu-wiki",
     "description": "home of the osu! wiki",
     "repository": "github:ppy/osu-wiki",
+    "scripts": {
+        "lint:wiki" : "npx remark wiki/*.md -qf --no-stdout --report=vfile-reporter-position --color"
+    },
     "dependencies": {
         "remark-cli": "8.0.0",
         "remark": "12.0.0",


sr229:
So getting an issue that remark is ignoring the configuration inside `news/` not sure if that's intended.

sr229:
Marking ready for review, going to fix the problem with news in another PR.

cl8n:
lint:wiki doesn't seem very useful because you're usually only running this on files you edit. also it forces all these options I wouldn't want to use every time. I think it's easy enough to run remark already if you want to

sr229:
> lint:wiki doesn't seem very useful because you're usually only running this on files you edit. also it forces all these options I wouldn't want to use every time. I think it's easy enough to run remark already if you want to

Its forcing an option per se, it just makes the linting part easier for the end user since they don't need to run `npx remark` everytime and remember the actual command line arguments. `yarn lint:wiki` is easier to remember vs `npx remark <arg> <arg>`.

cl8n:
why would u want to be using any of these options by default, only way I've ever needed to run this was `remark <files I'm editing>`

peppy:
How does this actually show linting results? I can't seem to see anything.

sr229:
@peppy it only shows errors when it sees one, otherwise it returns a zero exit code.

peppy:
I was testing with errors and saw no results. Is this working for others?

TPGPL:
Closing as it doesn't seem really useful in its current form as it doesn't errors directly in code, plus the branch was deleted

PR #3577, opened 2020-05-20
Add Music Theory and Beat articles

Ph0su:
Added Music Theory and Beat articles to the wiki


MegaApplePi:
I think the music theory page should be moved to where you have it now, rather than writing a very short page then telling readers to just to go the guides page.

cl8n:
yeah most of the existing music theory "guide" is hardly a guide and would fit nicely here instead.

PR #3588, opened 2020-05-25
Clarify what counts as unintended .osu edits in RC

cl8n:
- [x] depends on #3687


peppy:
I'm not sure about this. This looks like it's more lenient than before, correct? Allowing changes to .osu files apart from the two mentioned?

cl8n:
the original rule only concerned .osu changes regarding diff settings and break times. reasoning for this change is in https://osu.ppy.sh/community/forums/posts/7479350

so this PR is missing point 3 from that post, but that's because some guideline about CS will be added back in a more appropriate place as pishi said in the post afterwards. keeping this PR blocked until that can happen at the same time

PR #3649, opened 2020-06-06
[TR] Timing

Drops-P:


wiki/Beatmap_Editor/Timing/tr.md
```diff
@@ -0,0 +1,23 @@
+*Ayrıca bakınız: [Şarkılar nasıl zamanlanır](/wiki/Guides/How_to_Time_Songs)*
```
MegaApplePi, line 1: It looks like you are missing the title heading.

PR #3651, opened 2020-06-07
Add origin of osu!catch

paper-toaster:
Basically just added a bit of text below the osu!catch section saying that's it's based off of the EZ2DJ series. No one replied on the Discord channel, so I decided to just go ahead and submit this to see if it is true.

### DIFF
--- a/wiki/Game_Modes/en.md
+++ b/wiki/Game_Modes/en.md
@@ -21,7 +21,7 @@ Based on *[Taiko no Tatsujin](https://en.wikipedia.org/wiki/Taiko_no_Tatsujin "W
 
 *Main page: [osu!catch](osu!catch)*
 
-Collect falling fruits by moving the catcher with a plate. Keep dancing and do not miss the step of the beat.
+Based on the EZ2CATCH mode from the *[EZ2DJ series](https://ja.wikipedia.org/wiki/Ez2Dj/EZ2AC "Wikipedia")*. Collect falling fruits by moving the catcher with a plate. Keep dancing and do not miss the step of the beat!
 
 ## ![](/wiki/shared/mode/mania.png) osu!mania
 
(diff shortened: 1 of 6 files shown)


wiki/Game_Modes/en.md
```diff
@@ -21,7 +21,7 @@ Based on *[Taiko no Tatsujin](https://en.wikipedia.org/wiki/Taiko_no_Tatsujin "W
 
 *Main page: [osu!catch](osu!catch)*
 
-Collect falling fruits by moving the catcher with a plate. Keep dancing and do not miss the step of the beat.
+Based on the EZ2CATCH mode from the *[EZ2DJ series](https://ja.wikipedia.org/wiki/Ez2Dj/EZ2AC "Wikipedia")*. Collect falling fruits by moving the catcher with a plate. Keep dancing and do not miss the step of the beat!
```
MegaApplePi, line 24:
```suggestion
Based on the EZ2CATCH mode from the *[EZ2DJ series](https://en.wikipedia.org/wiki/EZ2DJ "Wikipedia")*. Collect falling fruits by moving the catcher with a plate. Keep dancing and do not miss the step of the beat!
```
Should link to the English article.

paper-toaster: I would, but the English article doesn't mention anything about EZ2CATCH

peppy:
When I made osu!catch, I wasn't aware of ez2dj , but it was loosely based around a minigame inside a beatmania simulator I previously played. Which was probably based on ez2dj.

Not sure if that changes what we want to say on the wiki.

paper-toaster:
> When I made osu!catch, I wasn't aware of ez2dj , but it was loosely based around a minigame inside a beatmania simulator I previously played. Which was probably based on ez2dj.
> 
> Not sure if that changes what we want to say on the wiki.

We could edit it to be more specific by saying that it's loosely based of the series. But in that case, the only reason for adding this to the wiki would be to inform curious osu!catch players where the "original" or "first of it's kind" came from.

wiki/Game_Modes/en.md
cl8n: it doesn't look like there are any good resources that talk about EZ2CATCH in English, and Japanese is even less helpful so I've just left the reference to EZ2DJ as the English Wikipedia article

not much we can do about that short of writing something ourselves. which probably isn't necessary

PR #3669, opened 2020-06-10
Remove or replace all puush mentions

### LINKED ISSUE #3629: Remove all puush mentions
Since puush isn't maintained anymore there is no reason why mentions should stay

lmnyx:
Resolves #3629.

### DIFF
--- a/wiki/Glossary/en.md
+++ b/wiki/Glossary/en.md
@@ -639,12 +639,6 @@ The total duration of a beatmap. This includes the lead-in time and the break ti
 
 The main ranking metric of the game.
 
-### puush
-
-A screen capturing and file uploading application developed by peppy. puush is a successor of sorts to upppy which is commonly used to upload and share images in osu!
-
-For more information about puush, see [puush / home](https://puush.me).
-
 ## Q
 
 ### QAT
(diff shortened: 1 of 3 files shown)


wiki/Glossary/en.md
```diff
@@ -639,12 +639,6 @@ The total duration of a beatmap. This includes the lead-in time and the break ti
 
 The main ranking metric of the game.
 
-### puush
```
MegaApplePi, line 642: I think you can keep the term in the glossary. Maybe add a note that it no longer accepts new users and is no longer in active development.

peppy: agreed. please don't erase for existence :)

wiki/Glossary/en.md
```diff
@@ -643,6 +643,8 @@ The main ranking metric of the game.
 
 A screen capturing and file uploading application developed by peppy. puush is a successor of sorts to upppy which is commonly used to upload and share images in osu!
```
MegaApplePi, line 644:
```suggestion
A screen capturing and file uploading application developed by peppy. puush was a successor of sorts to upppy which is still commonly used to upload and share images in osu!
```

PR #3728, opened 2020-06-19
[ID] Add Bot Account

fajar13k:
Add Bahasa Indonesia translation of `Bot Account` article


cl8n:
EN was rewritten

fajar13k:
Updated

PR #3758, opened 2020-06-25
Use LF for markdown files and check with remark

cl8n:
if you already have CRLF stuff checked out, and want to convert all to LF right now, this is what I did:

- make sure you have a clean working dir at index of this branch (or master when it's merged)
- `git ls-files -z '*.md' | xargs -0 rm -f && git reset --hard` - "delete all tracked files and check them out again"

there might be an easier way to do that with git but internet solutions didn't help me

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 * text=auto
-*.md text
+*.md text eol=lf
--- a/.remarkrc.js
+++ b/.remarkrc.js
@@ -16,6 +16,7 @@ module.exports = {
     ['lint-fenced-code-flag', {allowEmpty: true}],
     ['lint-final-definition', false],
     ['lint-final-newline'],
+    ['lint-linebreak-style', 'unix'],
     ['lint-list-item-indent', 'space'],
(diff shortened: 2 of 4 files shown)


bdach:
How about seeing `core.autocrlf = input` in `.git/config` for another level of safety? Should prevent `\r\n` from ever getting committed as far as I understand.

cl8n:
there's no way to force changes to git configs like that I think. almost everyone with git is using either `true` or `input` for that, and both will be fine here

bdach:
Oh, local config is lowest priority, ouch. This will have to do I guess, although I'm now wondering if we don't want to set it to `false` for CI so that git doesn't fool it into thinking things are fine by fixing stuff up...

peppy:
For git enforcing, please copy the .gitattributes from https://raw.githubusercontent.com/ppy/osu/master/.gitattributes rather than changing git config. It enforces commit-time line endings and recommends checkout-time line endings for users.

peppy:
Never mind, you already did. Thought that was overlooked when reading the above discussion.

PR #3760, opened 2020-06-25
Cleanup and unify the style of the MWC articles

fajar13k:
- Some maps are missing/deleted
- Convert beatmap links to `/beatmapsets/mapsetId#mania/mapId`
- Add badges
- Add groups (in Group Stages)
- Add flag name references
- Optimize image to use .jpg if there aren't Alpha Channels used


bdach:
FYI merging master back here should now resolve the CI issue with the file size check.

PR #3768, opened 2020-06-27
Fix or add default news header in old posts

cl8n:
old news posts didn't have "headers" all the time because they were posted as a tumblr blog. now, headers are created by eating the first image of each post, so those old posts ended up broken in some cases if the first image wasn't supposed to be a header.
[example](https://osu.ppy.sh/home/news/2013-07-31-easy-mod-how-and-why), [example 2](https://osu.ppy.sh/home/news/2013-09-15-the-adventures-of-clicking-circles-1-never)

I've fixed these cases by adding a header to the top, using the first option here that applied:

- an image in the post that looks good as a header
- an image outside of the post that would clearly fit as a header for it (ex. the corresponding OWC header for an OWC results post)
- the default news header

and if the first paragraph made sense as a post preview, I added the headers below that.


cl8n:
remark's useless for this one because old posts have so many issues lol

peppy:
Some of these are kinda the wrong size / filesize, but I guess it's fine considering how ancient the posts are 👍 .

PR #3773, opened 2020-06-27
Unify the unordered list style

TPGPL:
I used `^[+*]{1} ` and `^ +[+*]{1} ` to search for wrong list styles, then manually replaced every occurences per-file, so there shouldn't be any wrong regexes in the PR.

It should be ready to merge as it is.


cl8n:
i'll check this one but i rly think this effort would be better spent checking if remark-stringify has any more issues, if not we can just start using it

TPGPL:
🤷 well same thing has been said months ago, yet wiki's remark still hasn't reached usable state, so i decided to do it myself in <30 minutes.

cl8n:
have u checked though? I feel like nobody's actually tested it out following some more changes we made.

- take list of all files that remark-lint currently passes
- parse and stringify all of them
- if remark-lint still passes all of them, we're probably good...

peppy:
Seems okay, although there's still remaining issues with spacing on unordered lists.

`remark ./* --use lint-unordered-list-marker-style --no-config -q --ignore-pattern node_modules`

Shows no remaining issues, at least.

Also

HEAD: 18821 messages (✖ 52 errors, ⚠ 18769 warnings)
This PR: 16877 messages (✖ 52 errors, ⚠ 16825 warnings)

PR #3795, opened 2020-07-01
Add Summer Theme Park 2020 Fanart Contest newspost

Ephemeralis:
🌞


peppy:
Is this scheduled before the FA release?

Ephemeralis:
yes, it is. should be good to go ASAP and will chase it up with FA article shortly after

Issue #3802, opened 2020-07-02
Validate yaml in CI


cl8n:
all yaml files should be valid and not have duplicate keys, in particular breaking redirect.yaml has a chance to go unnoticed because it depends on editor+reviewer remembering to check earlier keys

PR #3803, opened 2020-07-02
Add YAML linting support

peppy:
- Closes https://github.com/ppy/osu-wiki/issues/3802


peppy:
Reduced the rules down to a subset which catches critical errors but not formatting issues.

PR #3807, opened 2020-07-02
Update AiMod wiki page

hypercyte:
Completes a task from https://github.com/ppy/osu-wiki/issues/3128

- Updated the aimod warning/error messages
- Included flaws + suggested alternative
- Better looking screenshot of the window with up-to-date warnings

I left various comments around for things I'm not sure about.

Could get some thoughts on the flaws too!


Pennek:
remind me to take a look at this if i havent reviewed in a week or so

wiki/Beatmap_Editor/AiMod/en.md
```diff
@@ -2,12 +2,22 @@
...
+![AiMod](AiMod.jpg "This is the AiMod window")
 
 AiMod typically completes its beatmap analysis in under a minute and subsequently displays warnings and errors. If no problems were found in your beatmap, then congratulations! The beatmap passed the AiMod analysis, but it still requires modding from the community. The modders can find other issues that AiMod failed to detect.
 
+## Flaws
+
+<!-- TO DO: I'm sure there are more flaws, just can't think of any more off the top of my head -->
+- **Cannot detect a large number of issues** that are considered unrankable by the [Ranking Criteria](/wiki/Ranking_Criteria), such as; potentially missnapped objects, delayed hitsounds and background filesize above 2.5MB etc.
+- **May warn you of non-issues**, such as; the mapset needing 2 standard diffs on a marathon map and Kiai time being toggled on for less than 15 seconds.
+- **If "Check distance snap" is enabled, it will not check the distance snap between new combos.**
```
cl8n, line 16: this is intentional, it checks per-combo

hypercyte: hmm was thinking of it as a flaw from the mapper's perspective, since aimod doesn't detect spacing inconsistencies between combos in normal/easy diffs (which it probably should) which leads into being one of the reasons to use mapset verifier to check these things

but if we want to stick with unintentional flaws of aimod i can remove this one

cl8n: maybe instead of "flaws" you could say they're differences from RC

cl8n: not really problem anymore cuz i said what "flaws" means in intro

PR #3818, opened 2020-07-04
[ES] Internet Relay Chat

MisticalNipple:
Translated the whole document to Spanish


MegaApplePi:
@TPGPL that full stop is in the English version of the article.

TPGPL:
@MegaApplePi the dots are in the other FAQ headers as well in the English article, but they were removed here, so i think it's fine? 🤷

MegaApplePi:
Then we should probably remove them from the English version to keep them consistent.

PR #3819, opened 2020-07-04
[TH] Add Main Page

RondoKnight:

PR #3820, opened 2020-07-04
[TH] Wiki Main Page

RondoKnight:


TPGPL:
tracking in #3819

RondoKnight:
> tracking in #3819 

#3819 has some wrong tramslation, so i have edited on this #3820

TPGPL:
commented what you should do [here](https://github.com/ppy/osu-wiki/pull/3819#issuecomment-653775829)

PR #3925, opened 2020-07-23
Tighten lists in wiki articles

TPGPL:
used remark and `^ {0,}- .[^\n]*\n {0,}\n- ` regex to look for loose lists

Branch comparison for `remark ./wiki`:
- master
  - ⚠ 10819 warnings
- this branch
  - ⚠ 9917 warnings

Known issues:
remark acknowledges an empty line, serving as a break between list items, as a list-item-spacing warning.

e.g.
```md
- item
  - subitem
  
  ![](image)
- item
```
returns a list-item-spacing warning

I had to remove indent for images and add backslash hardbreaks for normal text. We might want to add no-indent criterium for block elements in lists like tables, code blocks or images.


cl8n:
this indent stuff is wrong, if you remove the indent it's no longer part of the list. you're splitting lists apart when you put other blocks in the middle

why wouldn't you expect your example to give a warning? that's not a tight list

TPGPL:
So should i just tighten the lists without removing the indent for blocks? The reason why i removed it is because in
```md
- item
  - item

  ![](image)
```

the image is treated as a subitem without the bullet while in

```md
- item
  - item
  ![](image)
```
or
```md
- item
  - item\
  ![](image)
```
the image is treated as extension of the first subitem.

cl8n:
the idea behind tight lists is if you have 1 paragraph per item, you can choose to remove the default spacing between paragraphs. here you have more than 1 paragraph per item so it's not something you can make into a tight list

I'm not really following why you want to do that

TPGPL:
So should i leave the lists loose when there's more than one paragraph per item, then? Wouldn't that return a remark warning for list-item-spacing?

cl8n:
leave loose and fix remaining warnings

TPGPL:
will open a new pr then

PR #3928, opened 2020-07-24
Fix deadlink in Storyboarding

Awesome-Aasim:
Fix dead link


cl8n:
should add a redirect for this if it doesn't already exist

PR #3933, opened 2020-07-25
Update osu!mania RC

Feerum:
https://osu.ppy.sh/community/forums/topics/1067554
https://osu.ppy.sh/community/forums/topics/1079647

- Added a sentence to "Jump" definition. To stay consistent with "Hand" definition.
- Changes "Stream" definition according to RC thread.
- Changed 4K NM and 7K NM Guidelines according to RC thread. 
- Minor fix in 7K NM Guideline, missing formatting. 

The 4K/7K thing was further discussed with the mania BNs on the BN Server and we came to that agreement for now.


cl8n:
for future stuff try to wrap up the discussion on the forums instead, there's not much point in that forum being public if in the end it's only BNs who get to make decisions

PR #3938, opened 2020-07-25
Unescape lists in osu! wiki Contribution Guide

cl8n:
turned out to be all of the escaped lists from https://github.com/ppy/osu-wiki/issues/3310#issuecomment-650934611


TPGPL:
already done in #3925

cl8n:
that has nothing to do with tightening lists ಠ_ಠ

TPGPL:
had to tighten them too ¯\_(ツ)_/¯

PR #3939, opened 2020-07-26
Adjust list item indent for wiki articles

TPGPL:
another take on #3925 

Branch comparison for `remark .`:
- 5503 warnings - master
- 4914 warnings - this branch


peppy:
Can you explain the addition of newlines in this PR?

TPGPL:
Basically if you want to add a list item without a bullet, you need to add a blank line between the item and the previous one. Otherwise, those two will merge and something like
```md
- list item
- list item
  list item without bullet
```
will be rendered
```md
- list item
- list item list item without bullet
```
and if you add a blank line between those two items, remark requires you to add blank lines between all list items of that specific level

MegaApplePi:
cf60f23 looks like it moved some files back to their old name.

MegaApplePi:
We should try to keep lists tight. I think the website still parses tight and loose lists differently. In some of the files you changed, they could have been fixed by removing indentation from a block that really didn't need be indented (looking at BBCode) and make them their own paragraph/block.

For example, this:
```md
- Et vitae
  ![](image)
- Facere et
  - Dolorem ipsum
    ![](image)
- Voluptatem aut
```

should have been this:
````md
- Et vitae

![](image)

- Facere et
  - Dolorem ipsum

![](image)

- Voluptatem aut
````

Though, if we are at a point where images need to be part of a list's item, then it probably shouldn't be using a list in the first place.

---

BBCode has stuff like this:
````md
- Dedicated button: ![Bold button](img/bold.png)
- Notes:
  - Use sparingly. Overuse may create discomfort and make the post harder to read.
- Syntax:

  ```
  [b] ... [/b]
  ```
````
when it could have been this:
````md
- Dedicated button: ![Bold button](img/bold.png)
- Notes:
  - Use sparingly. Overuse may create discomfort and make the post harder to read.
- Syntax:

```
[b] ... [/b]
```
````

or just not have "Syntax" be the last item in the list and put it in its own paragraph:

````md
- Dedicated button: ![Bold button](img/bold.png)
- Notes:
  - Use sparingly. Overuse may create discomfort and make the post harder to read.

**Syntax**

```
[b] ... [/b]
```
````
---

I know this will cause lists to be broken in the middle, but I think this is okay because it maintains tight lists, even if they are broken in the middle.

cl8n (approved):

cl8n:
@MegaApplePi if the images are understood to be part of lists then I think we should attempt to keep them syntactically part of the lists too. @TPGPL did put the code blocks and tables within tight lists so that should be best of both worlds provided osu-web's styling doesn't mess it all up

either way, gonna merge this one because it fixes a lot of issues and leaves the controversial parts easy to search for

PR #3940, opened 2020-07-26
[FR] Update osu!api

almaloste:
Update of  french translation of osu!api page


TPGPL:
conflicts with #3885

TPGPL:
request further changes to the translation at #3885

PR #3952, opened 2020-07-29
minor fixes on taiko rc translation & test

hikizisa:
test


hikizisa:
ah that was mistake, pull request isn't ready.. sorry

PR #3983, opened 2020-08-05
[FR] Add Bot Account

Oedipsos:
- Add French translation for `Bot Account` article.
- Removed reference for flags since it's used only once in the file


TPGPL:
all flags have to use reference link style actually no matter the usage count

Oedipsos:
Didn't know about that, it's not specified in the ASC. Thanks for your remark!

PR #4002, opened 2020-08-08
Fix link in MBC news post

pishifat:
fixed


pishifat:
didnt see that Pennek got this in #4001

PR #4006, opened 2020-08-09
Update BN/Dev listing and Staff log

hypercyte:
- Moved [Bibbity Bill](https://osu.ppy.sh/users/4446810) and [Garden](https://osu.ppy.sh/users/2849992) to the full BN members listing (osu!standard)
- Removed [Kyousukee](https://osu.ppy.sh/users/8842107) from the listing (osu!mania)
- Updated [yaspo](https://osu.ppy.sh/users/4945926)'s entry in the staff log for consistency.


cl8n:
something should be under BN section for yaspo and Scotty, else you won't see that they moved out of those groups if you're only looking at the BN section

and then I did "Added to NAT" instead of moved because I didn't want to imply the "removed from bn" part twice, I don't know if that's important at all, I'm just trying to make this painless for poor souls grabbing data from this staff log (me 😔 ) and eventually importing to osu-web

PR #4010, opened 2020-08-10
[KO] Partially update osu!taiko ranking criteria

Suinagi:
added more


cl8n:
#3953 updated osu!taiko rc in korean around the same time, looks more up-to-date than this pr

PR #4025, opened 2020-08-12
Update osu! gameplay images

TPGPL:


MegaApplePi:
I think these are going to be stuck behind the cache. You will need to change their names or append a query string.

PR #4034, opened 2020-08-14
[ES] Update AiMod

Lalo611:
I sincerely don't know what I've changed. I've started from scratch with this translation.
Ther order is like in es.md.
I did not translate some parts, because the message that AiMod gives is in english, so for the people that don't know the language will be hard to identify their problem. In addition I added the spanish translation in the explanation.


Lalo611:
fixed some mistranslations and bad accentuation.

TPGPL:
The original article was updated recently, so make sure your translation is up-to-date with the English one

Lalo611:
Updated according to en.md. Didn't find any other update than "Object is offscreen!"

PR #4047, opened 2020-08-17
Fix no-inline-padding remark warnings

TPGPL:


peppy:
Was this another manual change?

TPGPL:
mostly, as there's no stringify available for this rule

peppy (approved):

PR #4062, opened 2020-08-19
Add ARForest Featured Artist newspost

Ephemeralis:
🎵

### DIFF
--- a/news/2020-08-19-new-featured-artist-arforest.md
+++ b/news/2020-08-19-new-featured-artist-arforest.md
@@ -1,0 +1,63 @@
+---
+layout: post
+title: "New Featured Artist: ARForest"
+date: 2020-08-19 08:00:00 +0000
+---
+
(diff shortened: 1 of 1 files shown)


Nivalyx:
IT IS FINALLY HERE AAAAAAAAAAAAA ;~;

for the record i was also working on [Infection](https://osu.ppy.sh/beatmapsets/1225001#osu/2559136) actually but idk whether it's a good fit for the newspost or not as the mapset in question is still wip (・・；)

Ephemeralis:
I don't think there's enough room to fit another showcase mapset (especially if it's WIP still) in, usually the limit is 2 + 2 existing + 1 new for preview stuff

Nivalyx:
ah, okie

Ephemeralis:
Added it in anyway, don't see the harm

peppy (approved):

PR #4092, opened 2020-08-23
Update BN listing and Staff log

hypercyte:
#### osu!standard Probationary BN additions
- [Ryuusei Aika](https://osu.ppy.sh/users/7777875)
- [Een Persoon](https://osu.ppy.sh/users/10959501)
- [NeKroMan4ik](https://osu.ppy.sh/users/11387664)
- [Karen](https://osu.ppy.sh/users/3143784)
- [Sylas](https://osu.ppy.sh/users/3906405)

PR #4099, opened 2020-08-26
Add missing group language key

TPGPL:


peppy:
How are the outdated rules applicable to the new language key?

cl8n:
#4092 didn't update the translations because they didn't know how to translate "Romanian". just forgot to add outdated tags there

PR #4156, opened 2020-09-06
Conclude Catch The Beat Indonesia Cup 2020

Nivalyx:


Nivalyx:
accidentally uploaded the wrong podium file, oops

will open new pr with fixed podium image later

PR #4157, opened 2020-09-06
Conclude Catch The Beat Indonesia Cup 2020

Nivalyx:


wiki/Tournaments/CIC/2020/en.md
```diff
@@ -55,6 +57,10 @@ CIC 2020 is run by various osu! community members predominantly hailing from Ind
 - [Challonge brackets](https://challonge.com/CIC2020)
 - [Livestream](https://www.twitch.tv/osuIndonesia/)
 
+## Podium
+
+![Champion: Netamaru, 2nd place: dika312, 3rd place: Galaxi](img/podium.jpg)
```
TicClick, line 62: reads better when it's an enumerated list, or a table with crowns and stuff. also, did you intentionally omit the 4th place?

TicClick (changes requested):

wiki/Tournaments/CIC/2020/en.md
Nivalyx: this might be broken if you're looking at this from the pull request's rich diff view (http://github.com/ppy/osu-wiki/pull/4157/files) but once it got merged it'll automatically link up with the podium.jpg image that's also being pushed through in the same pull request. here's how it'll look once this pull request has been merged :

image

as for the podium's alt text, i only went to follow the format that has been commonly done with [OIT](https://github.com/ppy/osu-wiki/blob/master/wiki/Tournaments/OIT/2020/en.md) and some other tournaments previously here so i suppose everything should be fine.

TicClick (approved):
oh! we're good to go, then

PR #4180, opened 2020-09-14
Update IDC 2020 for Semifinals

fajar13k:
- Add Quarterfinals match results
- Add Semifinals mappool

### DIFF
--- a/wiki/Tournaments/IDC/2020/en.md
+++ b/wiki/Tournaments/IDC/2020/en.md
@@ -139,6 +139,22 @@ The Indonesian Draft Cup 2020 is run by various community members.
 
 ## Match results
 
+### Quarterfinals
+
+Saturday, 12 September 2020
+
+| Team 1 |  |  | Team 2 | Match link |
+| --: | :-: | :-: | :-- | :-- |
+| **Brazil** | **6** | 1 | Skinny Breakfast | [#1](https://osu.ppy.sh/community/matches/66836351) |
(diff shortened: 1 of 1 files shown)


wiki/Tournaments/IDC/2020/en.md
```diff
@@ -139,6 +170,22 @@ The Indonesian Draft Cup 2020 is run by various community members.
...
 ## Match results
 
+### Quarterfinals
+
+Saturday, 12 September 2020
+
+| Team 1 |  |  | Team 2 | Match link |
+| --: | :-: | :-: | :-- | :-- |
+| **Brazil** | **6** | 1 | Skinny Breakfast | [#1](https://osu.ppy.sh/community/matches/66836351) |
+| **awan fanclub 2** | **6** | 2 | `random.org` | [#1](https://osu.ppy.sh/community/matches/66840961) |
```
MegaApplePi, line 180: Is this team name actually spelt this way or is it just in code because of it looking like a link?
```
`random.org`
```

fajar13k: Used tilde to avoid hyperlink

MegaApplePi: I think escaping the periods would prevent the parser from hyperlinking it. I also don't think the website hyperlinks stuff that looks like links. Though I guess it should be fine as it is now.

MegaApplePi (approved):

PR #4234, opened 2020-09-23
Add New Featured Artist: Epica newspost

Ephemeralis:
🎉 🎵 100 FEATURED ARTISTS 🎵 🎉


Raidencio:
preferably change the artist name to EPICA in all the document in order to be consistent with the already ranked sets (except the oldest ones)

peppy (approved):

PR #4260, opened 2020-09-27
Update shortcut reference

dorothy3242:
Add double right click and double left click to the list of mouse commands in the editor.

### DIFF
--- a/wiki/Shortcut_key_reference/en.md
+++ b/wiki/Shortcut_key_reference/en.md
@@ -195,7 +195,7 @@ These shortcuts work anywhere within the beatmap editor:
 | Shortcut | Action |
 | :-- | :-- |
 | `Right Click` | Toggle new combo (when placing new hit objects) or delete object (when selecting). |
-| `Double Right Click` | Jump to selected objects' position on the timeline from any point in the song. |
+| `Double Right Click` | Jump to selected objects' position on the timeline or playfield from any point in the song. |
 
 #### General
 
--- a/wiki/Shortcut_key_reference/en.md
+++ b/wiki/Shortcut_key_reference/en.md
@@ -195,6 +195,7 @@ These shortcuts work anywhere within the beatmap editor:
 | Shortcut | Action |
 | :-- | :-- |
 | `Right Click` | Toggle new combo (when placing new hit objects) or delete object (when selecting). |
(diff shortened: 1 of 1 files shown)


Joehuu:
Isn't it double left click?

hypercyte:
double right click is correct (it teleports you to any object selected)

wiki/Shortcut_key_reference/en.md
```diff
@@ -195,6 +195,7 @@ These shortcuts work anywhere within the beatmap editor:
 | Shortcut | Action |
 | :-- | :-- |
 | `Right Click` | Toggle new combo (when placing new hit objects) or delete object (when selecting). |
+| `Double Right Click` | Jump to selected objects' position on the timeline from any point in the song. |
```
Joehuu, line 198:
```suggestion
| `Double Right Click` | Jump to selected objects' position on the timeline or playfield from any point in the song. |
```

peppy:
I'm confused. It should be double left click? If double right-click works that's an oversight, and won't be supported in lazer editor (while double left-click will be).

dorothy3242:
Yes it is definitely currently double right click, which is strange since the first click would delete objects if your cursor is above them. If it gets changed now to double left click or gets changed in lazer this shortcut reference could be updated again then I'd think.

peppy:
I've checked and it is, beyond doubt, double left click.

peppy (changes requested):
is wrong.

dorothy3242:
In another build or?  I checked this several times and double left did not do anything. My mouse settings are not customized. It's double clicking the same mouse key that removed objects.

peppy:
Again, you must have some custom setup. I'm telling you what the shortcuts are, as the developer of the game.

dorothy3242:
Ok so this confusion is because **both** in fact are correct.
Double left click if you click the start of the objects in the selection 
Double right if you click anywhere else.
I'll update the PR soon.

peppy:
If double right-click works that's an oversight, and won't be supported in lazer editor (while double left-click will be).

As in it should not be listed. Remove it or this cannot be merged.

dorothy3242:
I see. Ok, I've now removed double right click and renamed double left click to just double click, as it appears left clicking is just referred to as clicking in other portions of the existing page.

peppy (approved):

PR #4269, opened 2020-09-29
[EN] Ranking Criteria\Timing\Rules\4th Bullet

c-sig:
changes

> "There must not be extra uninherited timing points in any difficulty. These may accidentally affect main-menu pulsing, add unwanted sounds to the Nightcore mod, or cause timing to shift. Acceptable uses include:
> -Aligning beats of the Nightcore mod with the start of musical sections.
> -Accomodating for objects in musical sections requiring unsupported beat snap divisors (e.g. 1/5, 1/7)."

to

> "There must not be extra uninherited timing points in any difficulty. These may accidentally affect main-menu pulsing, add unwanted sounds to the Nightcore mod, or cause timing to shift. An acceptable use is to align beats of the Nightcore mod with the start of musical sections."

due to recent editor updates supporting previously unsupported divisors


c-sig:
https://github.com/ppy/osu-wiki/pull/4277/commits takes care of this better.

PR #4274, opened 2020-09-30
Reorganize and clarify title markers section of RC

hypercyte:
---

- Renames title "tags" to "markers"
- Clarifies that unofficial cuts roughly matching official cuts should use the official markers
- Specifies how to handle multiple markers in a title


cl8n:
can you explain more what d171a4c4858824b164a1df66a513b7484689a8a1 is supposed to do, I don't see how the meaning would be any different (what other track is there?)

hypercyte:
image

was thinking of a way to word it then noffy explained it better :)

cl8n:
i think instead u should move L94 under unofficial cuts, and reword it to be clear that you don't need Cut Ver if u have an official-ish marker for that case

dorothy3242:
I feel like that solution would be a bit more convoluted as it then requires reading both rules to understand one, rather than being able to be read mostly on their own and understood.

cl8n:
don't really see why 94 belongs in its current place at all with this unofficial cuts rule existing, because 94 is all about unofficial cuts

wiki/Ranking_Criteria/en.md
```diff
@@ -92,7 +92,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
 - **When a fictional character is credited as the singer of a song, the artist field is to be formatted in a `Character (CV: Voice Actor)` format.** For live action, credit the voice actor only.
 - **If the song is used in a television program, web series, or direct-to-video series, such as an opening/ending/insert song, use a `(TV Size)` marker at the end of the current title.** If there is an existing `TV size` marker in the title, the `(TV Size)` marker would replace it. Additionally, songs with existing `Short Ver` or `Game Ver` or similar markers in their titles must use `(Short Ver.)` or `(Game Ver.)` markers.
   - Note: If a mapset's song contains matching sections to the `TV size`/`Short Ver`/`Game Ver` song, in the same order, and is roughly the same length as the official cut song, the mapset's edit will also count as a `TV Size`/`Short Ver`/`Game Ver` respectively. Covers and remixes do not count.
```
dorothy3242, line 94:
```suggestion
- **If the song is used in a television program, web series, or direct-to-video series, such as an opening/ending/insert song, use a `(TV Size)` marker at the end of the current title.** If there is an existing `TV size` marker in the title, the `(TV Size)` marker would replace it. 
- **If a song title contains short or game version markers, their titles must be standardized to `(Short Ver.)` and `(Game Ver.)` respectively.**
- **Unofficial cut versions of songs must add a `(Cut Ver.)` marker at the end of the current title string.** If length markers are in the title of the track which it is cut from, `(Cut Ver.)` would replace the existing length marker. This is to distinguish unofficial cuts of a song from full length versions. Songs that are shortened in ways that nearly match their official versions, and songs that are a full loop of a looping track will not be considered cut.
  - Note: If a mapset's song contains matching sections to the `TV size`/`Short Ver`/`Game Ver` song, in the same order, and is roughly the same length as the official cut song, the mapset's edit will also count as a `TV Size`/`Short Ver`/`Game Ver` respectively. Covers and remixes do not count.
```

wiki/Ranking_Criteria/en.md
```diff
@@ -92,7 +92,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
-- **Unofficial cut versions of songs must add a `(Cut Ver.)` tag at the end of the current title.** If length markers are in the title of the track, `(Cut Ver.)` would replace the existing length marker instead. This is to distinguish unofficial cuts of a song from full length versions. Songs that are shortened in ways that nearly match their official versions, and songs that are a full loop of a looping track will not be considered cut.
+- **Unofficial cut versions of songs must add a `(Cut Ver.)` marker at the end of the current title.** If length markers are in the title of the track which it is cut from, `(Cut Ver.)` would replace the existing length marker instead. This is to distinguish unofficial cuts of a song from full length versions. Songs that are shortened in ways that nearly match their official versions, and songs that are a full loop of a looping track will not be considered cut.
```
dorothy3242, line 95:
```suggestion
```

wiki/Ranking_Criteria/en.md
```diff
@@ -92,7 +92,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
 - **If a song has been edited to have a higher tempo, use a `(Sped Up Ver.)` marker at the end of the current title string.** If there is an existing `Sped Up Ver` marker in the title, the `(Sped Up Ver.)` marker would replace it. Sped up songs in Techno, Trance, Dance, or other similar genres should use a `(Nightcore Mix)` marker instead.
```
dorothy3242, line 96:
```suggestion
- **If a song has been edited to have a higher tempo, use a `(Sped Up Ver.)` marker at the end of the current title string.** If there is an existing `Sped Up Ver` marker in the title, the `(Sped Up Ver.)` marker would replace it. Sped up songs in Techno, Trance, Dance, or other similar genres should use a `(Nightcore Mix)` marker instead.
  - Note: For tracks which are both cut and sped up, combine their markers into `(Sped Up & Cut Ver.)` and `(Nightcore & Cut Ver.)` respectively.
```

dorothy3242 (commented):
After clarifying with clayton what he meant, here is a set of suggestions to address the issue.

wiki/Ranking_Criteria/en.md
```diff
@@ -90,10 +90,12 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
+- **If the song is used in a television program, web series, or direct-to-video series, such as an opening/ending/insert song, use a `(TV Size)` marker at the end of the current title.** If there is an existing `TV size` marker in the title, the `(TV Size)` marker would replace it.
+- **If a song title contains short or game version markers, their titles must be standardized to `(Short Ver.)` and `(Game Ver.)` respectively.**
+- **Unofficial cut versions of songs must add a `(Cut Ver.)` marker at the end of the current title string.** If length markers are in the title of the track which it is cut from, `(Cut Ver.)` would replace the existing length marker. This is to distinguish unofficial cuts of a song from full length versions. Songs that are shortened in ways that nearly match their official versions, and songs that are a full loop of a looping track will not be considered cut.
   - Note: If a mapset's song contains matching sections to the `TV size`/`Short Ver`/`Game Ver` song, in the same order, and is roughly the same length as the official cut song, the mapset's edit will also count as a `TV Size`/`Short Ver`/`Game Ver` respectively. Covers and remixes do not count.
```
dorothy3242, line 96:
```suggestion
  - Note: If a mapset's song contains matching sections to the `TV size`/`Short Ver`/`Game Ver` song, in the same order, and is roughly the same length as the official cut song, the mapset's edit will instead count as a `TV Size`/`Short Ver`/`Game Ver`. Covers and remixes do not count.
```

hypercyte:
👍👍👍👍

cl8n (approved):

PR #4293, opened 2020-10-03
Add linting rule for wiki links

MegaApplePi:


cl8n (changes requested):
made pr to u, also why is it excluded from news?

MegaApplePi:
I merged the PR.

I thought that news didn't really care about the wiki link formatting.

cl8n:
might as well enforce it since there's no downside. idk why it wasn't already

cl8n (approved):
good check, thanks

PR #4303, opened 2020-10-05
Clarify CV formatting rule in RC

dorothy3242:
Minor wording fix to make it clearly allowed/required that if an anime character name is placed in the title field instead, i.e. `Title feat. Anime Girl` the same thing still applies.


hypercyte:
would make it clear for this if anyone wants an example https://osu.ppy.sh/beatmapsets/1271261 :)

cl8n:
"clearly allowed/required" seems like a stretch. I don't really understand this trend of making tiny technical wording changes to RC in hopes of clarifying a misunderstanding. why not just say it applies to title as well?

dorothy3242:
It's not really a misunderstanding when RC specifically says to do one thing when people would in fact expect something else. Adding "oh applies to title too" sounds like even more of a bandaid.. updated PR with a slightly different version.

cl8n:
updated version does solve my comment 👍

cl8n (approved):

PR #4309, opened 2020-10-07
Encourage a manual way to update changes from wiki

sr229:
While Upriver is a great service some advanced users might want to learn some new techniques as well. For that purpose, I documented a way to do so using git rebase

### DIFF
--- a/wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
+++ b/wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
@@ -42,7 +42,7 @@ When you created your fork of the `osu-wiki` repo, you took a snapshot of the co
 
 ---
 
-Fortunately, there are many ways to resolve this:
+Fortunately, there are two ways to resolve this:
 
 ### Using Upriver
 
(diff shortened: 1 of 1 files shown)


Ephemeralis:
this is a good start, but the article would probably benefit from another section detailing the use of a git GUI tool like Git Extensions to make the process more palatable for people not use to CLI apps

cl8n:
don't really understand why we'd encourage people to rebase like this. `git pull upstream master`?

Ephemeralis:
i think to encourage fast-forward instead of having a billion merge commits fuzzing up the commit line? that's what tripped me up a lot when first using git

cl8n:
`pull` (and the underlying `merge`) fast-forwards by default. if you're on master, you shouldn't be committing to it anyway so it'll always fast-forward. if you're on a branch that diverged from master (almost always means an open PR, for osu-wiki), you shouldn't overwrite history

peppy:
I honestly think adding this kind of complexity is the wrong direction.

Maybe a guide on using the github desktop app would be useful? Or link to one, I'm sure github have that covered 10x over. There shouldn't be any need to rebase or do anything custom (also what @cl8n said – standard pull/merge should be preferred over rebase. if you rebase an already PR'd branch you risk rewriting history and force-pushing, which is something we'd want to avoid).

Joehuu:
https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork (for CLI)

and/or

https://docs.github.com/en/free-pro-team@latest/desktop/contributing-and-collaborating-using-github-desktop/syncing-your-branch#merging-another-branch-into-your-project-branch (GitHub Desktop).

sr229:
I'll redo everything to just point towards using the GitHub Desktop way. Though I accidentally deleted the branch associated with this so I'll just open a new one with updated changes.

sr229:
> i think to encourage fast-forward instead of having a billion merge commits fuzzing up the commit line? that's what tripped me up a lot when first using git

This is the main reason, I was encouraging a more standardized and proven way to do fast-forwarding, while `git pull` works, `git rebase` is the recommended way.

peppy:
the only useful time for fast-forwarding is before PRing, though.

sr229:
Superseded by #4314

PR #4313, opened 2020-10-08
Update macOS installation

pishifat:
apparently the way to get around "unidentified developers" has changed

### DIFF
--- a/wiki/Installation/macOS/en.md
+++ b/wiki/Installation/macOS/en.md
@@ -12,9 +12,9 @@ The macOS version of osu! is experimental. The game may not function properly or
 
 1. Go to the unofficial [download page](https://osu.ppy.sh/community/forums/topics/682197), download `osu!.app.zip` and unzip it.
 2. You will see the `osu!.app` folder, which will look like an application. Right click it, and select `Open`.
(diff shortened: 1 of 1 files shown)


peppy:
right click + open still does work here. the right click part is important.

Xinnoh:
unsure how it differs between different apps but that method doesn't work for some other apps, have to go via system preferences.
tried downloading a new client to get the developer message to appear but getting a damaged file so can't really test it.

I think it would be fine if both were included since the method via system preferences will always work.

peppy:
It 100% works for normal applications. The exception here may be that osu-stable is running via wine, which does some fuckery with wrapping the true executable. Probably fine to go with the updated method in that case, since it'll work 100% of the time.

peppy (approved):

peppy:
Would appreciate adding a mention of lazer on this page, and recommending it as the best macOS client choice these days (wine doesn't even run on the latest macOS build as far as I know).

Techno-coder:
Steps for opening bundles from unidentified developers have slightly changed. You now need to right click and open twice for the Open button to show up (however, osu! itself will still only work after removing the quarantine attribute). 

> wine doesn't even run on the latest macOS build as far as I know
Not sure if you're aware but my latest Wineskin wrapper works on both Catalina and Big Sur (confirmed by a user).

PR #4320, opened 2020-10-09
Update unplayed search filter usage

C0D3-M4513R:
The search filter unplayed doesn't work like described anymore, so I updated the part to reflect the actual behavior (as of build cuttingedge b20201003).

### DIFF
--- a/wiki/Interface/en.md
+++ b/wiki/Interface/en.md
@@ -162,7 +162,7 @@ Supported filters:
 - `mode`: Mode. Value can be `osu`, `taiko`, `catchthebeat`, or `mania`, or `o`/`t`/`c`/`m` for short.
 - `status`: Ranked status. Value can be `ranked`, `approved`, `pending`, `notsubmitted`, `unknown`, or `loved`, or `r`/`a`/`p`/`n`/`u`/`l` for short.
 - `played`: Time since last played in days
-- `unplayed`: Unplayed maps. No value (use `unplayed=` for unplayed maps or `unplayed!=` for played maps)
+- `unplayed`: Unplayed maps. No value. Only supported comparison is `=`. Any other comparison will result in the comparison of `=`.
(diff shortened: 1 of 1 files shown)


wiki/Interface/en.md
```diff
@@ -162,7 +162,7 @@ Supported filters:
 - `mode`: Mode. Value can be `osu`, `taiko`, `catchthebeat`, or `mania`, or `o`/`t`/`c`/`m` for short.
 - `status`: Ranked status. Value can be `ranked`, `approved`, `pending`, `notsubmitted`, `unknown`, or `loved`, or `r`/`a`/`p`/`n`/`u`/`l` for short.
 - `played`: Time since last played in days
-- `unplayed`: Unplayed maps. No value (use `unplayed=` for unplayed maps or `unplayed!=` for played maps)
+- `unplayed`: Unplayed maps. No value. Only supported comparison is `=`. Any other comparison will result in the comparison of `=`.
```
TPGPL, line 165:
```suggestion
- `unplayed`: Unplayed maps. Any comparison can be used, as long as it has no set value.
```
might be better

C0D3-M4513R: That would imply, that `!=`
works, which is not the case, and that `<`,`<=`,`>` and `>=` give meaningful and different search results.

TPGPL: what about
```suggestion
- `unplayed`: Unplayed maps. Any comparison with no set value can be used, however all of them will have the same result.
```

C0D3-M4513R: I feel like you would need to read that twice to understand, because the result mentioned isn't always connected to the Unplayed Maps.

What about:

```suggestion

-- `unplayed`: Unplayed maps. No value. Only supported comparison is `=`. Any other comparison will result in the comparison of `=`.
+- `unplayed`: Shows only Unplayed maps. A comparison with no set value must be used. The comparison itself is ignored.

```

cl8n (approved):

PR #4344, opened 2020-10-16
[RU] Add staff logs (2007 to 2020)

TicClick:


cl8n (approved):
seems fine, I don't want to encourage translations of this though (#2486) since it's updated so often and will be replaced soon by the website

wiki/Staff_Log/ru.md
```diff
@@ -0,0 +1,25 @@
+---
+tags:
+  - зачисление
```
TPGPL, line 3: shouldnt have the English tags as well?

TicClick: should it? I've thought tags are always translation-specific, according to ASC

TicClick:
putting this on hold until a post from a native

wiki/Staff_Log/ru.md
cl8n: it depends if you want to people using the Russian locale on web to come to RU or EN when searching using English words, since the RU/EN search results will be mixed together. but in practice the inclusion of English tags usually doesn't change anything, because even without English tags, the search boost on your own locale is super high and trumps over every kind of match for EN (e.g. searching "Developers" with Chinese locale has the ZH page 1st and EN page 3rd, even though that word appears nowhere in ZH and is the exact title of EN)

imo just follow ASC, there shouldn't be any problems with it, and it would be an issue with osu-web if there are. probably due for a checkup on wiki searches anyway bc the behavior has been weird for a long time. looking back I'm pretty sure I added English tags to a lot of translations without thinking this through 🤔

wiki/Staff_Log/2019/ru.md
```diff
@@ -0,0 +1,335 @@
...
+- [Namki](https://osu.ppy.sh/users/5248582) - некомпетентность (2019-04-01)
+- [alienflybot](https://osu.ppy.sh/users/636114) - уход на покой (2019-04-12)
+- [Yales](https://osu.ppy.sh/users/2377881) - уход на покой (2019-04-16)
+- [wonjae](https://osu.ppy.sh/users/5032045) - уход на покой (2019-04-25)
+- [DMT9 Nowa](https://osu.ppy.sh/users/5428909) - уход на покой (2019-04-27)
+- [\_WWWWWW\_WWWWWW\_](https://osu.ppy.sh/users/1981781) - неподобающее поведение (2019-04-30)
+
+## Май
+
+- **Quality Assurance Team became Nomination Assessment Team** (2019-05-06)
```
UristMcKobold, line 102: Не перевёл…

UristMcKobold (changes requested):

UristMcKobold (approved):
Looks sufficient enough.

TPGPL (approved):

PR #4369, opened 2020-10-24
Fix mascots links

Semro:
This PR fixes links to the mascots across all wiki.


TPGPL (approved):
adding the hyphens will not work for outdated translations but thats not something we should care about i think ¯\_(ツ)_/¯

PR #4383, opened 2020-10-29
Remove reference to Kroytz in recent Project Loved News article

Coppertine:
the only time Kroytz was stated in the ctb mapset was the hitsounds (which was actually done by DJPop as it might be assumed that the ctb mapset's hitsounds are from the ranked std/mania mapset)


cl8n:
I told him I'd credit the hitsounding here. is it not from kroytz set though? I was talking to him about this yesterday 🤔

Coppertine:
Well, it would have been from Kroytz's set, however (even though you were talking to him about it), from the tags of the ranked set, you could evaluate that the hitsounds were done by DJPop (even Kroytz said it himself on stream).

peppy (approved):

cl8n:
i wrongly assumed he was giving his permission because he made the hitsounds, when really he just had them in his maps. thanks

PR #4397, opened 2020-11-01
[ZH] OWC2020 last minute reschedule

Kotoki1337:


Joehuu:
Should be moved one line below, see https://github.com/ppy/osu-wiki/pull/4396/files.

Joehuu (approved):

PR #4409, opened 2020-11-05
Fix broken and redundant redirects

cl8n:
skipped "star" because it looks like that article is supposed to be a disambiguation one and will need to reorganize some others to make it work


peppy:
If you're going to rename the bancho page, how about also fixing it being outright incorrect (probably the reason so many people think bancho runs on IRC or some shit).

wiki/redirect.yaml
```diff
@@ -272,15 +270,11 @@
...
-"geki":                        "Glossary#geki"
-"elite_beat!":                 "Glossary#geki"
-"激":                           "Glossary#geki"
-"grade":                       "Glossary#grade"
-"grades":                      "Glossary#grade"
-"grid_snapping":               "Glossary#grid-snapping"
-"grid_snap":                   "Glossary#grid-snapping"
+"fc":                          "Full_combo"
+"elite_beat!":                 "Geki"
+"激":                           "Geki"
```
TPGPL, line 275:
```suggestion
"激":                          "Geki"
```

wiki/redirect.yaml
```diff
@@ -293,13 +287,12 @@
 "drop":                        "Glossary#juice-stream"
 "droplet":                     "Glossary#juice-stream"
 "juice_stream":                "Glossary#juice-stream"
-"katu":                        "Glossary#katu"
-"katsu":                       "Glossary#katu"
-"beat!":                       "Glossary#katu"
-"喝":                          "Glossary#katu"
+"katsu":                       "Katu"
+"beat!":                       "Katu"
+"喝":                           "Katu"
```
TPGPL, line 292:
```suggestion
"喝":                          "Katu"
```

peppy (commented):
guess we can just go with this for now

PR #4417, opened 2020-11-07
Update reporting from user profile section in Handling Foul Play

Trigonoculus:
Not sure if what's already written is correct or not, as nothing happens when I mouseover a user's profile picture. The change in this PR is based on this screenshot:

image


peppy (approved):
A screenshot may be helpful, but this reads correctly.

TPGPL (approved):

PR #4476, opened 2020-11-17
Remove another defunct track from FA article

Ephemeralis:


peppy:
For full disclosure: the artist did not have correct permissions to license this song to us.

PR #4511, opened 2020-11-22
[DE|FR|IT|JA|KO|PL|PT|TL] Update GMT listing

pimpG:
updated GMT listing of all outdated article languages, they probably need some area of focus tweaks but the member lists are now up to date.


sr229 (approved):
Looks like nothing major changed other than the GMT listings so this is fine.

TicClick (approved):

PR #4546, opened 2020-11-30
Update SOL2 for Round of 32

ItsMestro:


wiki/Tournaments/SOL/2020/en.md
```diff
@@ -181,7 +198,126 @@ More detailed qualifiers results can be accessed via this [spreadsheet](https://
 
 ### Group Stage
```
TPGPL, line 200: `*-win by default-*` -> `*win by default* for all cases

TPGPL (changes requested):

TPGPL (approved):

PR #4549, opened 2020-12-01
Add Spotlights autumn playlist D

cl8n:


TPGPL (changes requested):
square brackets need to be escaped

TPGPL (approved):

PR #4552, opened 2020-12-01
Remove escaping from square brackets

TPGPL:
- `\\\](?!\()` -> `]`
- `\\\[` -> `[`

apparently its not needed at all ¯\_(ツ)_/¯


wiki/BBCode/pl.md
```diff
@@ -298,7 +298,7 @@ Dodaje obraz z łącza.
 
 Jeżeli obraz nie znajduje się pod danym adresem, załączenie go nie będzie możliwe.
 
-**Nie dodawaj obrazów, używając lokalnej ścieżki dostępu!** Wskazanie `C:\Użytkownicy\[nazwa]\Obrazy\obraz.jpg` (lub innej podobnej) **nie zadziała**
+**Nie dodawaj obrazów, używając lokalnej ścieżki dostępu!** Wskazanie `C:\Użytkownicy[nazwa]\Obrazy\obraz.jpg` (lub innej podobnej) **nie zadziała**
```
peppy, line 301: Not sure if this is the only case but this looks to be incorrectly applied.

TPGPL: checked through ~200 files from diff (everything except tournament articles but they should be okay anyways) and found nothing wrong except that one

PR should be fine now i guess

peppy:
Have you checked the CI results for this? It sure as hell looks like it is needed in non-nested cases.

TPGPL:
Yeah for some reason not escaping square brackets returns no-undefined-references warnings, which makes no sense as the syntax for references is:

```md
[link text][link ref]
(...)
[link ref]: /wiki/Link
```

I guess we should probably ignore them for now and wait for the fix from remark, as there's nothing wrong with not escaping square brackets (as can be seen [here](https://babelmark.github.io/?text=Lorem+ipsum+dolor+sit+amet%2C+consectetur+adipiscing+elit.+Sed+ac+hendret+tortor.+Mauris+sed+mauris+maximus%2C+%5Btest%5D+tincidunt+risus+vel%2C+ultriceien+%5Btest%5D+ligula+sed+massa.+dapibus%2C+sit+amet+gravida+metus+tincidunt.+Nulla+pellenaoreet.+Maecenas+porta+eros+%5C%5Btest%5C%5D+in+dolor+amet.%0A))

cl8n:
this is my fault for not being clear in discord the other day, `[` and `]` *do* have special meaning outside of links whenever they're paired, because `[ref]` and `[ref][]` are equivalent to `[ref][ref]` (markdown drives me crazy sometimes...). I think that's a good warning to keep because it suggests you might have forgotten to write a definition. it's just within link text that brackets don't have to be escaped because you couldn't have meant anything else, aside from the special `](` case

TPGPL:
Closing until a figure out a good regex solution, its not anything urgent anyway ¯\_(ツ)_/¯

PR #4589, opened 2020-12-10
Fix broken table

kj415j45:


TicClick (approved):

cl8n:
iirc this was intentional to get writers to notice it's broken and decide which header is more appropriate. too subtle to be important i guess

anyway i dont recommend using this template now because it's pretty old (#2621). you'd be better off copying a recent tournament and editing from there, assuming you want to take the formulaic approach to writing these articles (you don't have to!)

PR #4596, opened 2020-12-11
Fix CI by updating ini 1.3.5 -> 1.3.8

TicClick:
example error: https://github.com/ppy/osu-wiki/runs/1539026316


package-lock.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "osu-wiki",
+    "lockfileVersion": 2,
     "requires": true,
-    "lockfileVersion": 1,
```
cl8n, line 4: revert, most people use LTS node which doesn't have support for this yet.

cl8n (changes requested):

cl8n (approved):

PR #4605, opened 2020-12-13
Remove Countries_that_participated_in_osu!_tournaments article

TPGPL:
I don't think there's a point in keeping an article like this on the wiki, as it requires constant updates, serves little/no purpose and provides no necessary information


cl8n:
I'd mark these all outdated and leave them here instead, it does provide a useful aggregation to whoever is looking for OWC statistics. I referred someone to this just last week because they were looking for countries that only played once in earlier OWC

it'd be nice if this (and anything similar) could live in some place where it doesn't need manual upkeep but I don't think that exists rn

TPGPL:
okay then, gonna update the listing in another PR

PR #4619, opened 2020-12-14
[RU] Lowercase languages in group-info

cl8n:


TicClick (approved):
bless you. was just about to fire up the editor

PR #4631, opened 2020-12-16
RC: Add detailed reference for modified hepburn.

dorothy3242:
Currently the modified hepburn rule for romanisation only links to the wikipedia page, which is mostly a very basic overview. While that is fine for most cases, it means there's no good link for a reference for more technical cases that would benefit from looking at the full rules. So I added the link to make it more accessible.


wiki/Ranking_Criteria/en.md
```diff
@@ -109,7 +109,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
-- **Songs with Japanese metadata must use the [Modified Hepburn romanisation](https://en.wikipedia.org/wiki/Hepburn_romanization#Features) method in romanised fields.** The same applies to the Source field if a romanised Source is preferred by the mapper. As a non-unicode field, long vowels such as `おう` and `うう` should be romanised into `ou` and `uu` to avoid macrons.
+- **Songs with Japanese metadata must use the [Modified Hepburn romanisation](https://en.wikipedia.org/wiki/Hepburn_romanization#Features) method in romanised fields.** An in-depth reference on the system can be found [here](https://www.loc.gov/catdir/cpso/romanization/japanese.pdf). The same applies to the Source field if a romanised Source is preferred by the mapper. As a non-unicode field, long vowels such as `おう` and `うう` should be romanised into `ou` and `uu` to avoid macrons.
```
TicClick, line 112: please don't use "here" as a url title, pick the title of a web page or the site at least

TicClick (changes requested):

wiki/Ranking_Criteria/en.md
```diff
@@ -109,7 +109,8 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
-- **Songs with Japanese metadata must use the [Modified Hepburn romanisation](https://en.wikipedia.org/wiki/Hepburn_romanization#Features) method in romanised fields.** The same applies to the Source field if a romanised Source is preferred by the mapper. As a non-unicode field, long vowels such as `おう` and `うう` should be romanised into `ou` and `uu` to avoid macrons.
+- **Songs with Japanese metadata must use the [Modified Hepburn romanisation](https://en.wikipedia.org/wiki/Hepburn_romanization#Features) method in romanised fields.** Further 
+reference can be found [on in-depth romanization tables](https://www.loc.gov/catdir/cpso/romanization/japanese.pdf). The same applies to the Source field if a romanised Source is preferred by the mapper. As a non-unicode field, long vowels such as `おう` and `うう` should be romanised into `ou` and `uu` to avoid macrons.
```
TicClick, line 113:
```suggestion
- **Songs with Japanese metadata must use the [Modified Hepburn romanisation](https://en.wikipedia.org/wiki/Hepburn_romanization#Features) method in romanised fields.** Further reference can be found on [in-depth romanization tables](https://www.loc.gov/catdir/cpso/romanization/japanese.pdf). The same applies to the Source field if a romanised Source is preferred by the mapper. As a non-unicode field, long vowels such as `おう` and `うう` should be romanised into `ou` and `uu` to avoid macrons.
```

TicClick (approved):

PR #4668, opened 2020-12-22
[ID] Add osu!catch Ranking Criteria

IdrusShahab:

PR #4669, opened 2020-12-22
[ID] Add Mapping and Modding Timeline

towakane:


Nivalyx:
checked this pr in a glance earlier with @rorre, it seems that this suffers from the same problem with [the other pr that has also been recently opened](http://github.com/ppy/osu-wiki/pull/4668#issuecomment-749512263) where things are ***being excessively copied from Google Translate*** -- there's a lot of things that went incredibly out of context in this pr as you're translating all the stuffs 1:1.

another things that are also worth mentioning include (but not limited to) :

- broken urls, most likely as you're copying stuff directly from Google Translate without manually (for example in Line 123 `[Performance Point](/wiki/Performance_)` should be `[Performance Point](/wiki/Performance_Points)`)
- translation inconsistencies, esp. when it comes to the more technical terms (for example in Line 184 you kept `Quality Assurance Team` as is but in Line 272 you had it changed and translated to `Blog Tim Jaminan Mutu` for no apparent reason)
- a LOT of forced translations on stuffs that shouldn't even be translated at all (and vice versa), such as `Dewan Kriteria Peringkat` and `Artis Unggulan`

as with the other pr, your effort is appreciated but unfortunately i really can't see this being approved by any of us other Indonesian language contributors without this being completely overhauled from the ground up.

TicClick:
closing this as of now to hide from the list of open PRs. don't hesitate to reopen it once all issues pointed out by Niva (translation quality, linting checks, ...) are addressed

PR #4689, opened 2020-12-27
[KO] Add Game modifier

Ge-n-z:


Ge-n-z:
well.. does anyone know how to fix this error?
[https://i.postimg.cc/J0gRVLss/dadasddasd.png](url)

Nivalyx:
for the first error, it is caused by the double blank line on Line 61 and 62. simply removing either one of them should fix this.

image

for the second error, it's because how you write the square brackets (`[` and `]`) over Line 20 :

image

in Markdown, by default square brackets are used to indicate a link ([more on that here](https://www.markdownguide.org/basic-syntax/)) so if you're trying to write something in square brackets *that is not intended to be a link* you have to write it as `\[` and `\]`  instead. in this case, changing `['점수 배수']` to `\['점수 배수'\]` should fix the problem.

Ge-n-z:
I really appreciate your comment. thanks <3

TicClick:
merging (2+ weeks without a review from a native speaker)

TicClick (approved):

TicClick:
sigh. I absolutely forgot I added this note on top: https://github.com/ppy/osu-wiki/blob/master/wiki/Game_modifier/en.md#L15
@Ge-n-z could you add this to the translation really quick, please?

Ge-n-z:
wait a min

Ge-n-z:
I'm not sure what I exactly have to do. So, I made 2 pull req. Check them :)

PR #4652, opened 2020-12-20
Rewrite Beatmaps and Beatmapsets

TicClick:

PR #4701, opened 2020-12-29
Add Beatmap category article and fix URLs across the wiki

TicClick:
a follow-up to #4652


TPGPL (commented):
shouldn't you delete the translations at `wiki/Beatmaps` if you fully moved them to `wiki/Beatmaps/Category`?

TicClick:
> shouldn't you delete the translations at `wiki/Beatmaps` if you fully moved them to `wiki/Beatmaps/Category`?

I considered that, but translations, although outdated, have a bit of information on the general topic of beatmaps, and are referred to by many other articles, so removal will only harm the wiki.

On the other hand, I though about removing just the part about categories, but it's kinda problematic when it's ~90% of the article, with remaining 10% that look like a stub despite osu! having been built around the concept of beatmaps, and you can't just put a *Main article: [Beatmap categories](/)* because it's definitely not the main article for what's left, neither it belongs to translations.

TPGPL:
so maybe lets remove the translations at `Beatmaps/Category`, I dont understand why would we duplicate those files instead of just leaving them outdated at `Beatmaps`

TPGPL (approved):

PR #4709, opened 2020-12-29
Create tablet driver stub

paper-toaster:
Saw some users talking abt tablet drivers in osu-chat a while ago, thought it might be useful to mention in the wiki

### DIFF
--- a/wiki/Gameplay/Tablet_driver/en.md
+++ b/wiki/Gameplay/Tablet_driver/en.md
@@ -1,0 +1,27 @@
+---
+stub: true
+tags:
+  - tablet
+  - driver
+  - hawku
+  - devocub
+  - opentabletdriver
+---
+
+# Tablet driver
+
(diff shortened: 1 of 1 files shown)


bdach:
Word of a warning, there might be pushback on merging this. I'm not sure that installing custom drivers is something that the official osu! wiki might want to be mentioning at all.

At the very least, this stub should probably include a mention about kernel-mode and non-kernel-mode "drivers", and the potential dangers that installing the former might inflict on a user's operating system. There are very real risks involved with them (including maybe-false-but-maybe-not flagging from anticheat, aside from the fact that installing a kernel-mode driver from a potentially unaudited source is, in the worst case, paramount to letting some random internet person have full access to pretty much everything).

Joehuu:
Yep, or list these drivers under https://github.com/ppy/osu-wiki/blob/master/wiki/Projects/en.md#community (gameplay section) where there's a caution hatnote already.

paper-toaster:
@bdach @Joehuu maybe something like this?

```
*Caution: osu! neither endorses nor supports any unofficial drivers made for osu!. Installing [kernel](https://en.wikipedia.org/wiki/Kernel_(operating_system) "Wikipedia")-level software from untrusted sources can be extremely dangerous; use at your own risk.*
```

peppy:
Not sure we need this. Let people share this stuff on other sites instead? We will be making these obsolete in a coming lazer update, either way.

PR #4711, opened 2020-12-29
Use singular noun as a title for "Beatmaps" and "Beatmapsets"

TicClick:


Ephemeralis:
looks okay on the legal side of things as far as i can see

wiki/Beatmap/Beatmapset/en.md
```diff
@@ -0,0 +1,15 @@
+---
+stub: true
+---
+
+# Beatmapset
```
peppy, line 5: this is wrong. this is a term which should not exist.

peppy (changes requested):

wiki/Beatmap/Beatmapset/en.md
peppy: a beatmap contains difficulties, a "beatmapset" is a legacy internal term we use to collect difficulties of a beatmap. it has no place in casual discussion. use the terms "beatmap" and "difficulties" like we do everywhere on osu-web.

TicClick: if it shouldn't exist, why don't we remove the article as a whole and make a redirect to "Beatmap" instead with a small note about its historical/internal value? also, what about beatmapID/beatmapsetID articles, as we have these, too?

as for the term usage, I'll address it in a while

peppy: These IDs should also be internal. They are not intended to be used or consumed by the user. They should be documented on the API wiki or in the code.

TicClick:
it's easier to close this PR and go step by step:
- [ ] move `Beatmaps` to `Beatmap` and fix links
  - [ ] `Beatmaps` -> `Beatmap`
  - [ ] `Beatmaps#category-name` -> `Beatmap/Category#category-name`
  - [ ] `Beatmapsets` -> `Beatmap`
- [ ] move everything from `Beatmapsets` up a level into `Beatmap` and fix links
- [ ] add a note about beatmaps being called "beatmapsets" in the past, and another note on where to see a beatmap's/difficulty's identifier
- [ ] move `Difficulties` to `Beatmap/Difficulty` and fix links

PR #4714, opened 2020-12-30
Clarify examples given for BN contribution rule

radarrrrrr:
Tweaks the rule for not nominating maps you contributed to to include the mention of slider velocity changes in osu!taiko and osu!mania, where slider velocity changes are crucial aspects of map design (adjusting scroll speeds is pretty impactful, as opposed to individual slider speed adjustment), and thus is made to be treated on a similar level as hitsounds, storyboards, or regular object placement. Currently, someone could create all the SV for an osu!taiko map and then proceed to nominate it, which feels kinda meh considering its already common practice to add SV'ers to the tags for osu!taiko maps if present

### DIFF
--- a/wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
+++ b/wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
@@ -36,7 +36,7 @@ Each mode has its own required minimum for monthly activity.
 ## Nomination
 
 - **Do not nominate beatmaps with unrankable issues.** The General Ranking Criteria and additionally all mode-specific Ranking Criteria apply to all sets with mode-specific difficulties.
-- **Do not nominate beatmapsets that you contributed to.** This applies to your own beatmaps, collaboration difficulties you participate in as well as beatmaps you made other major contributions to in the form of hitsounds, slider velocity changes (osu!taiko & osu!mania) and storyboards.
+- **Do not nominate beatmapsets that you contributed to.** This includes any major contribution such as mapping, hitsounding, storyboarding, skinning, or slider velocity editing.
 - **Do not nominate a beatmapset without properly checking all difficulties of the mode(s) you are nominating.** Improper checking includes just running a modding tool before nominating or not even downloading the beatmap.
 - **Do not nominate a beatmap if you are unsure of something.** In such cases, ask fellow nominators for assistance before making any rushed decisions. This includes unfamiliar game modes, suspiciously gray area/unrankable issues, quality issues or complex timing.
 - **Do not nominate a beatmap if you cannot reasonably judge it.** Being able to reasonably evaluate a beatmap, either through playtesting or modding experience, is a core part of being able to judge a beatmap’s quality. If you are deemed not to have sufficient playing skill or modding experience, we may call your ability to judge the beatmap into question.
--- a/wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
+++ b/wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
@@ -36,7 +36,7 @@ Each mode has its own required minimum for monthly activity.
 ## Nomination
 
 - **Do not nominate beatmaps with unrankable issues.** The General Ranking Criteria and additionally all mode-specific Ranking Criteria apply to all sets with mode-specific difficulties.
-- **Do not nominate beatmapsets that you contributed to.** This applies to your own beatmaps, collaboration difficulties you participate in as well as beatmaps that you hitsounded or made a storyboard for.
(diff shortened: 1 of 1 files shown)


cl8n:
> Currently, someone could create all the SV for an osu!taiko map and then proceed to nominate it

this isn't true, rule says "do not nominate beatmapsets that you contributed to". the sentence following tells some examples of what it applies to but doesn't exclude it from being as broad as the bold part (it would've said "applies to only" or something, in that case)

seems like good clarification anyway

wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
```diff
@@ -36,7 +36,7 @@ Each mode has its own required minimum for monthly activity.
 ## Nomination
 
 - **Do not nominate beatmaps with unrankable issues.** The General Ranking Criteria and additionally all mode-specific Ranking Criteria apply to all sets with mode-specific difficulties.
-- **Do not nominate beatmapsets that you contributed to.** This applies to your own beatmaps, collaboration difficulties you participate in as well as beatmaps that you hitsounded or made a storyboard for.
+- **Do not nominate beatmapsets that you contributed to.** This applies to your own beatmaps, collaboration difficulties you participate in as well as beatmaps you made other major contributions to in the form of hitsounds, slider velocity changes (osu!taiko & osu!mania) and storyboards.
```
cl8n, line 39:
```suggestion
- **Do not nominate beatmapsets that you contributed to.** This includes any major contribution to mapping, hitsounding, storyboarding, skinning, or slider velocity editing.
```

cl8n (changes requested):

wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
dorothy3242:
```suggestion
- **Do not nominate beatmapsets that you contributed to.** This includes any major contribution such as mapping, hitsounding, storyboarding, skinning, or slider velocity editing.
```

peppy: why not just say "this includes ANY CONTRIBUTION at all, no matter how small" and call it a day?

venix12: that would possibly result in long and pointless discussions of "what even contribution is" nature, newer people could also be confused with it if they don't know how modding systems and stuff works. being more straightforward makes things simpler and more clear i suppose

Ephemeralis: i'm not sure how this isn't already covered in the initial wording

if SV is as impactful as the OP says to taiko then it is already "contribution" at a base line level without needing to be explicitly stated in the ranking criteria

the fact this PR exists is indicative of shenanigans going on somewhere that needs resolution, probably not the RC itself

wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
```diff
@@ -49,7 +49,7 @@ Each mode has its own required minimum for monthly activity.
 - **Only 3 significantly different versions of a song may be qualified in the same mode at the same time.** Significantly different versions include remixes, compilations, and mashups.
 - **Do not qualify beatmaps of modes which you are not a full member of, unless they were nominated by someone who is.** This ensures that a full nominator is always part of the process for each mode.
 - **Do not qualify a beatmap directly without any communication with the mapper.** The mapper might have done self-changes after the last nomination, so nominating a beatmapset without communicating with the mapper can lead to pointless disqualifications.
-- **Each mode in a hybrid beatmapset must be nominated by two nominators proficient in the respective mode.** Due to the hybrid nomination process not being fully supported, nominators proficient in multiple modes can not approve more than one mode.
+- **Each mode in a hybrid beatmapset must be nominated by two nominators proficient in each respective mode.** Multi-mode nominators may nominate multiple modes in one beatmapset, so long as they are proficient in each.
```
peppy, line 52: this change has nothing to do with the PR, does it?

cl8n: moved to #5155

wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
cl8n: > Ephemeral: i'm not sure how this isn't already covered in the initial wording

it is, this PR is just clarifying that because the rule didn't do a good job of conveying that its examples were just examples. I removed the "rule change" tag and will apply Noffy's suggestion

> peppy: why not just say "this includes ANY CONTRIBUTION at all, no matter how small" and call it a day?

I think this would be a rule change because currently BNs can make small contributions and still nominate-- and arguably modding itself is a small contribution, depending on how it's done. it could also be a source of confusion like Venix said

PR #4722, opened 2021-01-04
osu!catch Ranking Criteria update

osuSpectator:
The rule "Basic dashes must not be used more than four times between consecutive fruits. Higher-snapped dashes can be used up to two times between consecutive fruits, provided there isn't a direction change between them." for Platters will be split into two different ones according to https://osu.ppy.sh/community/forums/topics/1188695

The rule "Basic hyperdashes must not be used more than four times between consecutive fruits. If higher-snapped hyperdashes are used, they must not be used in conjunction with other hyperdashes or higher-snapped dashes." for Rains will be split into two different ones according to https://osu.ppy.sh/community/forums/topics/1188695

### DIFF
--- a/wiki/Ranking_Criteria/osu!catch/en.md
+++ b/wiki/Ranking_Criteria/osu!catch/en.md
@@ -121,7 +121,8 @@ Difficulty-specific rules and guidelines do only apply to the difficulty level t
 
 - **Hyperdashes may be used if the time between the ticks of the desired snapping is 125 ms or higher.** As an example, 1/3 hyperdashes would be allowed at 160 BPM and below, whereas 1/4 hyperdashes would be allowed under 120 BPM.
 - **Hyperdashes cannot be used on individual drops and/or slider repititions.** The accuracy and control required is unreasonable at this level and can create a situation where the player potentially fails to read the sliderpath.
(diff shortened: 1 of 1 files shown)


peppy (approved):

Darius-Wattimena:
When will this be merged? 
Would love to move the thread to finalized

TicClick:
sorry for the delay. it probably wasn't because no one was quite sure about the exact RC changes flow. if you see a pull request sitting on its place for a couple of days, go ahead and ask in #osu-wiki about its status to resolve the ambiguity

Ephemeralis:
RC change flow is as follows:

- New proposal is raised in the Ranking Criteria subforum on the forums.
- The proposal is discussed until a consensus is met at the decision of the NAT with an obvious majority of support from all participants in the thread. No exact numbers are given. 
- If the proposal is not contentious or has ongoing discussion present, it can then pass.
- Once passed, a PR encoding the changes into the Ranking Criteria can be made by any member of the NAT (presumably the one involved in finalizing the decision).
- The change is then discussed. peppy should be assigned to all RC PRs to sanity check them.
- After a week or two with no further discussion (or after approval from peppy), the change can then be merged.

Sound reasonable?

Issue #4723, opened 2021-01-04
Refactor "Beatmaps", "Beatmapsets", and "Difficulties"


TicClick:
(based on peppy's feedback in #4711)
- [x] move `Beatmaps` to `Beatmap` and fix links
  - [x] `Beatmaps` -> `Beatmap` (#4724)
  - [x] `Beatmaps#category-name` -> `Beatmap/Category#category-name` (already done)
- [x] edit `Beatmap`, so that it's about beatmaps (formerly *beatmapsets*), and `Difficulties` is about difficulties
  - [x] rewrite the articles (#4773)
    - intro + difficulty
    - note about beatmaps being called *beatmapsets* in the past
    - another note on where to see a beatmap's/difficulty's identifier 
  - [x] delete `Beatmap_ID` and `Beatmapset_ID` and fix links and redirects (#4802)
  - [x] move stuff around (#4803)
    - move everything from `Beatmapsets` up a level into `Beatmap` and fix links
    - remove `Beatmapsets`
    - fix links: `Beatmapsets` -> `Beatmap`
    - fix redirects
- [x] move `Difficulties` to `Beatmap/Difficulty` and fix links and redirects (#4808)
- [ ] fix wording across the wiki (`beatmap` -> `difficulty` where applicable, `beatmapset` -> `beatmap`) (#5495 + #5530)

_Originally posted by @TicClick in https://github.com/ppy/osu-wiki/issues/4711#issuecomment-753738890_

DoryanR:
Where are you on the last remaining point? If you are not on it I can see what I can do.

TicClick:
I have it planned

PR #4724, opened 2021-01-04
Rename articles and fix links (Beatmaps -> Beatmap)

TicClick:
part of #4723


wiki/Beatmapping/HP_drain_rate/en.md
```diff
@@ -7,7 +7,7 @@ tags:
 
 # HP drain rate
 
-*For health as a gameplay mechanic, see: [Health](/wiki/Beatmapping/Health)*
```
TPGPL, line 10: i think there actually shouldnt be any whitespace between the backslash and text/Markdown syntax

should fix for all occurences

TPGPL (approved):
looks fine for me

TPGPL (approved):

peppy (approved):
Let's go with this. Haven't done a thorough review, but we can catch any regressions on a second pass.

PR #4694, opened 2020-12-28
Add Cemara Mania Cup #1 (2020) wiki article

Nivalyx:
wiki article for the recently concluded Cemara Mania Cup #1 (2020) tournament

PR #4729, opened 2021-01-04
Update osu! South East Asia Tournament 4 to Quarterfinals

Nivalyx:
also applied a few grammatical improvements that were suggested on my other tournament pr (#4694)


TicClick (approved):
I've squashed your commits and removed a couple of trailing whitespaces prior to the merge

PR #4740, opened 2021-01-07
Update Project Loved Wiki Page

huuishuu:
Add 30s drain time requirement to the eligibility criteria.

### DIFF
--- a/wiki/Project_Loved/en.md
+++ b/wiki/Project_Loved/en.md
@@ -28,6 +28,7 @@ There are some very minimal criteria that need to be met for beatmaps to be nomi
 - Its creator is neither banned nor restricted
 - At least one difficulty can be passed
 - It has at least 30 favourites (only applies to osu!standard maps)
+- Every difficulty of a beatmapset must have a minimum [drain time](https://osu.ppy.sh/wiki/en/Gameplay/Drain_time) of 30 seconds.
 
 However, despite nearly all [Pending and Work-in-progress](/wiki/Beatmaps/Category#work-in-progress-and-pending) maps meeting these criteria, few are voted into the Loved category due to the more selective nomination and voting requirements.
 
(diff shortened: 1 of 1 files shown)


wiki/Project_Loved/en.md
```diff
@@ -28,6 +28,7 @@ There are some very minimal criteria that need to be met for beatmaps to be nomi
 - Its creator is neither banned nor restricted
 - At least one difficulty can be passed
 - It has at least 30 favourites (only applies to osu!standard maps)
+- Every difficulty of a beatmapset must have a minimum [drain time](/wiki/en/Gameplay/Drain_time) of 30 seconds.
```
MegaApplePi, line 31:
```suggestion
- Every difficulty of a beatmapset must have a minimum [drain time](/wiki/Gameplay/Drain_time) of 30 seconds.
```
The link should not have the `en/` part. Also, are we also moving away from using beatmapset?

huuishuu: I copied and pasted this exact point from the [Ranking Criteria](https://osu.ppy.sh/wiki/en/Ranking_Criteria) page so I'm not entirely sure.

wiki/Project_Loved/en.md
```diff
@@ -28,6 +28,7 @@ There are some very minimal criteria that need to be met for beatmaps to be nomi
 - Its creator is neither banned nor restricted
 - At least one difficulty can be passed
 - It has at least 30 favourites (only applies to osu!standard maps)
+- Every difficulty of a beatmapset must have a minimum [drain time](/wiki/Gameplay/Drain_time) of 30 seconds.
```
MegaApplePi, line 31: Oh, I did not realise that. I think it should be phrased a bit differently to match the above points. Maybe something like this?
```suggestion
- Every difficulty has at least 30 seconds of [drain time](/wiki/Gameplay/Drain_time)
```
If you don't think so, then I'll go ahead and merge.

huuishuu: Sure, it looks more uniform that way.

MegaApplePi (approved):

wiki/Project_Loved/en.md
TicClick: > are we also moving away from using beatmapset?

we totally are, but I need to organize the existing `Beatmap`/`Beatmapset` articles first, then play find-and-replace

PR #4742, opened 2021-01-07
Remove 18+ warning rule from Ranking Criteria

pishifat:
website's nsfw warning covers this now


cl8n:
that todo should be moved to the spoiler warning rule, wording has been wrong on it too since spoiler was moved to another section

peppy (approved):

PR #4749, opened 2021-01-08
Add initial pull request template

### LINKED ISSUE #4622: Look into better ways to ensure content parity
[the content parity rule for translations](https://osu.ppy.sh/wiki/en/Article_Styling_Criteria#content-parity) is written with the right idea in mind, but in practice it gets messy when the original English article isn't fully up-to-date or written well. translators (rightly) choose to correct the mistakes and deviate from English a bit, leading to [situations the rule was designed to prevent](https://discord.com/channels/188630481301012481/218677502141399041/788156447375622216) if they don't also update the English article.

imo it's a problem that:

- we are effectively making translators responsible for maintaining English articles, which at best is not what they want to spend their time doing, and at worst is impossible because they may not have the required English writing skills
- deviations from English can easily go unnoticed, since many languages have little to no reviewers that are experienced with osu-wiki
- following English as close as possible (according to the rule) means translations adopt plenty of bad writing, structure, and formatting from English, basically just putting off work for far-future contributors to deal with

I don't know if there's any good way to prevent these problems, just opening this issue for now to see if anyone else has had thoughts about this

TicClick:
aims to resolve #4622, or at least reduce the chance of receiving translations that diverge from the original articles. post your suggestions

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,19 @@
+### General (all pull requests)
+
+- [ ] I have read the [contribution guide](https://osu.ppy.sh/wiki/en/osu%21_wiki_Contribution_Guide)
+
+This pull request:
+
+- [ ] conforms to the [formatting](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting) and [writing](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting) style criteria;
+- [ ] is related to an existing issue (specify which one).
+
+### Translations only
+
+This pull request:
+
+- [ ] does not introduce any original content that does not exist in the original article(s);
+- [ ] will be reviewed by native speakers.
+
+### Additional information
+
+Anything you would like to communicate to the reviewers and/or osu! wiki curators goes here.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,18 +1,13 @@
 ### General (all pull requests)
 
-- [ ] I have read the [contribution guide](https://osu.ppy.sh/wiki/en/osu%21_wiki_Contribution_Guide)
-
-This pull request:
-
-- [ ] conforms to the [formatting](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting) and [writing](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting) style criteria;
-- [ ] is related to an existing issue (specify which one).
+- [ ] The changes follow the [formatting](https://github.com/ppy/osu-wiki/tree/master/wiki/Article_styling_criteria/Formatting) and [writing](https://github.com/ppy/osu-wiki/tree/master/wiki/Article_styling_criteria/Writing) style criteria.
 
 ### Translations only
 
-This pull request:
-
-- [ ] does not introduce any original content that does not exist in the original article(s);
-- [ ] will be reviewed by native speakers.
+- [ ] This pull request does not add any original content missing from the original article. Exceptions include:
+  - language community-specific information;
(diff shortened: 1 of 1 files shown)


TicClick:
@ppy/wiki what do you think?

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
+### General (all pull requests)
+
+- [ ] I have read the [contribution guide](https://osu.ppy.sh/wiki/en/osu%21_wiki_Contribution_Guide)
```
bdach, line 3: Yeah I don't know about this one item especially. There is nothing stopping people from just checking this off without actually reading. Yeah it would be a crappy move on their behalf but this is mostly performative, and I'd get tired of checking this off every time *very* fast if I were PRing to wiki more often.

Sad thing is that generally speaking random new contributors don't (*or won't*) read and that's from speaking from experience over at the lazer repo. You can try issue templates, contributing guidelines - if it's long or technical enough people will just ignore anyway.

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+
+- [ ] conforms to the [formatting](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting) and [writing](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting) style criteria;
+- [ ] is related to an existing issue (specify which one).
+
+### Translations only
+
+This pull request:
+
+- [ ] does not introduce any original content that does not exist in the original article(s);
+- [ ] will be reviewed by native speakers.
```
bdach, line 15: Would probably specify when that is the case (guessing formatting/markup changes only)?

.github/pull_request_template.md
TicClick: sad but true, this is useless the way I use it (as a requirement)

.github/pull_request_template.md
TicClick: nah, this is the main reason for the PR: we want translations to almost always [follow the original article](https://github.com/ppy/osu-wiki/blob/master/wiki/Article_styling_criteria/Formatting/en.md#content-parity), with exception to language-specific information. thinking of how to phrase that better

bdach: Ah sorry brainfart, meant "when that is *not* the case".

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+### General (all pull requests)
+
+- [ ] The changes follow the [formatting](https://github.com/ppy/osu-wiki/tree/master/wiki/Article_styling_criteria/Formatting) and [writing](https://github.com/ppy/osu-wiki/tree/master/wiki/Article_styling_criteria/Writing) style criteria.
+- [ ] There is an existing relevant GitHub issue (specify which one).
```
TPGPL, line 4: i don't think this one is actually needed

it might also scare away potential contributors?

TicClick: good point

TicClick:
now that I think about, it's still a lot to ask. the formatting rules are ~700 lines long; the writing ones is around 300 lines. for me personally, it's hard to imagine everyone consulting with ASC (article styling criteria) before making a pull -- hell, I myself don't do that every so often.

one thing that would definitely help is having some sort of a contribution guide that would go over most points of the ASC and give a really short explanation, while referencing the relevant parts of ASC for those who need to see the reasoning, or examples of how things should be done.

this way, we'd have a checklist you can go over before submitting changes for review, and the criteria would only serve as a reference.

TicClick:
not to mention that some things in ASC are already checked for automatically and enforced by linter, you don't need to repeat that step manually and spend extra time.

TicClick:
will reopen it after #4888

PR #4772, opened 2021-01-12
Fix remaining beatmap links

TPGPL:


TicClick (approved):
thanks for covering me

PR #4784, opened 2021-01-16
Update miss count plot in performance points newspost

### LINKED ISSUE #4783: Miss penalty in latest Performance Points Updates Post is confusing
Hi. This chart used in [the new Performance Points Updates Post](https://osu.ppy.sh/home/news/2021-01-14-performance-points-updates) is really confusing in my eyes (and to the few others I talked to about it).
image
It is labeled with "Miss penalty" but seems to show the actual multiplier applied to the aim and speed pp. 
If it would be the penalty in my eyes it should start at 0 and go up (as the penalty grows with each miss), not like it's now, which starts at 1 and goes down. 

The chart also starts at 0 misses and shows 0.97 for the new algo even though there (I hope at least) isn't a penalty from not missing at all. It should really be starting at 1 miss or show a 1x multi for misses<1. 

I would pr a updated version but I don't have the tools that was used to make this chart

bdach:
Closes #4783

Should have figured starting at 0 wasn't too smart. (For what it's worth [no-miss *is* handled properly.](https://github.com/Xexxar/osu/blob/eb40d3a5ba2201cca111a7f0ac9415a82c4878e5/osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs#L95-L97))

Figure the scatterplot makes more sense as the domain is both discrete and narrow enough to allow it being scatterplotted.


Joehuu (approved):
Seems better.

PR #4794, opened 2021-01-18
Update 3TWC to Round of 16

fajar13k:
- Add Round of 24 match results
- Add Round of 16 mappool
- New Zealand -> Oceaning Alliance in line 209


TicClick (approved):
looks about right

PR #4802, opened 2021-01-18
Remove Beatmap_ID and Beatmapset_ID articles

TicClick:
part of #4723, they are replaced by /wiki/Beatmap#Identification now


Joehuu:
I feel like this is ruining this idea said in https://github.com/ppy/osu-wiki/issues/3143:

> A wiki thrives and lives from having each article be intertwined with each other. It lives from the idea that you can go to an article you find or look up and you can follow down the rabbit hole of information by clicking every important term and be sent to a new article that tells you more about it.

Not sure if this applies to "beatmap(set} id" as it is not really a common word, but still.

TicClick:
I'd agree with you, but it's more of an internal infomation that only needs to be mentioned briefly (besides, these two are really similar with the exception of the terms themselves and 0/-1 being default values): https://github.com/ppy/osu-wiki/pull/4711#discussion_r550854245

Joehuu (approved):

PR #4808, opened 2021-01-19
Set the correct title and a proper category for Difficulties

TicClick:
part of #4723


TicClick:
I don't get why TicClick/osu-wiki@fba08c6bfb1e5612f8828659c6c91b1bab7fb294 doesn't show up..

wiki/Beatmap/Difficulty/en.md
```diff
@@ -1,15 +1,14 @@
 ---
 tags:
-  - difficulty
   - difficulty name
```
Joehuu, line 3: probably add the plural noun in tags now

```suggestion
  - difficulties
  - difficulty name
```

TicClick: makes sense

Joehuu (approved):

PR #4820, opened 2021-01-20
Add exemption to spinner starts in rule about snapping

kitty-cloud:
for the partial version of this proposal accepted by naxess, which was discussed here: https://osu.ppy.sh/community/forums/topics/1214161

(first time making a PR, hope i'm doing it right)

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -19,7 +19,7 @@ Changes to the rules and guidelines in this document are proposed and discussed 
 
 ### Rules
 
-- **No two hit objects can be placed on the same tick.** This includes hit circles and the durations of sliders and spinners. osu!mania beatmaps are exempt from this.
+- **No two hit objects can be placed on the same tick.** This includes hit circles and the durations of sliders and spinners. osu!mania beatmaps are exempt from this. The beginning of spinners in osu!standard may be placed on the same tick as a hit circle, slider end, or spinner end, as long as the two are [ordered correctly in the .osu file](https://osu.ppy.sh/community/forums/posts/7869661)
 - **There must not be any obscene imagery in the background/storyboard/video content.** This includes nudity, near-nudity, sexual references, violence, drug abuse, etc. See [Visual Content Considerations](/wiki/Rules/Visual_Content_Considerations) for more detailed rules.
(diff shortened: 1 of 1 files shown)


TicClick:
looks okay, but needs a look from NAT (@Naxesss?)/peppy

> RC change flow is as follows:
> 
> - New proposal is raised in the Ranking Criteria subforum on the forums.
> - The proposal is discussed until a consensus is met at the decision of the NAT with an obvious majority of support from all participants in the thread. No exact numbers are given. 
> - If the proposal is not contentious or has ongoing discussion present, it can then pass.
> - Once passed, a PR encoding the changes into the Ranking Criteria can be made by any member of the NAT (presumably the one involved in finalizing the decision).
> - The change is then discussed. peppy should be assigned to all RC PRs to sanity check them.
> - After a week or two with no further discussion (or after approval from peppy), the change can then be merged.

_Originally posted by @Ephemeralis in https://github.com/ppy/osu-wiki/issues/4722#issuecomment-761845070_

Naxesss:
This will need other NAT to confirm that it's accessible enough before proceeding, as there are a few concerns:
- Changing order of objects in the .osu may be too technically demanding for the RC, which is aimed at even new mappers.
- It's (afaik) impossible to tell that the order of objects is wrong when looking purely in the editor.  As a result, this change may have a too high reliance on third-party map checking tools, e.g. Mapset Verifier, to follow/enforce reasonably.

Objects on the same tick is generally undefined behavior in standard. It may seem to work ok currently, but there's no guarantee things will stay that way. As such, you will also need to check with a dev (probably peppy, since he'll look at this anyway) to confirm that the change won't break anything.

kitty-cloud:
>     * It's (afaik) impossible to tell that the order of objects is wrong when looking purely in the editor.  As a result, this change may have a too high reliance on third-party map checking tools, e.g. Mapset Verifier, to follow/enforce reasonably.

it's not, as explained by the linked lewski post; it's easy to tell by looking at the objects on the timeline

peppy:
How is this any different?

peppy:
Two hit circles should *not exist at the same time value*. The *only exception for this is mania*.

This may change in the future but is how it is for now. Please place the spinner *closer* but *not at the same millisecond value* as another hit object.

PR #4822, opened 2021-01-21
[KO] Add Game modifier

Ge-n-z:


TicClick:
this tag indicates that [no one has taken a look at your translation](https://github.com/ppy/osu-wiki/blob/master/wiki/Article_styling_criteria/Formatting/en.md#translations-without-reviews) and it was merged as-is. it shouldn't be removed by you

PR #4837, opened 2021-01-23
Add the article about SOFT5 tournament

WalterToro:
Will this be the end of WalterToro?, we will see.


TicClick (changes requested):
add a link to the tournament to `/wiki/Tournaments/SOFT`, please

TicClick (approved):

PR #4838, opened 2021-01-24
(updated) Exempt some spinner-starts from rule about unsnapped objects

kitty-cloud:
for this proposal again: https://osu.ppy.sh/community/forums/topics/1214161
with this wording, spinners can be "closer but not at the same millisecond value as another hit object", as advised by peppy's reply to the older PR for this proposal [here](https://github.com/ppy/osu-wiki/pull/4820#issuecomment-766273849)


peppy:
1ms is still going to potentially cause issues with scoring. this should not be merged without proper testing. @MillhioreF want to weigh in on this as well (i think you know the slider edge cases the best).

peppy (changes requested):
Blocking until properly tested. Those proposing these changes don't necessarily understand the nuances of gameplay.

MillhioreF:
Spinners starting 1ms after a circle or sliderend shouldn't cause any problems unless said spinner is shorter than the current OD's hitwindow, but that's unrankable anyway right? There are potential playability issues in taiko/mania converts but I don't think they'd be breaking.

peppy:
Can't this rule just remain unchanged, and AiMod be changed to allow spinners? The wording here is just confusing and it seems like that would clear things up better?

Also we don't reference "osu!standard" anywhere else, so whoever came up with this change didn't really put much effort into it.

peppy:
This is hanging on a response to my proposal above. If no one is against this direction, shall I make changes to AiMod?

kitty-cloud:
yes imo

peppy:
I've got a change in for the next cutting edge release which suppresses notices about rare snap divisors for spinners when they occur under 50ms away from the last hitobject. Please confirm that this is good enough to resolve this issue (and mean that we can close this PR as the change is no longer required).

PR #4853, opened 2021-01-26
Actualize font family for image annotations

TicClick:
if I understand correctly, Exo 2 is not used anymore in osu! and on the website


Nivalyx:
iirc there has been more fonts that have been added to the osu! environment since the last time this article was updated tho, with the most recent being the "Prompt" font for Thai language ([osu-web #7076](https://github.com/ppy/osu-web/pull/7076))...

these are all the fonts listed in that specific pr btw :

image

peppy:
I think this is fine as an initial fix. We shouldn't be using exo2 anywhere.

peppy (approved):

PR #4852, opened 2021-01-26
Add Beatmap Spotlights Season 3: Winter 2021 Playlists B

Nivalyx:
\+ a fix for a broken hyperlink url that i noticed on [the Beatmap Spotlights Season 3: Winter 2021 newspost](http://osu.ppy.sh/home/news/2021-01-18-beatmap-spotlights-season-3-winter-2021)

PR #4857, opened 2021-01-27
Fix link in Spotlights Winter 2021 news post

pishifat:


pishifat:
oh looks like #4852 included this change, will merge that

PR #4873, opened 2021-01-30
[KO] Update Registration

BamgoeSN:
Previously the file said "At most 8 characters long" for the password section


TicClick (approved):
thanks for catching that

PR #4904, opened 2021-02-03
Add accurate formula to determine score multiplier

LeaPhant:
The currently described way to determine the score multiplier for scorev1 std scoring is inaccurate, especially in edge cases.


TicClick:
I confess I have no idea whether the changes proposed here are true, but I think I know someone who does

peppy:
I think I would have to check the code to confirm this one, not 100% on it.

peppy:
Looks to be accurate:

IMG_0086

peppy (approved):

PR #4913, opened 2021-02-05
[KO] Add Hidden

Ge-n-z:


Ge-n-z:
I can't find what is the problem. Someone help me.

TicClick:
you missed links on lines 60 and 74, because `[text]` is a part of markdown syntax (and it is expected to have `(link)` after square brackets). if that wasn't your intention, you need to escape square brackets like this: `\[text\]`

Ge-n-z:
thx <3

TPGPL (approved):

PR #4699, opened 2020-12-29
Clarify rhythm incarnate tier info on Beatmap Spotlights

venix12:

PR #4914, opened 2021-02-05
Keep Beatmap_Spotlights up to date

TTTaevas:
Note that for some reason, the Winter 2021 season lacks image files (badges) with names finishing by "_3", at least compared to Summer & Autumn 2020. Due to that, I chose to let the pages show the Autumn badges rather than the Winter ones.


venix12:
the removal of ``_3`` badges is intentional, they are not a thing anymore

Nivalyx:
a lot of things in the pt-br locale seems to be outdated tho. for example the playlists for each season [have long been moved to a separate article to avoid congestion](http://osu.ppy.sh/wiki/en/Beatmap_Spotlights/Seasons) and quite a good bunch of other stuffs seems to have either been or deleted, reworded, or moved elsewhere :

image

image

i think rather than updating/fixing the pt-br article it would be wiser to simply add an "outdated" tag into it tho, so that it can be properly fixed up later by a native speaker in a future pr...

TTTaevas:
Alright, then I can update which badges the pages should show, with the exception of the Brazilian page.
Do you think I should revert pt-br.md back to its original state, so a native speaker can fully take care of updating it?

Nivalyx:
i believe so, yeah

TTTaevas:
Unless I'm mistaking, pt-br already features the "outdated" marker:
Screenshot_2021-02-05 ppy osu-wiki

Nivalyx:
ah, yeah, i think you are right 👀

i think everything should be good to go except for the one missing space between `当前赛季为` and `2021` in the zh.md :

image

Nivalyx (approved):
looks good on my end

peppy:
Should previous seasons not be listed on the page as well? Or are we happy to throw out the historical context?

Nivalyx:
hmm, i'm not the one who writes the current English article but iirc the documentation of all the seasons (both previous and current) are also on the wiki tho albeit on a slightly different page (http://osu.ppy.sh/wiki/en/Beatmap_Spotlights/Seasons) :

image

image

((as written above this is presumably done to avoid putting too many information in the main Beatmap Spotlights article))

TTTaevas:
#4699 had created a conflict, a96d186 should resolve it. My bad for the commit's name

peppy:
I see. No worries then.

TicClick (commented):
not sure why this is not merged yet -- apologies

TicClick (approved):

PR #4931, opened 2021-02-09
Add FringeRemover program

RoanH:
I [just released](https://osu.ppy.sh/community/forums/topics/1244643) a new program and the projects wiki page mentions that it is fine to add your own projects. I assumed that projects in each section are listed in alphabetical order (amended the commit because I managed to mess that up...).


peppy (approved):
cool stuff. osu! tries its best to fix these but there's edge cases which it can't easily, i guess.

PR #4933, opened 2021-02-09
[ID] Update Do you really want to ask peppy?

AncuL001:


hikikuman (changes requested):
You may want to check line 15, there's a word that is written twice `kepada kepadanya`.

Nivalyx (approved):

TicClick (approved):

PR #4992, opened 2021-02-25
Update FL4T to Quarterfinals

fajar13k:
- Add Group Stage match results
- Add Round of 16 mappool
- Add Round of 16 match results
- Add Quarterfinals mappool

### DIFF
--- a/wiki/Tournaments/FL4T/1/en.md
+++ b/wiki/Tournaments/FL4T/1/en.md
@@ -77,6 +77,26 @@ The osu!mania Flashlight 4K Tournament is run by various community members.
 
 ## Mappools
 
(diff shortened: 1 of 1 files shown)


TicClick:
forgive me my curiosity, but what is `Rice` in context of osu!mania?

fajar13k:
Quoting from the other staff:
> In osu!mania map pools, there's two overarching classifications of patterning: LN and Rice. LN (long note) maps being maps where the prominent difficulty comes from the sliders aka the long notes in the map and Rice meaning maps where the difficulty comes from non-sliders

TicClick (approved):
🍚

PR #4997, opened 2021-02-27
fixed a player name in the German team

marshallracer:
Azosial -> Asozial (account banned now, multiaccounting and cheating)


peppy (approved):
Not sure they deserve to have their name fixed

PR #5001, opened 2021-02-28
[PT-BR] osu!tourney

daviirodrig:
Add pt-br translation of osu!tourney articles


maotovisk (changes requested):
The grammar is really good and well-written, just a minor typo here and there.
I would suggest taking a look over the italic names rule, name such as trademarks, product names etc. doesn't need to be in italic.
Besides that everything seems to be okay.

daviirodrig:
@maotovisk Thanks for the review! I applied everything because I thought it was appropriate

By the way, next time it's better if you do the review [suggesting changes](https://haacked.com/archive/2019/06/03/suggested-changes/) directly in the comment, that way I can apply the changes in a more efficiently and you receive the co-authoring in the commit.

maotovisk (approved):

TicClick (approved):

PR #5013, opened 2021-03-02
Finalize A Labour of Love teams list

pishifat:


peppy (approved):
RIP teams that didn't submit :(

PR #5031, opened 2021-03-08
Add guideline about crediting media to RC

cl8n:
https://osu.ppy.sh/community/forums/topics/1191040

also fixes the TODO comment in the list above that didn't refer to anything (original referenced rule got deleted, I think)

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -30,6 +30,7 @@ Changes to the rules and guidelines in this document are proposed and discussed 
 
 ### Guidelines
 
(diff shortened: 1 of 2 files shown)


peppy:
This should only be added if it references the part where you should have explicit permission (or correctly licenced) media in the first place. The licensing for it should detail how attribution should be done.

peppy:
I think this reads better now 👍🏻 .

peppy (approved):

peppy:
Is this in a draft status for a reason?

cl8n:
wanted to make sure hobbes2 (OP of the rc thread) thought this wording was good too, just got their approval

PR #5041, opened 2021-03-08
Update CGTS 2021 to Regional Finals

fajar13k:
- Add each region's Semifinals match results
- Add each region's Finals mappool

### DIFF
--- a/wiki/Tournaments/GTS/CGTS_2021/en.md
+++ b/wiki/Tournaments/GTS/CGTS_2021/en.md
@@ -221,6 +221,36 @@ The Continental Global Taiko Showdown 2021 is run by various community members.
 
 ### Asia
 
+#### Finals
+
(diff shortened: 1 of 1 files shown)


TicClick (approved):
why are there so many commits when you can fit this into a single one?

fajar13k:
> 
> 
> why are there so many commits when you can fit this into a single one?

It's the best practice I learn and use from my workpalce, so the habit's going here too. ;-;

TicClick:
to me, the best practice is that two separate commits include two logically separated sets of changes -- I can hardly imagine, for example, the EU mappool being submitted but the NA one rolled back; if things are at the point of being 100% correct and finalized, _I personally_ see no point in splitting them.

on the other hand, it's harmless for osu! wiki anyway

fajar13k:
> to me, the best practice is that two separate commits include two logically separated sets of changes -- I can hardly imagine, for example, the EU mappool being submitted but the NA one rolled back; if things are at the point of being 100% correct and finalized, _I personally_ see no point in splitting them.
> 
> on the other hand, it's harmless for osu! wiki anyway

Will short it to 2 commits in the next update. (pools and results) ✨

PR #5053, opened 2021-03-12
[VI] Add Rules and Account Restrictions

TamamoNoMae13:


TicClick (approved):
thank you for the contribution. sadly, I don't think there's anyone to review it, so it'll be merged as-is.

PR #5067, opened 2021-03-14
[RU] Circle size

wlaxru:
Ready for review.


TicClick (approved):
well, I have already reviewed it one way or another, so

PR #4600, opened 2020-12-12
Cleanup BBCode

paper-toaster:
Cleanup on the BBCode article. Although its more of a rewrite than a cleanup, it still aims to get rid of the forum-style "bulleted list" descriptions, and also adds some extra information wherever possible. This pr does not change any images aside from the addition of the "Quote reply" button

PR #5074, opened 2021-03-15
Fix Spoilerbox not being treated as a tag in BBCode

TTTaevas:
Currently, https://osu.ppy.sh/wiki/en/BBCode#spoilerbox (link that can be found under both "Spoiler" and "Box") does not redirect to the Spoilerbox part of the article.
Furthermore, Spoilerbox doesn't show up in the table of contents on the left side of the screen.
I believe this commit/PR should fix this!

Below is a screenshot of the current table of contents, you can see that Spoilerbox is not featured in it.

Screenshot_2021-03-15 BBCode · knowledge base osu

### DIFF
--- a/wiki/BBCode/en.md
+++ b/wiki/BBCode/en.md
@@ -119,7 +119,7 @@ The tag is most commonly used to hide large walls of text and images that may bl
 
 Toolbar button: ![Box button](img/spoilerbox.png "Box")
 
-#### Spoilerbox
+### Spoilerbox
 
 ```
 [spoilerbox]text[/spoilerbox]
--- a/wiki/BBCode/es.md
+++ b/wiki/BBCode/es.md
@@ -1,3 +1,7 @@
+---
+outdated: true
+---
+
 # BBCode
 
 ![La caja de edición en los foros](img/editor.jpg)
--- a/wiki/BBCode/fr.md
+++ b/wiki/BBCode/fr.md
@@ -1,3 +1,7 @@
+---
+outdated: true
+---
+
 # BBCode
 
 ![L'éditeur de sujet du forum avec ces boutons](img/editor.jpg "la boite à outil d'édition du forum")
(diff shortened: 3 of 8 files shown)


Joehuu:
The heading level is kinda intended as a spoilerbox is:
> a special type of BBCode box

Not showing on the TOC or linking is also [intended](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#sections):
> Notice: On the website, heading levels 4 and 5 will not appear in the table of contents. They cannot be linked to directly either.

---

I think the spoilerbox link can be removed as it is already in the same section as box.

TTTaevas:
Ohhh, it didn't seem intentional as it is inconsistent with the versions of that article in other languages!
Then, would it be better if I close this PR, then create another one to get rid of the links, and make the article in other languages consistent with the English version?

Joehuu:
English should be the updated one. Gonna ping @MilkyIQ first as they changed the format and heading level in https://github.com/ppy/osu-wiki/pull/4600.

I would just continue this PR, instead of closing. Also would just mark the other translations as outdated, since the cleanup PR I linked didn't do so.

MilkyIQ:
> Ohhh, it didn't seem intentional as it is inconsistent with the versions of that article in other languages!

When I did a rewrite on BBCode I wanted to make spoilerboxes stylistically separate from the rest of the article because documentation was sparse and they didn't really behave like other BBCode tags. So yeah, this was intentional


> Then, would it be better if I close this PR, then create another one to get rid of the links, and make the article in other languages consistent with the English version?

As Joehuu said, it'd be best to just continue this pr as an update to the translations if you're at all able to. All translations should follow the original language the article was written in (which is almost always English)

TicClick:
I'm all for making it a level three heading, so that it is linkable

TTTaevas:
After thinking about it, I believe it should be a level three heading, as it isn't too different from the other BBCode tags.
So, I guess this PR is ready for reviewing!

TicClick (approved):

Issue #5090, opened 2021-03-17
Add a GitHub-native way of syncing back to ppy tree


sr229:
As we have provided a guide for [syncing back to the upstream](https://github.com/ppy/osu-wiki/blob/master/wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md#my-branch-is-out-of-date) for contributions, I figured that instead of using Upriver for a web-based upstream re-sync would be based on GitHub Actions via the `workflow_dispatch` event. This will be manually-initiated and would require the user to manually sync their repos by their own using `master` as the basis.

![image](https://user-images.githubusercontent.com/14976516/111454178-539e9500-874f-11eb-8aaa-b7099bc6e72d.png)

This will make the workflow stay at GitHub and reduce or diminish our reliance to third party sync providers like Upriver.

## Limitations

The initial proposal will focus on *destructively* force-pushing updated history from ppy upstream in the click of the button. In the near future I may add a less-obtrusive way to sync back to `master` if the user decides to use `master` for their working tree.

## References 

- [GitHub Documentation on `workflow_dispatch`](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/)

PR #5092, opened 2021-03-17
GH-5090: Add GitHub-native upriver-style upstream sync

sr229:
Per #5090, this adds an upriver-style syncing for forks. All they need to do is go to the Actions Tab and Run Workflow.

Resolves GH-5090.


sr229:
You can preview how the workflow works [here](https://github.com/sr229/osu-wiki/runs/2130891260?check_suite_focus=true).

Currently, it will look like this, you will have to choose your branch and decide if you want to overwrite your hard-earned changes.

image

image

sr229:
Documentation is complete so this should be ready to be merged now.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,90 @@
...
+           # Obviously this is pretty dangerous, this needs to be properly fixed later on.
+           echo "::warning::Non-overwriting syncing is experimental, use with caution!"
+
+           sleep 5
+
+           # Make a backup then rebase all changes
+           git branch $BACKUP_REPO_NAME
+           git reset --hard upstream/master
+           git rebase upstream/master
+           git merge --squash $BACKUP_REPO_NAME -m "Merge changes from Upstream to $(BRANCH_NAME)"
```
gagahpangeran, line 63:
```suggestion
           git merge --squash $BACKUP_BRANCH_NAME -m "Merge changes from Upstream to $(BRANCH_NAME)"
```

gagahpangeran: Also what happen if there's merge conflict? This workflow just failed?

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,90 @@
...
+         }
+
+         #Upstream repository URL, this is where everything ties down.
+         UPSTREAM_REPO="https://github.com/ppy/osu-wiki"
+         BACKUP_REPO_NAME="nellytmp-$(git branch --show-current)"
+         BRANCH_NAME="$(git branch --show-current)"
+
+         git remote add upstream $UPSTREAM_REPO
+         git fetch upstream
+
```
gagahpangeran, line 52: Maybe here, create and push the backup branch to user's remote repo. So user still has the backup in their repo in case they want to revert it, especially with the Force-sync command bellow that overwrite user's change.

For example

```bash
git branch $BACKUP_REPO_NAME
git push origin $BACKUP_REPO_NAME
```

So now the backup branch is already in user's remote repo.

gagahpangeran (changes requested):
This a bit destructive, but let's see....

.github/workflows/manual-sync.yml
sr229: We're not going to push the backup branch, it's a intermediate to allow us to sync updates from the original branch, then merge the changes back in again.

.github/workflows/manual-sync.yml
sr229: The workflow assumes there won't be any major conflicts here. If there is, the workflow will fail as designed.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,89 @@
...
+           git reset --hard upstream/master
+           git rebase upstream/master
+           git merge --squash $BACKUP_REPO_NAME -m "Merge changes from Upstream to $(BRANCH_NAME)"
+
+           git push origin +$BRANCH_NAME
+         else
+           echo "::warning::This will overwrite your changes. Unless you're really really sure you want to, please cancel this workflow task."
+           echo "::warning::Force-sync will occur in 5 seconds."
+           # Give them enough time to rethink their decision
+           sleep 5
```
bdach, line 70: I don't think this is a good idea. In my opinion this should just fail if the trees have diverged.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,90 @@
...
+
+# Workflow dispatch can take inputs like any other github.event
+# As part of #5090 - we should evaluate adding inputs and outputs to refine this.
+on:
+  workflow_dispatch:
+    inputs:
+      force:
+        description: "Overwrite any changes in the target repository? (true/false)"
+        required: true
+        default: "true"
```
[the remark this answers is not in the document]
sr229: Good catch, I'll make this `false` by default .

.github/workflows/manual-sync.yml
sr229: It's designed to fail if it does diverge, but this line of code is for the force push conditional. I added a sleep timer here to make sure if they really meant to overwrite their branch using the latest history from `master`.

.github/workflows/manual-sync.yml
gagahpangeran: The `BACKUP_BRANCH_NAME` suggested change not applied yet.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,82 @@
...
+
+         git remote add upstream $UPSTREAM_REPO
+         git fetch upstream
+
+         if [ ! $(boolean ${{ github.event.inputs.force }}) ]; then
+           # Obviously this is pretty dangerous, this needs to be properly fixed later on.
+           echo "::warning::Non-overwriting syncing is experimental, use with caution!"
+
+           # Make a backup then rebase all changes
+           git branch $BACKUP_REPO_NAME
```
gagahpangeran, line 58:
```suggestion
           git branch $BACKUP_BRANCH_NAME
```

gagahpangeran (changes requested):
Dont forget to change all `BACKUP_REPO_NAME` to `BACKUP_BRANCH_NAME`

sr229:
Requesting another RFC, suggestions and concerns were addressed now @bdach @gagahpangeran

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,81 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Configure Git to use your name
+        shell: bash
+        run: |
+          set -eo pipefail
+
+          #FIXME: We're using the noreply email of GitHub for now. Let's figure out a way to get the committer's email.
```
TicClick, line 25: I don't think this is meaningful. if you really want to go for it, try using `github.token` in conjunction with [`GET /users/{username}`](https://docs.github.com/en/rest/reference/users#get-a-user) and `json_pp`, but it's not worth the hassle.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,81 @@
...
+         #Upstream repository URL, this is where everything ties down.
+         UPSTREAM_REPO="https://github.com/ppy/osu-wiki"
+         BACKUP_BRANCH_NAME="backup-$(git branch --show-current)"
+         BRANCH_NAME="$(git branch --show-current)"
+
+         git remote add upstream $UPSTREAM_REPO
+         git fetch upstream
+
+         if [ ! $(boolean ${{ github.event.inputs.force }}) ]; then
+           # Obviously this is pretty dangerous, this needs to be properly fixed later on.
```
TicClick, line 54: why not fix it now?

wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
```diff
@@ -48,6 +48,25 @@ Fortunately, there are a few ways to resolve this problem depending on which bra
 
 `master` is the main branch of your fork repository. It should contain a clean copy of `osu-wiki` repository's `master` branch contents without any personal edits. There are two ways to update the `master` branch of your fork:
 
+#### Using GitHub
```
TicClick, line 51: while I appreciate you going into details (not kidding), it'd be easier if this was a go-to set of instructions as conscise as possible. besides, if we have this, we don't need upriver at all.
* * *
also, don't forget to outdate the existing translations!

TicClick (changes requested):
no objections to the workflow itself

.github/workflows/manual-sync.yml
sr229: Oh this should be fixed now (I hope), forgot to remove this.

wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
sr229: I managed to make it concise as possible and make it idiot-proof, however, you may try your hand on making it more understandable (it's pretty hard to make it understandable enough to most people (I applaud those people who make user guides for this purpose)).

.github/workflows/manual-sync.yml
sr229: Oh the current fix works just as well so I'll remove this comment later

Ephemeralis:
i'm amazed something as "simple" as this requires this much elbow grease to get moving

sr229:
> i'm amazed something as "simple" as this requires this much elbow grease to get moving

Unfortunately yes since we'd need to account for two user situations and more to follow after the initial version, but the idea here is we want our contributors to not rely so much on external tools and have their stuff be on GitHub (IIRC Upriver sometimes has a wonky behavior with non-destructive repo synchronization, that's something we can fix with a live Linux environment, whereas Upriver just uses GitHub API commands which is a hit and miss).

wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
```diff
@@ -48,6 +48,25 @@ Fortunately, there are a few ways to resolve this problem depending on which bra
...
+1. Go to the Actions tab of your fork.
+2. In "Workflows", look for "Sync from osu! upstream."
+3. Click "Run workflow". You will see the following options:
+
+![GitHub Actions Workflow - Run Workflow](img/github-actions-run-workflow.png "GitHub Actions Workflow - Run Workflow")
+
+ - **Use workflow from** - This will run the workflow on your selected branch. By default it is set to master.
+ - **Overwrite any changes in the target repository** - This accepts a boolean that is either true or false. Setting it to true will overwrite your changes in your target branch and force push the changes. Setting it to false will tell the workflow to back up your changes, sync your branch to the latest copy, then merge your changes back in.
+
+ **NOTE: Setting "Overwrite any changes in the target repository" value to false is experimental. Your may or may not lose your changes. If you encounter issues, please report it on the [osu-wiki issues board](https://github.com/ppy/osu-wiki/issues).**
```
bdach, line 64: Bit weird to be saying the default is experimental now, isn't it? If that is true, the entire thing should probably be considered experimental at this juncture if anything.

wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
```diff
@@ -48,6 +48,25 @@ Fortunately, there are a few ways to resolve this problem depending on which bra
...
 
+#### Using GitHub
+
+The wiki team has written a tool to allow you to synchronize your fork to the repository without ever leaving GitHub itself. The tool uses GitHub's continuous delivery feature called "GitHub Actions" and is self-contained within the repository.
+
+1. Go to the Actions tab of your fork.
+2. In "Workflows", look for "Sync from osu! upstream."
+3. Click "Run workflow". You will see the following options:
+
+![GitHub Actions Workflow - Run Workflow](img/github-actions-run-workflow.png "GitHub Actions Workflow - Run Workflow")
```
bdach, line 59: These images are a bit small, I'd say.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,78 @@
...
+         git fetch upstream
+
+         if [ ! $(boolean ${{ github.event.inputs.force }}) ]; then
+           # Make a backup then rebase all changes
+           git branch $BACKUP_BRANCH_NAME
+           git reset --hard upstream/master
+           git rebase upstream/master
+           git merge --squash $BACKUP_BRANCH_NAME -m "Merge changes from Upstream to $(BRANCH_NAME)"
+
+           git push origin +$BRANCH_NAME
```
bdach, line 60: I'm reading this again and I'm thoroughly confused.

* What's the rebase doing there? `git reset --hard` moves the branch pointer to the given ref, does it not? There *is* nothing to rebase onto at that point...?
* `git merge --squash` [is documented to not merge](https://git-scm.com/docs/git-merge#Documentation/git-merge.txt---squash), so I'm not sure what the push is supposed to achieve?

I've tried these commands locally and it doesn't look like this does what it claims to be. How does this work?

wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
sr229: Thanks for letting me know, this was during the initial development so I should reword this.

wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md
sr229: They're taken in a 1080p screen so it could use much more better screenshots in the near future.

.github/workflows/manual-sync.yml
sr229: I've tried out some of these commands locally so I was a bit sure how they'll work. Looks like I have to redo this again, thanks for the catch.

sr229: The idea behind these is 

- we make a backup branch to keep our changes
- we reset this tree based on `upstream/master`
- Call a rebase to verify if we do have the current HEAD set to `upstream/master`. If not, rebase will rebase all the changes in.
- merge the changes in, squashed (though I'll just make this a merge commit now in the new iteration).

bdach: I'm still lost. What is the goal of this tool? Is the goal to sync the master branch, or is it to help users update their "feature" branches with changes in master?

If it's the first one, this should be no more than:

```
git checkout master // to make sure, i guess
git branch backup
git pull --ff-only upstream master // refuse to update if user has new commits that never were in upstream/master
```

Alternatively, depending on what we want, that last one should be `git reset --hard upstream/master`, if the goal is to just bring the user back to a good state and not care that they had their own stuff merged into master. After all they have a backup branch.

If it's the second one, this should be no more than:

```
git merge upstream/master
```

sr229: > I'm still lost. What is the goal of this tool? Is the goal to sync the master branch, or is it to help users update their "feature" branches with changes in master?
> 
> If it's the first one, this should be no more than:
> 
> ```
> git checkout master // to make sure, i guess
> git branch backup
> git pull --ff-only upstream master // refuse to update if user has new commits that never were in upstream/master
> ```
> 
> Alternatively, depending on what we want, that last one should be `git reset --hard upstream/master`, if the goal is to just bring the user back to a good state and not care that they had their own stuff merged into master. After all they have a backup branch.
> 
> If it's the second one, this should be no more than:
> 
> ```
> git merge upstream/master
> ```

it's designed to sync their changes regardless of what branch they are in, so we want to avoid relying too much on `master` if such a case we have some web editor people using master as their working tree.

The reason I want the branch in a clean state for the non-destructive flow is to make sure the changes are integrated "safely", so I did a reset first then merge the changes in from the backup branch. Though i can see the `no-ff` part more feasible but I fear it may break easily and leave our tree in a very undesirable state.

sr229: @bdach I recommend you suggest a change for this part that works much more better, but as far as it goes I think this is the safest I can go :D

bdach: I'm not sure what more to suggest beyond what I already have. I'm still lost on why the current code is doing a fetch, then a hard reset and finally a pull. The pull is entirely redundant (pull = fetch + merge; the fetch has already happened manually, and the merge does nothing as the branch pointer was already moved to that same ref being merged, so I fail to see why it's "safer").

kj415j45 (dismissed):

.github/workflows/manual-sync.yml
sr229: IIRC during my testing inside GitHub Codespaces Git doesn't pull the changes from `upstream/master`'s HEAD, though I could be wrong, I'll probably refactor this to use your suggestions and see if it'll bite us in the long run

sr229: 1eb3fa4 should have this refactored now, thanks for your insight, please RFC again if you're not satisifed @bdach :D

sr229:
Latest revision should now include backups! This provide a way to recover someone's work if the tool doesn't work as expected.

A live demonstration can be found [here](https://github.com/sr229/osu-wiki/actions)

image
image

sr229:
Marking a draft for now as I found a bug that would destructively force push upstream's changes even when set to false, will lift this once I fixed it.

sr229:
Marking ready to review again. Looks like the workflow would break if we use shallow clones or shallow fetches, it needs the **ENTIRE** history to work properly so let's not do that.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,95 @@
...
+          git version
+          # clone without blobs; don't checkout to avoid fetching them anyway
+          git clone --filter=blob:none --no-checkout ${REPOSITORY} .
+          git config --local gc.auto 0
+
+          # set up sparse checkout
+          git sparse-checkout init
+          git sparse-checkout set '**/*.md' '**/.remark*' '**/*.json' '**/*.yaml'
+
+          git checkout --progress --force -B $BRANCH refs/remotes/origin/$BRANCH
```
bdach, line 39: What's with the unexplained sparse checkout that wasn't here before? I don't think it needs to be used here. It's only used in wiki CI to speed it up by avoiding fetching blobs that don't need to be fetched (because they're not inspected anyway).

This case is different, as pretty much any blob changed on a topic branch will conflict. What will happen in that case if this checkout is sparse?

No reason to not go with a default checkout, I'd say.

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,95 @@
...
+         git fetch upstream
+
+         # User has no confidence in our git skills, so let's backup when needed!
+         if [ $(boolean ${{ github.event.inputs.backup }}) != false ]; then
+          git branch $BACKUP_BRANCH_NAME
+          git push -u origin +$BACKUP_BRANCH_NAME
+         fi
+
+         if [ $(boolean ${{ github.event.inputs.force }}) != true ]; then
+           git merge upstream/master --commit -m "Merge upstream/master changes to $BRANCH_NAME.";
```
bdach, line 79: Does the merge commit message need to be specified here? I think the default would do just fine.

bdach (commented):
Looks closer to sanity, but still some weirdness.

(By the way, I'm not sure if you're aware, but RFC expands to "request for comments". It doesn't really make sense to ask me for a RFC in the context of a pull. If anything, the *pull* is the RFC.)

.github/workflows/manual-sync.yml
```diff
@@ -0,0 +1,95 @@
...
+         #Upstream repository URL, this is where everything ties down.
+         UPSTREAM_REPO="https://github.com/ppy/osu-wiki"
+         BACKUP_BRANCH_NAME="backup-$(git branch --show-current)"
+         BRANCH_NAME="$(git branch --show-current)"
+
+         git remote add upstream $UPSTREAM_REPO
+         git fetch upstream
+
+         # User has no confidence in our git skills, so let's backup when needed!
+         if [ $(boolean ${{ github.event.inputs.backup }}) != false ]; then
```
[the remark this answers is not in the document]
sr229: Ah, I blame POSIX for this case, as we use POSIX-compatible syntax for this part of the script. The previous way to check for bools had a problem acknowledging that when it's set to `true`, we do not run this particular part of the script.

.github/workflows/manual-sync.yml
sr229: Right, this was a haphazard copy after 3 hours of bashing my head because `actions/checkout` was breaking the pipeline continuously. Good catch.

.github/workflows/manual-sync.yml
sr229: Should be removed now in the latest rev.

sr229:
[According to the latest test](https://github.com/sr229/osu-wiki/runs/2151335808?check_suite_focus=true), this should be "production-ready" now. Thanks for helping out with cleaning up the script @bdach.

sr229:
All fixes and changes are applied on this PR now. @TicClick, I suggest doing the re-phrasing of the documentation as a separate PR, but you can do a follow-up commit before merging.

TicClick:
looks ok with some minor text quirks I'll fix later today. sorry for not taking a closer look at the script before.

TicClick:
@sr229 if what I've done works for you, please squash your workflow-related changes as much as possible and we're done here

sr229:
No changes needed on my end so the documentation was the only change that was needed @TicClick

Ephemeralis:
passed this by peppy and he said "looks safe from my end, as far as security is concerned"

bdach:
The clone is still done quite awkwardly (clones without blobs but then checks them out) but /shrug. I'll fix it myself post-merge if push comes to shove.

sr229:
> The clone is still done quite awkwardly (clones without blobs but then checks them out anyway) but /shrug. I'll fix it myself post-merge if push comes to shove.

Should be resolved now, as far as cloning is concerned. I can test further later on.

PR #5147, opened 2021-03-30
[FR] Add "Downbeat"

DoryanR:
Proposal for the French translation of the wiki page "Downbeat".

**Notify me if the PR is merge without review, that I self-review before.**

### DIFF
--- a/wiki/Downbeat/en.md
+++ b/wiki/Downbeat/en.md
@@ -7,6 +7,6 @@ tags:
 
 # Downbeat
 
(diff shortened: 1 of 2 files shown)


Satsukel (approved):

TicClick:
@DoryanR why are there so many merge commits? there aren't even any changes in the master branch that'd affect the files you are editing

DoryanR:
I used to think it was useful to do it, but in the end it wasn't. I realize it much less now, but from time to time to overcome conflicts that I could not see otherwise.

TicClick (approved):
removed empty merge commits

PR #5150, opened 2021-03-31
update additional column mania RC for a new key

mousesalt:


Feerum:
Oh yeah this should have been changed as well now

cl8n (approved):

PR #5152, opened 2021-03-31
Add Project Loved: March 2021 news post

cl8n:


Ephemeralis (approved):
there's some little foibles i might ordinarily suggest changes for but i'm happy to pin them off as stylistic choices this time

the general formatting + quality of all the blurbs is so much better than usual, props to all the writers for that

Ephemeralis (approved):

PR #5171, opened 2021-04-04
Remove BaAR_Vendel from Users with unique titles

Nivalyx:
in accordance with [BaAR_Vendel's recent resignation](http://osu.ppy.sh/users/8679346) (as the "Elite Nominator" title is revoked upon someone's resignation from BN/NAT)

image


Joehuu (approved):
Next time, please combine these kinds of changes into one commit by using an external editor.

PR #5220, opened 2021-04-07
[EN] Game mode / osu!mania | remove invalid link news page reference.

mousesalt:
[EN] [Game mode / osu!mania](https://osu.ppy.sh/wiki/en/Game_mode%2Fosu!mania) remove the link due to non-refer. add a new to woc2006's profile.

users with non-authorized can not access the page.

### DIFF
--- a/wiki/Game_mode/osu!mania/en.md
+++ b/wiki/Game_mode/osu!mania/en.md
@@ -347,10 +347,6 @@ Enable the `Co-op mode` button to force *Co-Op* game modifier when played, with 
 
 ![osu!mania logo in the Special Modes](img/Mania_logo.png "osu!mania logo in Special Mode")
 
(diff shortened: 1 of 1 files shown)


peppy:
Probably best to actually leave the announcement reference there and refer to it, rather than removing it. Seems like something that could be useful to someone visiting this page and looking for more information to continue to.

mousesalt:
that is a good idea, however, the link is referred to the forbidden page which the users may not access.
another way probably rewrites to a new link with same announcement article, so it can be updated and see by everyone.
image

peppy:
Oh, that's a mistake. I'll move it somewhere it can be accessed again, give me a day or two.

peppy:
I've fixed the issue. Sorry for the delay, phpbb is a pain to work with.

TicClick:
closing this, as the whole announcements forum is publicly accessible now

PR #5223, opened 2021-04-08
Remove song restrictions from qualification BN rules

VINXIS:
The discussion forum of this is here: https://osu.ppy.sh/community/forums/topics/1282982 
^ Hopefully all questions/concerns are also resolved here regarding this (outside of the "vouch"/"agree" spam)

The likes of players, mappers, BNs, and NATs have provided their input and support (to some degree). If there are any more concerns/questions here or any more discussions to be had I'll try to respond when I can or hopefully someone else who is free otherwise this is good to go I think


cl8n (approved):

Ephemeralis:
don't fastmerge RC changes in the future plz, let them wait at least a day or two so those who aren't following the forum can chip in during PR phase

PR #5245, opened 2021-04-11
Add "Featured artist showcase beatmaps" wiki article

Nivalyx:
*((this took unexpectedly longer that i thought it would be whew ;~;))*

ok so i always had this idea of documenting all of the fa showcase beatmaps that have been made by the members of the Mappers' Guild in one page in a [Staff Log](http://osu.ppy.sh/wiki/Staff_Log)-kind of style, and... after putting together the many pieces of information that are scattered all around here is the said article at last, yeah 🧙

PR #5246, opened 2021-04-12
Rewrite all instances of "osu!standard" as "osu!" in the Article Styling Criteria

Nivalyx:
this was discussed earlier in one of the comments in my other open pr : http://github.com/ppy/osu-wiki/pull/5245#pullrequestreview-633145435

in short the Article Styling Criteria as it stands is still promoting the use of `osu!standard` despite the fact that `osu!` has been the much preferred term for a long while, as confirmed by @peppy himself :

image

image

this pr makes changes to the Article Styling Criteria so that `osu!` can formally be considered as the proper way of referring the game mode in question going forward.

### DIFF
--- a/wiki/Article_styling_criteria/Writing/en.md
+++ b/wiki/Article_styling_criteria/Writing/en.md
@@ -182,7 +182,7 @@ Game modes must be written as follows:
 - `osu!catch`
 - `osu!mania`
 
-Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note however that citing the osu! game mode as `osu!standard` is *not* recommended — unless necesary, `osu!` should be used instead.
+Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note however that citing the osu! game mode as `osu!standard` is *not* recommended.
 
 ## osu!
 
--- a/wiki/Article_styling_criteria/Writing/en.md
+++ b/wiki/Article_styling_criteria/Writing/en.md
@@ -83,7 +83,7 @@ When writing out game modifiers for tournament articles, they must instead use u
 Gameplay elements must not be capitalised, unless they act as a title for a link that points to the article. The following is an example:
 
 ```markdown
-In osu!, beatmaps are composed of three different gameplay elements: hit circles, sliders, and spinners.
+In the osu! game mode, beatmaps are composed of three different gameplay elements: hit circles, sliders, and spinners.
 ```
 
 ### Language names
--- a/wiki/Article_styling_criteria/Writing/en.md
+++ b/wiki/Article_styling_criteria/Writing/en.md
@@ -83,7 +83,7 @@ When writing out game modifiers for tournament articles, they must instead use u
 Gameplay elements must not be capitalised, unless they act as a title for a link that points to the article. The following is an example:
 
 ```markdown
-In osu!standard, beatmaps are composed of three different gameplay elements: hit circles, sliders, and spinners.
+In osu!, beatmaps are composed of three different gameplay elements: hit circles, sliders, and spinners.
 ```
 
 ### Language names
@@ -177,16 +177,12 @@ October 25, 2016 (11:45 UTC)
 
 Game modes must be written as follows:
 
-- `osu!standard` (unofficial, but used to prevent ambiguity)
+- `osu!`
 - `osu!taiko`
 - `osu!catch`
 - `osu!mania`
 
-*Notice: `osu!standard` is used to maintain consistency when referring to the game mode. Folder names, however, must use `osu!` even if it is referring to the game mode.*
-
-Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name.
-
-Articles such as the [Ranking Criteria](/wiki/Ranking_Criteria) may use `osu!` instead of `osu!standard`.
+Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note however that citing the osu! game mode as `osu!standard` is *not* recommended — unless necesary, `osu!` should be used instead.
 
 ## osu!
 
--- a/wiki/Article_styling_criteria/Writing/en.md
+++ b/wiki/Article_styling_criteria/Writing/en.md
@@ -182,7 +182,7 @@ Game modes must be written as follows:
 - `osu!catch`
 - `osu!mania`
 
-Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note however that citing the osu! game mode as `osu!standard` is *not* recommended.
+Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name.
 
 ## osu!
 
--- a/wiki/Article_styling_criteria/Writing/es.md
+++ b/wiki/Article_styling_criteria/Writing/es.md
@@ -1,4 +1,5 @@
 ---
+outdated: true
 no_native_review: true
 ---
 
--- a/wiki/Article_styling_criteria/Writing/ru.md
+++ b/wiki/Article_styling_criteria/Writing/ru.md
@@ -1,3 +1,7 @@
+---
+outdated: true
+---
+
 # Содержание статей
 
 *См. также: [Оформление статей](/wiki/Article_style_criteria/Formatting)*


wiki/Article_styling_criteria/Writing/en.md
```diff
@@ -177,16 +177,12 @@ October 25, 2016 (11:45 UTC)
...
 - `osu!taiko`
 - `osu!catch`
 - `osu!mania`
 
-*Notice: `osu!standard` is used to maintain consistency when referring to the game mode. Folder names, however, must use `osu!` even if it is referring to the game mode.*
-
-Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name.
-
-Articles such as the [Ranking Criteria](/wiki/Ranking_Criteria) may use `osu!` instead of `osu!standard`.
+Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note however that citing the osu! game mode as `osu!standard` is *not* recommended — unless necesary, `osu!` should be used instead.
```
TicClick, line 185: please define "unless necessary" if you use this sentence. personally I think we may be able to rewrite the wiki in a way that it would always be clear which of the two entities (the game, or the main mode) is mentioned, so no such sentence is needed

Nivalyx: ah, okok, removed

DoryanR: > please define "unless necessary" if you use this sentence. personally I think we may be able to rewrite the wiki in a way that it would always be clear which of the two entities (the game, or the main mode) is mentioned, so no such sentence is needed

On this point I was thinking of making a disambiguation for the term `osu!`, or if existing include this new point.

Because I'm thinking of renaming `osu!standard` > `osu!`, so I'm also thinking of a solution to not confuse the two.

TicClick: nah, what I meant is that you didn't even need to mention "osu!standard" as an alternative spelling if it will be removed (and I guess it will)

wiki/Article_styling_criteria/Writing/en.md
```diff
@@ -177,16 +177,12 @@ October 25, 2016 (11:45 UTC)
...
 - `osu!taiko`
 - `osu!catch`
 - `osu!mania`
 
-*Notice: `osu!standard` is used to maintain consistency when referring to the game mode. Folder names, however, must use `osu!` even if it is referring to the game mode.*
-
-Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name.
-
-Articles such as the [Ranking Criteria](/wiki/Ranking_Criteria) may use `osu!` instead of `osu!standard`.
+Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note however that citing the osu! game mode as `osu!standard` is *not* recommended.
```
MilkyIQ, line 185: "reccomended" doesn't really seem to be the best word to describe "not standard". Here's a few options:
```suggestion
Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note, however, that `osu!standard` should *not* be used to refer to the osu! game mode unless speaking about it's colloquial nature.
```
```suggestion
Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note, however, that citing the osu! game mode as `osu!standard` is *not* prohibited unless it is speaking of its colloquial nature.
```
```suggestion
Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note, however, that `osu!standard` should *not* be used to refer to the osu! game mode.
```
```suggestion
Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do note, however, that citing the osu! game mode as `osu!standard` is *not* prohibited.
```

peppy: Why are there four suggestions here?

wiki/Article_styling_criteria/Writing/en.md
peppy: You can call it "osu! (ruleset)" or "osu! mode" or anything. But don't call it osu!standard. It's not how I want it referenced and it sounds bad.

wiki/Article_styling_criteria/Writing/en.md
MilkyIQ: I wasn't sure if the exception to the rule should be specified, so it's just different ways of saying the same thing. Only one will be applied tho, so ¯\\\_(ツ)_/¯

Walavouchey: All of these seem either too verbose or flip the statement around with "is not recommended" -> "is not prohibited", which to me doesn't make much sense. The previous sentence should make that clear enough.

How about
```suggestion
Referring to the old game mode names (i.e. `Catch the Beat`, `Taiko`, and `Mania`) may be done if discussing said game mode's previous name. Do not otherwise cite the osu! game mode as `osu!standard`.
```
(yes, i'm adding another suggestion)

MilkyIQ: The previous sentence does not refer to osu!standard at all, so why "otherwise"?. Doesn't really matter, though, cus this works fine, too. As long as the rule surrounding the usage of "osu!standard" is understandable and made clear, then it should be fine; just needed to get rid of "recommended".

Walavouchey: Hmm, right, and although a direct "do not cite" may be correct, it doesn't flow well in the paragraph. I'd go with your third suggestion in that case.

Nivalyx: o thanks for all the suggestions guys but i guess i'll just go with TicClick's solution in the other comment chain to delete this problematic sentence altogether :x

wiki/Article_styling_criteria/Writing/en.md
Nivalyx: @TicClick ah i see, i guess i was thinking too much when i wrote that sentence down then ;~;

removed the mention of the alternative spelling, should be good now!

MilkyIQ (approved):
No issues as far as I can see. LGTM

wiki/Article_styling_criteria/Writing/en.md
```diff
@@ -83,7 +83,7 @@ When writing out game modifiers for tournament articles, they must instead use u
 Gameplay elements must not be capitalised, unless they act as a title for a link that points to the article. The following is an example:
 
 ```markdown
-In osu!standard, beatmaps are composed of three different gameplay elements: hit circles, sliders, and spinners.
+In osu!, beatmaps are composed of three different gameplay elements: hit circles, sliders, and spinners.
```
DoryanR, line 86:
```suggestion
In osu! mode, beatmaps are composed of three different gameplay elements: hit circles, sliders, and spinners.
```

wiki/Article_styling_criteria/Writing/en.md
```diff
@@ -177,17 +177,13 @@ October 25, 2016 (11:45 UTC)
 
 Game modes must be written as follows:
 
-- `osu!standard` (unofficial, but used to prevent ambiguity)
+- `osu!`
```
DoryanR, line 180:
```suggestion
- `osu!` (add `mode` when will be needed)
```

DoryanR (changes requested):
I'm doing the necessary fix, because I think the PR will be merge before the one I'll put online in a few days.

wiki/Article_styling_criteria/Writing/en.md
Walavouchey: Now needs definite article because English is weird
"In the osu! mode"

wiki/Article_styling_criteria/Writing/en.md
Walavouchey: ``add `mode` when needed``

wiki/Article_styling_criteria/Writing/en.md
DoryanR: Yes, it does look better.

Nivalyx: decided to go with `the osu! game mode` as most pages (the Ranking Criteria, the World Cup articles, etc) are writing it this way instead

image

DoryanR: That's what I did here: #5255 

However, in some cases this is not applicable, so I left `osu! mode`.

wiki/Article_styling_criteria/Writing/en.md
Nivalyx: hmm i genuinely don't think that an instruction \*this\* specific is needed tho. i think needless to say all writers by default are already expected to simply apply common sense on a case-by-case basis here whether to write `osu!`, `osu! mode`, `osu! game mode`, or any variations of it in a sentence :(

DoryanR: I have already found variates like you describe in the articles, each one has its own way of using the term `osu!` to describe the main game mode, to avoid confusion with the game name.

In the end I would say don't worry too much about it, and formulate it the way you think it should be so that it can be understood.

Without using the term `standard`, because indeed I found it quite misplaced. Indeed we have more chances to confuse the two, but if we talk about spinners or hit objects, we can't go wrong I think.

Honestly, it's really case by case how to name the game mode, we can't define a universal term, because it won't fit well in some cases.

In our case, I think we can leave it as it is, and then I think (with practice), we will know how to use it, maybe it's just a matter of adaptation.

TicClick (approved):

PR #5251, opened 2021-04-13
Update Universal_Offset Stub

THATotaku:
- Capitalizations fixes (that are hopefully correct)
- Added small info on how positive and negative values effect things, as requested in the TODO
- added link to wiki guide on how to set universal offset, as requested in the TODO

### DIFF
--- a/wiki/Universal_offset/en.md
+++ b/wiki/Universal_offset/en.md
@@ -10,4 +10,4 @@ stub: true
 
 Each player's optimal universal offset value differs due to variations in their systems. A Universal Offset, other than `0`, should only be used when **all beatmaps** have a noticeable delay, otherwise a [local offset](/wiki/Glossary/Local_song_offset) should be used instead. An incorrectly determined Universal Offset will cause significant timing problems on all beatmaps.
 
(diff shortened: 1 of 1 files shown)


wiki/Universal_offset/en.md
```diff
@@ -2,12 +2,12 @@
...
+
+Each player's optimal universal offset value differs due to variations in their systems. A Universal Offset, other than `0`, should only be used when **all beatmaps** have a noticeable delay, otherwise a [local offset](/wiki/Glossary/Local_song_offset) should be used instead. An incorrectly determined Universal Offset will cause significant timing problems on all beatmaps.
 
-Each player's optimal universal offset value differs due to variations in their systems. An incorrectly determined universal offset will cause significant timing problems on all [beatmaps](/wiki/Beatmap).
+Follow [this guide](/wiki/Guides/How_to_Use_the_Offset_Wizard) for step-by-step instructions on how to set your Universal Offset to the correct value using the [offset wizard](/wiki/Options/Offset_Wizard).
```
MilkyIQ, line 13: This would be better formatted as a hatnote imo. Either way, "this guide" shouldn't be used as the title text and "your" should be removed

THATotaku: Yeah, my bad. I'm not sure if i like it being a hatnote but if anyone else thinks it should be, ill change it :) thanks again

peppy (approved):

PR #5263, opened 2021-04-16
Allow deletion for accounts outside of EU

osuShiro:
Removes mention of "If you are in the EU" in the Privacy article

### DIFF
--- a/wiki/Legal/Privacy/ru.md
+++ b/wiki/Legal/Privacy/ru.md
@@ -1,7 +1,6 @@
 ---
 legal: true
 outdated: true
-outdated_since: e95d1bf509dcb730b46f93ef70215e84f57bc37f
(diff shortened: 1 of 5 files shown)


wiki/Legal/Privacy/ru.md
```diff
@@ -1,6 +1,7 @@
 ---
 legal: true
 outdated: true
+outdated_since: e95d1bf509dcb730b46f93ef70215e84f57bc37f
```
TicClick, line 4: uhh no, if it's already outdated, then you don't know how many commits behind it is, and the diff may be (and will be) misleading. same goes for `zh.md`

TicClick (changes requested):

peppy (approved):

PR #5306, opened 2021-04-23
Make Boolean Matching Case-Insensitive For Manually-Triggered Actions

sr229:
This will make boolean matching case-insensitive in the manual actions script. This also removes the disclaimer in the OWCG guide as this script is mostly considered production-ready by most cases.

### DIFF
--- a/.github/workflows/manual-sync.yml
+++ b/.github/workflows/manual-sync.yml
@@ -51,8 +51,8 @@ jobs:
          # https://stackoverflow.com/questions/14700579/bash-converting-string-to-boolean-variable
          function boolean () {
             case $1 in
-              true) echo true ;;
-              false) echo false ;;
+              [Tt][Rr][Uu][Ee]) echo true ;;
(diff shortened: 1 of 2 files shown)


.github/workflows/manual-sync.yml
```diff
@@ -51,8 +51,8 @@ jobs:
          # https://stackoverflow.com/questions/14700579/bash-converting-string-to-boolean-variable
          function boolean () {
             case $1 in
```
TicClick, line 53: how about this instead?
```sh
case `echo $1 | tr '[:upper:]' '[:lower:]'` in
```

sr229: I was leaning on K&R but I'll see if that fits better

sr229: I'll keep this as it is for now since I'm using Bash-native patterns here which should reduce the point of failure.

TicClick (approved):
well

PR #5309, opened 2021-04-23
Update "Users with unique titles" listing

zeusminus:
- Add TWC2021 winners under the **osu!taiko Champion** listing
- Add [Mao](https://osu.ppy.sh/users/2204515) under the **Elite Mapper** listing


TPGPL (approved):
looks fine i guess

PR #5319, opened 2021-04-25
Add A Labour of Love contest winners to "Users with unique titles" wiki article

Nivalyx:


Nivalyx:
ok so

i was talking w/ Mao (https://osu.ppy.sh/users/2204515) earlier about this and we found out that apparently there are quite a lot of stuffs in this article that are missing/needs updating aside from the recently crowned A Labour of Love winners :x

closing this pr, will open a new one in a bit

PR #5324, opened 2021-04-26
Prevent beatmap stealing in Ranking Criteria

Naxesss:
This adds measures to prevent stolen beatmaps from being ranked. Reason we're adding this now is that we have been informed not to remove stolen beatmaps anymore. Because of this they are much more likely to reach ranked, especially as Beatmap Nominators may not be aware that a beatmap is included without permission, much less taken from elsewhere in the first place.

We now enforce explicit permission if a beatmap has been taken from another set, ensuring that ranked maps are not at risk of being DMCA claimed by its original creator for not having said permission given.


NoffyNoffeh:
`[Guest difficulties](/wiki/Beatmap/Guest_difficulty) (with involvement from the guest) are ok; this is more aimed at preventing potentially stolen beatmaps from being ranked. `

I think this part can be removed, it doesn't seem to contribute much to understanding the rule and could serve as a point of confusion by elongation it. It's already pretty clear in the main rule that having clear involvement = it's fine

peppy:
I'd leave out the mention of "stolen". It's a stupid term and shouldn't be in the ranking criteria.

wiki/Ranking_Criteria/en.md
```diff
@@ -58,6 +58,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
 - **A [beatmapset host](/wiki/Beatmap/Beatmap_host) and [guest difficulty](/wiki/Beatmap/Guest_difficulty) creators can make changes to their respective difficulties as they wish.** If there is a disagreement between both, the beatmapset host must delete the guest contribution upon request. If a guest difficulty creator cannot be contacted for a month, they will be assumed to agree with any changes.
+- **Beatmapsets cannot contain significant portions of difficulties belonging to other users without their explicit involvement or permission.** If you find the original beatmap(s), and there is no permission given, link those in the [discussion of the offending beatmapset](/wiki/Beatmap_discussion) and wait for a resolution.
```
peppy, line 61:
```suggestion
- **Beatmapsets cannot contain significant portions of difficulties belonging to other users without their explicit involvement or permission.** [Guest difficulties](/wiki/Beatmap/Guest_difficulty) (with involvement from the guest) are fine.
```

No need to mention the process when it's the same as any other ranking criteria issue.

peppy (changes requested):

peppy:
> I think this part can be removed, it doesn't seem to contribute much to understanding the rule and could serve as a point of confusion by elongation it. It's already pretty clear in the main rule that having clear involvement = it's fine

Tend to agree that whole thing can go, yeah.

Naxesss:
Should we perhaps add what counts as permission or is that obvious enough already?

peppy (approved):

peppy:
I'd say it's fine as-is. It's not going to come up much; if it does and there's confusion we can iterate on the wording.

Ephemeralis (approved):
Looks good to me as well.

PR #5350, opened 2021-05-01
Changed mistake in section "Grade"

af2111:
The text stated that the grade F exists, which isn't true, instead it is the the grade D.

### DIFF
--- a/wiki/Glossary/en.md
+++ b/wiki/Glossary/en.md
@@ -303,8 +303,7 @@ Geki(激), or *Elite Beat!*, is a scoring term used when you complete a [combose
 
 A grade, usually confused with "rank", is a letter-grade that represents the player's performance after playing a beatmap. It accompanies the total score at the result screen and appears in all lists where scores appear.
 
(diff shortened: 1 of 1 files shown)


peppy (approved):

wiki/Glossary/en.md
```diff
@@ -303,8 +303,7 @@ Geki(激), or *Elite Beat!*, is a scoring term used when you complete a [combose
 
 A grade, usually confused with "rank", is a letter-grade that represents the player's performance after playing a beatmap. It accompanies the total score at the result screen and appears in all lists where scores appear.
 
-From lowest to highest, the possible grades are F, C, B, A, S, Silver S, SS, and Silver SS.
-
+From lowest to highest, the possible grades are D, C, B, A, S, Silver S, SS, and Silver SS.
```
sr229, line 306: In my opinion, F exists as a rank, is this no longer the case?

af2111: No, the lowest rank is D, at least in osu!standard.

peppy: I'm not sure one person's opinion changes fact.

PR #5366, opened 2021-05-04
CWC2021 Round of 16 updates

WalterToro:


peppy:
As an aside, we should probably have rules against players changing their names during an active tournament, or at least not update them in the tournament documentation on each change.

peppy (approved):

PR #5376, opened 2021-05-05
Fix missing digit from Mentorship newspost filename

Ephemeralis:
This probably needs a remark rule to avoid similar things happening in the future.


peppy:
Keep in mind this will break any links to the article. I guess it's old enough that that's not a huge deal, but something to keep in mind if it happens to be pinned/linked anywhere.

peppy (approved):

PR #5272, opened 2021-04-18
[DE] Add Main Page

HeyMika:
German translation for the main wiki page. I used the english markdown for the main page so it is just the translation for that.

PR #5377, opened 2021-05-05
[CS] Main Page

Okarisu:


DoryanR:
I don't speak the language, so I can't help you with that, but I think you can use #5272 to improve the translation, because I notice that you translate some terms that are not usually translated, for example `Globální moderátoři` for `Global Moderation Team`

For the error in the checks, this is due to problems that have normally been resolved, I think that [updating the branch](https://github.com/ppy/osu-wiki/blob/master/wiki/osu!_wiki_Contribution_Guide/Common_Issues/en.md#my-branch-is-out-of-date) should solve the problem.

I hope this will help you.

Okarisu:
Thank you @DoryanR for advice, I'll go through it and improve the translation (and myself, too).

PR #5391, opened 2021-05-08
[JA] Add Disambiguation/Bancho

ViaSnake:
Add `Disambiguation/Bancho` in Japanese.


peppy:
@ViaSnake thanks for your contributions. for the future, please avoid using force push (we don't mind multiple commits). it makes it harder to review when commits go missing.

peppy (approved):

ViaSnake:
Ok, in the future I will try not to use Force push and try to keep all commits. Thanks for the review and merge.

PR #5402, opened 2021-05-10
Add Project Loved: May 2021 news post

cl8n:
- [x] post/link forum topics


news/2021-05-10-project-loved-may-2021.md
```diff
@@ -0,0 +1,235 @@
...
+165 BPM, 2:34 | 3.67★ – 6.15★\
+*description written by [Penguinosity](https://osu.ppy.sh/users/10235296)*
+
+Mipha-'s Battle Against a True Hero is here to show who the strongest players truly are! This mapset includes two very different difficulties, each with their own respective rates. Starting with the "Challenge" difficulty, this one offers some clean rice patterns that test players ability to stay patient with lots of short breaks between the relatively comfortable patterns. The final section on this difficulty, introduces some heavy hand usage and still keeps these same breaks from before, so your accuracy will be put to the test. Now as for the "Howling" difficulty, the charting takes a turn and becomes full LN. At OD8, your release accuracy abilities are definitely strained. The first half or so you will be coordinating a barrage of short LNs alongside ones that extend further. As for the final leg of the map, hands are also introduced like the challenge diff but in full LN form this time. Your reading is truly put to the test to try and properly hit this last tricky section of the map. Give Mipha- your support by voting!
```
Joehuu, line 205: players ability -> players' ability
```suggestion
Mipha-'s Battle Against a True Hero is here to show who the strongest players truly are! This mapset includes two very different difficulties, each with their own respective rates. Starting with the "Challenge" difficulty, this one offers some clean rice patterns that test players' ability to stay patient with lots of short breaks between the relatively comfortable patterns. The final section on this difficulty, introduces some heavy hand usage and still keeps these same breaks from before, so your accuracy will be put to the test. Now as for the "Howling" difficulty, the charting takes a turn and becomes full LN. At OD8, your release accuracy abilities are definitely strained. The first half or so you will be coordinating a barrage of short LNs alongside ones that extend further. As for the final leg of the map, hands are also introduced like the challenge diff but in full LN form this time. Your reading is truly put to the test to try and properly hit this last tricky section of the map. Give Mipha- your support by voting!
```

Ephemeralis:
I've applied a big ol' chonky reword/reformat commit to this directly, you might want to inspect the diff for changes and make sure they're all okay.

osu!taiko descriptions are really good this time around mostly, I barely had to do anything with most of them. I'd suggest having some sort of writing guidelines to apply to these posts before PRing them - the monthly dance of splitting up map descriptions with line breaks is starting to get a bit grating after a while.

Ephemeralis:
otherwise, ping me on discord when threads are up presuming it is still Friday UTC (so within the next 6 hours), or we'll have to wait until a news slot opens up on Sunday to avoid clipping with FA announcements

cl8n:
some of this review is less "review" than using ur own words for what they were trying to say--- I'll keep it here cuz it all reads fine but in future I think we should preserve the captains' original wording as much as possible. having "description by [name]" followed by something where a lot of the words aren't theirs seems weird.

Ephemeralis:
Preserving their original wording is always the goal, sometimes it is difficult when 70% of a paragraph is tangle in the wording rather than actual content. I'll just opt to cut in the future instead of spending the effort to incorporate what I think they're trying to say.

Ephemeralis (approved):

PR #5405, opened 2021-05-10
Fix CI checks by upgrading remark, remark-cli, remark-frontmatter, and linting plugins to the next major version

TicClick:
- http://npmjs.com/advisories/1700
- example error: https://github.com/ppy/osu-wiki/runs/2548764241

`package-lock.json` generated with:
```
nvm install --lts  # after that, edit package.json
nvm exec --lts npm install
```


TicClick:
the change is incomplete, because the new version scoffs at the yaml header most of `.md` files have:

```bash
blah% head -n 5 wiki/Main_Page/en.md
---
layout: main_page
---

<!-- Do not add any empty lines inside this div. -->
blah% npx remark wiki/Main_Page/en.md --no-stdout
wiki/Main_Page/en.md
  2:1-3:4  warning  Headings should use atx               heading-style           remark-lint
  2:1-3:4  warning  Missing blank line before block node  no-missing-blank-lines  remark-lint

⚠ 2 warnings
```

package.json
```diff
@@ -4,9 +4,9 @@
     "description": "home of the osu! wiki",
     "repository": "github:ppy/osu-wiki",
     "dependencies": {
-        "remark": "12.0.0",
-        "remark-cli": "8.0.0",
-        "remark-frontmatter": "2.0.0",
+        "remark": "^13.0.0",
+        "remark-cli": "^9.0.0",
+        "remark-frontmatter": "^3.0.0",
```
bdach, line 9: I personally wouldn't go as far as to add carets here. Consider the case of a contributor that does `npm install` to run remark locally, but doesn't really understand much about what they're doing, and then accidentally bumps and commits the lockfile to a PR, because one of these packages has a new minor/bugfix release.

Back when I hard-pinned these the explicit intention was to use specific known-good versions and combine that with npm audits to check security advisories about them, to preserve as much reproducibility in the setup as is possible.

TicClick: true. besides, remarkjs/remark-lint@7791db9450da83e402455c902b6b202d178133ba looks like I need to update a couple more packages

TicClick:
bumped a few more packages from release notes https://github.com/remarkjs/remark-lint/releases/tag/8.0.0

bdach (approved):
not that it matters from me here, but here's an ACK

package.json
cl8n: maybe I'm misunderstanding something but this is exactly what `package-lock` is for-- `package` specifies what's compatible with your project, `package-lock` pins it to something so that every contributor's `npm install` result is the same. caret is the default and standard for almost everything on npm bc no breaking changes are made within major versions, I don't see why it wouldn't go here too

TicClick:
didn't check for that, only basic stuff. if you've got a spare minute, it might be faster to test these yourself

package.json
bdach: I thought that running `npm install` *will rewrite* `package-lock.json` if it determines that it can (and from limited experience, I did witness it doing that). Googling around gives mixed results - some claim that the above behaviour is fixed - but I still feel hard pinning is safest. Especially that the circumstances here are special: remark is a tool here, and not a library dependency.

TicClick:
> do the plugins megaapplepi and I wrote still work here?

yes

```
blah% nvm exec --lts npx remark wiki/Main_Page/en.md --no-stdout
Running node latest LTS -> v14.16.1 (npm v6.14.12)
wiki/Main_Page/en.md
  16:154-16:188  warning  Beatmap URLs should use the /beatmapsets/{beatmapset-id}#{mode}/{beatmap-id} format  links       remark-lint-osu
  16:191-16:216  warning  Wiki links must not include the file name                                            wiki-links  remark-lint-osu

⚠ 2 warnings
```

as for the package pinning, I noticed that `npm i` only changes `package-lock.json` if you experiment with the contents of `package.json`, or install an extra package with `npm i whatever@1.2.3`. plain `npm i` or `npm ci` doesn't change anything (at least for a list of dependencies that I've made for this PR), so it's your call whether to pin these or not

cl8n:
continued in #5411, which contains all the changes here + brings the new remark config in line with how it behaved in previous version

PR #5411, opened 2021-05-11
Upgrade remark and clean up options and linting rules

cl8n:
instead of #5405


TicClick:
@Ephemeralis could you please take a look?

peppy (approved):

PR #5419, opened 2021-05-11
Add keymode rules to MBC May news post

David-Ottertun:
Rule change so that 4K and 7K entries will only be allowed. This is so that antimeta keymodes being 5K, 6K, 8K, 9K, and 10K, won't be biased against as there is a lack of judges capable of handling every keymode.


pishifat (approved):
please name your commits in the future!

PR #5451, opened 2021-05-13
Move "Ouendan" to "Disambiguation/Ouendan"

ViaSnake:
Move `/wiki/Ouendan` to `/wiki/Disambiguation/Ouendan`.
I don't know what the page was originally intended for, but I think it should be in `/wiki/Disambiguation`, not `/wiki` root, at least for now.

- EN and ID have been updated to match the other pages.
- TH added outdated because there are no other pages in Disambiguation.


TicClick (approved):
that's right

PR #5457, opened 2021-05-13
Move "Star" to "Disambiguation/Star"

ViaSnake:
Move `/wiki/Star` to `/wiki/Disambiguation/Star`.
I think the Star should be in `/wiki/Disambiguation`, not `/wiki` root.


TicClick (approved):
and you are correct

PR #5474, opened 2021-05-15
Remove anchor tags from news headers and fix relevant links

cl8n:


cl8n:
all markdown links in /news were verified by program but I just realised there's some manual `a` tags not covered in the check-- fixing those now

TPGPL (approved):
skimmed through the changes, and looks fine i guess

pishifat (approved):
is there any context for why this is all being changed at once? i feel like i missed something

cl8n:
was part of https://github.com/ppy/osu-web/pull/7544, turns out that not being able to link to headers in news was a bug and not intentional 🤔

pishifat:
ah ty

PR #5534, opened 2021-05-21
Split Help Centre into separate articles

### LINKED ISSUE #1145: Add separate pages for help centre how-to's
The main page will only have Q&A while the separate pages show how to do it.

TPGPL:
Resolves #1145 

Splits Help Centre into seperate articles for each section


MilkyIQ:
I'm not sure if creating a new article for each individual question is the right direction here...

iirc, the discussion centered around separating it by sections (e.g., Account issues, Troubleshooting, Website, etc. etc.), that way there could be an easy way to view and link to all the questions directly instead of using ctrl-f

TicClick:
my take on it was going to be a separate article for every second level heading, that is,

```
Help centre
\___ Account
\___ Client
\___ ...
```

to make cross-linking easier, in every article I could've added a short anchor for every small section:

```
# Account

(possibly some short and small preface goes here)

## Supporter

### I've paid for a supporter tag, but still have not received it! {#no-supporter-tag}

**Depending on what payment service you used (PayPal or Xsolla), there may be a slight delay in the processing of your payment before it reaches us.**
```

wiki/Help_Centre/Client/en.md
```diff
@@ -0,0 +1,491 @@
...
+This is a completely harmless (albeit ugly!) effect that happens because the game's refresh rate is not synced up with your monitor's refresh rate.
+
+You can also try playing in Windowed or Borderless mode, which will use your operating system's native VSync instead.
+
+Turning on VSync or playing in Windowed or Borderless mode will introduce a *slight* amount of input lag, unfortunately. For most people, this will not cause significant issues.
+
+Later NVidia graphics cards have a global option in the NVidia Control Panel to set the global refresh rate to a setting called 'Fast', which will also fix this issue.
+
+## Online features
+
```
MilkyIQ, line 274: Assuming you haven't made any changes to the actual text under the headers, but while we're here, might as well add thematic break between these "section changes"

wiki/Help_Centre/Account/en.md
```diff
@@ -0,0 +1,212 @@
+# Account
```
MilkyIQ, line 1: iirc ephemeral wanted the articles to have lots of tags so that they could be easily searchable

TicClick: +1

TicClick:
make sure to apply #5601

TPGPL:
> make sure to apply #5601

done

peppy:
Let's get this in because the longer it stays open the more patches that are going to have to be applied to keep it up-to-date.

PR #5537, opened 2021-05-22
Update Random Game Modifier

THATotaku:
- Update old and/or incorrect information (score multiplier & note about randomisation)
- Remove 1 mention of osu!mania that was previously twice in one sentence (making it less wordy).

PR #5540, opened 2021-05-22
[FR] Update "Game modifier/Random"

DoryanR:
Proposal to update the French translation of the wiki page `Random`.

In accordance with #5537


TicClick (approved):
as it's super small and I've seen the original change, I don't think it needs an additional review step

PR #5575, opened 2021-05-25
Add catch Hit Objects stubs

Darius-Wattimena:
TicClick asked me if I could help out so here it is. Added stubs for catch hit objects, issue #3152

The issue also mentions 'Edge Dash' although this is a term that is used for a far dash and is not a hit object by itself

A PR with catch glossary and implementing this in the catch RC will follow once this is merged

PR #5586, opened 2021-05-26
Add glossary stubs for osu!catch

Darius-Wattimena:
All of them will be replacing the glossaries found in the catch RC, most of them still need some extra attention with some good example screenshots.

Depends on #5575


Darius-Wattimena:
Will create a new PR as my git client decided to not like me rebasing the whole PR

PR #5588, opened 2021-05-26
Correct the Half Time mod description for osu!catch

Darius-Wattimena:
The description of the halftime mod for osu!catch isn't correct as it only reduces the speed of the beatmap and the catcher.

### DIFF
--- a/wiki/Game_modifier/Half_Time/en.md
+++ b/wiki/Game_modifier/Half_Time/en.md
@@ -38,7 +38,7 @@ As a result, using the Half Time mod will lead to an increase in maximum possibl
 
 ### osu!catch
 
-In [osu!catch](/wiki/Game_mode/osu!catch), the BPM and the speed of the catcher is lowered by the same factor as in other modes. All the [fruits](/wiki/Hit_object/Fruit), [drops](/wiki/Hit_object/Juice_stream#drop), [droplets](/wiki/Hit_object/Juice_stream#droplet) and [bananas](/wiki/Hit_object/Banana) stay untouched.
+In [osu!catch](/wiki/Game_mode/osu!catch), the BPM and the speed of the catcher are lowered by the same factor as in other modes. All the [fruits](/wiki/Hit_object/Fruit), [drops](/wiki/Hit_object/Juice_stream#drop), [droplets](/wiki/Hit_object/Juice_stream#droplet) and [bananas](/wiki/Hit_object/Banana) stay untouched.
 
 ## Trivia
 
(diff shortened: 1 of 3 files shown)


RedcXca (changes requested):
For the trivia section I think the wording of "decreases the speed of beatmaps by 75%" can be misleading. I would rewrite it as "...misnomer, as the HT mod does not actually change the speed of a beatmap to 50% of the original but rather to 75%"

Darius-Wattimena:
> For the trivia section I think the wording of "decreases the speed of beatmaps by 75%" can be misleading. I would rewrite it as "...misnomer, as the HT mod does not actually change the speed of a beatmap to 50% of the original but rather to 75%"

The current wording in the trivia section should be fine, also I rather want to keep this PR small to fix just the description of catch

wiki/Game_modifier/Half_Time/en.md
```diff
@@ -38,9 +38,7 @@ As a result, using the Half Time mod will lead to an increase in maximum possibl
 
 ### osu!catch
 
-In [osu!catch](/wiki/Game_mode/osu!catch), the BPM is lowered by the same factor as in other modes. Though it *also* decreases the player character's speed so normal fruits without any mods *may* turn into hyperdash fruits.
-
-In addition to this, the leniency for hyperdashes is increased, which makes it much more difficult to stop underneath the next fruit when performing a hyperdash.
+In [osu!catch](/wiki/Game_mode/osu!catch), the BPM is lowered by the same factor as in other modes. Additionally, the speed of the catcher is also decreased. All the [fruits](/wiki/Hit_object/Fruit), [drops](/wiki/Hit_object/Juice_stream#drop), [droplets](/wiki/Hit_object/Juice_stream#droplet) and [bananas](/wiki/Hit_object/Banana) stay untouched.
```
MilkyIQ, line 41:
```suggestion
In [osu!catch](/wiki/Game_mode/osu!catch), the BPM and the speed of the catcher is lowered by the same factor as in other modes. All the [fruits](/wiki/Hit_object/Fruit), [drops](/wiki/Hit_object/Juice_stream#drop), [droplets](/wiki/Hit_object/Juice_stream#droplet) and [bananas](/wiki/Hit_object/Banana) stay untouched.
```
sounds a bit better imo

TPGPL:
```suggestion
In [osu!catch](/wiki/Game_mode/osu!catch), the BPM and the speed of the catcher are lowered by the same factor as in other modes. All the [fruits](/wiki/Hit_object/Fruit), [drops](/wiki/Hit_object/Juice_stream#drop), [droplets](/wiki/Hit_object/Juice_stream#droplet) and [bananas](/wiki/Hit_object/Banana) stay untouched.
```

TicClick (approved):

Issue #4804, opened 2021-01-19
Clean up references to Glossary


TicClick:
context: #3824

1. the glossary is still in place
2. a lot of redirects point to it
3. a lot of articles still point to the glossary

what needs to be done:
- [ ] redirects must point to the new articles;
- [x] direct links to the glossary must point to the new articles;
- [ ] the glossary needs to be an [index article](https://github.com/ppy/osu-wiki/blob/master/wiki/Article_styling_criteria/Formatting/en.md#index-articles)

_Originally posted by @TicClick in https://github.com/ppy/osu-wiki/issues/3824#issuecomment-762525385_

DoryanR:
Is anyone on the job? If not, I can take care of it

TicClick:
go for it. start with small, scoped changes, so that we can ensure your direction is correct

DoryanR:
I have already started, I will post a PR to see if I can continue then

PR #5589, opened 2021-05-26
Fix glossary references in `redirect.yaml`

DoryanR:
part of #4804


wiki/redirect.yaml
```diff
@@ -228,155 +226,147 @@
...
+"test_mode":                   "Beatmap_Editor/Test_mode"
+"time_signature":              "Beatmapping/Time_signature"
+"timing_section":              "Beatmapping/Timing_section"
+"touchpad":                    "Gameplay/Input_device/Graphics_tablet"
+"touchscreen":                 "Gameplay/Input_device/Graphics_tablet"
+"touhosu!":                    "Glossary/Touhosu!"
 "wip":                         "Beatmap/Category#work-in-progress-and-pending"
 "work_in_progress":            "Beatmap/Category#work-in-progress-and-pending"
-"z/x":                         "Glossary#z/x"
+"z/x":                         "Play_style"
```
TicClick, line 369: I guess that's a good alternative

TicClick (changes requested):

TPGPL:
The redirects with `_glossary`/`glossary_` in name should be removed i guess

TicClick (dismissed):
the removed redirects, such as AFK, are not osu!-specific, or are too specific (like `#looping`, which links to a 30s-themed cut from the ranking criteria).

on top of that, my wild guess is that we don't need any `*_glossary` (or `glossary_*`) redirects either, as they duplicate links to existing articles and were unlikely made for meaningful usage.

need someone else to give it a read, but here's an approval from me

wiki/redirect.yaml
```diff
@@ -213,7 +213,7 @@
 "xk":                          "Game_modifier/xk"
 
 "acc":                         "Accuracy"
-"accuracy_glossary":           "Accuracy"
+"accuracy":                    "Accuracy"
```
TicClick, line 216: this is unnecessary, osu! wiki picks up any case

TicClick (changes requested):

wiki/redirect.yaml
DoryanR: I had forgotten to remove it, as for `nc`

TPGPL (dismissed):
lgtm

TPGPL (approved):

PR #5594, opened 2021-05-26
Change oddly structured sentences in Skinning/Interface

Wrongwayscout:
"select ___ for this to see." -> "select ___ to see."


Joehuu (approved):

peppy:
I still don't understand what these sentences are saying. Should it read something like "This element is only visible in [mode]"??

peppy (changes requested):
still reads wrong.

PR #5604, opened 2021-05-27
Fix editing mistake on skinning contest badge

RockRoller01:
for some reason one row of pixel was missing on the right and bottom..


Walavouchey:
it honestly looked intentional

PR #5609, opened 2021-05-28
Update osu!supporter (images, amount of hearts in a badge)

geshii:
o/
I updated the supporter page and also changed the images since these were not up to date

Should I add images to show the different badges ?


Walavouchey (changes requested):

geshii:
I think we are ready to merge, if there's nothing else to change

wiki/osu!supporter/en.md
```diff
@@ -1,6 +1,3 @@
----
-outdated: true
```
TicClick, line 2: you can't remove the outdated marker if you don't address the issues mentioned in TODO, and I don't feel that what you've changed so far is enough for that

TicClick (changes requested):

wiki/osu!supporter/en.md
geshii: should I add the marker back or else what should I add to the page?

TicClick: I've just skimmed through the article again and didn't find many things to improve. I think that for now you can replace `outdated: true` with `needs_cleanup: true`. 

@cl8n blame says the TODO is yours — any details on cleanup ideas other than replacing the images?

geshii: I'll change that as soon as I'm at home

geshii:
I think we're good now

TicClick (approved):
alright

wiki/osu!supporter/en.md
cl8n: reading over the article some sentences sound odd due to mixing formal and casual style, and the author sounds uncertain about details, which is really strange for a wiki article. I imagine all I meant by my TODO comment was that it needed a quick pass by a native speaker. if these are too nitpicky of reasons for `needs_cleanup` and/or my comment, feel free to remove

TicClick: they aren't, thanks for clarification

PR #5610, opened 2021-05-28
Add 2nd and 3rd place badges to the Skinning Contest results news post

RockRoller01:


zeusminus:
The badges for 2nd and 3rd places seem to be mixed as a result of misnaming the files.

PR #5701, opened 2021-06-10
CWC2021 Conclusion update

WalterToro:
Closes #5694 as is not relevant.


TPGPL (commented):
L11 needs past tense too

peppy:
The podium is the wrong image format and size, as usual. Would appreciate if you can fix your process so I don't have to fix the image each tournament.

peppy (changes requested):
As stated

WalterToro:
> The podium is the wrong image format and size, as usual. Would appreciate if you can fix your process so I don't have to fix the image each tournament.

There is no process, the podium is the image im given on ops, but no issues, i already changed it.

peppy (approved):

PR #5713, opened 2021-06-15
Fix wmfchris' flag in the TWC 2012 article

TheDefaultGuy:
[wmfchris](https://osu.ppy.sh/users/7401) is incorrectly listed as Korean when in reality he's from Hong Kong.


TicClick (approved):
confirmed by wmfchris

PR #5767, opened 2021-07-01
[FR] Add `History_of_osu!/2012`

Chateauvision:


DoryanR:
La PR n'est pas terminé pour moi :  

- La page interface à besoin de plus de changement que ça, de plus, quelqu'un du nom de Dela__ sur Discord est déjà dessus (sauf si c'est toi) 
- La page pour l'history of osu!, elle n'est pas complètement traduite, et également @Airwam est déjà dessus.

DoryanR:
@Chateauvision In my opinion, in view of what is stated in my last message, it is better that you take care of another article than the ones you have started.

Also, I would say to you to take the English version already present, to copy it on the French file, and to translate directly on it, it will be simpler and you will not have to seek what must be updated or not.

Don't be discouraged, every contributor is important, and makes the translation on the French side progress.

TPGPL:
Reopen the PR when your translation is in near-complete or complete state

PR #5769, opened 2021-07-01
Minor staff and schedule readjustment in `ATT 2021` article

fajar13k:
Some staff addition and addition of Grand Finals that was left in the tournament schedule


TPGPL (commented):
add US flag to the references

TicClick (approved):

PR #5774, opened 2021-07-01
Copy Normal maximum SV guideline to Easy in osu! RC

NoffyNoffeh:
As discussed in https://osu.ppy.sh/community/forums/topics/1285933 

Currently there has been a bit of a hole in RC here, so the PR fixes it as well as the misleading "1.3x" instead of "1.3" label.
Using the same limit for Easy and Lowest Diff Normal so that effectively nothing is changed, just clarified.


cl8n (approved):
I feel it may be nonobvious that 1.3 needs to be adjusted at different BPMs than 180, but at least this fixes the multiplier confusion.

PR #5778, opened 2021-07-03
 Update the information about star priority in Feature Requests forum

### LINKED ISSUE #4946: Star priority wiki appears to be outdated
Star priority Wiki article seems to be outdated: https://github.com/ppy/osu-wiki/blob/master/wiki/Modding/Star_priority/en.md

It says: `Users that have supporter status active when voting will shoot a pink star (Kudosu) that increases Star Priority by 2 each, whilst a user without supporter will shoot a blue star (Star) that is worth only 1 each. Users may be able to shoot multiple stars depending on how many votes they have left.`, but other parts of a website say that only supporters are able to star features, and it doesn't show the colour of a star at all:
image

TPGPL:
Resolves #4946


Walavouchey (commented):
There should probably be a mention of how this system is outdated/unused. Also, is all the other information in this article accurate nowadays?

TicClick (approved):
works for me, but /wiki/Disambiguation would be a better place, along with the split for each meaning of the two -- guess that could be done in a separate pull

PR #5791, opened 2021-07-06
Beatmap Nominators: Updated outdated information

godoffours:
Moved new full BNs from the Probationary Beatmap Nominators section to the Full Beatmap Nominators section, updated names for BNs that have received a name change, and removed previous BNs that no longer hold a position on the team.


zeusminus:
Thanks for your contribution! However, we update these listings along with translations in addition to the Staff Log on a semi-regular basis (usually happens biweekly or monthly). I'm already going to push an update covering these changes on top of the recent GMT & Project Loved additions/removals in the forthcoming days. I think it would be better to update them all at once than separately.

TPGPL:
Superseded by #5797

PR #5831, opened 2021-07-12
RC - Disallow featured artist tags in maps of unlicensed tracks

NoffyNoffeh:
Based on discussion at https://osu.ppy.sh/community/forums/topics/1293405

Reworded the main rule a bit so the two can read well together


cl8n:
this is one "if and only if" rule split into two, I'm not sure if that helps with reading comprehension or not

cl8n (changes requested):
link "featured artist library" to the listing on website or relevant wiki page if it exists

NoffyNoffeh:
Yeah it is and I do have a way to word it as one rule, but when asking around several people agreed the fully separate is clearer, especially for non native speakers. Added links.

peppy:
The tag should be added based on the artist name if anything?

A "featured artist" is an "artist" not a track.

AncuL001:
Maybe do the discussion about the proposal itself in the forum post instead? So people don't have to keep track of two separate discussions

peppy:
As per my extended discussion in the thread, I'd propose just removing this requirement, or leaving it as a suggestion to include "featured artist" if the artist is such.

PR #5891, opened 2021-07-26
Update `Users with unique titles`

hikikuman:
Changed username (EN/ID):
- Stevie Blunder > im a fancy lad
- [224]Hyperw7 > CTB Rushia1
- andreymc > Amerom
- Remu > Punnies

Deleted from `By name` (EN/ID):
- Motion
- R-18/Petit
- Rocma
- T s u m i

Added to `By name` (EN/ID)
- Abstract-
- Berea
- CTB Rushia1
- DreStar
- Rells

Added to `Contributors` and `By name` (ID):
- Azer
- ChillierPear
- LeoFLT


Nivalyx:
oh fwiw whenever this article were to be updated the update usually comes for all languages tho (not just any specific language(s)) to avoid any possible discrepancy in the future :

---

image

TPGPL (changes requested):
needs to be done in other translations as well

hikikuman:
Updated

TPGPL (approved):

PR #5902, opened 2021-07-27
RC - Add guideline about 5:4 aspect ratio

Dopaminos:
I have 16:9 screen, but using native 5:4 since i need switch to another screen sometimes. Recently i saw offscreen at one ranked map (https://osu.ppy.sh/beatmaps/2356116 (tomatsu expert, 02:58:626 (3) https://osu.ppy.sh/ss/16879237/c7fd) ) what created me a reading difficulty and i somehow missed there. Since 5:4 (1280x1024) are being used at official osu! client for 16:9 (https://media.discordapp.net/attachments/312267283563151361/869541148031610920/unknown.png?width=502&height=399), i'd suggest to add 5:4 guideline.


peppy:
Please don't merge this without direct approval for me. Does not sounds correct.

cl8n:
if anything I feel like clients taller than 4:3 (playfield ratio) should prevent the playfield from scaling larger than it would appear on 4:3 of the same width. it doesn't seem like RC's problem that exceptionally small aspect ratio has problems

also see [How to propose Ranking Criteria changes](https://osu.ppy.sh/community/forums/topics/720532).

peppy:
I think this can safely be closed. I am willing to add a warning in the osu!(stable) settings advising against using 5:4 ratios if that's seen as required. Given that this is the first time it's been mentioned in years I'm not so sure. 5:4 monitors are very uncommon these days.

PR #5941, opened 2021-08-07
Add ScoreV2 and Mirror mod to `.osr` description

abstrakt-osu:
Saw them listed on https://github.com/ppy/osu-api/wiki#mods and https://github.com/ppy/osu/blob/7654df94f6f37b8382be7dfcb4f674e03bd35427/osu.Game/Beatmaps/Legacy/LegacyMods.cs


peppy (dismissed):

abstrakt-osu:
Added two spaces because lint check failed

TPGPL (approved):

PR #5943, opened 2021-08-08
[DE] Add `Graveyard`

Bloeckchengrafik:
It's another Addition to the german Wiki!

### DIFF
--- a/wiki/Graveyard/de.md
+++ b/wiki/Graveyard/de.md
@@ -1,0 +1,14 @@
+---
+stub: true
+tags:
+  - dead
+---
+
+# Graveyard
+
+Der **Graveyard** (auf Deutsch: Friedhof) ist eine Unterform von [Beatmaps](/wiki/Beatmap), welche nicht mehr von ihren Erstellern (Mappern) beachtet werden. 
+Solche Maps zählen nicht zum maximalen Uploadlimit des Benutzers. Der [Beatmap-Host](/wiki/Beatmap/Beatmap_host) kann diese "wiederbeleben", 
+indem er diese bearbeitet. Hierfür wird ein freier Map-Slot benötigt. Danach wird die Map in 
+["Work in Progress (in Bearbeitung)"](/wiki/Beatmap/Category#work-in-progress-and-pending) verschoben.
+
+Wenn eine [Beatmap](/wiki/Beatmap) für **28 Tage** nicht von ihrem Ersteller bearbeitet wird, landet diese automatisch "bei den Toten" auf dem **Graveyard**.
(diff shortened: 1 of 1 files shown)


wiki/Graveyard/de.md
```diff
@@ -0,0 +1,15 @@
+---
+stub: true
+no_native_review: true
+tags:
+  - dead
+---
+
+# Graveyard
+
+Der **Graveyard** (auf Deutsch: Friedhof) ist eine Unterform von [Beatmaps](/wiki/Beatmap), welche nicht mehr von ihren Erstellern (Mappern) beachtet werden.
```
Phippe, line 10: I wouldn't translate _subforum_ as _Unterform_ (maybe this was an inadvertent mistake anyway?). The literal translation sounds alright to me (_Unterforum_), but I think _Unterkategorie_ would be better.

Also, it's probably clear what _Ersteller_ means, so the _Mappern_ in brackets feels a bit redundant.

wiki/Graveyard/de.md
```diff
@@ -0,0 +1,15 @@
...
+tags:
+  - dead
+---
+
+# Graveyard
+
+Der **Graveyard** (auf Deutsch: Friedhof) ist eine Unterform von [Beatmaps](/wiki/Beatmap), welche nicht mehr von ihren Erstellern (Mappern) beachtet werden. 
+Solche Maps zählen nicht zum maximalen Uploadlimit des Benutzers. Der [Beatmap-Host](/wiki/Beatmap/Beatmap_host) kann diese "wiederbeleben", 
+indem er diese bearbeitet. Hierfür wird ein freier Map-Slot benötigt. Danach wird die Map in 
+["Work in Progress (in Bearbeitung)"](/wiki/Beatmap/Category#work-in-progress-and-pending) verschoben.
```
Phippe, line 13:
```suggestion
indem er diese bearbeitet. Hierfür wird ein freier Map-Slot benötigt. Danach wird die Map in den
["Work in Progress (in Bearbeitung)"](/wiki/Beatmap/Category#work-in-progress-and-pending)-Status zurückversetzt.
```

Phippe (commented):
There doesn't exist consensus about whether to use the translated or the English term for graveyard on [crowdin](https://crowdin.com/translate/osu-web/all/en-de?filter=advanced&value=12&sort_method=1&sort_ascending=0#q=graveyard). I'm leaning towards translating it, but I guess both are alright (just make sure to update the "wrong" term on crowdin after merging this for consistency).

If you decide to translate it, you could still add something like "eng.: Graveyard" in brackets for more clarity.

Everything here is more of a suggestion than being absolutely necessary :D

TicClick:
don't want to disappoint you, but this stub is being removed in #5984. please put your effort in translating /wiki/Beatmap/Category instead

PR #5962, opened 2021-08-15
[ID] Add `Help_Centre/Installation_and_registration`

Nivalyx:
Indonesian translation of the `Help_Centre/Installation_and_registration` wiki page (http://osu.ppy.sh/wiki/en/Help_Centre/Installation_and_registration).

Also included a quick fix for an error in the English article that I found, where the download button on the main page was incorrectly referred to as `Download` instead of `Download now`.

image

---

edit : why did GitHub automatically sent a review request to both peppy and ephemeral out of the blue O_o

### DIFF
--- a/wiki/Help_Centre/Installation_and_registration/en.md
+++ b/wiki/Help_Centre/Installation_and_registration/en.md
@@ -19,7 +19,7 @@ This section is dedicated to issues with installing osu! and creating your own a
 
 ### How do I download osu!?
 
-**Simply click `Download` found at the center of the [official osu! website homepage](https://osu.ppy.sh/home). You may also find the link through the `home` dropdown section found at the navigation bar on any page.**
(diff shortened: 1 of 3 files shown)


mousesalt (approved):
overall looks good on ID

TicClick (dismissed):

TPGPL (dismissed):

TicClick:
> edit : why did GitHub automatically sent a review request to both peppy and ephemeral out of the blue O_o

your pull request affects files and/or directories that are protected and require a review from their owners: https://github.com/ppy/osu-wiki/blob/master/.github/CODEOWNERS. bundling them with a regular translation usually hides them and it's not always obvious that your PR needs an extra approval

TicClick (approved):
if anything, I've reverted the `Download now` change for _now_

Nivalyx:
ah, i see 👍

PR #5861, opened 2021-07-19
Clean up formatting in tournament articles

TPGPL:
part 3 (and possibly the last big one?) of tournaments cleanup

PR #5966, opened 2021-08-15
Corrections to tables in `OWC2016` tournament article

Bazzadwarf:
I spotted some errors in the OWC2016 article while grabbing some data from them, just some flags and a score; otherwise, it looks to be fine.


TicClick (commented):
shouldn't cause conflicts with #5861, so here you go. love these kinds of mistakes

TicClick (approved):

PR #5967, opened 2021-08-15
[PT-BR] Add `Skinning/osu!mania`

Icaro1447:


Sebola3461 (dismissed):
good job

TicClick:
before this is merged, you've got a couple of headings, like `Lighting`, or `Long Notes`, untranslated — is that intended?

Icaro1447:
> before this is merged, you've got a couple of headings, like `Lighting`, or `Long Notes`, untranslated — is that intended?

gotta fix it, but long note is a basic term and in portuguese would be really weird and players would not understand it properly (unless its intended to be all translated, but i checked french and there is english terms there aswell). same for other stuff on the translation.

TicClick:
> unless its intended to be all translated

it isn't, many terms in other languages are borrowed as-is

TicClick (approved):

PR #5976, opened 2021-08-18
[CS] Add `Registration`

Ondra-D:
Translated the Registration page to czech


Matej-Hajek (approved):
I think that translation is good

TPGPL (approved):

Issue #5440, opened 2021-05-12
Make rules for usage of recently-added markdown features


cl8n:
those features being:
- specifying custom ID (wiki only)
  ```markdown
  ## Header {#with-custom-id}
  
  > Block quote
  {#with-custom-id-2}
  ```
- headers in news posts now generate ID like they do in wiki
- headers at level >=4 now also generate an ID like <4 (news and wiki)
- infoboxes (wiki only)
  - please only header >=4 within this, anything larger looks dumb
  ```markdown
  {{{ Infobox
  <!-- normal markdown here -->
  }}}
  ```
  
<details><summary>more involved infobox example, it shows all of the styling that changes (note table header especially)</summary>

```markdown
# Infobox test

{{{ Infobox

#### Heading

![](https://a.ppy.sh/6616586?1434426025.png "Someone's avatar")

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.

##### Smaller heading

| | |
| :-- | :-- |
| Property | It's a cat |
| Color | Blue and also here is more text sed ut perspiciatis unde omnis iste natus error sit voluptatem accusantium |
| More prop | ... |

}}}

Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Dolor magna eget est lorem ipsum dolor sit amet consectetur. Morbi tempus iaculis urna id volutpat lacus laoreet. Consectetur lorem donec massa sapien faucibus et molestie ac feugiat. Etiam sit amet nisl purus in. Sit amet nulla facilisi morbi tempus iaculis urna. Ultrices neque ornare aenean euismod elementum nisi quis eleifend quam. Odio tempor orci dapibus ultrices in iaculis nunc sed. Sed risus pretium quam vulputate dignissim suspendisse in est ante. Lectus proin nibh nisl condimentum id venenatis a condimentum vitae. Lacus sed turpis tincidunt id aliquet. Potenti nullam ac tortor vitae. Tortor condimentum lacinia quis vel. Morbi non arcu risus quis varius quam quisque id. Orci eu lobortis elementum nibh tellus molestie nunc non. Fames ac turpis egestas sed tempus urna et pharetra. Lobortis scelerisque fermentum dui faucibus in ornare. Sed pulvinar proin gravida hendrerit. Orci ac auctor augue mauris augue neque gravida.

[...]

PR #5987, opened 2021-08-19
Establish usage of custom section identifiers in article styling criteria

TicClick:
part of #5440. the first article to adapt this should be DAMN HELP CENTRE, finally

### DIFF
--- a/wiki/Article_styling_criteria/Formatting/en.md
+++ b/wiki/Article_styling_criteria/Formatting/en.md
@@ -314,6 +314,30 @@ Section headings must not skip a heading level (i.e. do not go from a level 2 he
 
 *Notice: On the website, heading levels 4 and 5 will not appear in the table of contents. They cannot be linked to directly either.*
 
+### Custom identifiers
+
+It is possible to redefine a section's identifier, which is used for linking to it directly:
+
+```markdown
+## My cooldown has passed. How do I appeal? {#appeal}
+
+## Common restriction reasons and cooldowns {#common-reasons}
+```
+
+Custom identifiers should be used in case the automatically generated ones are too long or contain tricky punctuation marks. For uniform experience, headings in a translated article must use the same identifiers as these in the original article:
+
+```markdown
+## Chat rules <!-- original heading -->
+## Правила чата {#chat-rules} <!-- translated heading -->
+```
+
+This feature can also be used for tagging a specific part of the article which doesn't have a heading. Use it sparingly:
+
+```markdown
+> That's it! You're well on your way to becoming an osu! rhythm champion!
+{#tutorial-quote}
+```
+
 ## Lists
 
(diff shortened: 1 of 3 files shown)


TicClick:
of course I wrote it as I saw fit -- do comment on whether it's sensible or not. for example, requiring translations to use the same anchors as in the originals will make section links less prone to breakage on refactoring/translation.

cl8n:
> For uniform experience, headings in a translated article must use the same identifiers as these in the original article

this doesn't seem all that useful without special handling for outdated articles/sections. if EN updates the content of `## Some section` and a translation doesn't, the `/wiki/Article#some-section` link without locale specified becomes ambiguous, so you'd have to send `/wiki/en/Article#some-section` instead, defeating any use the cross-locale ID would've had. this issue isn't *caused by* cross-locale ID (e.g. just `/wiki/Article` has same issue), but it feels more critical to me here because opening a section link will scroll right past the "outdated" warning

so I think usage of this should be limited to
- making long/obnoxious IDs prettier
- adding IDs to elements which otherwise wouldn't generate them (when needed for a link)
- cross-locale sync *only if* the EN article can be gauranteed to not make breaking changes to the meaning of its sections

TicClick:
> cross-locale sync only if the EN article can be guaranteed to not make breaking changes to the meaning of its sections

unlikely possible, not to mention that headings can also be changed.

all in all, I wanted not only to get rid of long identifiers, but solve that issue where you need to cross-check all references with anchors in a translation to make sure it doesn't link to a page in general (just because someone forgot that the article is or is not translated, or didn't specify the exact ahcnor, and then you have fucking images which leave a hyphen, so it's not `#osu`, but `#-osu`).

whew.

well, I guess it's the wrong tool for this purpose

wiki/Article_styling_criteria/Formatting/en.md
```diff
@@ -314,6 +314,23 @@ Section headings must not skip a heading level (i.e. do not go from a level 2 he
...
+
+It is possible to redefine a section's identifier, which is used for linking to it directly:
+
+```markdown
+## My cooldown has passed. How do I appeal? {#appeal}
+
+## Common restriction reasons and cooldowns {#common-reasons}
+```
+
+Custom identifiers should be used in case the automatically generated ones are too long, contain tricky punctuation marks or images. This feature can also be used for tagging a specific part of the article which doesn't have a heading. Use it sparingly:
```
TPGPL, line 327:
```suggestion
Custom identifiers should be used in case the automatically generated ones are too long, contain tricky punctuation marks or images. 

This feature can also be used for tagging a specific part of the article which doesn't have a heading. Use it sparingly:
```

TicClick: yes, you are right. in this case I'd rather attach the first sentence to the one before the example

TPGPL (approved):

PR #6004, opened 2021-08-23
Update file formats to include official MIME

peppy:
As discussed in https://github.com/ppy/osu-web/issues/7202.

This also drops the weird "executable" terminology, which made no sense.


peppy:
I've intentionally dropped the referenced to `.db` type because it's not intended that users really deal with these ever.

wiki/osu!_File_Formats/en.md
```diff
@@ -2,26 +2,13 @@
...
-| File format | Usage |
-| :-- | :-- |
-| [.osr][Osr Link] | Executable replay file (cannot be decompressed). |
+| File format | MIME type | Usage |
+| :-- | :-- | :-- |
+| [.osz][Osz Link] | `x-osu-beatmap-archive` | osu! beatmap archive |
+| [.osk][Osk Link] | `x-osu-skin-archive` | osu! skin archive |
+| [.osu][Osu Link] | `x-osu-beatmap` | osu! [beatmap](/wiki/Beatmap) |
+| [.osb][Osb Link] | `x-osu-storyboard` | osu! [storyboard](/wiki/Storyboard_Scripting) |
+| [.osr][Osr Link] | `x-osu-replay` | osu! replay |
```
Joehuu, line 11:
```suggestion
| [.osr][Osr Link] | `x-osu-replay` | osu! [replay](/wiki/Replay) |
```

TicClick:
> I've intentionally dropped the referenced to `.db` type because it's not intended that users really deal with these ever.

I object: there are a couple of tools out there which work with the `. db` files and help you export collections, or [analyze replays (this one is very helpful)](https://github.com/ThereGoesMySanity/osuMissAnalyzer), or otherwise explore a list of installed maps. they wouldn't have been possible without the `. db` specification written on the wiki.

otherwise, looks good to me

peppy:
I dunno. It's wrong as it stands. There are like 3 different .db files we use, none have a standard format, none have a mime type, and none will be used going forward. It's fine to document them, but they should not be listed under file formats in the same place as user facing files.

How about we move that documentation to the github wiki?

Joehuu (approved):

TicClick:
> I dunno. It's wrong as it stands. There are like 3 different .db files we use, none have a standard format, none have a mime type, and none will be used going forward. It's fine to document them, but they should not be listed under file formats in the same place as user facing files.
> 
> How about we move that documentation to the [github wiki](https://github.com/ppy/osu/wiki)? It will fit well.

I suppose. In this case, the same treatment should be applied to `.osr`, and we need to put up a note somewhere for these willing to explore that'd point them to the wiki hosted on GitHub.

peppy:
this page is intended to list formats the user interacts with. .osr is one of those. if you mean describing the internals then potentially we can move that too, but the *extension* should be listed on this page still.

peppy:
> .db should also be removed in the main page.
> 
> https://github.com/ppy/osu-wiki/blob/50e18a0628f6f047dde8c874e2b769219f827e39/wiki/Main_Page/en.md#L31

I'm not doing that in this PR. I've opened an issue tracking this: https://github.com/ppy/osu-wiki/issues/6007

Joehuu (approved):

TicClick (approved):

PR #6013, opened 2021-08-25
[RU] Use alphanumeric identifiers for the headings of `Help Centre/Client`

TicClick:
unicode doesn't work yet.

also using this opportunity to suggest the same identifiers for the English version of `Help Centre/Client` — they are clear and conscise (or at least I've tried making them that)


TicClick:
to make it clear, no review from a native is required here

PR #6016, opened 2021-08-25
[KZ] Add `flag-references` and `group-info`

Elnurk:
Created kz.md in flag-reference folder. Created kz.yaml in group-info folder


TicClick:
looks like it's blocked due to ppy/osu-web#8060, and even then, what you have translated is pieces for the user group wiki articles. they don't have any impact until an article from that cycle is translated.

that being said, your pull request is *probably* harmless (= won't take the site down once submitted) but doesn't do anything, you may want to keep it for the better times.

p.s. if you know written Russian very well, you could contribute to its wiki segment instead — [come join us](https://osu.ppy.sh/community/forums/topics/566978)

Elnurk:
Should I close pull request untill better times?

TicClick:
I think so. it's not required, but if it isn't going to be merged in the foreseeable future, I see no point in keeping it open (especially because you can always reopen it)

PR #6043, opened 2021-09-02
Add `Notelock`

TicClick:
proofreading and fact-checking wanted. not that I've encountered this thing often


wiki/Gameplay/Note_lock/en.md
```diff
@@ -0,0 +1,32 @@
...
+---
+
+# Note lock
+
+![](img/notelock.gif "Example of note lock: the second note is inaccessible")
+
+**Note lock** is a gameplay mechanic of [osu!](/wiki/Game_mode/osu!) which may prevents a player from clearing a hit object. It happens if **two** conditions are met at the same time:
+
+1. The [timing windows](/wiki/Beatmapping/Overall_difficulty#timing) of two hit objects overlap.
+2. The first object of those two was either missed, or ignored by the player, and is still present on the play field.
```
peppy, line 15: missed objects don't block future objects. only ignored.

peppy:
Worth noting that "note lock" is a community defined term, not something we use internally. Not sure if that matters for the wiki (I guess we don't mention that for other terms so maybe irrelevant).

wiki/Gameplay/Note_lock/en.md
Walavouchey: pretty sure the "miss" here refers to clicking outside the hit object (and moving on to the next one), while i think you mean "miss" as in the hit window running out (which happens later). something like this should be correct:

```suggestion
2. The first object of those two was not clicked and is still present on the play field.
```

wiki/Gameplay/Note_lock/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+In this case, the second object is said to be *locked* behind the first one, which makes osu! ignore the player's input on it. Note lock only becomes a problem if the player is unable to return to the previous object and interact with it properly, thus removing the lock. In this case, a locked note may cause a cascading failure, forcing the player to miss more and more subsequent objects, until they run out of [health](/wiki/Beatmapping/Health) and fail the beatmap.
```
Walavouchey, line 17:
```suggestion
In this case, the second object is said to be *locked* behind the first one, which makes osu! ignore the player's input on it until the first object's hit window has passed. Note lock only becomes a problem if the player is unable to return to the previous object and interact with it properly, thus removing the lock. In this case, a locked note may cause a cascading failure, forcing the player to miss more and more subsequent objects, until they run out of [health](/wiki/Beatmapping/Health) and fail the beatmap.
```

wiki/Gameplay/Note_lock/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+When osu! rejects input, note lock visually manifests as a jittering hit circle (sliders and spinners are exempt from the visual effect of note lock).
+
+## Cause
+
+Note lock is a part of osu!'s timing system and happens when the timing windows of two objects overlap. It occurs more often on beatmaps with low [OD](/wiki/Beatmapping/Overall_difficulty) values, because the timing windows of hit objects overlap more frequently. High [BPM](/wiki/Beatmapping/Beats_per_minute) combined with low or medium OD values may also expose note lock.
```
Walavouchey, line 23:
```suggestion
Note lock is a part of osu!'s timing system and happens when the timing windows of two objects overlap. It occurs more often on beatmaps with low [OD](/wiki/Beatmapping/Overall_difficulty) or high [BPM](/wiki/Beatmapping/Beats_per_minute) values, because timing windows may overlap more frequently.
```

wiki/Gameplay/Note_lock/en.md
peppy: An alternative way of saying all this would be to combine both 1 and  2 into a single statement. Something like "if there's a previous object which has not been hit or missed (ie. not judged)"

wiki/Gameplay/Note_lock/en.md
```diff
@@ -0,0 +1,32 @@
...
+![](img/notelock.gif "Example of note lock: the second note is inaccessible")
+
+**Note lock** is a gameplay mechanic of [osu!](/wiki/Game_mode/osu!) which may prevents a player from clearing a hit object. It happens if **two** conditions are met at the same time:
+
+1. The [timing windows](/wiki/Beatmapping/Overall_difficulty#timing) of two hit objects overlap.
+2. The first object of those two was either missed, or ignored by the player, and is still present on the play field.
+
+In this case, the second object is said to be *locked* behind the first one, which makes osu! ignore the player's input on it. Note lock only becomes a problem if the player is unable to return to the previous object and interact with it properly, thus removing the lock. In this case, a locked note may cause a cascading failure, forcing the player to miss more and more subsequent objects, until they run out of [health](/wiki/Beatmapping/Health) and fail the beatmap.
+
+When osu! rejects input, note lock visually manifests as a jittering hit circle (sliders and spinners are exempt from the visual effect of note lock).
```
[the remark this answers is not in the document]
TicClick: yeah. was struggling with this one

wiki/Gameplay/Note_lock/en.md
TicClick: fixed, should be better now

wiki/Beatmapping/Overall_difficulty/en.md
```diff
@@ -23,7 +22,7 @@ Higher OD values mean shorter time windows to clear [hit objects](/wiki/Hit_Obje
 | 100 | `280ms - 16ms * OD` |
 | 300 | `160ms - 12ms * OD` |
 
-If the player hits outside of the 50's hit window, it will count as a miss.
+If the player hits outside of the 50's hit window, it will count as a miss. In case the hit windows of two objects overlap, the second object will be inaccessible due to [note lock](/wiki/Gameplay/Note_lock) until the first object is cleared.
```
TPGPL, line 25:
```suggestion
If the player hits outside of the 50's hit window, it will count as a miss. In case the hit windows of two objects overlap, the second object will be inaccessible until the first object disappears due to [note lock](/wiki/Gameplay/Note_lock).
```

wiki/Gameplay/Note_lock/en.md
```diff
@@ -0,0 +1,32 @@
+---
+tags:
+  - notelock
+  - jittering notes
+  - shaking notes
+---
+
+# Note lock
+
+![](img/notelock.gif "Example of note lock: the second note is inaccessible")
```
TPGPL, line 10:
```suggestion
![](img/notelock.gif "An example of note lock: the second note is inaccessible")
```

wiki/Gameplay/Note_lock/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Note lock** is an informal term for a gameplay mechanic of [osu!](/wiki/Game_mode/osu!) which may prevent a player from clearing a hit object. It happens if **two** conditions are met at the same time:
+
+1. The [timing windows](/wiki/Beatmapping/Overall_difficulty#timing) of two hit objects overlap.
+2. The first object of those two has not been judged yet (hit or missed).
+
+In this case, the second object is said to be *locked* behind the first one, which makes osu! ignore the player's input on it until the first object's hit window has passed. Note lock only becomes a problem if the player is unable to return to the previous object and interact with it properly, thus removing the lock. In this case, a locked note may cause a cascading failure, forcing the player to miss more and more subsequent objects, until they run out of [health](/wiki/Beatmapping/Health) and fail the beatmap.
```
TPGPL, line 17:
```suggestion
In this case, the second object is said to be *locked* behind the first one, which makes osu! ignore the player's input on it until the first object's hit window has passed. Note lock only becomes a problem if the player is unable to return to the previous object and interact with it properly, thus removing the lock. In this case, a locked note may cause a cascading failure, forcing the player to miss more and more subsequent objects, until they run out of [health](/wiki/Beatmapping/Health), and fail the beatmap.
```

TPGPL (changes requested):

wiki/Gameplay/Note_lock/en.md
TicClick: this is a regular sequential part of the sentence, no serial comma required IMO

TPGPL (approved):
lgtm

TPGPL (approved):

peppy (approved):

PR #6049, opened 2021-09-02
Mention new cutoff height for `scorebar-colour.png` in RC and `Interface`

### LINKED ISSUE #5945: Add scorebar-colour height limit
its now limited to 120px (SD) for both animated and non-animated scorebar-colours.

(confirmation from peppy: https://github.com/ppy/osu-stable-issues/issues/809)

TicClick:
closes #5945 — @RockRoller01 is that right?


RockRoller01:
yes, thats correct

TPGPL (approved):

PR #6012, opened 2021-08-25
Add tooltips to links

TPGPL:
used `\((http|https)://(?!(assets|osu|status|docs|old|s)\.ppy\.sh)(?!.[^ ]* ")` to find links without tooltips

*i hope it didnt miss anything 🙄*

PR #6060, opened 2021-09-05
Remove tooltips from links

TPGPL:
contrary to #6012

since the only reason for link tooltips to exist is making the external links distinguishable from regular wikilinks, we decided that it will be better to remove them completely and leave the matter for osu-web to add a native indicator for such links like on Wikipedia

image

used `\((http|https)://(?!(assets|osu|status|docs|old|s)\.ppy\.sh).[^ ]* "` to find the links with tooltips


TicClick:
two points:
- please make a relevant issue for osu-web
- we still need the http -> https replacement

TPGPL:
created an issue; about http link conversion, i will do it in a separate PR as it turns out there's many more links to be converted in the translations and `/news/` directory, and it would be an overkill for this PR

wiki/History_of_osu!/2015/en.md
```diff
@@ -115,14 +115,14 @@ A few days later (19 November 2015), peppy updated users on the situation in [an
 > - All users which were targeted so far have used the same password for 3+ years.
 > - Many used a very simple password, or shared the password across other services.
 > - All users compromised have had at least one other account compromised in the past.
->
+
```
peppy, line 118: Isn't this going to change the visual style of this?

TPGPL: yeah it will be split into three parts, but there's no other way to fix the warning, because for some reason remark doesn't view lists in citation blocks correctly

probably we could return it to the original state and ignore the error, as it's not an issue on our side

peppy: I think I'd prefer leaving it.

cl8n:
adding and removing link titles by script seems defeating of their purpose. they can signal an external link, sure, but that's because they're used to give more context about a link in general.

for example these are pointless, and probably shouldn't have been added in the first place:

```diff
-For forthcoming updates, [follow us on Twitter](https://twitter.com/osubwc "Twitter") and [join our Discord server](https://discord.gg/CZp4bNx "Discord")!
+For forthcoming updates, [follow us on Twitter](https://twitter.com/osubwc) and [join our Discord server](https://discord.gg/CZp4bNx)!
```

while these give useful context about the links (just the site name here, there might be better examples), and shouldn't be removed:

```diff
-[**Please fill out this form in order to apply as a judge!**](https://forms.gle/he4sT7DeX5UVr4j27 "Google Forms")
+[**Please fill out this form in order to apply as a judge!**](https://forms.gle/he4sT7DeX5UVr4j27)
```

```diff
-A full overview of judge results can be found [here](https://docs.google.com/spreadsheets/d/e/2PACX-1vTts0zpdjKdCUS3oCP4XZ18vh5ckEDqL_Jvy2eh0B4kC8je09bCGwYcpOWgjaXXy0251c3f33u7BDjv/pubhtml "Google Sheets")
+A full overview of judge results can be found [here](https://docs.google.com/spreadsheets/d/e/2PACX-1vTts0zpdjKdCUS3oCP4XZ18vh5ckEDqL_Jvy2eh0B4kC8je09bCGwYcpOWgjaXXy0251c3f33u7BDjv/pubhtml)
```

the proposed display feature for external links seems good to me, but it doesn't replace the function of link titles and shouldn't affect the wiki's content

TicClick:
I mean, alright, okay, *maybe* there are links that have the page's actual and detailed title as their alt text, but that could be included in the title as well, without everyone having to fill one more field. if anything, we could use wiki-like footnotes for most, if not all links that need such a description, like Wikipedia does:

image

cl8n:
> I disagree. what you describe should be achieved by the link's title (`Please fill out this form in order to apply as a judge` in your second example; `here` and `this article` is what should also be replaced by proper descriptions), not the alt text.

well yes in these cases I'd consider that sufficient too, but it wasn't done here 🤔

(also just to clarify I think by "alt text" you mean the title and by "title" you mean the content, links don't have "alt text")

> knowing that `https://forms.gle/he4sT7DeX5UVr4j27` is `Google Forms`, and `https://docs.google.com/spreadsheets/...` is `Google Spreadsheets` is useless and brings me as a reader absolutely nowhere. it's not even the form's name.

not sure I'm following, I know these titles aren't very descriptive, but at least they tell me the name of the service. this change (so far) removes the titles and intends to have them replaced by an indicator for external links, which won't tell me anything about the linked sites other than they aren't osu

TicClick:
> also just to clarify I think by "alt text" you mean the title and by "title" you mean the content, links don't have "alt text"

thanks, edited my comments to avoid confusion.

on topic of your reply, I don't think that telling the name of a service alone has any use for you as a reader.

suppose that now you know the form is hosted on Google Forms, what does it change? granted, in absence of the link's content (`download it [here](http://shady.website)`) it is somewhat useful, but otherwise you are primarily interested in filling out the form, or staying away from it, and IMO this is determined by the link's content, otherwise you most likely know the website's domain well enough anyway. the title itself is barely descriptive is what I want to say, and in vast majority of cases it is even beaten by a URL alone

TicClick (approved):
changes look good to me. @peppy merge at your discretion, together with #6069 and #6070.

ps we need to prune `here`s from the links' content as well, but that's for another time

PR #6065, opened 2021-09-06
Corrections to video encoding guide & add FFmpeg instructions.

nathanielcwm:


nathanielcwm:
btw is there a reason why the framerate in the guide is set to 30 and not just left at the default if it's not hfr?

nathanielcwm:
> btw is there a reason why the guide states to set the framerate to 30 and not just left at the source if it's not hfr?

welp there are ranked maps that have 60 fps video so ig not 
anyway it would be weird for ffmpeg to not handle those framerates properly / handle them worse than 30 fps (especially since those framerates are so common)

wiki/Guides/Compressing_files/en.md
```diff
@@ -29,27 +29,49 @@ This section will show you how to remove audio from videos using [Handbrake](htt
...
 
 ![Setting the video codec and quality in Handbrake](img/codecquality-handbrake.jpg "Setting the video codec and constant quality")
 
-5. Set the framerate to `30`.
+6. Set the framerate to be the same as the source and set it to constant framerate.
 
 ![Setting the video framerate in Handbrake](img/framerate-handbrake.jpg "Setting the framerate")
 
-6. To resize the image of the video file, go to the `Dimensions` tab and change the width to `1280` and change the height to `720`.
+7. To resize the image of the video file, go to the `Dimensions` tab and change the width to `1280` and change the height to `720`.
```
peppy, line 42: Does this only downscale?

wiki/Guides/Compressing_files/en.md
```diff
@@ -62,3 +84,16 @@ The [Ranking Criteria](/wiki/Ranking_Criteria#audio) has a rule noting that anyt
...
 ![Export settings](img/exportsettings-audacity.jpg "Export settings")
+
+### Using FFmpeg
+
+Sample command: ```ffmpeg -i input -c:a libmp3lame -q:a 4 -vn -sn -map_metadata -1 -map_chapters -1 output.mp3```
+
+Explanation:
+
+* `-i input`: your source file.
+* `-c:a libmp3lame`: specifies that you want to use the lame MP3 encoder *tbh this isn't necessary as lame should be the default*.
```
Walavouchey, line 95:
```suggestion
* `-c:a libmp3lame`: Specify that the audio should be encoded using the LAME MP3 encoder.
```

wiki/Guides/Compressing_files/en.md
```diff
@@ -62,3 +84,16 @@ The [Ranking Criteria](/wiki/Ranking_Criteria#audio) has a rule noting that anyt
...
+* `-q:a 4`: uses the same variable bitrate range as the Audacity example (lower numbers are higher bitrate). (use `-b:a 128k` for a constant 128k bitrate instead)
```
Walavouchey, line 96:
```suggestion
* `-q:a 4`: Use the same variable bitrate range as in the Audacity example, where a lower number means higher bitrate. If you want constant bitrate, you would instead use for instance `-b:a 128k` for a constant 128kps bitrate.
```

wiki/Guides/Compressing_files/en.md
```diff
@@ -62,3 +84,16 @@ The [Ranking Criteria](/wiki/Ranking_Criteria#audio) has a rule noting that anyt
...
+* `-vn -sn`: removes video & subtitles if present.
+* `-map_metadata -1 -map_chapters -1`: removes metadata & chapters if present.
```
Walavouchey, line 98:
```suggestion
* `-vn -sn`: Remove video and subtitles if present.
* `-map_metadata -1 -map_chapters -1`: Remove metadata and chapters if present.
```

wiki/Guides/Compressing_files/en.md
```diff
@@ -29,27 +29,49 @@ This section will show you how to remove audio from videos using [Handbrake](htt
 
 ![Removing audio tracks from the audio tab in Handbrake](img/removeaudio-handbrake.jpg "Removing the audio tracks")
 
-4. Go into the `Video` tab and set the video codec as `H.264 (X264)`. Change the `Constant Quality` to between 20–25. Smaller values will produce smaller file sizes at the cost of the video's quality.
+4. Go into the `Video` tab and make sure the video codec is set as `H.264 (X264)`. Change the `Constant Quality` to between 20–25. Smaller values will produce larger file sizes but with a higher video quality.
+
+5. If you're willing to spend more time encoding slower Encoder Presets deliver better video quality and may also reduce the video file size, but do not go down to placebo as it is prohibitively slow and often performs worse than VerySlow.
```
Walavouchey, line 34:
```suggestion
5. If you are willing to spend more time encoding, change the `Encoder Preset` under `Encoder Options`. Slower presets deliver better video quality and may also reduce video file size, but do not go down to placebo as it takes much longer than `VerySlow` for very little improvement in quality.
```

wiki/Guides/Compressing_files/en.md
Walavouchey: apparently not (anymore?). seems like these images are very outdated

image

Walavouchey (changes requested):
i'm unsure on how much detail and explanation should be included, as well as on some of the formatting, so before looking at this again i'll wait for the potential opinions of others.

wiki/Guides/Compressing_files/en.md
nathanielcwm: lol I still have the old UI for this... *same as the old image and not like above*

ig I would need to redo some of the screenshots and update the handbrake that I have locally.

nathanielcwm: Looks like this new UI was released as part of the handbrake 1.4 update (dates back to July 2021).

ig it might be a good idea to pop in a notice to check for updates if the ui looks different?

nathanielcwm:
Alternatively for handbrake I've created an osu! preset using the settings in the guide.

[preset.txt](https://github.com/ppy/osu-wiki/files/7272843/preset.txt)

TicClick:
what's the status of this PR? are you going to add more information, or are you waiting for extra reviews?

nathanielcwm:
> what's the status of this PR? are you going to add more information, or are you waiting for extra reviews?

Mix of both, I'm currently a bit busy with school so it might be a while until I can have a look at this again.

Walavouchey:
this is all ok now except for the osu! preset in this image, which shouldn't be there

i don't think there needs to be a preset provided by the article anyway

image

Walavouchey (approved):

TPGPL (approved):
looks good i think

PR #6067, opened 2021-09-06
Adjust existing articles to the infobox syntax change

TicClick:
- details: https://github.com/ppy/osu-wiki/issues/5440#issuecomment-913170841
- depends on: https://github.com/ppy/osu-web/pull/8087 (please only merge after it's deployed)


gagahpangeran:
Hmm, the new syntax is `:::`.

gagahpangeran:
This can be merge because https://github.com/ppy/osu-web/pull/8094 as intermediate step for transition is already deployed.

nanaya (approved):

PR #6070, opened 2021-09-06
Remove tooltips from links in news posts

TPGPL:


TicClick (approved):

peppy:
Any reason this is split up into specific sections of articles?

TicClick:
it's to speed up the review process. the first pull covers only regular articles which aren't protected by the code owners file (there are about 450 files affected), while these two (news/help centre) are <=20 files total

peppy:
I dunno, I see it as the opposite. If it's all in one place with one command I can run and diff/compare against it's less effort for me. And if I'm disagreeing with this, it's going to be on all or none, not just on new posts for example.

Whatever works i guess.

Ephemeralis:
Why are we removing these, exactly? Do they break something?

TicClick:
> Why are we removing these, exactly? Do they break something?

from #6060:

> contrary to #6012
> 
> since the only reason for link tooltips to exist is making the external links distinguishable from regular wikilinks, we decided that it will be better to remove them completely and leave the matter for osu-web to add a native indicator for such links like on Wikipedia
> 
> image

TicClick (approved):

PR #6071, opened 2021-09-06
Update `ASC` with recent changes to link tooltips

TPGPL:


TicClick (approved):
oh, really? this is pretty slim, I expected more for some reason

PR #6079, opened 2021-09-08
Clean up name order RC

NoffyNoffeh:
The change makes it so it would largely be handled the same way as song titles are already, making it more consistent to understand how to apply the rules across the board.

see thread: 
https://osu.ppy.sh/community/forums/topics/1415070?n=1

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -106,6 +106,7 @@ Changes to the rules and guidelines in this document are proposed and discussed 
 
 #### Romanisation
 
+- **An artist's name must romanised in the order it is printed in the unicode field if there is no official romanisation available.** If an official romansiation is available, either the direct romanisation or official may be used.
 - **Loan words from other languages have to use the original words in their stead when attempting to romanise them.**
 - **When a song uses repeat words in the title or in the artist where one is in unicode, and the other as a basic romanisation, the romanised field must use the provided romanisation only and remove the duplicate word.**
 - **Umlauts must be romanised into two-letter equivalents: `ü` to `ue`, `ö` to `oe`, `ä` to `ae` and `ß` to `ss`.**
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -106,7 +106,6 @@ Changes to the rules and guidelines in this document are proposed and discussed 
 
 #### Romanisation
 
-- **An artist's name is to be romanised in the order it is printed in the unicode field, unless an alternative order for the romanised name is provided by the artist.**
 - **Loan words from other languages have to use the original words in their stead when attempting to romanise them.**
 - **When a song uses repeat words in the title or in the artist where one is in unicode, and the other as a basic romanisation, the romanised field must use the provided romanisation only and remove the duplicate word.**
 - **Umlauts must be romanised into two-letter equivalents: `ü` to `ue`, `ö` to `oe`, `ä` to `ae` and `ß` to `ss`.**
@@ -146,8 +145,7 @@ Changes to the rules and guidelines in this document are proposed and discussed 
 This category contains explicit allowance statements of concepts and rules that are not commonly straightforward even after reading this whole section of the ranking criteria.
 
 - **For songs where the composer(s) and singer(s) are different people, the singer(s) may be listed after the composer(s) or circle/group name following a `feat.` indicator.**
-- **If an artist has provided an official translation for their name, this may be used in the romanised artist field.** Official romanisation may be used for the spelling of an artist's name, but the name order must follow the related rule.
-- **If a Unicode Song title has either an official translation or romanisation provided by the artist, either or may be used in the romanised title field.**
+- **If a Unicode song title or artist has either an official translation or romanisation provided by the artist, either or may be used in the respective romanised field.**
 - **If a beatmap's track was contributed to by multiple artists, they may be listed with commas inbetween.** If there are 3 or more contributing artists and they are not part of one officially labelled group, `Various Artists` or other descriptive artist labels may be used instead.
 - **For Remixes/Covers, the original artist may be used in the artist field, as long as the title field is modified to clearly show that the song is remixed.** This marker should all be in parentheses and contain the Remix/Cover artist followed by descriptor.
 
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -106,7 +106,7 @@ Changes to the rules and guidelines in this document are proposed and discussed 
 
 #### Romanisation
 
-- **An artist's name must romanised in the order it is printed in the unicode field if there is no official romanisation available.** If an official romansiation is available, either the direct romanisation or official may be used.
+- **An artist's name must romanised in the order it is printed in the unicode field if there is no official romanisation or translation available.** If an official romanisation is available, either option may be used in accordance with the relevant allowance.
 - **Loan words from other languages have to use the original words in their stead when attempting to romanise them.**
 - **When a song uses repeat words in the title or in the artist where one is in unicode, and the other as a basic romanisation, the romanised field must use the provided romanisation only and remove the duplicate word.**
 - **Umlauts must be romanised into two-letter equivalents: `ü` to `ue`, `ö` to `oe`, `ä` to `ae` and `ß` to `ss`.**
@@ -143,7 +143,7 @@ Changes to the rules and guidelines in this document are proposed and discussed 
 
 ### Allowances
 
-This category contains explicit allowance statements of concepts and rules that are not commonly straightforward even after reading this whole section of the ranking criteria.
+This category contains explicit allowance statements of concepts and rules that are not commonly straightforward even after reading this whole section of the ranking criteria. In cases where there are multiple options available for a song's metadata, the above standardization rules for the Ranked and Loved sections takes priority.
 
 - **For songs where the composer(s) and singer(s) are different people, the singer(s) may be listed after the composer(s) or circle/group name following a `feat.` indicator.**
 - **If a Unicode song title or artist has either an official translation or romanisation provided by the artist, either or may be used in the respective romanised field.**


wiki/Ranking_Criteria/en.md
```diff
@@ -106,7 +106,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
 
 #### Romanisation
 
-- **An artist's name is to be romanised in the order it is printed in the unicode field, unless an alternative order for the romanised name is provided by the artist.**
+- **An artist's name must romanised in the order it is printed in the unicode field if there is no official romanisation or translation available.** If an official romanisation is available, either option may be used in accordance with the relevant allowance.
```
Ephemeralis, line 109:
```suggestion
- **An artist's name must be romanised in the order it is printed in the unicode field if there is no official romanisation or translation available.** If an official romanisation is available, either option may be used in accordance with the relevant allowance.
```

Ephemeralis:
This looks fine to me. I'll let it stew for a few days before approving in case anyone else wants to ogle at it.

wiki/Ranking_Criteria/en.md
```diff
@@ -106,7 +106,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
 
 #### Romanisation
 
-- **An artist's name is to be romanised in the order it is printed in the unicode field, unless an alternative order for the romanised name is provided by the artist.**
+- **An artist's name must be romanised in the order it is printed in the unicode field if there is no official romanisation or translation available.** If an official romanisation is available, either option may be used in accordance with the relevant allowance.
```
peppy, line 109: what does "relevant allowance" mean? i can't understand this.

NoffyNoffeh: Same one changed in the PR, line 149 which allows people to choose if they have good reason. Ideally it would be cross linked but not sure on how that would be done.

TicClick: I think peppy refers to the complexity of that sentence and that it could be simplified (I honestly barely understand it either)

wiki/Ranking_Criteria/en.md
```diff
@@ -143,11 +143,10 @@ Changes to the rules and guidelines in this document are proposed and discussed
 
 ### Allowances
 
-This category contains explicit allowance statements of concepts and rules that are not commonly straightforward even after reading this whole section of the ranking criteria.
+This category contains explicit allowance statements of concepts and rules that are not commonly straightforward even after reading this whole section of the ranking criteria. In cases where there are multiple options available for a song's metadata, the above standardization rules for the Ranked and Loved sections takes priority.
```
Walavouchey, line 146:
```suggestion
This category contains explicit allowance statements of concepts and rules that are not commonly straightforward even after reading this whole section of the ranking criteria. In cases where there are multiple options available for a song's metadata, the above standardisation rules for the Ranked and Loved sections take priority.
```

wiki/Ranking_Criteria/en.md
NoffyNoffeh: Reworded, please follow up if still unclear.

wiki/Ranking_Criteria/en.md
```diff
@@ -106,7 +106,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
 
 #### Romanisation
 
-- **An artist's name is to be romanised in the order it is printed in the unicode field, unless an alternative order for the romanised name is provided by the artist.**
+- **An artist's name must be romanised in the order it is printed in the unicode field if there is no official romanisation or translation available.** When an official romanisation is available, this method is optional.
```
peppy, line 109: this still reads shockingly. what is optional? why is it optional? why is the official romanisation not preferred? it should be preferred no?

what's a "method"? you mean order?

wiki/Ranking_Criteria/en.md
```diff
@@ -143,11 +143,10 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
+This category contains explicit allowance statements of concepts and rules that are not commonly straightforward even after reading this whole section of the ranking criteria. In cases where there are multiple options available for a song's metadata, the above standardisation rules for the Ranked and Loved sections take priority.
 
 - **For songs where the composer(s) and singer(s) are different people, the singer(s) may be listed after the composer(s) or circle/group name following a `feat.` indicator.**
-- **If an artist has provided an official translation for their name, this may be used in the romanised artist field.** Official romanisation may be used for the spelling of an artist's name, but the name order must follow the related rule.
-- **If a Unicode Song title has either an official translation or romanisation provided by the artist, either or may be used in the romanised title field.**
+- **If a Unicode song title or artist has either an official translation or romanisation provided by the artist, either or may be used in the respective romanised field.**
```
peppy, line 149: either or? the artist is specifying two? am i english?

peppy:
As a heads up, generally you'd discuss an alternative in the review conversation before committing it. As seen here, your new change is no better, so it would be best to discuss and reach a consensus rather than just committing again and again.

wiki/Ranking_Criteria/en.md
NoffyNoffeh: Forcing it in a rule would be a regression as some official romanizations are extremely obscure and to require someone to find it to follow the RC. Think a single occurence on one album 12 years ago or buried agency pages. Defining what counts or not for this would be a headache in its own right.

I mean order.

wiki/Ranking_Criteria/en.md
NoffyNoffeh:
```suggestion
- **If a Unicode song title or artist has either an official translation or romanisation provided by the artist, it may be used in the respective romanised field.**
```
?

wiki/Ranking_Criteria/en.md
peppy: Shouldn't it still be preferred? If you want to guard against effort being expended, something like:

"If an official romanisation is readily available, it should be preferred. If not, an artist's name must be a direct romanisation of the unicode artist name (including order)"

Walavouchey:
flowchart based on [discussion](https://discord.com/channels/188630481301012481/218677502141399041/885569615055425586) in `#osu-wiki`:

Romanisation-flow-chart

hopefully this clears any ambiguities when trying to review.

wiki/Ranking_Criteria/en.md
NoffyNoffeh: With the allowances in place there's a tendency to prefer what's prevalent and recognizable already, as has been the case for song titles. Your rewording would work similarly to what's currently here in the PR but...

However, that wording "should be preferred" would make the comment into its own guideline. There are currently no guideline section for romanization as all relevant parts have been treated as hard rules, formatting wise it would separate this clause from the rest of the romanization sections, making it a bit more obscure to notice in RC.

wiki/Ranking_Criteria/en.md
NoffyNoffeh: Song titles often do have both, such as https://www.youtube.com/watch?v=XZnmfDByGWU or many others (first example that came to mind). It's less common for artists.

peppy: the flow chart changes nothing. it's just splitting the sentence into boxes. the point is that it is ambiguous to someone reading it

"oh either may be used? i think i'll choose the one that is different from 100 ranked maps just to be difficult!"

is this not something we want to avoid? the sentence should say which is preferred to give the reader direction. if the preference is to follow suite then that should be mentioned.

NoffyNoffeh: `If the same song exists in the Ranked or Loved sections already, the metadata should be followed unless it breaks other rules in the ranking criteria or the official sources state something completely different.` and 
`**Artist names should be consistent between different songs from the same person or group in the Ranked or Loved sections.** This does not apply if the person or group intentionally uses a different alias for different song or album releases. ` 

are in standardization guidelines already since consistency across sets relates to any meta rules or cases where there's options and not just romanisation.

peppy: So this is the case where this is the first beatmap by the artist that is ranked? In that case, shouldn't the official romanisation be used?

NoffyNoffeh: If that's the case, that expands beyond just how romanisation is handled and is out of scope of this rule. The topic of pushing what options for metadata in general is preferred would need a seperate community discussion first and creating it as its own separate guideline, to make sure it's something agreed upon and easily understood. This would be instead of putting it in solely for artist name romanisation.

wiki/Ranking_Criteria/en.md
NoffyNoffeh: With https://github.com/ppy/osu-wiki/pull/6079#discussion_r706344648 in mind,

Does 
```suggestion
- **When romanising an artist's name yourself, it must be romanised in the order it is printed in the unicode field.** This must be done if no official romanisation is available.
```

work? This forces no specific method when official options are given, but still leaves the option of romanising it this way open and not breaking any rules in cases where "official" is obscure and it is romanised by the mapper or modder.

Preference for official/recognizable over all should be in its own guideline as mentioned in the other comment.

wiki/Ranking_Criteria/en.md
Walavouchey: whether this should explicitly be a choice or preferred one way, i'll just put this here to get the sentence into a readable state in case this does go through.

```suggestion
- **If a Unicode song title or artist has an official translation or romanisation provided by the artist, either may be used in the respective romanised field.**
```

wiki/Ranking_Criteria/en.md
Walavouchey: note that since you removed "or translation" and moved the part outside like that, it would imply that if there is no official romanisation but there is a translation, the artist name must be romanised anyway (and the translation discarded), which conflicts with the allowance

shuffling it around like this prevents that

```suggestion
- **When romanising an artist's name yourself and no official romanisation is available, it must be romanised in the order it is printed in the unicode field.**
```

this doesn't specify whether the the translation or the official romanisation should be picked at all or which is preferred, and isolates the clause to say exactly one thing only. if other rules after discussion will eventually specify what is preferred, this way would be cleaner.

peppy: This reads fine, but the larger issue I touched on probably still stands (and how things are currently done sounds pretty weird). I don't have the throughput for dealing with that can of worms though.

wiki/Ranking_Criteria/en.md
peppy: I still don't know what "either" is referencing here. There's no two items in the adjacent text. Is it talking about the "feat" term?

NoffyNoffeh: Either the translation or the romanisation, as mentioned some songs and artists have and use both translation and romanisation interchangeably.

Walavouchey: "either [of them]" is referring to the "translation or the romanisation" in the same sentence, and to me that reads fine, but here's an alternative suggestion:

```suggestion
- **Official translations or romanisations of the Unicode song title or artist name may be used in their respective romanised fields.**
```

peppy: "either" in english can only be used if there are two optiond at hand. please re-read my issue.

If this is talking about a case where the artist/title has two options, this sentence needs to mention what the two options are.

Walavouchey: hmm. maybe i've misunderstood what this clause was supposed to convey. @NoffyNoffeh is this clause supposed to only cover the case where both of these options are available?

```suggestion
- **If a Unicode song title or artist has both an official translation or romanisation provided by the artist, either of them may be used in the respective romanised field.**
```

peppy: @Walavouchey did you mean "official translation *and*"? because at that point the sentence validates for me.

NoffyNoffeh: Saying and doesn't make sense, that implies both title and artist have to be translated/romanized to be able to use one.

Plenty of artists have translated/romanized names but not song titles.


I don't understand your issue with the original sentence, EITHER translation OR romanisation when both exist. Both options are in the same sentence as when `either` is ever mentioned.

Walavouchey: 1) this only implies that there is no preference between any official translation or romanisation if both are available. it doesn't say anything about any other case.

2) if the issue is with the usage of "either", then this clause simply needs to be reworded to be a bit longer to cover both cases

NoffyNoffeh: The clause covers using official options when they available.

Its main function is telling you that you can use official options, and mentions if there's both romanisation/translation, use either romanisation or translation. Choosing between the two is not the main point.


I don't understand why a completely different clause about `feat.` is being mentioned when it's totally unrelated and is its own point.

Walavouchey: we really are getting lost in the wording. i hope this satisfies all sides, and if not, i'll leave this for other reviewers to opinionate over.

```suggestion
- **If a Unicode song title or artist has an official translation or romanisation provided by the artist, it may be used in the respective romanised field. If both are available, either may be used.**
```

peppy: I'm only getting lost because I know others will in the same way. Not trying to be difficult.

So without reading the previous discussion: This is implying that either the "translation" or "romanisation" may be used?

NoffyNoffeh: Yes

peppy: I've reworded the suggestion slightly to be more explicit.

wiki/Ranking_Criteria/en.md
```diff
@@ -106,7 +106,7 @@ Changes to the rules and guidelines in this document are proposed and discussed
 
 #### Romanisation
 
-- **An artist's name is to be romanised in the order it is printed in the unicode field, unless an alternative order for the romanised name is provided by the artist.**
+- **When romanising an artist's name yourself and no official romanisation is available, it must be romanised in the order it is printed in the unicode field.**
```
peppy, line 109:
```suggestion
- **When romanising an artist's name yourself (ie. no official romanisation is available), it must be romanised in the order it is printed in the unicode field.**
```

I think you wouldn't be romanising yourself if there was an official one, correct?

Walavouchey: Implying that you shouldn't romanise the artist name yourself if there is an official one available contradicts with the allowance, since it states that picking the official romanisation or translation is optional. Again, any preference towards which option to pick is out of scope for this proposal as [Noffy stated](https://github.com/ppy/osu-wiki/pull/6079#discussion_r706344648).

peppy: so you are allowed to romanise differently to the artist's preference?? that seems very counter-intuitive.

we should just fix that with the proposed wording change here. @Ephemeralis second opinion on this weirdness please.

NoffyNoffeh: I'm not ephemeral but opened PR https://github.com/ppy/osu-wiki/pull/6149 to cover this instead. I think this current PR should be good to go as is once that is merged.

mousesalt:
please exclude the id.md file, we still has a PR open on #6118

TicClick:
I'd rather keep it to inform others that it's not up to date: even if you and me are aware of it, others aren't

PR #6090, opened 2021-09-11
Add clause about osu! source usage in Ranking Criteria

Ephemeralis:


peppy (approved):
Kinda hoping this wouldn't need to be specifically mentioned, but no harm.

PR #6108, opened 2021-09-15
Fix typo on Cychloryn's name

mangomizer:
Cychlorin -> Cychloryn


WalterToro:
Already resolved by peppy on commit 179093f

PR #6109, opened 2021-09-15
Update  `Medals/Unlock_requirements/Dedication`

SiderealArt:


TicClick (approved):
thanks (checked by visiting [aabc271](https://osu.ppy.sh/users/155707)'s profile)

PR #6112, opened 2021-09-16
Update `OWC 2021`

LeoFLT:
Fixes a few incorrect links that pointed to a non-existent profile and adds some missing information on the screening process and updates both the wiki article and the news post with the new prize distribution.


news/2021-09-16-osu-world-cup-2021-registrations-now-open.md
```diff
@@ -26,11 +26,11 @@ For more details, refer to the **[osu! World Cup 2021 wiki article](/wiki/Tourna
 
 ## Prizes
 
-The osu! World Cup 2021 offers a $5,000 prize pool, along with limited edition goods!
+The osu! World Cup 2021 offers a $5,000 prize pool, along with limited edition goods! Prize amounts for each placement have been slightly adjusted from 2020 for a more fair distribution.
 
-- **1st:** 50% of the prize pool, exclusive limited-edition merch, profile badge, "osu! Champion" user title for one year
-- **2nd:** 34% of the prize pool, exclusive limited-edition merch, profile badge
-- **3rd:** 16% of the prize pool, exclusive limited-edition merch, profile badge
+- **1st:** 48% of the prize pool, exclusive limited-edition merch, profile badge, "osu! Champion" user title for one year
```
peppy, line 31: Just for documentation purposes, is this objectively fairer? Is it based off other tournaments or some discussion that came up?

LeoFLT: The main drive behind this change was so that 3rd place finalists get a value that is more "acceptable" at a negligible cost, seeing as the only difference between the second and third place finishers is one match:

1st place goes from $312.50 for each player to $300
2nd from $225 each to $200
3rd from $87.50 each to $125

LeoFLT: There are a few examples of other recent tournaments doing this as well, [corsace open](https://open.corsace.io/info) for instance (while the percentages are not exactly the same as ours, the overall idea stays the same: make the distinction between 2nd and 3rd place finalists less extreme)

peppy: Any reason to not make it 50/30/20 in that case? I was more curious about the first place reduction.

peppy: one could argue the proposal is punishing first place too, i dunno what that logic is. but i guess as long as the community can understand this even if i can't?

will leave this PR open for a day or two in case anyone else wants to chime in

wiki/Tournaments/OWC/2021/en.md
```diff
@@ -102,7 +102,7 @@ The osu! World Cup 2021 is run by the osu! team and various community members.
    - Captains are allowed to exclude themselves from the team list.
    - Captains are allowed to conduct and coordinate "tryouts", tests to gauge player aptitude, on their own terms.
    - Captains may choose, with the approval of the Tournament Management, at any time, to delegate their role to other prospective team members. Once given, only the new captain may assign the captaincy to another player.
-3. To ensure valid registrations, every prospective participant will be manually checked by the Tournament Management, in a manner similar to the [tournament screening](/wiki/Tournaments/Official_support#tournament-screening) that is offered to community tournaments.
+3. To ensure valid registrations, every prospective participant will be manually checked by the [Account Support Team](/wiki/People/The_Team/Account_support_team), in a manner similar to the [tournament screening](/wiki/Tournaments/Official_support#tournament-screening) that is offered to community tournaments.
```
TPGPL, line 105:
```suggestion
3. To ensure valid registrations, every prospective participant will be manually checked by the [account support team](/wiki/People/The_Team/Account_support_team), in a manner similar to the [tournament screening](/wiki/Tournaments/Official_support#tournament-screening) that is offered to community tournaments.
```

TPGPL (changes requested):

peppy (approved):

PR #6128, opened 2021-09-18
Update `Ranking Criteria`

abraker95:
Discussion at https://osu.ppy.sh/community/forums/topics/1418542 is in agreement with the proposal.

Removes the requirement of having hitsound additions in osu!mania beatmaps.


peppy:
Is it possible that we can reword this to give rationale? Something like "To allow for easier approachability to osu!mania mappers of different upbringings, hitsounds are not required. It is still highly recommended to add them to improve the feel of your beatmaps."

abraker95:
Added rationale. Very minor change from suggested to preserve "**rule** details" format.

peppy:
Hold off on merging this until I have a chance to read through the thread. It doesn't seem 100% cut and dry, so want to make sure I'm on the same page.

One thing to keep in mind is that hitsounds are used for converting maps, which in lazer would be used for conversions from mania to taiko, for instance. Food for thought that dropping the requirement would make such conversions no longer play well.

juankristal:
I was not aware that a conversion of mania to taiko was a thing in lazer. In that case I would argue it is not that important of a thing though considering you already have the STD to any mode conversion. And to be fair, conversions never really played well from STD to mania or taiko (therefore it would probably be the same for mania to taiko if it is converted in a similar way)

Kominaru:
If worries about implementing this change are mania->taiko conversions (which to be fair I don't think are necessary anymore, taking into account the amount of content we have in the game now, as Juan says), there are way better ways that do not rely on hitsounds (one possibility is by using chord sizes to differentiate kats and dons).

Anyway, perhaps not the right place to discuss this but I feel something like automatic converts shouldn't halt the implementation of this change, mainly seeing that we already have this same rule for IX and X spreads.

peppy:
Forget I said anything bout converts. No further response here is required for now.

PR #6136, opened 2021-09-19
[EN] Update `Guides/Cropping with Complex Backgrounds`

DoryanR:
Update of the article to change the phrasing, and remove the informal tone.


TicClick:
the wording hasn't really changed and is still too convoluted, do you want to take on the whole article, or leave that to some other person instead? in case of the latter, I'd suggest closing this PR, as it doesn't make difference

DoryanR:
I prefer to let someone else do it, I'm not very comfortable in rewriting an article in English.

PR #6149, opened 2021-09-22
Add preferred metadata guideline to `Ranking_criteria`

NoffyNoffeh:
From https://osu.ppy.sh/community/forums/topics/1417008 which initially started from discussion at https://github.com/ppy/osu-wiki/pull/6079 .

Makes the common sense thing be written down so it can be more commonly understood.


peppy:
I guess this works, although it's adding many more words to an already wordy explanation.

NoffyNoffeh:
The meaning would be the same with just the bold I think, should I cut the other half off? Or are you imagining something else for wordiness reduction.

peppy:
I dunno I think it's best just going with this until we get people interested in rewriting the whole document to not be an essay.

peppy (approved):

PR #6150, opened 2021-09-22
Update source rule in `Ranking_Criteria`

NoffyNoffeh:
From thread https://osu.ppy.sh/community/forums/topics/1417764

Based on the general understanding shown in the thread, with bullet points for some particular source types.


Purplegaze:
"such as commissioned tiebreaker tracks" as an example for using osu! as a source seems to go against the earlier guideline talking about "songs specifically made for events", and people in the thread seemed to prefer using the particular tournament in the source field as it's consistent with other non-osu-tournament events. are both allowed under these rules? if tournament names are to be preferred, i feel like it may make more sense to give an example such as "main menu theme" for osu! originals instead.

apollo-dw:
it'd be worth mentioning explicitly that songs created for osu! (like for tournaments or contests, including community-ran) can use "osu!" as it's source, according to peppy: https://github.com/ppy/osu-wiki/commit/5153b14b756d9d9d207b9559b6d3916ab549e6ae#commitcomment-56327830

NoffyNoffeh:
@apollo-dw this is already in the RC so it wasn't in the PR, look at line 86 , unless you meant something else. 

@Purplegaze I'm not sure I understand what you mean but: The rule says when multiple valid sources exist you can choose from those, so osu! and the event would each be valid and they don't conflict.

wiki/Ranking_Criteria/en.md
```diff
@@ -78,7 +78,10 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
 - **The artists of a song must be traceable to existing people.** If there is no existing person recorded to be the artist, then `Unknown Artist` is to be used. A song's sole artist cannot be a fictional character or program unless it is the artist's alias.
-- **You must use the Source field if the song comes from or is directly tied to another media such as a video game, movie, series, web site, community etc.** If the song was featured or tied to a media after it was released, the source field should not be used. If a song has multiple potential sources, any option is valid. For remixes, arrangements, or covers based upon the original song, the source(s) of the original song would apply in the same way.
+- **You must use the Source field if the song comes from or is directly tied to another media such as a video game, movie, series, etc.** If the song was featured or tied to a media after it was released, the source field is optional. If a song has multiple potential sources, any options are valid.
```
peppy, line 81: why is this reverted to "optional"? i think this is wrong?

wiki/Ranking_Criteria/en.md
```diff
@@ -78,7 +78,10 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
 - **The artists of a song must be traceable to existing people.** If there is no existing person recorded to be the artist, then `Unknown Artist` is to be used. A song's sole artist cannot be a fictional character or program unless it is the artist's alias.
-- **You must use the Source field if the song comes from or is directly tied to another media such as a video game, movie, series, web site, community etc.** If the song was featured or tied to a media after it was released, the source field should not be used. If a song has multiple potential sources, any option is valid. For remixes, arrangements, or covers based upon the original song, the source(s) of the original song would apply in the same way.
+- **You must use the Source field if the song comes from or is directly tied to another media such as a video game, movie, series, etc.** If the song was featured or tied to a media after it was released, the source field is optional. If a song has multiple potential sources, any options are valid. 
+  - For remixes, arrangements, or covers based upon the original song, the source(s) of the original song would apply in the same way.
```
peppy, line 82:
```suggestion
  - For remixes, arrangements, or covers based upon the original song, the source of the original song should be used.
```
is this i'm the intended meaning?

wiki/Ranking_Criteria/en.md
```diff
@@ -78,7 +78,10 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
+  - Songs specifically made for events such as contest entries or event theme songs may use the event name as a source.
```
peppy, line 83: why not just add the word "event" to the first line, alongside "video game"?

peppy (changes requested):

wiki/Ranking_Criteria/en.md
NoffyNoffeh: many older songs are obscure and hard to figure out if the media or song came first, and many songs that were tied to media later become prominently known for that media even if it wasn't the initial source. Disallowing it would make researching source much more difficult for little benefit. Similar opinions on this are expressed in the thread

wiki/Ranking_Criteria/en.md
NoffyNoffeh: will do

wiki/Ranking_Criteria/en.md
```diff
@@ -78,7 +78,9 @@ Changes to the rules and guidelines in this document are proposed and discussed
...
 - **The artists of a song must be traceable to existing people.** If there is no existing person recorded to be the artist, then `Unknown Artist` is to be used. A song's sole artist cannot be a fictional character or program unless it is the artist's alias.
-- **You must use the Source field if the song comes from or is directly tied to another media such as a video game, movie, series, web site, community etc.** If the song was featured or tied to a media after it was released, the source field should not be used. If a song has multiple potential sources, any option is valid. For remixes, arrangements, or covers based upon the original song, the source(s) of the original song would apply in the same way.
+- **You must use the Source field if the song comes from or is directly tied to another media such as a video game, movie, series, event, etc.** If the song was featured or tied to a media after it was released, the source field is optional. If a song has multiple potential sources, any options are valid.
```
peppy, line 81: Am I the only one that finds it weird that we're adding an "optional" clause under "rules"?

Rewording since I still disagree with the version in this PR:

```suggestion
- **You must use the Source field if the song comes from or is directly tied to another media such as a video game, movie, series, event, etc.** This also applies if the song was featured or tied to a media after it was released, but is primarily known due to that source. If a song has multiple potential sources, any option is valid. 
```

peppy (changes requested):

wiki/Ranking_Criteria/en.md
NoffyNoffeh: Oh, I understand better what your issue with the wording was now. I'll apply this change as it is fair and fits to what the rule was initially going for.

peppy (approved):

PR #6162, opened 2021-09-25
Update `Countries that participated in osu! tournaments` with `MWC 4K 2021` results

Bazzadwarf:


wiki/Tournaments/Countries_that_participated_in_osu!_tournaments/en.md
```diff
@@ -258,49 +258,54 @@ This article lists countries that have participated in official [osu!](/wiki/Gam
 
 ### ![][osu!mania] osu!mania World Cup
```
Bazzadwarf, line 259: This section should probably be considered for splitting up into 4K and 7K sub-sections as the table is *rather* big.

TicClick (approved):
looks ok from what I see. maintaining this must be painful, but as long as it's updated after every tourney..

wiki/Tournaments/Countries_that_participated_in_osu!_tournaments/en.md
TicClick: maybe the whole article should

PR #6196, opened 2021-10-06
Fix autogenerated identifiers of the headings with images

TicClick:
otherwise the alt text makes its way both into the section's identifier and ToC:

image


TPGPL (approved):
pretty sure this was not a case before 🤔

PR #6223, opened 2021-10-13
Update en.md

Rice-wine-o:
- Rule #4 is made redundant by rule #3.
- The word 'douche' in this context lacks class.


Ephemeralis:
Rule four serves a sacred, ineffable purpose. It is not quite the same as rule three and serves a distinct subset of people that we encounter unfortunately all too often.

As an aside, changes to the community rules are generally only made by the osu!team, so this PR is a bit out of scope. Thanks for your contribution all the same - there's likely other non-critical articles you could review with issues if you're still interested in helping out.

PR #6226, opened 2021-10-14
[BG] Improvement of `Main_Page`

YDu8VchU:
Thanks for the good reviews and tips. I added the latest url updates, check if they are right.


TicClick:
briefly checked for formatting/structure, looks ok to me with one note: names like `Project Loved`, or `Beatmap Spotlights`, or `Featured Artist`, are generally not translated, in the same way you wouldn't translate `osu!`, unless it's an actual phrase used in the community. for example, in Russian translations we provide a short explanation of what the term means, but only work with the original name throuhgouht the article

if that isn't an issue for you, and if you don't think there will be any reviewers, I can merge this

TicClick:
please add changes from https://github.com/ppy/osu-wiki/pull/6244/files#diff-29991329496d7bd10fac16a36067e0882737f40f29fc936710e449089f8385d4 and remove `outdated`/`outdated_since` markers, and I'll merge this

YDu8VchU:
I tested to add some tooltips over the [urls](http://example "tooltip example"). Are you approve them? If not i will delete them.

TicClick:
tooltips are not needed for the wiki links (and for regular ones either), so no

YDu8VchU:
Ok, im removing them and the file is done.

TicClick (approved):

PR #6238, opened 2021-10-16
Revamp osu!mania Spread requirements from `Ranking criteria/osu!mania`

Kominaru:
After several weeks of data collection, opinion surveying and discussion among the BNG/NAT and with the community, we have arrived at a [consensual proposal](https://osu.ppy.sh/community/forums/topics/1440933) for revamping how beatmap spreads are defined in the osu!mania game-mode. 

To avoid cluttering the General RC page we've decided to define it in the mania-specific RC and redirect readers to it. Please do feel free to suggest more "wiki-like" ways to structure the necessary information of the revamp if this one isn't satisfactory.


Kominaru:
Applied some changes after clayton's review, should be ready to go now

PR #6239, opened 2021-10-16
Add `Modding/Bubble`

TicClick:
part of #3163


TicClick:
thanks, that helps. all comments addressed

PR #6277, opened 2021-10-27
Update `Projects`

daud-ahn:
Proposed deletion is commented out along with the reason.


cl8n:
not sure what you're referring to, for danser I changed it to match the formatting of other projects (name links to a forum post if it exists, other links are listed below)

daud-ahn:
> not sure what you're referring to, for danser I changed it to match the formatting of other projects (name links to a forum post if it exists, other links are listed below)

right my bad, i get it now

PR #6289, opened 2021-10-29
Improve explanation of image formatting in `Article styling criteria/Formatting`

Samaoo:
- edit examples for `Images`
- add explanation for text between square brackets and quotes


wiki/Article_styling_criteria/Formatting/en.md
```diff
@@ -613,19 +613,23 @@ There are two types of image links: inline and reference. Examples:
 **Inline style:**
 
 ```markdown
-![](/wiki/shared/flag/AU.gif)
+![Australian flag](/wiki/shared/flag/AU.gif "Australia")
```
cl8n, line 616: flags are kinda bad examples at this point because they're supporting icons that don't need alt text.

I think a good example would be the placement crowns from tournaments, since they showcase the difference between alt & title text

wiki/Article_styling_criteria/Formatting/en.md
```diff
@@ -613,19 +613,23 @@ There are two types of image links: inline and reference. Examples:
...
+![Australian flag][flag_AU]
 
 [flag_AU]: /wiki/shared/flag/AU.gif "Australia"
 ```
 
 Images should use the inline linking style. References to reference links must be placed at the bottom of the article.
 
+The (first pair of) square brackets should contain a description of the image. It will be used by screen readers or when the image fails to load.
+
+The text between the quotation marks will be displayed below the image and when hovering the mouse cursor over the image.
```
cl8n, line 631:
```suggestion
The text in the first pair of square brackets (*[alternative text](#alt-text)*) is optional and should describe the image literally. It is used by screen readers or when the image fails to load.

The text in the quotation marks (*title text*) is optional and should give additional context to the image or indicate its meaning. It is displayed when hovering over the image and set as the [caption](#captions) if applicable.
```

cl8n, line 631: not sure if it matters but this feels a little repetitive with the `Alt text` and `Captions` sections. maybe this should stick to explaining the formatting only, and suggested usage goes in the other sections? maybe combine all the sections into this one? maybe it's fine as-is?

I'll just wait for another review...

cl8n (changes requested):

cl8n:
also try to use commit names that describe what you're changing, rather than just the file name.

wiki/Article_styling_criteria/Formatting/en.md
TicClick: imo merging all three sections into a single one makes more sense. it would also allow for a shorter explanation, such as:

```markdown
Images are blah blah and have the following format:

![Alt text](/wiki/path/to/image.png "Title text")

Where:

- `Alt text` is ... . It is optional and does ... and ... .
- `Title text` is ... . It is optional and does ... and ... . It will be rendered as the image's caption if ... and ... .
```

TicClick (changes requested):

wiki/Article_styling_criteria/Formatting/en.md
TicClick: +1

cl8n:
will update this one myself, noticing other things in this section I want to write better

wiki/Article_styling_criteria/Formatting/en.md
cl8n: didn't do exactly this but I think I made it read well from top to bottom now

TicClick (approved):

TicClick:
that's better, thank you both. @cl8n I am fine with the changes, wanna merge?

TPGPL (dismissed):
lgtm

cl8n (approved):

PR #6295, opened 2021-10-30
Fixed incorrect year in "Featured artist showcase beatmaps"

mousesalt:
:penguin: nobody realized it


pishifat (approved):
time stopped in march 2020 so you're the crazy one here

PR #6297, opened 2021-10-30
Update to reflect changes in timezones due to summer time ending

jthistle:
On 31st October (the last Sunday in the month) all countries in Europe turn the clocks back an hour. This means that some of the stated timezones and local times are incorrect. Daylight saving time also ends in Mexico for most regions. The one other country in the Sunday list that uses DST is Chile, but it ended there in early September.

Also, next Sunday, DST will end in Canada.

(Sorry for such an inconsequential PR, but when I noticed I couldn't help myself :))


peppy (dismissed):

peppy:
May be worth informing the teams to check their time zones in addition to updating the wiki. I can imagine some may get confused.

gagahpangeran:
I thought UTC is not adjusted for daylight saving time :thinking: 

From https://en.wikipedia.org/wiki/Coordinated_Universal_Time

> Coordinated Universal Time or UTC is the primary time standard by which the world regulates clocks and time. It is within about 1 second of mean solar time at 0° longitude and is not adjusted for daylight saving time. It is effectively a successor to Greenwich Mean Time (GMT).

jthistle:
>I thought UTC is not adjusted for daylight saving time 

You are correct. But local times are - these are the only times these PR updates.

jthistle:
@ppy apologies, had to rebase, requires re-approval

PR #6302, opened 2021-10-30
Vietnamese translate for Hybrid play style

Jackie-Potato:


DoryanR:
Why delete the English file? It must be kept, otherwise the article will no longer be available.

Jackie-Potato:
> Why delete the English file? It must be kept, otherwise the article will no longer be available.

oh sorry um, this is my first time doing something like but um how can i do it

DoryanR:
In your case I can advise you to take the content of the English file, to recreate it and to paste its contents on the freshly created file.
This is the simplest solution, which comes to my mind immediately.

Jackie-Potato:
> In your case I can advise you to take the content of the English file, to recreate it and to paste its contents on the freshly created file. This is the simplest solution, which comes to my mind immediately.

thanks

PR #6303, opened 2021-10-30
[VI] Add `Play style/Hybrid`

Jackie-Potato:
Re-do because last time deleted the en.md instead of duplicate.


TicClick (approved):

PR #6305, opened 2021-10-30
Improve tags for game mod articles 

TicClick:
helps you find the articles if you search for `hardrock` and the likes.
in addition, there's no point in repeating the article's name in the tags, as titles are matched against the query anyway

### DIFF
--- a/wiki/Game_modifier/No_Fail/ko.md
+++ b/wiki/Game_modifier/No_Fail/ko.md
@@ -3,7 +3,6 @@ stub: true
 outdated: true
 outdated_since: 77180e4fde6272da6bfab8e8337dcb379b598b64
 tags:
-  - 노페일 모드
   - no fail
   - nofail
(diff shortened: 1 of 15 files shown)


wiki/Game_modifier/No_Fail/ko.md
```diff
@@ -5,6 +5,7 @@ outdated_since: 77180e4fde6272da6bfab8e8337dcb379b598b64
 tags:
   - 노페일 모드
```
cl8n, line 6: should be removed as well?

cl8n (changes requested):

wiki/Game_modifier/No_Fail/ko.md
TicClick: yes

cl8n (approved):

PR #6336, opened 2021-11-07
Fix wrong flag in LVD Team page & add missing translations in ALM page locales

zeusminus:
- Fix wrong flag on Project Loved Team page
- Add missing translations for ztrot's role on Alumni page


TicClick (approved):
thanks, this looks better

PR #6337, opened 2021-11-08
Add Performance Points & Star Rating Updates article 

Ephemeralis:
Some changes still required, just putting it up into PR to facilitate further reviews from wiki contributors and others.


Ephemeralis:
I've attempted to shuffle things around a bit to improve general reader usability with the last commit. Some sanity checking to ensure the technical details remain correct would be appreciated.

The osu!catch section is still a bit messier than I'd like.

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+- [Instant spinners no longer give an insane amount of strain](https://github.com/ppy/osu/pull/14217), thanks to [**StanR**](https://osu.ppy.sh/users/7217455). This fixes maps [such as this one](https://osu.ppy.sh/beatmapsets/814850#osu/1901200), which abused this. 
+- Joz spotted that the initial strain for skills started at 1, not 0. [This has been fixed](https://github.com/ppy/osu/pull/15014).
+- In preparation for osu!lazer, difficulty calculation for [the Blinds mod](https://github.com/ppy/osu/pull/14921) and [the Relax mod](https://github.com/ppy/osu/pull/14942) has been added by [**Apo11o**](https://osu.ppy.sh/users/9558549).
+  - These could be potentially enabled in the future once osu!lazer leaderboard submissions go live, since the concept of "ranked mods" will be no more.
+
+---
+
+Massive thanks to the star rating developers for all the changes above, the community members for providing continuous feedback, and [Apo11o](https://osu.ppy.sh/users/9558549), [Xexxar](https://osu.ppy.sh/users/2773526) and [bastoo0](https://osu.ppy.sh/users/4864877) for writing this post.
```
apollo-dw, line 130: May wanna remove bastoo0 since his contributions were related to the osu!catch section.

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+Here are a few examples:
+
+- [DragonForce - Through the Fire and Flames (Ponoyoshi) [Myth]](https://osu.ppy.sh/beatmapsets/382400#osu/1001682) 6.1* -> 6.32*
+- [Panda Eyes & Teminite - Immortal Flame (feat. Anna Yvette) (Sotarks) [Eternal Blaze]](https://osu.ppy.sh/beatmapsets/703957#osu/1489207) 7.05* -> 6.94*
+
+As for the formula change itself, you can compare and contrast between these:
+
+Old formula: ![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/old-sr-white.png)
+New formula: ![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/new-sr-white.png)
```
cl8n, line 83: if the images are to be kept in the article they should be figures instead of inline. "\_ formula" can be the captions

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+- [Instant spinners no longer give an insane amount of strain](https://github.com/ppy/osu/pull/14217), thanks to [**StanR**](https://osu.ppy.sh/users/7217455). This fixes maps [such as this one](https://osu.ppy.sh/beatmapsets/814850#osu/1901200), which abused this. 
+- Joz spotted that the initial strain for skills started at 1, not 0. [This has been fixed](https://github.com/ppy/osu/pull/15014).
+- In preparation for osu!lazer, difficulty calculation for [the Blinds mod](https://github.com/ppy/osu/pull/14921) and [the Relax mod](https://github.com/ppy/osu/pull/14942) has been added by [**Apo11o**](https://osu.ppy.sh/users/9558549).
+  - These could be potentially enabled in the future once osu!lazer leaderboard submissions go live, since the concept of "ranked mods" will be no more.
+
+---
+
+Massive thanks to the star rating developers for all the changes above, the community members for providing continuous feedback, plus [Apo11o](https://osu.ppy.sh/users/9558549) and [Xexxar](https://osu.ppy.sh/users/2773526) for writing this post.
+
+—osu!team
```
cl8n, line 132:
```suggestion
Massive thanks to the star rating developers for all the changes above, and the community members for providing continuous feedback.

—Apo11o and Xexxar
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,171 @@
+---
+layout: post
+title: "Performance Points & Star Rating Updates"
+date: 2021-11-08 10:00:00 +0000
+---
+
+![](/wiki/shared/news/2021-07-27-performance-points-star-rating-updates/star-rating.jpg)
+
+Thanks to the efforts of various star rating contributors, we are excited to announce that a new series of difficulty calculation changes will be going live shortly. This newspost will go into the what, the who, the when and the how, so read on to find out more!
```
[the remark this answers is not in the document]
cl8n: also for "star rating developers" on L130

cl8n (changes requested):

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+We will be covering only changes made to the **osu!** game mode in this post.
+
+## osu! changes
+
+As a reminder of the various core values in the difficulty calculation algorithm:
+
+- **Aim**, which considers how difficult it is to consistently hit notes back-to-back during play (affected by things like approach rate, jumps, Flashlight, Hidden, Hard Rock and more)
+- **Speed**, which considers how quickly new elements are presented to the player during the course of the map (affected by things like the number of hit objects over time, BPM, Double Time and Half Time)
+- **Accuracy**, which considers the player's ability to stay on time rhythmically and how difficult this is to achieve (affected by things like Hidden, Hard Rock, Flashlight and more)
+- **Strain**, which considers how often and how long a player has to endure intense or technically demanding patterns during play (affected by things like streams, fast jumps and more)
```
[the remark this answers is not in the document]
bdach: This is mostly an introductory recap. [Was present on past posts, too.](https://osu.ppy.sh/home/news/2021-07-27-performance-points-star-rating-updates)

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+- [DragonForce - Through the Fire and Flames (Ponoyoshi) [Myth]](https://osu.ppy.sh/beatmapsets/382400#osu/1001682) 6.1* -> 6.32*
+- [Panda Eyes & Teminite - Immortal Flame (feat. Anna Yvette) (Sotarks) [Eternal Blaze]](https://osu.ppy.sh/beatmapsets/703957#osu/1489207) 7.05* -> 6.94*
+
+As for the formula change itself, you can compare and contrast between these:
+
+Old formula: ![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/old-sr-white.png)
+New formula: ![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/new-sr-white.png)
+
+Interested in what each term means? [**This document**](https://docs.google.com/document/d/10DZGYYSsT_yjz2Mtp6yIJld0Rqx4E-vVHupCqiM4TNI/edit?usp=sharing) written up by [Naitsirk](https://osu.ppy.sh/users/8202998) goes into the nitty gritty details regarding the terms and the derivation itself, for those more mathematically inclined.
```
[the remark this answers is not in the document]
peppy: Also found this weird. If the formulas can be replaced with a layman's version that the average user might be able to parse then that would be best.

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+## osu! changes
+
+As a reminder of the various core values in the difficulty calculation algorithm:
+
+- **Aim**, which considers how difficult it is to consistently hit notes back-to-back during play (affected by things like approach rate, jumps, Flashlight, Hidden, Hard Rock and more)
+- **Speed**, which considers how quickly new elements are presented to the player during the course of the map (affected by things like the number of hit objects over time, BPM, Double Time and Half Time)
+- **Accuracy**, which considers the player's ability to stay on time rhythmically and how difficult this is to achieve (affected by things like Hidden, Hard Rock, Flashlight and more)
+- **Strain**, which considers how often and how long a player has to endure intense or technically demanding patterns during play (affected by things like streams, fast jumps and more)
+
+To help facilitate faster review and uptake of incoming changes, a Performance Points Committee has been formed, consisting of [**StanR**](https://osu.ppy.sh/users/7217455), [**emu1337**](https://osu.ppy.sh/users/2185987), [**MBmasher**](https://osu.ppy.sh/users/4498616) and [**Apo11o**](https://osu.ppy.sh/users/9558549).
```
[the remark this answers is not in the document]
bdach: The first public reference I can find about it is [this discord message](https://discord.com/channels/188630481301012481/380598781432823815/883167719942414366).

I believe the rationale for forming it is just to ensure consensus and coherent movement with the regard to how changes are made, to avoid back-and-forths.

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+Good luck setting new, impressive top plays that showcase your skills!
+
+### Total star rating adjustments
+
+In a nutshell, osu! difficulty calculation processes a map and produces skill rating values. These skill values are then manipulated in two different ways to produce two different numbers that you’ll be familiar with: star rating and performance points. 
+
+The correlation between these two numbers has now been improved, following the mathematical derivation work from [**Naitsirk**](https://osu.ppy.sh/users/8202998) in [this pull request](https://github.com/ppy/osu/pull/13986). 
+
+To summarize these changes briefly, maps which focus on one skill will now receive lower star rating than ones which focus on both. **These changes do not affect performance points.**
```
bdach, line 73: While true, this feels like it's beside the point of the change. The main point I'd emphasise here is that the old SR didn't really care about what the *composition* of a star rating was - as in how large the individual skill values are; it just summed the constituent parts and left it at that, while the new one combines them in a way analogous to how pp does. The statement that one-skill maps will be nerfed is a corollary of the above.

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+As for the formula change itself, you can compare and contrast between these:
+
+Old formula: ![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/old-sr-white.png)
+New formula: ![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/new-sr-white.png)
+
+Interested in what each term means? [**This document**](https://docs.google.com/document/d/10DZGYYSsT_yjz2Mtp6yIJld0Rqx4E-vVHupCqiM4TNI/edit?usp=sharing) written up by [Naitsirk](https://osu.ppy.sh/users/8202998) goes into the nitty gritty details regarding the terms and the derivation itself, for those more mathematically inclined.
+
+### New Flashlight difficulty skill
+
+Flashlight in difficulty calculation is now revamped, as it moves from an object count-based multiplier on the Aim skill to a brand new skill of its own.
```
bdach, line 89:
```suggestion
Flashlight in difficulty calculation is now revamped, as it moves from an object count-based multiplier applied to the Aim skill to a brand new skill of its own. 
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+Interested in what each term means? [**This document**](https://docs.google.com/document/d/10DZGYYSsT_yjz2Mtp6yIJld0Rqx4E-vVHupCqiM4TNI/edit?usp=sharing) written up by [Naitsirk](https://osu.ppy.sh/users/8202998) goes into the nitty gritty details regarding the terms and the derivation itself, for those more mathematically inclined.
+
+### New Flashlight difficulty skill
+
+Flashlight in difficulty calculation is now revamped, as it moves from an object count-based multiplier on the Aim skill to a brand new skill of its own. 
+
+The difficulty that Flashlight adds to a map is now measured much more accurately, thanks to [**MBmasher**](https://osu.ppy.sh/users/4498616)’s change in [this pull request](https://github.com/ppy/osu/pull/14217).
+
+Previously, the object count-based method posed a number of issues. Flow aim (measured when playing streams) is mostly represented by the speed skill, not aim. Additionally, the aim skill (surprisingly) only measures the aim difficulty, not the flashlight difficulty. To put this simply, there will be types of jumps that are hard to aim and quite easy to read with flashlight, and vice versa.
```
bdach, line 93:
```suggestion
Previously, the object count-based method posed a number of issues. Flow aim (used for hitting streams) is mostly represented by the speed skill, not aim. Additionally, the aim skill, unsurprisingly, only measures the aim difficulty, not the flashlight difficulty. To put this simply, there will be types of jumps that are hard to aim but quite easy to read with flashlight, and vice versa.
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+It is also worth noting that the Hidden mod applies a 1.3x multiplier on the Flashlight skill. The visibility of hitobjects are decreased significantly due to fading of circles and the absence of approach circles.
+
+Find yourself curious about the details? [MBmasher](https://osu.ppy.sh/users/4498616)’s write up about the implementation will very likely scratch that itch, so feel free to [**indulge in some Flashlight theory**](https://docs.google.com/document/d/11FBEEPXK8yI2d5uTZuebJC3S9nVUKgMD8WdvblERyYA/edit).
+
+### Speed hard cap removal
+
+The speed hard cap in osu! has been removed by [**emu1337**](https://osu.ppy.sh/users/2185987) and [**Apo11o**](https://osu.ppy.sh/users/9558549) in [this pull request](https://github.com/ppy/osu/pull/14617), allowing 300+ BPM streams to be weighed accurately. 
+
+The speed cap was previously set to 300 BPM to prevent abuse cases, which means that calculating higher BPMs requires consideration into those cases. Two factors are implemented with regard to this:
```
bdach, line 109: kinda reads weird. maybe this works better?

```suggestion
The speed cap was previously set to 300 BPM to prevent abuse cases, which means that calculating difficulty on beatmaps with higher BPM requires special consideration. Two adjustments were made to account for that:
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Speed is nerfed on doubles that are doubletappable. If the time since the last object (deltatime) for the current object is smaller than the 300 hit window, we instead interpolate between that and the deltatime for the previous object.
```
bdach, line 111: I wouldn't bring "deltatime" into this at all, it's not easily understandable to readers. The interpolation detail also doesn't really matter. I would refer to hit windows instead as that seems more reasonable.

news/2021-11-08-performance-points-star-rating-updates.md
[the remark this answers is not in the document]
bdach: Can only agree with the predecessors. I'd agree with snipping the formulas entirely. This plot from the google doc may get the point across better, I think (but only with explanation as to what is being shown):

2021-11-08-082811_808x563_scrot

I can theme it better using my jupyter notebooks I have stashed from the past newsposts if you think that'll work ok.

news/2021-11-08-performance-points-star-rating-updates.md
bdach: These posts are generally always credited to `osu!team` anyhow.

peppy: Yeah, if anything the previous version read better (leaving `osu!team` as the sign-off). Xexxar is already mentioned above so could probably be left to just Apo11o, or omit completely as this post is going to be a community effort once all these reviews are applied.

news/2021-11-08-performance-points-star-rating-updates.md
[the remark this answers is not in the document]
peppy: I'd just use the term "community contributors" or "developers in the community" or something like that.

news/2021-11-08-performance-points-star-rating-updates.md
cl8n: I commented to point out that it literally says who wrote the post, followed by a sign-off that is different from who wrote the post. I don't have an opinion who counts as the author, just that it shouldn't contradict itself

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,130 @@
...
+
+Flashlight in difficulty calculation is now revamped, as it moves from an object count-based multiplier applied to the Aim skill to a brand new skill of its own. 
+
+The difficulty that Flashlight adds to a map is now measured much more accurately, thanks to [**MBmasher**](https://osu.ppy.sh/users/4498616)'s change in [this pull request](https://github.com/ppy/osu/pull/14217).
+
+Previously, the object count-based method posed a number of issues. Flow aim (used for hitting streams) is mostly represented by the speed skill, not aim. Additionally, the aim skill, unsurprisingly, only measures the aim difficulty, not the flashlight difficulty. To put this simply, there will be types of jumps that are hard to aim but quite easy to read with flashlight, and vice versa.
+
+For example, let’s consider a familiar type of jump pattern: a back and forth. This is an example of a jump that can have high aim difficulty, but relatively low flashlight difficulty. The distance between the first object and the third object in a back and forth is typically small enough for it to be within the flashlight radius, thus allowing a player to read the position of that circle very easily.
```
Walavouchey, line 93:
```suggestion
For example, let's consider a familiar type of jump pattern: a back and forth. This is an example of a jump that can have high aim difficulty, but relatively low Flashlight difficulty. The distance between the first object and the third object in a back and forth is typically small enough for it to be within the Flashlight radius, thus allowing a player to read the position of that circle very easily.
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,130 @@
...
+
+The speed cap was previously set to 300 BPM to prevent abuse cases, which means that calculating difficulty on beatmaps with higher BPM requires special consideration. Two adjustments were made to account for that:
+
+- Speed is nerfed on doubles that are doubletappable. If the time since the last object (deltatime) for the current object is smaller than the 300 hit window, we instead interpolate between that and the deltatime for the previous object.
+- Speed is nerfed if successive objects are faster than the 300 hit window. This is intended to target cheesing of faster bursts and streams if the hit window is way larger than the time between notes.
+
+The intention behind this change is to accommodate for future improvement, as there are a handful of players who have proven themselves to be competent at such high speeds. 
+
+As an added bonus, the star ratings of absurdly fast maps are now measured more accurately, which means you can gawk as much as you’d like at the now even-higher star ratings on them.
```
Walavouchey, line 114:
```suggestion
As an added bonus, the star ratings of absurdly fast maps are now measured more accurately, which means you can gawk as much as you'd like at the now even-higher star ratings on them.
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Minor changes
+
+Alongside all those big changes above, there are some changes on the smaller side of things:
+
+- [Sliderbreaks are now approximated into the total miss count](https://github.com/ppy/osu/pull/15086) in performance calculation, following [**StanR**](https://osu.ppy.sh/users/7217455)’s pull request. This is guessed by comparing the achieved combo to the maximum combo possible on a map.
```
Walavouchey, line 120:
```suggestion
- [Sliderbreaks are now approximated into the total miss count](https://github.com/ppy/osu/pull/15086) in performance calculation, following [**StanR**](https://osu.ppy.sh/users/7217455)'s pull request. This is guessed by comparing the achieved combo to the maximum combo possible on a map. 
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,130 @@
...
+
+For example, let’s consider a familiar type of jump pattern: a back and forth. This is an example of a jump that can have high aim difficulty, but relatively low flashlight difficulty. The distance between the first object and the third object in a back and forth is typically small enough for it to be within the flashlight radius, thus allowing a player to read the position of that circle very easily.
+
+![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/visible-fl-area.png)
+
+The creation of a Flashlight skill makes it possible to have it contribute to star rating. [This has been enabled](https://github.com/ppy/osu/pull/14753) following the results of the recent osu! performance points and star rating survey, which revealed that players desire star rating to be a metric of difficulty, and that they want subjective skills included. Ask and thou shall receive!
+
+It is also worth noting that the Hidden mod now applies a 1.3x multiplier on the Flashlight skill. The visibility of hitobjects are decreased significantly due to fading of circles and the absence of approach circles.
```
Walavouchey, line 99:
```suggestion
It is also worth noting that the Hidden mod now applies a 1.3x multiplier on the Flashlight skill. The visibility of hit objects are decreased significantly due to the fading of circles and the absence of approach circles.
```

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,130 @@
...
+
+Interested in what each term means? [**This document**](https://docs.google.com/document/d/10DZGYYSsT_yjz2Mtp6yIJld0Rqx4E-vVHupCqiM4TNI/edit?usp=sharing) written up by [Naitsirk](https://osu.ppy.sh/users/8202998) goes into the nitty gritty details regarding the terms and the derivation itself, for those more mathematically inclined.
+
+### New Flashlight difficulty skill
+
+Flashlight in difficulty calculation is now revamped, as it moves from an object count-based multiplier applied to the Aim skill to a brand new skill of its own. 
+
+The difficulty that Flashlight adds to a map is now measured much more accurately, thanks to [**MBmasher**](https://osu.ppy.sh/users/4498616)'s change in [this pull request](https://github.com/ppy/osu/pull/14217).
+
+Previously, the object count-based method posed a number of issues. Flow aim (used for hitting streams) is mostly represented by the speed skill, not aim. Additionally, the aim skill, unsurprisingly, only measures the aim difficulty, not the flashlight difficulty. To put this simply, there will be types of jumps that are hard to aim but quite easy to read with flashlight, and vice versa.
```
Walavouchey, line 91:
```suggestion
Previously, the object count-based method posed a number of issues. Flow aim (used for hitting streams) is mostly represented by the speed skill, not aim. Additionally, the aim skill, unsurprisingly, only measures the aim difficulty, not the Flashlight difficulty. To put this simply, there will be types of jumps that are hard to aim but quite easy to read with Flashlight, and vice versa.
```

news/2021-11-08-performance-points-star-rating-updates.md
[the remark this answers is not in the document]
Ephemeralis: can try it if you want, if you think it's descriptive enough?

news/2021-11-08-performance-points-star-rating-updates.md
Ephemeralis: attempted a reword to resolve/clarify this, not sure i have, am very open to suggestions on how to word this

news/2021-11-08-performance-points-star-rating-updates.md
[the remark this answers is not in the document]
bdach: Yeah I was just about to start preparing something to resolve my unresolved reviews. Sorry for the wait.

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+As an added bonus, the star ratings of absurdly fast maps are now measured more accurately, which means you can gawk as much as you'd like at the now even-higher star ratings on them.
+
+### Minor changes
+
+Alongside all those big changes above, there are some changes on the smaller side of things:
+
+- [Sliderbreaks are now approximated into the total miss count](https://github.com/ppy/osu/pull/15086) in performance calculation, following [**StanR**](https://osu.ppy.sh/users/7217455)'s pull request. This is guessed by comparing the achieved combo to the maximum combo possible on a map. 
+- [Instant spinners no longer give an insane amount of strain](https://github.com/ppy/osu/pull/14217), thanks to [**StanR**](https://osu.ppy.sh/users/7217455). This fixes maps [such as this one](https://osu.ppy.sh/beatmapsets/814850#osu/1901200), which abused this. 
+- Joz spotted that the initial strain for skills started at 1, not 0. [This has been fixed](https://github.com/ppy/osu/pull/15014).
```
smoogipoo, line 124:
```suggestion
- [**Joz**](https://osu.ppy.sh/users/10644596) spotted that the initial strain for skills started at 1, not 0. [This has been fixed](https://github.com/ppy/osu/pull/15014).
```

smoogipoo (changes requested):

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+Find yourself curious about the details? [MBmasher](https://osu.ppy.sh/users/4498616)'s write-up about the implementation will very likely scratch that itch, so feel free to [**indulge in some Flashlight theory**](https://docs.google.com/document/d/11FBEEPXK8yI2d5uTZuebJC3S9nVUKgMD8WdvblERyYA/edit).
+
+### Speed hard cap removal
+
+The speed hard cap in osu! has been removed by [**emu1337**](https://osu.ppy.sh/users/2185987) and [**Apo11o**](https://osu.ppy.sh/users/9558549) in [this pull request](https://github.com/ppy/osu/pull/14617), allowing 300+ BPM streams to be weighed accurately. 
+
+The speed cap was previously set to 300 BPM to prevent abuse cases, which means that calculating difficulty on beatmaps with higher BPM requires special consideration. Two adjustments were made to account for that:
+
+- Speed is nerfed on doubles that are double-tappable. If the time since the last object for the current object is smaller than the hit window for a 300, we instead take the difference between that object and the hit window of the last one.
```
Walavouchey, line 111: Ah, no, I have no idea what this wording is trying to convey.

[The change in question for reference](https://github.com/ppy/osu/pull/14617/files#diff-2877a1e4555d4b301a40010c0693e8a8ce3f9bcd8759f1bc22b887d62a03b459R55-R56):

```cs
double speedWindowRatio = strainTime / greatWindowFull;

// Aim to nerf cheesy rhythms (Very fast consecutive doubles with large deltatimes between)
if (osuPrevious != null && strainTime < greatWindowFull && osuPrevious.StrainTime > strainTime)
                strainTime = Interpolation.Lerp(osuPrevious.StrainTime, strainTime, speedWindowRatio);
```

By inspection of the code, for a tight double, the time difference between its notes will be considered as if it was closer to the time difference of the gap before the double.

For a double with notes 25 ms apart (the minimum in calculations) and hit window of 50 ms (about AR 9.6 I think), the ratio of this weighted sum would be 0.5. For lower AR, doubles are nerfed harder and vice versa.

```suggestion
- Speed is nerfed on doubles that are double-tappable. If the time gap between two notes in a double is smaller than the hit window for a 300, the speed skill calculation will consider it as if it was closer to the time gap before the double.
```

... or whatever you find most illustrative.

Walavouchey (changes requested):

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+With [a new change](https://github.com/ppy/osu/pull/14395) submitted by [**Xexxar**](https://osu.ppy.sh/users/2773526), rhythm complexity will now play a more important role in determining the star rating of a map. 
+
+This change introduces an algorithmic assessment of the complexity of note patterns and will introduce a more accurate assessment for maps that feature atypical combinations of triples, doubles, quints and much more!
+
+Consider these two rhythms:
+
+![](/wiki/shared/news/2021-11-08-performance-points-star-rating-updates/rhythm-complexity.png)
+
+Before this difficulty calculation change, the first pattern would have always had greater strain than the second.  In reality, the second can be as demanding or even harder than the first depending on the map. With these new changes, the algorithm now more accurately reflects this.
```
Walavouchey, line 55:
```suggestion
Before this difficulty calculation change, the first pattern would have always had greater strain than the second. In reality, the second can be as demanding or even harder than the first depending on the map. With these new changes, the algorithm now more accurately reflects this.
```

news/2021-11-08-performance-points-star-rating-updates.md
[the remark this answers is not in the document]
bdach: Ok, so after some sanity loss due to failing to algebra properly, here's the best graph I can give for use instead of the formulas above. I'd just replace them wholesale with this graph.

sr_pp_correlation

So the way this graph works, is that the axes are ratings for individual skills. The lines represent the combinations of aim & speed skills that a map needs to have to get a final rating of 7*. I don't know whether this is understandable, but I have no idea as to what could be better. Please advise whether you would like me to expand the description further.

Unfortunately I don't think using the ones from the original doc is viable as they are kind of confusing and arbitrary (the axes there don't really represent anything tangible). So I don't have any good idea how to graphically present the "more closely related to pp" part well.

I'd say the link to the doc can stay if someone _really_ wants to look at the mathematical derivation properly.

news/2021-11-08-performance-points-star-rating-updates.md
bdach: The above seems fine, but I'd probably just omit/skip over the interpolation details, mostly. I'm not sure it's worth the attention. I'd probably leave it at something like

```suggestion
- A double can be considered double-tappable if the time interval between both notes of the double is small enough that their hit windows overlap, which makes it harder to miss both when a key is pressed twice in the overlap period. Those sorts of patterns are nerfed by artificially increasing the time gap between them.
```

Walavouchey: I can agree with something like this, but I want to turn it around to front-load the information

```suggestion
Quick doubles with overlapping hit windows are nerfed. With a small enough time interval between two notes, they can still be perfectly hit by simply double-tapping during the overlap of their hit windows. This is now accounted for in calculations by artificially increasing the time gap between the notes.
```

news/2021-11-08-performance-points-star-rating-updates.md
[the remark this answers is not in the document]
peppy: New graph needs to be added to the article.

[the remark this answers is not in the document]
Ephemeralis: I think this does the trick far better than the formula ever did.

bdach (approved):
Not much more to add, reads pretty good here. Thanks for bearing with my comments. :+1:

news/2021-11-08-performance-points-star-rating-updates.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Speed hard cap removal
+
+The speed hard cap in osu! has been removed by [**emu1337**](https://osu.ppy.sh/users/2185987) and [**Apo11o**](https://osu.ppy.sh/users/9558549) in [this pull request](https://github.com/ppy/osu/pull/14617), allowing 300+ BPM streams to be weighed accurately. 
+
+The speed cap was previously set to 300 BPM to prevent abuse cases, which means that calculating difficulty on beatmaps with higher BPM requires special consideration. Two adjustments were made to account for that:
+
+- Speed is nerfed on doubles that are double-tappable. If the time since the last object for the current object is smaller than the hit window for a 300, we instead take the difference between that object and the hit window of the last one.
+- Speed is nerfed if successive objects are faster than the 300 hit window. This is intended to target cheesing of faster bursts and streams if the hit window is way larger than the time between notes.
```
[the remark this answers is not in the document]
Walavouchey: Would like this addressed. Main concern I guess is that "speed is nerfed" doesn't make much sense as a statement. "Speed pp" would also be less to-the-point. Can be a separate point still, but it could use some less redundant wording.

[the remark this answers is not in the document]
Ephemeralis: fair point, was holding off on this since I'm not 100% sure on either wording, but we can go with yours for the time being

Walavouchey (dismissed):
Read throught it again, looking good

peppy:
I've added the initial release schedule, which mostly draws from the previous post. Reviews welcomed.

news/2021-11-09-performance-points-star-rating-updates.md
```diff
@@ -12,6 +12,49 @@ If you find yourself scratching your head when reading, consider consulting the
 
 We will be covering only changes made to the **osu!** game mode in this post.
 
+## Release Schedule
+
+Deployment of these changes will as usual, take some time to fully complete. As this batch of changes affects all game modes, it will likely take longer than normal. If you are interested in following the progress we recommend you bookmark this page and check back on this section, as it will be updated as forward progress is made.
```
smoogipoo, line 17:
```suggestion
Deployment of these changes will as usual, take some time to fully complete. If you are interested in following the progress we recommend you bookmark this page and check back on this section, as it will be updated as forward progress is made.
```

news/2021-11-09-performance-points-star-rating-updates.md
```diff
@@ -12,6 +12,49 @@ If you find yourself scratching your head when reading, consider consulting the
...
+
+### 🏃 Star rating updates \[2021-11-09 - 2021-11-11\]
+
+An initial pass over all beatmaps to update their star ratings. Usually we wouldn't pause pp updates during this stage, but due to new additions it is required this time.
```
Biokax, line 21: what do you mean by "pause pp updates during this stage" ? does the score submission will not be possible during the stage ? maybe more clarification about that.

news/2021-11-09-performance-points-star-rating-updates.md
```diff
@@ -12,6 +12,49 @@ If you find yourself scratching your head when reading, consider consulting the
...
+- Inactive users (anyone that has not played the game in 3 months) will be delisted.
+- Scores in the "best performance" profile section may look to be out of order, or not visible at all during this phase. This will resolve completely during the reindexing phase listed below.
+
+Estimated to take 4 days for the majority of active users, and longer for lower ranking users. The initial run is targeting all users above 1,000 total pp, after which we will begin to start a longer running update that may run over a week or so.
+
+### ⌛ Re-enabling of global rank history updates \[2021-11-15\]
+
+Rank history graphs updates will be enabled and updated again. At this point, all users' global leaderboard ranks will be stable going forward.
+
+UPDATE: Updates have been re-enabled and will continue as per normal.
```
smoogipoo, line 42:
```suggestion
```

smoogipoo (changes requested):

news/2021-11-09-performance-points-star-rating-updates.md
Walavouchey: Pretty sure this refers to freezing rank history graphs

peppy: This means freezing *all* updates to the point I'm not sure we will ever catch up.

OfficialWhiffy: > Pretty sure this refers to freezing rank history graphs

People will understand this better than pause pp updates, I guess

peppy: Pending further consideration. I don't think it's going to work as we have it.

peppy: It has nothing to do with rank graph updates.

Biokax: so it do not pause the pp update because it does not start at this point if im not wrong. You can say that it start by updating the star rating and after that the pp update begin. Is that what you mean ?

megahello: In my opinion saying that freezing updates to pp makes a lot more sense.
Just in case add a extra note what it means as people may get confused.

respektive: I think having a note, reassuring that this won't affect score submission would probably good, cause that's always peoples biggest fear.

OfficialWhiffy: I understand it now. If you add that it contains all pp updates, people might get less confused.

Nukrid: If I understood it correctly, it means that when you submit a play, your pp won't update because the updates are currently paused due to the rework process.

If I am correct on that, the note should say something like: "Any plays you make during that time window will be submitted, but any pp values won't be updated."

Biokax (dismissed):

smoogipoo (approved):

peppy (approved):

Walavouchey (approved):

PR #6339, opened 2021-11-08
[EN] Fix missing 'the' in `osu! wiki`

The-Last-Cookie:
I noticed a missing 'the' when working on the German translation. Since there is a comment there, I assume that this article will be edited in the future. But I guess I should mention it nonetheless.

## Self-check


TicClick (approved):
yeah I am guilty of doing this often

PR #6343, opened 2021-11-09
add spreadsheet link in the Performance Points post

Biokax:
Instead of just adding it in the comment section I think its better to have the spreadsheet available for all in the original post.

### DIFF
--- a/news/2021-11-09-performance-points-star-rating-updates.md
+++ b/news/2021-11-09-performance-points-star-rating-updates.md
@@ -165,6 +165,11 @@ Alongside all those big changes above, there are some changes on the smaller sid
 - In preparation for osu!lazer, difficulty calculation for [the Blinds mod](https://github.com/ppy/osu/pull/14921) and [the Relax mod](https://github.com/ppy/osu/pull/14942) has been added by [**Apo11o**](https://osu.ppy.sh/users/9558549).
   - These could be potentially enabled in the future once osu!lazer leaderboard submissions go live, since the concept of "ranked mods" will be no more.
 
+### Most impacted maps
+
+In order to test the result of the rework a spreadsheet with the most 10000 impacted maps has been generated.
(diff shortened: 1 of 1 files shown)


news/2021-11-09-performance-points-star-rating-updates.md
```diff
@@ -165,6 +165,11 @@ Alongside all those big changes above, there are some changes on the smaller sid
 - In preparation for osu!lazer, difficulty calculation for [the Blinds mod](https://github.com/ppy/osu/pull/14921) and [the Relax mod](https://github.com/ppy/osu/pull/14942) has been added by [**Apo11o**](https://osu.ppy.sh/users/9558549).
   - These could be potentially enabled in the future once osu!lazer leaderboard submissions go live, since the concept of "ranked mods" will be no more.
 
+### Most impacted maps
+
+In order to test the result of the rework a spreadsheet with the most 10000 impacted maps has been generated.
+You can find it [here](https://docs.google.com/spreadsheets/d/1XONSMA-LTUMuCw3VpZQM2-fo9ZcEsgMp2ekwjZ2QMCA/edit?usp=sharing) and see which map gains or loses the most of star rating or pp.
```
apollo-dw, line 171: The spreadsheet here is not useful at all here. `sr_master` values are from a partly-merged version of the aim overhaul, so it's useless for comparisons for pretty much everyone.

apollo-dw (changes requested):

news/2021-11-09-performance-points-star-rating-updates.md
Biokax: do you have the values of the full before/after merge ?

peppy: i adddition to the average user it's near unreadable. a sheet should be generated with mods on separate pages (using acronyms not numbers) and aspire stuff excluded.

peppy:
Unless someone is going to generate something somewhat presentable let's leave this as a "should happen next time" thing.

PR #6353, opened 2021-11-11
Update en.md

Heinzen:
Changing my own (old) name to current.

## Self-check


peppy:
I don't think usernames need to be updated on historical pages like this. It should probably represent the state at that point in time. Also please read the contribution guide on how to submit changes if you plan on submitting anything else.

PR #6381, opened 2021-11-17
OWC2021: fix: semifinals -> finals

jthistle:
This is the schedule for the finals - the semifinals have already happened


TicClick (approved):
..yeah

PR #6390, opened 2021-11-18
Update difficulty spectrum

mcendu:


ZeroPyrozen:
Can you also update [Star rating](https://github.com/ppy/osu-wiki/blob/master/wiki/Beatmapping/Star_rating/en.md) article image?

TicClick:
I may be wrong, but osu! wiki may not support SVGs

gagahpangeran:
Now I check osu-framework doesn't support load svg, see https://github.com/ppy/osu-framework/issues/1956

So for now this image can't be displayed in osu!(lazer) wiki page.

mcendu:
> Now I check osu-framework doesn't support load svg, see [ppy/osu-framework#1956](https://github.com/ppy/osu-framework/issues/1956)
> 
> So for now this image can't be displayed in osu!(lazer) wiki page.

Also I used `@font-face` with `data:` URLs instead of converting the text to path for ease of editing and translation, which could complicate things.

peppy:
Please use PNG.

mcendu:
Should I still keep the SVG (used to generate the PNG) in case we need another update or for translations?

peppy:
I'd say that should be in figma, not in an svg. But I'm 100% certain @arflyte is not updating that. So let's keep the svg here as a reference.

peppy:
You CANNOT embed the torus font. It is a commercial font and that goes against its licence.

Please rebase / force push to avoid committing it to this repository.

mcendu:
Switching to Exo 2.

peppy:
No. Don't embed any fonts.

peppy:
If you need fonts then just nuke the svg from this repo.

peppy (approved):

PR #6394, opened 2021-11-18
Fix style errors in `History of osu!/2015`

TicClick:
## Self-check


TicClick:
ok maybe next time, as it doesn't like either option

PR #6397, opened 2021-11-18
[DE] Add `Gameplay/Drain_time`

respektive:
Would like some opinions on whether to leave `health drain` in english as an osu term of sorts or to translate it to something like `Gesundheits-Drain`. I personally think the later sounds a bit weird and too forced maybe.
## Self-check


The-Last-Cookie:
The only word I can come up with is `Gesundheitsreduktion`. But the English version is still better in my opinion.

respektive:
i put the `health drain` link in quotation marks, i think that looks good.

The-Last-Cookie (approved):

TicClick (approved):

PR #6406, opened 2021-11-20
Fix one OWC 2021 match result

LevFlavien:
## Self-check


TicClick (approved):

cl8n:
well now the bold is wrong.
also I think the scores were already correct, just names in wrong order, since Chile is the higher seed

TicClick:
why it is wrong? bold = winner's score, no?
and, according to both the linked match results, and the challonge bracket, France has won: https://challonge.com/OWC_2021

cl8n:
the bold of the country name, i mean
and yes France won but the names are in wrong order, so the score was actually the correct thing and the name order was wrong

ChillierPear:
MP link is displayed as France vs Chile so the page should follow convention. France won the match 6-5, as such their name should be highlighted alongside their score

cl8n:
is the "convention" not that higher seed goes on left? I thought the whole point is that you can see upsets by finding bold names on the right. maybe I'm out of touch with how newer tournament articles go, but it seems weird to obscure that info

Bazzadwarf:
> is the "convention" not that higher seed goes on left? I thought the whole point is that you can see upsets by finding bold names on the right. maybe I'm out of touch with how newer tournament articles go, but it seems weird to obscure that info

afaik there is no set way, I try to follow the bracket when possible so that most of the time higher seeds end up on the left and lower seeds on the right.

LevFlavien:
oops

Issue #6007, opened 2021-08-24
System file formats should not be documented on this wiki


peppy:
Specifically, the https://osu.ppy.sh/wiki/en/osu%21_File_Formats/Db_%28file_format%29 page should be moved to the developer wiki, over at https://github.com/ppy/osu/wiki. Once it is added there, references should be removed from this wiki, apart from some text (maybe at the bottom of https://osu.ppy.sh/wiki/en/osu%21_File_Formats?) linking to the page for developers wanting information on system file types.

cl8n:
why would a database format for stable go onto the lazer wiki? I don't think anyone would guess to look there, or know that it exists in the first place, if they aren't working on lazer.

peppy:
That's the point. It should not be consumed for any other purpose. Documenting it on this wiki makes no sense as it is not supposed to be a publicly documented file nor something a user would be reading or writing.

Issue #2709, opened 2019-10-25
Rewrite file format articles


cl8n:
following the style of #2665

luckily .osu is the worst it gets, the others seem decent already

SisypheOvO:
I would like to ask: why do we need rewrites for `.osk` & .`osz` ? These two files are just `.zip` files tbh. Are they using compression algorithms like Deflate or not? I'm not very sure.

cl8n:
osz/osk/osb all seem to be pointless stubs rn and should either be expanded with more details abt their contents + the archive format as you mention, or removed if it feels better to talk about those only in the main file formats article.

when I opened this issue it was probably referring to .osr and the .db files (which have since been moved to lazer github wiki), the checklist was edited in later. i'll just close bc it looks like an old personal todo list item and I left no details about what to change

PR #6408, opened 2021-11-20
Remove system file formats

The-Last-Cookie:
This is a first step to [#6007](https://github.com/ppy/osu-wiki/issues/6007). The issue [#2709](https://github.com/ppy/osu-wiki/issues/2709) is also kinda loosely related.

The links to the page were removed from every file. I didn't find any mention of `wiki/osu!_File_Formats/Db_(file_format)` elsewhere in the wiki (I searched for `Db_(file_format)`).

I don't know how to open a PR for the wiki at the osu repository, so I would need help with that in case I should add the new files there.


TicClick:
instead of removing any mention of the `.db` format, I'd suggest simply redirecting the links to the ppy/osu wiki when the content is there

The-Last-Cookie:
After writing a small [python script](https://github.com/The-Last-Cookie/treesearch/blob/master/treesearch.py), I found a few other occurences of `.db`. Just adding this for visibility for now.
When the article on the osu repository is implemented, I will change every file to your suggestion because I think this is better than just removing the links. 👍

Walavouchey:
Just for the sake of being explicit, please hold off on merging this until someone with access to the `ppy/osu` wiki has moved the documentation over.

peppy:
https://github.com/ppy/osu/wiki/Legacy-database-file-structure

peppy (approved):

PR #6414, opened 2021-11-21
Remove remaining links to `osu!_File_Formats\Db_(file_format)`

The-Last-Cookie:
Removes all remaining links (that I could find) to the `.db` file format page (or sets the file to `outdated`).

I probably should have added this as a task, because I originally planned to do this together with #6408 .


DoryanR:
Is it possible to change the links on other articles, instead of outdating them? For us translators, it's a lot of work for nothing, so I think it's more logical to change the links on all the articles concerned.

wiki/Beatmap/Online_offset/en.md
```diff
@@ -13,6 +13,6 @@ It can be set by members of the [Nomination Assesment Team](/wiki/People/The_Tea
 
 ![online offset on web](img/online_offset.jpg)<!--TODO: add image of the new webpage once online offset gets added there-->
 
-All online offset values are stored locally in the [osu!.db](/wiki/osu!_File_Formats/Db_(file_format)) file for later use (as `short`s). Storing online offset values in the osu!.db allows the values to be applied even if a player is playing offline as long as they have been connected to the internet prior to importing or playing the beatmap.
+All online offset values are stored locally in the [osu!.db](https://github.com/ppy/osu/wiki/Legacy-database-file-structure) file for later use (as `short`s). Storing online offset values in the osu!.db allows the values to be applied even if a player is playing offline as long as they have been connected to the internet prior to importing or playing the beatmap.
```
peppy, line 16: doesn't need to be mentioned. just remove the whole reference to the db file.

peppy (dismissed):

Walavouchey:
additionally, there should be a note at the bottom of the file formats article pointing to the documentation, as per https://github.com/ppy/osu-wiki/issues/6007

peppy:
There doesn't need to be. They are not intended for consumption and don't need to be mentioned on this wiki.

wiki/Beatmap/Online_offset/en.md
Walavouchey:
```suggestion
All online offset values are stored locally for later use. This allows the values to be applied even if a player is playing offline as long as they have been connected to the internet prior to importing or playing the beatmap.
```

The-Last-Cookie:
> Is it possible to change the links on other articles, instead of outdating them? For us translators, it's a lot of work for nothing, so I think it's more logical to change the links on all the articles concerned.

I also thought about this, but since the links are sometimes embedded in a text (see `Online_Offset` for example), it's hard to just remove the link. You would need to rewrite that paragraph anyway.

wiki/Beatmap/Online_offset/fr.md
```diff
@@ -1,4 +1,5 @@
 ---
+outdated: true
 tags:
   - online song offset
 ---
```
DoryanR, line 5: L17: ```Toutes les valeurs de online offset sont stockées localement. Le stockage des valeurs de online offset dans le fichier osu!.db permet d'appliquer les valeurs même si un joueur joue hors ligne, à condition qu'il ait été connecté à Internet avant d'importer ou de jouer la beatmap.```

(which will avoid outdating this article)

wiki/Main_Page/fr.md
```diff
@@ -1,4 +1,5 @@
 ---
+outdated: true
 layout: main_page
 ---
```
DoryanR, line 4: L30: ```[Formats de fichiers](/wiki/osu!_File_Formats) : [.osz](/wiki/osu!_File_Formats/Osz_(file_format)) • [.osk](/wiki/osu!_File_Formats/Osk_(file_format)) • [.osr](/wiki/osu!_File_Formats/Osr_(file_format)) • [.osu](/wiki/osu!_File_Formats/Osu_(file_format)) • [.osb](/wiki/osu!_File_Formats/Osb_(file_format))```

The-Last-Cookie:
If I commit the change by DoryanR, I will also remove the .db link in every `Main_Page` which I set as outdated. I won't change any pages that were already outdated before opening this PR. Otherwise, this PR is ready for merge.

Is it okay to commit the changes by DoryanR?

Walavouchey:
should be ok

The-Last-Cookie:
Okay, this is now ready for merge.

TicClick (approved):

Issue #6233, opened 2021-10-15
Automation/CI checks wanted


TicClick:
- [x] footnotes: https://github.com/ppy/osu-wiki/issues/6232
- [x] ~~use of redirects~~ no: https://github.com/ppy/osu-wiki/pull/6645#discussion_r777794205
- [x] broken wiki links https://github.com/ppy/osu-wiki/pull/6645
  - [x] including references to the translated subsections, and reference-style links https://github.com/ppy/osu-wiki/pull/7288
- [x] `outdated: true` and `outdated_since: ...` missing from the translations when the original articles are being updated ([via](https://discord.com/channels/188630481301012481/218677502141399041/888446144512921641); this one could be a bot instead): https://github.com/ppy/osu-wiki/pull/6428
- [x] `yamllint` on individual articles to avoid things like #6619: https://github.com/ppy/osu-wiki/pull/6629
- [ ] `outdated_since` hash validation to avoid things like #6747

TicClick:
#3310 could be a source of ideas for the remark plugins

cl8n:
moved to #2275

PR #6428, opened 2021-11-23
Add CI step for checking if translations are properly marked as outdated

TicClick:
# what

this check protects us from the situations when someone has edited an original article (`en.md`), but forgot to check its translations and mark them as outdated by adding `outdated: true` and `outdated_since: {commit hash}` markers to their front matters.

for cases which can't be handled automatically (say, you're fixing a typo which is only present in English article), it can be silenced by adding `SKIP_OUTDATED_CHECK` to the pull request's message (it may appear everywhere).

example: https://github.com/TicClick/osu-wiki/pull/2. see CI output on individual commits, as well as its status for the whole PR (forced to OK by the tag):
- [error](https://github.com/TicClick/osu-wiki/pull/2/checks?sha=cea95b0a8c752d7b914bde5f8c9728011f657f03) | screenshot: https://user-images.githubusercontent.com/4686101/142972040-358bc307-e6c2-42e9-bae9-89c812c324af.png
- [clean pass](https://github.com/TicClick/osu-wiki/pull/2/checks?sha=5b2017f4b172287c0c12f1607179c883173e9b3f) | screenshot: https://user-images.githubusercontent.com/4686101/142972098-c7be1404-8a39-4d91-a32d-4cef13251972.png
- [suppressed](https://github.com/TicClick/osu-wiki/runs/4295034186) | screenshot: https://user-images.githubusercontent.com/4686101/142972138-aac62d32-4f50-4849-8201-4289125bb319.png

# where

it's... uh, inlined in the `continuous-integration.yml` file. could be moved into a separate script if necessary. if deemed useful, could be sped up by throwing out unnecesary steps like linters. could be moved to a separate workflow and sped up by [caching the repository checked out](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#caching-dependencies).

# when

The script is triggered on [default `pull_request`-related events](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request) **plus** PR message edits to catch the (dis)appearance of `SKIP_OUTDATED_CHECK`.

# why

## ...the check?

part of #6233. prevents articles expiring silently because of random maintainers' occasional lazy eye syndrom (at least #6197, #6189, #6111, #4651 -- found by me only).

## ...is this a GitHub Action?

- needs to be run on every commit
- should block merge
- may be overriden
- the source is easily accessible and editable
- I honestly don't want to host a bot, or an app, not to mention that it'd have been a bit more complicated

## ...do you think people will follow it?

it prints errors AND direct suggestions on how to fix them. besides, it's important to know which translations are really up to date.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -7,6 +7,11 @@ on:
   pull_request:
     branches:
       - master
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
 
 jobs:
   ci:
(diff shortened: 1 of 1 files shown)


peppy:
The error message should probably better explain where the `SKIP_OUTDATED` should be added. It's a bit unclear to users that first see it in that message.

DoryanR (approved):

The-Last-Cookie:
Might it be a good idea to put a small comment containing `SKIP_OUTDATED` into the PR template?
I don't want to spend extra time adding this tag in the PR message after the PR has been opened.

TicClick:
I am against this. the situations where you need to willingly ignore the check are much more rare compared to a missing `outdated` tag here and there — you only need this knowledge occasionally and not in advance. the description will also be much longer than the rest of the points from the PR template comment, partly because what's happening is hard to describe in a few words, and will lenghten it even more, which will lead to less people reading the template as a whole at all

peppy:
Let's give this a try and see how it works. The script looks sane enough.

peppy (approved):

Issue #6080, opened 2021-09-08
Add several new categories and shut down `Glossary`


TicClick:
we have quite a few top-level categories already (such as `Beatmap`, `Beatmapping`, `Gameplay`, `Modding`, etc.)  which contain *most* of the relevant articles. I have a few ideas on where to go further:

- [x] the new top-level `Client` category: https://github.com/ppy/osu-wiki/pull/6627
  <details>
  <summary>what would be inside</summary>

  - [`Beatmap_Editor`](https://github.com/ppy/osu-wiki/tree/master/wiki/Beatmap_Editor) (12 articles)
  - [`Chat_Console`](https://github.com/ppy/osu-wiki/tree/master/wiki/Chat_Console)
  - [`Interface`](https://github.com/ppy/osu-wiki/tree/master/wiki/Interface)
  - [`Options`](https://github.com/ppy/osu-wiki/tree/master/wiki/Options)
  - [`osu!_Program_Files`](https://github.com/ppy/osu-wiki/tree/master/wiki/osu!_Program_Files) (1 article)
  - `osu! file formats` (5 articles or something)
  - see Walavouchey's comments down below
  </details>

- [x] ~~mode-specific categories for `Hit object`s (right now they are dumped together: [`Hit_object`](https://github.com/ppy/osu-wiki/tree/master/wiki/Hit_object))~~ after second, third, and fourth thoughts this doesn't look like it's really needed
- [x] disassemble `Glossary`. it's a blanket category anyway, and all of the articles in [`Glossary`](https://github.com/ppy/osu-wiki/tree/master/wiki/Glossary) may have their new homes already
- [x] the new `Music theory` category. this will include musical terms from the top level, such as `Beat`/`Downbeat`, and stuff from the current `Music theory` article (which should be split into several small articles as well), AND the timing-related articles/guides? need to hone this more
- [x] ~~a new `Series` category: articles such as `osu!academy`, `osu!talk`, `osu!mapping` will live here~~ a better idea: #6438
  - [x] check if the main page's community section should be updated accordingly https://github.com/ppy/osu-wiki/pull/8101
- [x] ~~(uncertain) move everything out of `The Team` category to `People`~~ no as per the last point of https://github.com/ppy/osu-wiki/issues/1698#issuecomment-917912450, or at least "not in foreseeable future"

### thoughts?

[...]

PR #6438, opened 2021-11-25
Add the `Community` category

TicClick:
part of #6080. structure (all these articles were top-level previously):

```
(NEW) Community
├── Community_Mentorship_Program
├── How_you_can_help!
├── Internet_Relay_Chat
├── Mappers_Guild
├── Organisations
│   └── osu!_UCI
├── osu!_Community_Meetings
├── osu!dev_Discord_server
├── Project_Loved
├── Projects
└── (NEW) Video_series
    ├── osu!academy
    ├── osu!mapping
    └── osu!talk
```

## Self-check


TicClick:
(I'll fix old links in other articles later, in one sweep)

wiki/Community/en.md
```diff
@@ -0,0 +1,32 @@
...
+
+This page lists ways for people from all around osu! to get in touch with each other. If you're looking for contribution opportunities, but don't know where to start, make sure to read [How you can help!](/wiki/Community/How_you_can_help!)
+
+## Places
+
+<!-- TODO: add forums -->
+
+- [Internet Relay Chat](/wiki/Community/Internet_Relay_Chat) (IRC)
+- [osu!dev Discord server](/wiki/Community/osu!dev_Discord_server)
+- Offline [organisations](/wiki/Community/Organisations)
```
Walavouchey, line 11:
```suggestion
- [Offline organisations](/wiki/Community/Organisations)
```

TPGPL:
I don't think the `Video_series` directory is necessary tbh; I'd rather have these articles placed directly in the `Community` category

Walavouchey:
Just like `Organisations` is its own directory, so should `Video_series`

wiki/Community/Video_series/en.md
```diff
@@ -0,0 +1,5 @@
+# Community video series
+
```
Walavouchey, line 2: afaik all index pages have a summary paragraph (and probably should)

```suggestion

Index page for video series run by community members.

```

Walavouchey (changes requested):
looks pretty ok

wiki/Community/en.md
```diff
@@ -0,0 +1,32 @@
...
+## Projects
+
+- [Community mentorship program](/wiki/Community/Community_Mentorship_Program)
+- [Mappers' Guild](/wiki/Community/Mappers_Guild)
+- [osu! wiki](/wiki/osu!_wiki)
+- [Project Loved](/wiki/Community/Project_Loved)
+
+## Initiatives
+
+- [osu! community meetings](/wiki/Community/osu!_Community_Meetings)
```
Walavouchey, line 22: file should be `osu!_community_meetings` now that i think about it

Walavouchey (changes requested):

wiki/Community/en.md
Walavouchey: there are other articles with seemingly incorrect casing now that i think about it, so you may ignore this since it should be its own issue

TicClick: renamed it anyway

Walavouchey (approved):

The-Last-Cookie:
If `Internet_Relay_Chat` is in the new category, `Chat_Console` should be in there too in my opinion. Nevertheless, this is a good idea 👍 

`Projects` is also kinda related to this, since it has a `Community` section.

TicClick:
`Chat Console` is something that belongs to the client itself, but yes, it _may_ deserve a mention here.
`Projects`... good call, actually!

DoryanR:
Looking at #6080, I don't see a place for the `Tournaments` and `Contests` folders, and since this is about the community, why not put them in the new `Community` folder? 
In that sense, there may be other folders that could fit into this type of category.

TicClick:
tournaments and contests are very large categories on their own, with a lot going on inside. in some sense, a lot of things may be shoved inside `Community`, but I don't think doing that is necessary.

OTOH, `Projects` are just a listing of tools and ideas, so..

wiki/redirect.yaml
```diff
@@ -785,8 +795,9 @@
 "user_configuration":                         "osu!_Program_Files/User_Configuration_File"
 "user_config":                                "osu!_Program_Files/User_Configuration_File"
 
-"academy":                     "osu!academy"
-"announcements/osu!academy":   "osu!academy"
+"academy":                     "community/video_series/osu!academy"
```
Joehuu, line 798: The casing should usually match the folder's, else web has to do an extra url redirection for that.

TicClick: fair point. fixed

Joehuu (dismissed):
LGTM

peppy (approved):

PR #6446, opened 2021-11-27
Update difficulty icons

TicClick:
- icons are 16x16, with transparent background, colors taken from osu-web's [`variables.less`](https://github.com/ppy/osu-web/blob/398dab41d18381e14111de133ffb095a58611bc1/resources/assets/less/variables.less#L305-L313)
- rendered with imagemagick: https://gist.github.com/TicClick/1f56a41c7b43b2fec7563c8bc3ea3692

I have neither renamed the files, nor changed the links (there are simply too many of them), so osu-web's image cache needs to be dropped after this is merged


TicClick:
@cl8n are the colors I used correct, or should I take those from https://github.com/ppy/osu-web/pull/8317?

peppy:
You would hope the colours would match that in the PR, no?

TicClick:
took me some time, but I understood the source of my confusion: difficulty stars from clayton's PR don't match these mentioned in the spectrum image, hence the difference in color. I think that's not relevant for the wiki icons, as they attempt to represent a part of the spectrum at once rather than some fixed value

cl8n:
the image you linked is the old colors. I don't think I'm following what the dilemma here is anyway--- colors for the icons should be the middle of each diff's range, right? it can be calculated from the gradient on osu-web

TicClick:
yeah, I got that. the spectrum image may be fixed (again) later, I think

TicClick:
bump

Walavouchey:
so is this waiting on corrections or a merge?

TicClick:
colors appear to be correct; the only thing I am concerned about is cache flush as an alternative to us pointlessly dragging images across the disk (read: renaming) every time that needs to be performed.

so,
- this can be merged, but the images won't update for ~2mo, OR
- I can rename the images and have the update reflected immediately, but that's, like, really unnecessary, and I belive there's got to be another way

Walavouchey:
it's probably even more pointless to wait all the way until https://github.com/ppy/osu-web/issues/7678 gets resolved, so adding a query string like peppy suggests in that thread should be ok for now

peppy:
query string is the preferred way. just put `?20211215` on the end.

TicClick:
should be ready now

Walavouchey:
`na-c.png`, `na-m.png`, `na-o.png` and `na-t.png` have 1 row of pixels removed from the bottom, making the images smaller and off-centre to the left

comparison for `na-m.png`:

image

others are smaller and off-centre to the top left as well, but that i'll assume is intentional™️ 

image

TicClick:
thanks for catching this. re-rendered, all dimensions should match now.

Walavouchey (approved):
seems good

PR #6456, opened 2021-11-28
[ZH-TW] Update `osu!_Community_Meetings`

SiderealArt:
## Self-check


TicClick (approved):
this is a minor change, I don't think it should be held back by lack of reviews

PR #1375, opened 2018-05-19
Change "kanal" to "channel" on Bahasa Indonesia localisation

fajar13k:
### Description

- Change tons of "kanal" word to "channel"

### Status

- [x] Finished

### Additional Notes 

**Merge this PR LAST**, after #1366 #1371 #1372 has been merged.

Co-Authored-By: Ganendra Afrasya <gsculerlor@hotmail.com>
Co-Authored-By: acylla_prime <acylla@users.noreply.github.com>

PR #6469, opened 2021-11-30
[ID] Fix Typo in `Bot_account`

ZeroPyrozen:
## Self-check


Nivalyx (commented):
should be okay but i'm not sure about leaving "channel" as is without having it translated as `kanal` or written in italics as `*channel*` (due to it being a foreign word) tho

ZeroPyrozen:
iirc, there is standardization on #1375 but it wasn't italicized. I also think it should be italicized because foreign word.

Nivalyx:
yeah i think writing them down in italic is the way to go @ZeroPyrozen

TicClick (approved):

PR #6477, opened 2021-12-01
Move `Guides/Making_Properly_Centered_Spinners` to British English

daud-ahn:
## Self-check

@TicClick said this article may use a completely different name.


peppy:
Not sure I agree with this one, even as an Australian. But let's go with it for consistency.

PR #6478, opened 2021-12-01
Update `Users with unique titles` listing

Nivalyx:
this is an unusually big update for this article whew, hopefully i didn't miss anything 🐇

...yeah, after some considerations i decided to omit the `By Name` section from the article as unfortunately i don't think it's feasible to retain the list with the recent addition of over a hundred names to the page. perhaps it could be written in an article of its own instead? dunno 😦

---


Ephemeralis (approved):
reasonable re: by name removal, i think the current categorization handles it fine

PR #6493, opened 2021-12-04
Update casing of lazer tournament client custom asset folder names

bdach:
In https://github.com/ppy/osu/pull/15847 (which has been included in a public lazer release - 2021.1202.0, to be exact), I've changed the tournament custom asset folder names to be uppercased rather than lowercased. For most users (read: Windows/macOS users), this will not have any discernible impact on anything; the only case where this matters is Linux. In fact a Linux report was what spurred this change to begin with (https://github.com/ppy/osu/discussions/15837).

This pull request therefore updates the lazer tournament client article to reflect the new casing.

## Self-check

  - This pull also touches the French translation of the article. While the changes concern only casing of a few folders, and I have a rudimentary enough understanding of French to be somewhat confident that the changes are correct, it may be helpful to have a second pair of eyes here. I've made this change mostly to avoid outdating the translations for no reason.


DoryanR (approved):
The French side is good.

TPGPL (approved):

TPGPL (approved):

PR #6507, opened 2021-12-06
Add installation guide for Linux

daud-ahn:
## Self-check

{insert GNU/Linux copypasta here}


ZeroPyrozen:
Add missing line, list item and remove unnecessary `:` per [this action](https://github.com/ppy/osu-wiki/runs/4426962245?check_suite_focus=true) run

peppy:
I dunno about this. I'd rather link to a third party source with installation instructions rather than make this something official.

wiki/Installation/Linux/en.md
```diff
@@ -0,0 +1,497 @@
...
+pulse.min.req = 32/48000 # 0.67ms
+pulse.min.quantum = 32/48000 # 0.67ms
+}
+```
+
+4. Save with Ctrl+O and exit with Ctrl+X.
+5. Reboot your PC.
+6. Enable compatibility mode on osu!.
+
+### - PP Counter for OBS: gosumemory
```
peppy, line 434: This kind of stuff most *definitely* cannot be on the official wiki.

daud-ahn: if you are talking about those additional info, it can just be removed

peppy: No applications, links to executables, etc. can be in this guide.

I would propose this is just a stub leading to off-site guides to getting stuff working. I don't believe it should be present in the official wiki. Officially we do not support linux, and don't intend to offer support that isn't via osu!(lazer).

In fact I'd argue that this guide should be about osu!(lazer) instead.

daud-ahn: how about, move this page content to my gist (or maybe forum post) and then link that to this page.
about osu!lazer

peppy: As long as it doesn't link to custom osu/wine packages. It should be instructions on how to make things work with no external binaries.

peppy:
Going to close this PR for now because I don't want the commits in git history regardless of how this is changed going forward. Feel free to open an issue to discuss this further.

Issue #6431, opened 2021-11-23
Rework score-related sections and individual articles


TicClick:
the score articles look cumbersome and are a bit hard to read (even more if you open them from the wiki, as it has less horizontal space available). here's what I propose to make them better:

- [ ] a disambiguation page for score, which would mention individual/total/ranked scores, as well as scoring mechanics (perhaps some of them should have their own articles)
- [ ] another structure. instead of a single page on score, there's an article per game mode, with V1 and V2 sections.
  ```
  Score/
  ├─ osu!
  ├─ osu!taiko
  ├─ osu!catch
  ├─ osu!mania
  ```
- [ ] the game mode articles link to the **articles** from the above tree
- [ ] the new score articles should probably be rewritten a bit to make them easier to read
- [ ] fact checking:
  - [ ] https://github.com/ppy/osu-wiki/issues/2025
  - [ ] https://github.com/ppy/osu-wiki/issues/2026
  - [ ] https://github.com/ppy/osu-wiki/issues/2528
  - [ ] https://github.com/ppy/osu-wiki/issues/3430
  - [ ] https://github.com/ppy/osu-wiki/issues/4448

<details>
<summary>for the record, here's everything we have on score in the wiki</summary>

| Location | Contents |
| :-- | :-- |
| https://github.com/ppy/osu-wiki/blob/master/wiki/Score/en.md | - base values of different hit objects, per game mode<br/> - note about score v2<br/> - play score calculation, per game mode (formulas + lengthy explanations)<br/> - "other scores" (an attempt to make a disambiguation page) |
| https://github.com/ppy/osu-wiki/blob/master/wiki/Score/ScoreV1/en.md | direct copypaste of the play score calculation section from the above article |
| - https://github.com/ppy/osu-wiki/blob/master/wiki/Game_mode/osu!/en.md#scoring<br/> - https://github.com/ppy/osu-wiki/blob/master/wiki/Game_mode/osu!taiko/en.md#scoring<br/> - https://github.com/ppy/osu-wiki/blob/master/wiki/Game_mode/osu!catch/en.md#scoring<br/> - https://github.com/ppy/osu-wiki/blob/master/wiki/Game_mode/osu!mania/en.md#scoring | for every game mode:<br/> - grades<br/> - accuracy formula and explanation<br/> - play score calculation (direct copypaste from the `Score` page?)<br/> - judgement<br/> - combo multiplier mechanics |

</details>

Walavouchey:
[...]

PR #6515, opened 2021-12-07
Fix image path in `Score/osu!catch`

SiderealArt:
## Self-check

### DIFF
--- a/wiki/Score/osu!catch/en.md
+++ b/wiki/Score/osu!catch/en.md
@@ -10,12 +10,12 @@ needs_cleanup: true
 
 | Image | Name | Effect |
 | :-: | :-: | :-- |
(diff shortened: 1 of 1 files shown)


Walavouchey (dismissed):

Walavouchey:
actually no wait the images should be moved to the same folder

ZeroPyrozen:
I think all game modes score hit values have inconsistent path. osu! hit values are from interface folder, osu!catch and osu!mania are from skinning, and osu!catch are from parent score folder.

Walavouchey:
rewriting these articles is already part of https://github.com/ppy/osu-wiki/issues/6431
closing in favour of discussion there, but you're welcome to make the change afterwards

PR #6523, opened 2021-12-10
[EN] Add `Forum` article in `Community`

DoryanR:
Added the article `Forum` in the index of the `Community` folder.

@TicClick Need to know if you need to detail the whole tree or not, since you are the one who added the TODO.


TicClick (approved):
no, that would be all

PR #6546, opened 2021-12-14
Change link to CC in `Rules/Content_Usage_Guidelines`

kj415j45:
~~Non-commercial use (NC) is an optional right of CC, and the original link limit the CC to CC-BY-NC-SA 3.0 which include NC. So remove the link for better wording.~~

↓ Check discussion below ↓


peppy:
Wouldn't it make sense to update the link to the non-NC version instead?

kj415j45:
Then it should link to [About CC Licenses](https://creativecommons.org/about/cclicenses/)?

peppy:
Might be best. It's intended to give a curious user a starting point to look into available licences further. A page which explains all the "free for use" licences would also work just fine, I think.

peppy (approved):

PR #6589, opened 2021-12-20
Add osu!lazer troubleshooting article

huupoke12:
I have also added a troubleshooting index article and an empty osu!stable article to prevent "Page not found" error.

## Self-check

### DIFF
--- a/wiki/Troubleshooting/osu!lazer/en.md
+++ b/wiki/Troubleshooting/osu!lazer/en.md
@@ -1,0 +1,25 @@
+# osu!lazer troubleshooting
+
+## No audio
+First, make sure that audio volumes (in-game volume, system volume, speaker/headphone volume) are high enough.
+
+Then, check if the problem only happens with a specific skin or beatmap. If that is the case, the culprit is usually the use of wrong file extension of the audio files (example: the file contains AAC audio data but the extension is `.mp3`). The problem should be reported to the creator in this case.
+
+If the problem happens for the whole system, it might be a driver issue or hardware issue. If it happens for the whole game, check if the correct audio device has been selected in the settings (try selecting the `Default` device on the top).
+
+If osu! has audio while the other applications do not or vice-versa, or this error appear: `BASS faulted with error code 46: Busy`, the problem is usually osu! directly using the audio device (most audio devices only support one stream). The solution depends on the operating system:
+
+### Windows
+1. Open `Control Panel` → `Hardware and Sound` → `Sound`.
+2. Select the `Playback` tab, select the preferred audio device, then click `Properties`.
+3. Select the `Advanced` tab, then uncheck the option “Allow applications to take exclusive control of this device”.
+
+### Linux
+Since BASS (the audio library osu! is using) only supports ALSA backend on Linux, and ALSA is a backend to access audio hardware devices, this problem will happen if osu! is using ALSA directly.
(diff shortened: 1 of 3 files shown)


peppy:
I'd rather not have the term "lazer" anywhere on the wiki. It won't be anything more than a release stream.

wiki/Troubleshooting/osu!lazer/en.md
```diff
@@ -0,0 +1,25 @@
+# osu!lazer troubleshooting
+
+## No audio
+First, make sure that audio volumes (in-game volume, system volume, speaker/headphone volume) are high enough.
+
+Then, check if the problem only happens with a specific skin or beatmap. If that is the case, the culprit is usually the use of wrong file extension of the audio files (example: the file contains AAC audio data but the extension is `.mp3`). The problem should be reported to the creator in this case.
```
peppy, line 6: nah

wiki/Troubleshooting/osu!lazer/en.md
```diff
@@ -0,0 +1,25 @@
...
+
+If the problem happens for the whole system, it might be a driver issue or hardware issue. If it happens for the whole game, check if the correct audio device has been selected in the settings (try selecting the `Default` device on the top).
+
+If osu! has audio while the other applications do not or vice-versa, or this error appear: `BASS faulted with error code 46: Busy`, the problem is usually osu! directly using the audio device (most audio devices only support one stream). The solution depends on the operating system:
+
+### Windows
+1. Open `Control Panel` → `Hardware and Sound` → `Sound`.
+2. Select the `Playback` tab, select the preferred audio device, then click `Properties`.
+3. Select the `Advanced` tab, then uncheck the option “Allow applications to take exclusive control of this device”.
```
peppy, line 15: wha? where are you geting this info

wiki/Troubleshooting/osu!stable/en.md
```diff
@@ -0,0 +1 @@
+# osu!stable troubleshooting
```
peppy, line 1: ?

peppy:
Where is this requirement and text coming from? I'm very confused.

wiki/Troubleshooting/osu!stable/en.md
huupoke12: I thought it's necessary to prevent not found pages.

Walavouchey:
If there is going to be a lazer troubleshooting article, then it should be its own article in `Help_Centre` (and also written in a similar style). About half of the few bits of information here is probably not even lazer-specific, meaning this addition would be pretty bare-bones on its own (although perhaps a good start once the information has been sorted out).

The best way to continue would be to open a discussion first, so that information can be discussed, aggregated, and verified before haphazardly writing a minimally helpful article.

peppy:
Also we tend to keep all information on lazer on the lazer github for now. When it is pushed as an update the existing article will get deprecated in favour of what lazer becomes (a major update).

huupoke12:
Yeah, what I wanted is a kind of a centralised place for troubleshooting answer (or FAQ) for lazer, where the answers can be more high quality (community contributed) and avoid duplicate work. The current discussion tab on GitHub situation is just like a ton of duplicate questions, and the answers can be fragmented.

Walavouchey:
Opened an [issue](https://github.com/ppy/osu-wiki/issues/6592) in favour of this for now.

peppy:
@huupoke12 we welcome additions to the wiki on the osu-repo. but as above, we wouldn't include such general items as the ones you have (which can apply to any application, not lazer-specific).

PR #6596, opened 2021-12-21
Move `Score` to `Gameplay/Score`

TicClick:
part of #6431, slightly related to #6080


Walavouchey (approved):
new catch images added in https://github.com/ppy/osu-wiki/pull/6597

PR #6619, opened 2021-12-27
[ZH] Fix front matter in `Help_Centre/Account/zh.md`

Walavouchey:
Small fix for front matter that broke in #6618. There was already an outdated tag.

## Self-check


TicClick (approved):

TicClick:
I've added a prevention item to #6233

PR #6641, opened 2022-01-03
Fix broken links where possible

TicClick:
- links found with Walavouchey's [`find_broken_wikilinks.py`](https://github.com/Walavouchey/osu-wiki/blob/find-all-borked-links/.github/scripts/ci/find_broken_wikilinks.py)
- this touches even outdated articles for sake of completeness
- redirects will be replaced by direct links in another pr

## Self-check


TicClick:
(this doesn't touch the newsposts now)

Walavouchey (approved):

Issue #2104, opened 2019-03-04
Don't link wiki redirects from within the wiki


cl8n:
still quite a few cases of this left, I'll look into it at some point ~

PR #6645, opened 2022-01-03
Add ci test to check for broken wikilinks in articles and news posts

Walavouchey:
- part of https://github.com/ppy/osu-wiki/issues/6233
- [clean output 1](https://user-images.githubusercontent.com/36758269/148000663-0b292332-a7b8-4895-a3c2-11837d99916b.png) | [clean output 2](https://user-images.githubusercontent.com/36758269/147963320-f3734293-1d7b-4f35-9a3b-b217027f28c8.png) | [error example 1](https://user-images.githubusercontent.com/36758269/147963505-10147756-7757-4440-9d34-588de0e01837.png) | [error example 2](https://user-images.githubusercontent.com/36758269/147963549-08f6ceb0-18e8-4854-a8b3-a7cd512b885c.png)

## Remarks

This adds .png, .jpg, and .gif files to the sparse checkout (the only ones ever used, with the exception of some old news posts), which is required for the script to check image links locally. Test times are still only at around or less than 1 minute (about +10 to 15 seconds from quick observation), probably because all images are very small anyway.

This does not check for reference-style (`[text][link]`) or in-article links (`[text](link#heading)`) yet, which I'll probably add later (could also use some more verbose notes that explain why a link is broken).

Non-wiki links are assumed to start with `http`, `mailto:` or `irc:`, which are the only cases currently in the wiki.

All links in master are currently fixed by @TicClick using this in <https://github.com/ppy/osu-wiki/pull/6641>. 

ASC, readme and template articles, as well as links in comments are excluded from the test (they usually contain examples or link to other repo files).

The error message is written in the assumption that you're not a maintainer, so it doesn't actually tell you to edit `redirect.yaml`.

Suggestions on output colours/formatting are welcome.


TicClick:
(force push because I committed suggested improvements directly to the PR branch)

.github/scripts/ci/find_broken_wikilinks.py
```diff
@@ -0,0 +1,173 @@
...
+    global note
+    note = ""
+    if link.startswith("/wiki/"):
+        # absolute wikilink
+        if os.path.exists(link[1:]):
+            return True
+        else:
+            # may have a redirect
+            return check_redirect(child(link))
+    elif not (link.startswith("http") or link.startswith("mailto:") or link.startswith("irc:")):
```
TicClick, line 55: to protect against `httpp://` and the likes, I'd use `://` for HTTP(S) (also, I don't think we need `irc:`)

```python
elif not any(link.startswith(prefix) for prefix in ("http://", "https://", "irc:", "mailto:"))
```

.github/scripts/ci/find_broken_wikilinks.py
```diff
@@ -0,0 +1,173 @@
...
+            # may have a redirect
+            return check_redirect(f"{directory}/{link}")
+    else:
+        # some other link; don't care
+        return True
+
+def is_in_comment(s, start, end):
+    return s.rfind("<!--", 0, start) != -1 and s.find("-->", end, -1)
+
+def find_link(s, index=0):
```
TicClick, line 69: would be nice to have a short docstring on what it does

.github/scripts/ci/find_broken_wikilinks.py
```diff
@@ -0,0 +1,173 @@
...
+        results.append(match)
+        match = find_link(s, match['pos'][3] + 1)
+    return results
+
+def print_error():
+    print(f"{red('Error:')} Some wiki or image links in the files you've changed have errors.\n")
+    print("This can happen in one of the following ways:\n")
+    print("- The article or image that the link points to has since been moved or renamed (make sure to match capitalisation)")
+    print("- The image that the link points to has the wrong extension (must be one of .png, .jpg, or .gif)")
+    print("- The link simply contains typos or formatting errors")
```
TicClick, line 131: this may be replaced by something along the lines of:
```suggestion
    print("- The article or file does not exist, or you've made a typo")
```

.github/scripts/ci/find_broken_wikilinks.py
```diff
@@ -0,0 +1,173 @@
...
+    print("- The link works, but contains locale selection (e.g. /wiki/en/Article_styling_criteria instead of /wiki/Article_styling_criteria)")
+    print("- The link works, but contains URL-escaped characters (https://en.wikipedia.org/wiki/Percent-encoding). This only applies for links to articles and images inside the wiki.")
+    print("- The link works, but incurs multiple redirects. Use a direct link instead.")
```
TicClick, line 134: this is too specific but I don't know how to shorten it or put better, perhaps... (considering we wanted to ban redirects: #2104)

```
The link must not contain explicit locale or be a redirect
```

.github/workflows/continuous-integration.yml
```diff
@@ -89,6 +89,11 @@ jobs:
           pip install yamllint
           python .github/scripts/ci/run_yamllint.py --config .yamllint.yaml
 
+      - name: find broken wikilinks
+        run: |
+          ARTICLES=$( git diff --diff-filter=d --name-only ${{ github.sha }}^ ${{ github.sha }} "wiki/**/*.md" "news/*.md" )
+          python .github/scripts/ci/find_broken_wikilinks.py $ARTICLES
```
TicClick, line 95: FYI it also detects regular broken links, so I'd just call it `find_broken_links.py`

TicClick (changes requested):
thanks, this is nice.
my comments follow, apart from https://github.com/Walavouchey/osu-wiki/pull/8

TicClick:
on another note, we have two link checkers which _may_ be eventually replaced by this:
- https://github.com/MegaApplePi/remark-lint-osu-wiki-links
- https://github.com/cl8n/remark-lint-osu-links

.github/scripts/ci/find_broken_wikilinks.py
Walavouchey: ok i've identified the only `irc:` links as being in outdated translations of `Community/Internet_Relay_Chat`

.github/scripts/ci/find_broken_wikilinks.py
Walavouchey: i know that we just fixed a bunch of links, but the source of an error may not come from the pr itself

.github/workflows/continuous-integration.yml
Walavouchey: it's not intended for regular links, and even skips any that it thinks are broken links. we can't check for regular links apart from format issues unless we try ping that link

TicClick: I know it's not a universal liveness checker. what I wanted to say was that it caught https://github.com/ppy/osu-wiki/pull/6641/files#diff-c5cdbe755d54c6909a866faa59684ae07e9be41c799e0675f7de9e4efa66bcbe

.github/scripts/ci/find_broken_wikilinks.py
Walavouchey: if you mean `# Finds one link in a string` then i don't think that's necessary.. ?

.github/workflows/continuous-integration.yml
```diff
@@ -89,6 +89,11 @@ jobs:
           pip install yamllint
           python .github/scripts/ci/run_yamllint.py --config .yamllint.yaml
 
+      - name: find broken wikilinks
+        run: |
+          ARTICLES=$( git diff --diff-filter=d --name-only ${{ github.sha }}^ ${{ github.sha }} "wiki/**/*.md" "news/*.md" )
+          python .github/scripts/ci/find_broken_wikilinks.py --target $ARTICLES
```
TicClick, line 95: I forgot to handle the case when there's nothing to check
```suggestion
          [[ -n "$ARTICLES" ]] && python .github/scripts/ci/find_broken_wikilinks.py --target $ARTICLES
```

TicClick (changes requested):

.github/workflows/continuous-integration.yml
Walavouchey: the script itself prints a notice when that's the case

TicClick: which means that the check will always fail if no articles are edited in a PR, for example, if I replace an image

Walavouchey: oh you're right it throws if you don't provide anything to --target

Walavouchey: though if you do rename an image, it wouldn't have anything to check 🤔

.github/scripts/ci/find_broken_wikilinks.py
Walavouchey: also if someone does add an image with a `.PNG` extension with a working link to it, then they'd have no idea why it was failing the test

TicClick: fair enough

peppy:
Is there any reason to not include all file types in the checkout? Is there some specific file you're trying to avoid?

.github/scripts/ci/find_broken_wikilinks.py
Walavouchey: if we are banning redirects, then i can change it to error when it has to follow any redirect (and you'd have to fix all links when moving around articles)

```suggestion
    print("- The link contains explicit locale selection (e.g. /wiki/en/Article_styling_criteria instead of /wiki/Article_styling_criteria)")
    print("- The link contains URL-escaped characters (https://en.wikipedia.org/wiki/Percent-encoding)")
    print("- The link uses redirects")
```

.github/workflows/continuous-integration.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
           # set up sparse checkout
           git sparse-checkout init
-          git sparse-checkout set '**/*.md' '**/.remark*' '**/*.json' '**/*.yaml' '.github/scripts'
+          git sparse-checkout set '**/*.md' '**/.remark*' '**/*.json' '**/*.yaml' '.github/scripts' 'wiki/**/*.png' 'wiki/**/*.jpg' 'wiki/**/*.gif'
```
Walavouchey, line 36: (also it seems the news part got lost in some commit)

```suggestion
          git sparse-checkout set '**/*.md' '**/.remark*' '**/*.json' '**/*.yaml' '.github/scripts' 'wiki/**/*' 'news/**/*'
```

.github/scripts/ci/find_broken_wikilinks.py
```diff
@@ -0,0 +1,262 @@
...
+        results.append(match)
+        match = find_link(s, match.link_end + 1)
+    return results
+
+
+def print_error():
+    print(f"{red('Error:')} Some wiki or image links in the files you've changed have errors.\n")
+    print("This can happen in one of the following ways:\n")
+    print("- The article or image that the link points to has since been moved or renamed (make sure to match capitalisation)")
+    print("- The image that the link points to has the wrong extension (must be one of .png, .jpg, or .gif)")
```
Walavouchey, line 200:
```suggestion
```

Walavouchey (commented):
not really. i suppose the sparse checkout was there to lower test times. but one reason for limiting to specific types is to encourage people to use consistent extensions (`.jpg` instead of `.jpeg`, `.png` instead of `.PNG`, or any use of other weird file formats), but that can be done in a (separate, also filename-checking) ci script if desired

.github/workflows/continuous-integration.yml
peppy: If you're looking to improve times (i don't think it's really required) you could also clone using a max depth of 1 see https://www.perforce.com/blog/vcs/git-beyond-basics-using-shallow-clones

.github/scripts/ci/find_broken_wikilinks.py
TicClick: on a second thought, this would suck. I think it's better if we agree to prune redirects from time to time

.github/scripts/ci/find_broken_wikilinks.py
```diff
@@ -0,0 +1,261 @@
...
+    else:
+        # some other link; don't care
+        return (True, "")
+
+
+def is_in_comment(s, start, end):
+    return s.rfind("<!--", 0, start) != -1 and s.find("-->", end, -1)
+
+
+def find_link(s: str, index=0) -> typing.Optional[Link]:
```
peppy, line 134: I'm just going to pretend I didn't read this code since it seems to work 🙈

.github/scripts/ci/find_broken_wikilinks.py
peppy: There's no need to keep redirects up-to-date (easing that overhead is what they're there for).

Walavouchey: wdyt about wording?

peppy (approved):

.github/scripts/ci/find_broken_wikilinks.py
Walavouchey: it's high-tech python that i'm not quite familiar with either

TicClick (approved):

PR #6646, opened 2022-01-04
Fix various links across the wiki

Walavouchey:
Fixes casing on various links, while also fixing links that incur more than one redirect

Specifically, this intends to fix most errors, that might have been missed in https://github.com/ppy/osu-wiki/pull/6641, as reported by the new and shiny [wikilink checker](https://github.com/Walavouchey/osu-wiki-tools/blob/master/find-broken-wikilinks.py) when set to check all files

## Self-check

- [x] `SKIP_OUTDATED_CHECK`


wiki/Rules/de.md
```diff
@@ -45,7 +45,7 @@ Diese Regeln beziehen sich ausschließlich auf Beatmaps, die über das [Beatmap
...
 2. **Komm in [Beatmapdiskussionen](/wiki/Beatmap_Discussion) nicht vom Thema ab.** Die Threads und Diskussionen gelten der Beatmap und sonst nichts. Wenn du ein Problem mit etwas hast, was sich nicht direkt auf die Beatmap bezieht, dann veröffentliche es stattdessen im richtigen Forum.
 3. **Plagiiere oder stehle nicht die Arbeit anderer Nutzer.** Lade keine Inhalte anderer Nutzer ohne deren expliziten Erlaubnis hoch, einschließlich Skins und Guest Difficulties.
-4. **Folge den [Regeln für Liedinhalte](/wiki/Song_Content_Rules) und den [Empfehlungen für visuelle Inhalte](/wiki/Rules/Visual_Content_Considerations).**
+4. **Folge den [Regeln für Liedinhalte](Song_Content_Rules) und den [Empfehlungen für visuelle Inhalte](Visual_Content_Considerations).**
```
peppy, line 48: What's the case where the `/wiki/` prefix is not required? ie. why is it removed here but not in `2.` above?

Walavouchey: these two as changed are relative links (`Song_Content_Rules` and `Visual_Content_Considerations` are in the `Rules` folder). `/wiki/` means absolute link, so as it was, it required a redirect.

The only reason this was reported was because of the redirect in `redirect.yaml` having the wrong casing (see 1b6f4b9), so it incurred two redirects. I left this fix on this line here while at it.

peppy: Not sure about relative links in general, I think absolute reads better, but I guess that discussion is one to be had elsewhere.

peppy (approved):

PR #6673, opened 2022-01-06
Make the freshness check standalone

TicClick:
also, don't fail on merge commits (example: https://github.com/ppy/osu-wiki/commit/cc5b3afae57fa34e876edcd52668b6bb3234bb4e)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -103,10 +103,7 @@ jobs:
             if ${{ github.event_name == 'pull_request' }}; then
               ${{ contains(github.event.pull_request.body, env.PULL_REQUEST_TAG) }}
             elif ${{ github.event_name == 'push' }}; then {
-              commits_url="${{ github.event.push.repository.commits_url }}"
-              suffix='{/sha}'
-              desired_suffix='/${{ github.sha }}/pulls'
-              pulls_url="${commits_url/$suffix/$desired_suffix}"
+              pulls_url='http://api.github.com/repos/${{ github.event.push.repository.full_name }}/commits/${{ github.sha }}/pulls'
               [[ $( curl -sS "$pulls_url" | jq -r '.[0].body' ) =~ $PULL_REQUEST_TAG ]]
             }
             fi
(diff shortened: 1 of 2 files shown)


Walavouchey:
seems to work ([error](https://github.com/Walavouchey/osu-wiki/runs/4737689997?check_suite_focus=true), [suppressed](https://github.com/Walavouchey/osu-wiki/runs/4737703497?check_suite_focus=true), [merge](https://github.com/Walavouchey/osu-wiki/runs/4737716718?check_suite_focus=true))

.github/workflows/continuous-integration.yml
```diff
@@ -97,61 +97,29 @@ jobs:
...
+              ${{ contains(github.event.pull_request.body, env.PULL_REQUEST_TAG) }}
+            elif ${{ github.event_name == 'push' }}; then {
+              commits_url="${{ github.event.push.repository.commits_url }}"
+              suffix='{/sha}'
+              desired_suffix='/${{ github.sha }}/pulls'
+              pulls_url="${commits_url/$suffix/$desired_suffix}"
+              pr_body=$( curl -sS "$pulls_url" | jq -r '.[0].body' || true )
+
+              # don't fail on network errors
+              [[ -z $pr_body ]] || [[ $pr_body =~ $PULL_REQUEST_TAG ]]
```
Walavouchey, line 113: Is this also related to pushes? I don't agree that it should pass on a network error.

Walavouchey: also shouldn't the script run on every commit, not just when pr'ed? just making sure i understand this correctly

TicClick:
I actually broke it in my attempts to make it as portable as possible, because `-E` only works for BSD `grep`, and `-P` for GNU `grep` (formatting with escape codes also seemed to work different, painting all lines on macOS and only the first one on Linux, which is observed in the PR you linked), but that is fixed in 427cddb.

as for network errors, I just wanted to make the check as stable as possible. that part is obsolete anyway since I've added `set -o pipefail`, which will propagate failures from `curl`

.github/workflows/continuous-integration.yml
TicClick: yeah it runs on every commit, just like other checks from `continuous-integration.yaml`

TicClick:
going to close the PR for now, as a) I acknowledge the changes don't do much, and b) I have slightly different plans

PR #6678, opened 2022-01-07
Use direct links with exact case instead of redirects

TicClick:
does not include:
- news
- help centre
- article styling criteria

generated with: https://gist.github.com/TicClick/8e160434511093616f33876493458aa1 (with a couple of `git checkout master -- wiki/directory/to/be/excluded` runs afterwards)

`SKIP_OUTDATED_CHECK` since some translations already have correct links


Walavouchey (approved):
replace

Walavouchey (approved):

PR #6679, opened 2022-01-07
Use direct links with exact case instead of redirects in the newsposts

TicClick:
- continuation of #6678
- generated with https://gist.github.com/TicClick/8e160434511093616f33876493458aa1


peppy:
If this is something that is 100% automatic (which it should be), it would be best to add it as a github workflow action with no trigger. We could either run it manually every few months or year, or setup a schedule for it.

Walavouchey (approved):

TicClick:
I'll look into it. currently it has a few shortcomings which, but more automation sounds like a good end goal

peppy (approved):

PR #6681, opened 2022-01-07
Update Discord channel list

Joehuu:
Reused the article title translations if they had one.

## Self-check


Walavouchey:
doesn't this need review for translations?

Joehuu:
This page would be updated accordingly when those articles get translated, I think that's how it should work in this case.

TicClick:
I'm ok with merging this if you add `no_native_review: true` to the translations which use English link titles

Walavouchey (approved):

PR #6692, opened 2022-01-09
Update `Countries that participated in osu! tournaments` for `MWC 7K 2022`

Bazzadwarf:
Does what it says in the title, so here's a related fun fact, MWC 7K 2022 marks the first Indian entrance into a World Cup!

## Self-check

### DIFF
--- a/wiki/Tournaments/Countries_that_participated_in_osu!_tournaments/en.md
+++ b/wiki/Tournaments/Countries_that_participated_in_osu!_tournaments/en.md
@@ -8,7 +8,7 @@ This article lists countries that have participated in official [osu!](/wiki/Gam
 
 ## Rankings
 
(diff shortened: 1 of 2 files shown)


Walavouchey:
it would be best to split the reordering of the rows into a separate commit, because the diff is unreadable rn which makes it take very long to check

Walavouchey:
nvm there was no reordering, github is just terrible at doing diffs
image

Bazzadwarf:
> it would be best to split the reordering of the rows into a separate commit, because the diff is unreadable rn which makes it take very long to check

Makes sense for the next time the page gets updated but yeah, no reordering is being done, github's diffs aren't terribly helpful for this kind of update.

Walavouchey (approved):

PR #6715, opened 2022-01-11
[DE] Update `People/The_Team` and `People/The_Team/osu!_Alumni`

respektive:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #6721, opened 2022-01-12
[DE] Add group info

respektive:
## Self-check


meta/group-info/de.yaml
```diff
@@ -0,0 +1,67 @@
+separator: ", "
+alumni:
+  roles:
+    academy: osu!academy-Gründer
+    admin: Administration
+    chat: Chat Moderator
```
The-Last-Cookie, line 6:
```suggestion
    chat: Chat-Moderator
```

meta/group-info/de.yaml
```diff
@@ -0,0 +1,67 @@
...
+    dev: osu!dev
+    media: osu!media Erstellung
```
The-Last-Cookie, line 8:
```suggestion
    media: osu!media-Erstellung
```

meta/group-info/de.yaml
```diff
@@ -0,0 +1,67 @@
...
+    support: Support-Team
+    tournaments: Verwaltung von Wettbewerben
+gmt:
+  all_mods: alle Moderatoren
+  areas:
+    beatmap: Beatmap-Moderation
+    chat: Chat-Moderation
+    forum: Forum-Moderation
+    team: offizielle Präsenz des Teams
+    dev: osu! Entwicklung
```
The-Last-Cookie, line 18:
```suggestion
    dev: osu!-Entwicklung
```

meta/group-info/de.yaml
```diff
@@ -0,0 +1,67 @@
...
+    player: Spieler-Support
+    support: Technischer Support
+    tournaments: Verwaltung von Wettbewerben
+    wiki: Wiki-Administration
+    mapping: Moderation der Mapping/Modding Community
```
The-Last-Cookie, line 23:
```suggestion
    mapping: Moderation der Mapping-/Modding-Community
```

meta/group-info/de.yaml
```diff
@@ -0,0 +1,67 @@
...
+nat:
+  areas:
+    communication: Kommunikation
+    evaluation: Beurteilung
+    moderation: Moderation
+    structural: strukturell
```
The-Last-Cookie, line 29: I've looked at the `NAT` article and there seems to be only one case where this needs to have another grammar form and that is the text which describes what this word actually means. I guess the mention in the text can be translated manually. The only thing that I would change is `Strukturell` instead of `strukturell`. You may want to take a look at the article yourself (`People\The_Team\Nomination_Assessment_Team`).

meta/group-info/de.yaml
```diff
@@ -0,0 +1,67 @@
...
+languages:
+  partial: irgendeine <Sprache>
```
The-Last-Cookie, line 31:
```suggestion
  partial: etwas <language>
```
In some articles, I saw that the `some` is used when someone speaks a language only a bit (maybe not a native but fluent enough for communicating). In tables, it would look like this probably:
```md
| ![][flag_PH] [osu user](user id) | some Filipino |
```
So I think `etwas` fits here more.

What I'm not sure about is the _key_ on the right, `<language>`. I don't know if the code implementation requires this to not be changed. Might need an answer from a dev here.

The-Last-Cookie (changes requested):

meta/group-info/de.yaml
respektive: i checked the other language files and noticed it has to stay `<language>` but forgot to change back, my bad

respektive:
I agree with your changes for the hyphens, i will have adjust them accordingly in https://github.com/ppy/osu-wiki/pull/6715

meta/group-info/de.yaml
respektive: Agree, I'll capatalize it.

The-Last-Cookie (approved):

cl8n (approved):

PR #6728, opened 2022-01-14
[SV] Clean up `Legal`

Walavouchey:
## Self-check

- ~~*(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)~~

I highly doubt anyone Swedish is around to review this

Some remarks:

- The alternative to using hyphens is to conjugate "osu!" to the genitive case, which has this wonderfully simple and atrocious syntax of "osu!:s copyright policy", "osu!:s musiklicensiering", etc.
- The other alternative is to translate it as "copyright policy in osu!" ("copyright policy i osu!") which doesn't quite make sense (although perhaps only for "musiklicensiering i osu!").
- Although hyphens might stylistically be a good compromise, the best alternative is to just remove the "osu!" prefix from the list entirely, but no other language seems to do that and I'm hesitant to make it inconsistent. The same could be done for "osu!-hemsidan".
- "copyright policy" (although ok) can be translated as "upphovsrättspolicy" and that'd be good and proper, but the site footer says "Copyright (DMCA)" in the Swedish locale (which again is ok, though it could also be "Upphovsrätt (DMCA)") so it'd be weird to deviate. Leaving as-is causes the weird "osu!-copyright policy" which I suppose is still technically correct.

### DIFF
--- a/wiki/Legal/sv.md
+++ b/wiki/Legal/sv.md
@@ -1,12 +1,8 @@
----
-no_native_review: true
----
+# Juridisk information
 
-# Legal
+Indexsida för alla juridiska artiklar. Dessa länkar finns även i sidfoten längst ned på osu!-hemsidan.
 
-Indexsida för alla juridiska artiklar. Dessa länkar finns också i sidfoten på osu! hemsidan.
-
-- [osu! Copyright Policy](Copyright)
-- [osu! Musiklicencser](Music_licensing)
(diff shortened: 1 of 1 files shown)


cl8n (approved):
i guess this PR is the "review" for no_native_review now

PR #6734, opened 2022-01-15
Fix broken link in `Main_Page`

ZeroPyrozen:
## Self-check


ZeroPyrozen:
I'm going to take a look on another translation files which haven't contained `osu! music licensing` article to make sure all link work

peppy:
What's the point of this? Unless it's going to be applied using a script to the whole wiki (which I would run myself) this doesn't seem productive.

ZeroPyrozen:
Can you check on this?

peppy:
This is due to a missing redirect. No idea how this happened, someone has fucked up bad.

PR #6745, opened 2022-01-16
Add LastExceed to `Users with unique titles`

ZeroPyrozen:
## Summary

Add notable osu!mania Fade In player and contributor of Fade In and Hidden implementation on osu!lazer ppy/osu#9538

## Self-check

- [x] Get confirmation from @WalterToro and @Ephemeralis


cl8n:
pretty sure this is more of a gameplay-related one, they're one of the only ppl that consistently play FI at high level

hopefully someone can confirm exactly when/why the title was given though.

peppy:
cc @Ephemeralis

Walavouchey:
if the lazer contributions were the reason for the title, then it should be listed below `Contributors` imo

peppy:
If you just need to know whether it's gameplay related then yes, it is.

ZeroPyrozen:
> If you just need to know whether it's gameplay related then yes, it is.

Is there any explanation about what LastExceed achieved to get this title? Like "Have the highest osu!mania performance point count which all played with Fade In mod as of ${dateTitleIssued}"

Ephemeralis:
The todo note (notable fade-in/hidden player & implemented fade-in in osu!lazer) about covers it. The dude is bonkers.

Ephemeralis (approved):

PR #6750, opened 2022-01-18
Fix active, passive, and storyboard hitsound definitions

Naxesss:
Currently, the wiki makes it seem as if "active" and "passive" determine the properties of the sound played, when it is rather determined by what part of a hit object it plays on.

E.g. an active hitsound is not defined by immediately reaching its peak impact, but rather by playing when the player clicks. The former is just a requirement of active hitsounds from the Ranking Criteria, and if it were defined as such we wouldn't need to enforce it.

Also separating out storyboarded hitsounds because while "passive" in some sense, they are often not referred to as such (in my experience) and behave differently enough for its own header.

(Doing `SKIP_OUTDATED_CHECK` as the ranking criterias themselves are still up to date.)

---

Mostly just clearing this up because I've seen people confused about it on multiple occasions, myself included.


wiki/Beatmapping/Hitsound/en.md
```diff
@@ -21,13 +21,15 @@ Beatmappers can replace any default sound samples with custom samples using a [b
 
 ## Active hitsound
 
-An active hitsound correlates to a player's clicking by reaching its peak impact immediately when played. Drum sound samples are the most frequently used active hitsounds because they provide clear and immediate feedback.
+A hitsound is considered *active* if it is on a clickable hitobject edge, for example a circle or slider head. These help the player judge whether they clicked early or late by producing a clear impact sound at the time of clicking. If clicked correctly, the sound should match up with the beats in the song the objects are mapped to.
```
cl8n, line 24:
```suggestion
A hitsound is considered *active* if it is on a clickable part of a hit object, for example a hit circle or slider head. These help the player judge whether they clicked early or late by producing a clear impactful sound at the time of clicking. If played correctly, the sound should match up with a beat in the song.
```
first change is more important i think, others are w/e just my preference

wiki/Beatmapping/Hitsound/en.md
```diff
@@ -21,13 +21,15 @@ Beatmappers can replace any default sound samples with custom samples using a [b
...
 
 ## Passive hitsound
 
-A passive hitsound does not correlate to a player's clicking. These are usually sound effects that accompany active hitsounds.
+If a hitsound is on a non-clickable edge, for example a slider repeat or spinner tail, it is considered *passive*. These always play at the same time as long as the player is interacting with the hitobject, unlike active hitsounds. Passive hitsounds neither need to have a clear impact nor be audible, but often are when over beats in the song to help the player maintain rhythm.
 
-Hitsounds can be added through storyboarding, but because they don't correlate to a user's clicking, they are also classified as passive.
+## Storyboarded hitsound
+
+Sound samples may be added through storyboarding to mimic hitsounds on objects. Unlike actual hitsounds, however, these always play at the given time regardless of whether the player hit the object. For that reason, they are often not recommended outside of osu!mania.
```
peppy, line 32: Why is mania a special case here?

wiki/Beatmapping/Hitsound/en.md
```diff
@@ -21,13 +21,15 @@ Beatmappers can replace any default sound samples with custom samples using a [b
 
 ## Active hitsound
 
-An active hitsound correlates to a player's clicking by reaching its peak impact immediately when played. Drum sound samples are the most frequently used active hitsounds because they provide clear and immediate feedback.
+A hitsound is considered *active* if it is on a clickable hitobject edge, for example a circle or slider head. These help the player judge whether they clicked early or late by producing a clear impact sound at the time of clicking. If clicked correctly, the sound should match up with the beats in the song the objects are mapped to.
 
 ## Passive hitsound
 
-A passive hitsound does not correlate to a player's clicking. These are usually sound effects that accompany active hitsounds.
+If a hitsound is on a non-clickable edge, for example a slider repeat or spinner tail, it is considered *passive*. These always play at the same time as long as the player is interacting with the hitobject, unlike active hitsounds. Passive hitsounds neither need to have a clear impact nor be audible, but often are when over beats in the song to help the player maintain rhythm.
```
cl8n, line 28:
```suggestion
If a hitsound is on a non-clickable part of a hit object, for example a slider repeat or spinner tail, it is considered *passive*. These always play at the same time as long as the player is interacting with the hit object, unlike active hitsounds. Passive hitsounds neither need to have a clear impact nor be audible, but often are when over beats in the song to help the player maintain rhythm.
```
also I've never heard people say "spinner tail". if im not just out of touch id pick a more common example, or reword that one

cl8n (changes requested):
looks good, main thing I noticed is that "edge" is a weird description for any place u can put a hitsound

wiki/Beatmapping/Hitsound/en.md
ZeroPyrozen: spinner end would be good

wiki/Beatmapping/Hitsound/en.md
cl8n: can probably split off the part abt mania into a second small paragraph with an explanation of why it's only recommended there (i dont know why myself but assuming it's relevant). otherwise it feels pretty random to include here

wiki/Beatmapping/Hitsound/en.md
Walavouchey, line 28: This last part reads a little awkwardly

```suggestion
If a hitsound is on a non-clickable edge, for example a slider repeat or spinner tail, it is considered *passive*. These always play at the same time as long as the player is interacting with the hitobject, unlike active hitsounds. Passive hitsounds need neither have a clear impact nor be audible, but often do in order to help the player maintain rhythm.
```

Walavouchey (changes requested):

wiki/Beatmapping/Hitsound/en.md
Walavouchey: Storyboarded hitsounds are way more common in mania than osu (from observation), and i think it's to counteract many hitsounds playing in short succession due to players not hitting multiple notes quite simultanously or quick stairs quite in time, etc (can sound bad on fast maps even if you're holding an SS)

peppy: Yeah I'm not sure. In mania you could just attach hitsounds to one column, that's usually the standard way. I'd probably drop the mania part.

wiki/Beatmapping/Hitsound/en.md
Naxesss: It is commonly used throughout the Ranking Criteria and is synonymous with "end", but I do think you have a point because its usage is a bit sporadic. Made [another PR to address that](https://github.com/ppy/osu-wiki/pull/6752).

wiki/Beatmapping/Hitsound/en.md
Naxesss: Mainly based this off of

image

but after asking a few mappers/modders from mania, it turns out they more or less just use them as a last resort when regular hitsounding is either not possible (e.g. to fill the gap between notes) or would require more than one object (e.g. for complex hitsounding or keysounding where multiple samples are used); probably for the reason Walavouchey described.

Either way, dropped the mania part.

wiki/Beatmapping/Hitsound/en.md
```diff
@@ -21,13 +21,15 @@ Beatmappers can replace any default sound samples with custom samples using a [b
 
 ## Active hitsound
 
-An active hitsound correlates to a player's clicking by reaching its peak impact immediately when played. Drum sound samples are the most frequently used active hitsounds because they provide clear and immediate feedback.
+A hitsound is considered *active* if it is on a clickable part of a hit object, for example a hit circle or slider start. These help the player judge whether they clicked early or late by playing a clear impactful sound at the time of clicking. If clicked correctly, the sound should match up with a beat in the song.
 
 ## Passive hitsound
 
-A passive hitsound does not correlate to a player's clicking. These are usually sound effects that accompany active hitsounds.
+If a hitsound is on a non-clickable part of a hit object, for example a slider repeat or spinner end, it is considered *passive*. These always play at the same time as long as the player is interacting with the hit object, unlike active hitsounds. Passive hitsounds do not need to have a clear impact or be audible, but often use the same sounds as active hitsounds when over beats in the song in order to help the player maintain rhythm.
```
Walavouchey, line 28: part of my concern was this wording of "when over beats in the song" that can be expressed more cleanly

```suggestion
If a hitsound is on a non-clickable part of a hit object, for example a slider repeat or spinner end, it is considered *passive*. These always play at the same time as long as the player is interacting with the hit object, unlike active hitsounds. Passive hitsounds do not need to have a clear impact or be audible, but often use the same sounds as active hitsounds when they land on distinct beats in order to help the player maintain rhythm.
```

Walavouchey (dismissed):
lgtm apart from the dilemma of whether to keep "spinner tail"

wiki/Beatmapping/Hitsound/en.md
cl8n: didn't expect my comment about "spinner tail" to turn into another PR, imo language about sliders is not related & doesn't matter cuz in different contexts people say different things

- "head" usually preferred over "start" for gameplay
- "end" *always* used for gameplay (you drop slider "ends", not "tails")
- "start"/"end" used for parts of skinning
- "start" relatively uncommon in mapping but "head" "end" "tail" all used

and ofc that^ changes over time and with different groups of people. just my experience

but across everyone I've talked to on osu I'm pretty sure I've never heard "spinner tail" which is why i think it should be "spinner end" or something. doesn't have to be consistent with slider language cuz they arent the same.

anyway, it's very minor detail, if u think it's fine as-is then w/e

peppy (dismissed):

cl8n:
I'm also gonna clean up some related links before merging this one, but I'll wait to see if naxess wants to make any more update or not

Naxesss:
Am good. Was a little unsure of whether to perhaps outdate the RCs after speaking with Eph so the hyperlink is eventually updated, but with the translations not actually being outdated it'd seem a bit strange to tag it as such.

cl8n (approved):

PR #6752, opened 2022-01-18
Consistently use "head/tail" in the Ranking Criteria

Naxesss:
I've noticed we use these terms synonymously throughout all Ranking Criterias and it's probably best to stick to one of the two for consistency.

(Doing `SKIP_OUTDATED_CHECK` as the change doesn't affect the meaning of the word.)

---

Stems from [this comment](https://github.com/ppy/osu-wiki/pull/6750#discussion_r786514724).


Walavouchey:
is "slider start" at all widely used? that one jumps out to me the most, but i'm not really a mapper

Coppertine:
Not really, it's more of the former than the latter.

Walavouchey (approved):

Joehuu (approved):

PR #6779, opened 2022-01-23
Update `MWC/7K_2022` with Quarterfinals information

LeoFLT:
## Self-check


LeoFLT:
Should be good to merge

Walavouchey (approved):

PR #6789, opened 2022-01-25
move and update "Timing_Songs_With_8-Signatures"

### LINKED ISSUE #6198: `Timing songs with #/8-signatures` should be in Guides instead of RC
it shouldn't mention RC at the beginning either, it doesn't set any rules and this article can stand on its own.
the first reference to it in RC is also misleading and can probably just be deleted

Gamingzone123:
closes #6198 

reference for the addition to lazer: https://github.com/ppy/osu/issues/2316

edit: github seems to have added more commits than i made here, i dont know why or how to fix


peppy:
Commit a fresh history and there's no need to use the "lazer" branch name. That's like saying "cutting edge will support in the future" when it's goign to be on stable anyway.

Gamingzone123:
can't figure out how to do fresh history

peppy:
Leave it to someone else to make the change in that case.

Gamingzone123:
figured it out

PR #6828, opened 2022-02-01
Update and simplify active hitsound rules in the Ranking Criteria

Naxesss:
For https://github.com/ppy/osu-wiki/pull/6750 I didn't realize that some rules in the Ranking Criteria used the terms under the old definition. This resulted in some confusion in [this mod discussion on a qualified map](https://osu.ppy.sh/beatmapsets/1585145/discussion/-/generalAll#/2871056) due to these rules implying _all_ active hitsounds must provide proper feedback when it should be _at least one_ (as it was before https://github.com/ppy/osu-wiki/pull/6750).

This both fixes that and simplifies wording such that looking up "active hitsound" is no longer necessary to understand these rules.

- [x] will fix outdating of translations in a bit
- [x] also going to ask a fellow nat to sanity check


wiki/Ranking_Criteria/en.md
```diff
@@ -182,9 +182,10 @@ This category contains explicit allowance statements of concepts and rules that
...
-- **All clicked objects must have audible [active hitsounds](/wiki/Beatmapping/Hitsound#active-hitsound).** osu!mania difficulties are exempt from this because of the mode's rhythm construction.
-- **[Active hitsounds](/wiki/Beatmapping/Hitsound#active-hitsound) must use the `.wav` or `.ogg` file formats.** `.mp3` files have slight delays, and therefore are reserved only for longer [passive hitsounds](/wiki/Beatmapping/Hitsound#passive-hitsound), such as ambient noises.
-- **[Active hitsounds](/wiki/Beatmapping/Hitsound#active-hitsound) cannot have a delay more than 5 milliseconds.** "Delay" accounts for the peak of a sound rather than its introduction. This ensures hitsound feedback is synchronised to a song effectively. The default skin's `normal-hitfinish.wav` has a slight delay, but still can be used as a custom hitsound.
+- **All clicked parts of objects must have at least one hitsound which both...**
```
TicClick, line 185: a passerby comment: the `clicked parts of objects` deserves a link to an article explaining what an active object is (and it's not `Glossary/Active object`, which is osu!catch only)

TicClick (approved):
looks good to me, feel free to have it merged by whomever checks it next

wiki/Ranking_Criteria/en.md
Walavouchey: same goes for similar references in `Beatmapping/Hitsound`

Walavouchey (approved):

Walavouchey:
if i understand this correctly, this reverts a rule change by technicality caused by #6750, and the combined change by both of these is just clarification of terms and removing ambiguity, correct? wording seems fine to me otherwise

Naxesss:
yeah #6750 + this = the same as before, #6750 just accidentally changed rules it wasn't supposed to change, which this fixes

Ephemeralis (approved):

PR #6834, opened 2022-02-02
Conclude `osu! South East Asia Tournament 5`

Nivalyx:
yup

also added all the missing entries for [CIC](http://osu.ppy.sh/wiki/en/Tournaments/CIC) in the Tournaments index page while i'm at it 🐇 


ZeroPyrozen (approved):

Walavouchey (approved):
force-pushed to fix misleading commit name

PR #6840, opened 2022-02-05
Ranking Criteria: Attempt to clarify song compilation mixing/transition guideline

Ephemeralis:
Ambiguity about what "properly mixed" means is causing people to ignore or avoid attempting to enforce this guideline. 

I have attempted to add a basic expression of expectation (aka: medley tracks having proper planned transitions over cross-fades where possible) and added a general expectation for overall cross-fade length when used (less than 5 seconds).


peppy:
I probably wouldn't even have the last sentence, but whatever works.

peppy (approved):

Ephemeralis:
The last sentence being ```Cross-faded transitions are acceptable, but should be used sparingly and generally be no longer than 5 seconds. Beat-matched and well-planned transitions are always a better option than cross-fading.``` I presume? Would you prefer that we avoid cross-fades entirely or?

Walavouchey:
Unless peppy actually meant the last two sentences, I can understand how the last one is slightly redundant since the preceeding one conveys essentially the same thing. That said, I don't have a particular opinion on whether the general recommendation in the last sentence should be there.

Walavouchey:
I see. The point of it is to clarify what "cleanly mixed with proper transitions" means. Imo the whole thing as-is works pretty well in that regard.

PR #6844, opened 2022-02-06
Update EBA description to be accurate

Flux3on:
Very very minor edit.

### DIFF
--- a/wiki/Storyboard/en.md
+++ b/wiki/Storyboard/en.md
@@ -1,7 +1,7 @@
 # Storyboard
 
 ![](img/Sbpassing.png "This scene is shown while playing Walkie Talkie Man; however, it is only shown when you get an Elite Beat 300")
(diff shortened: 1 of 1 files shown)


Walavouchey (commented):
Note that this indeed does apply to storyboards in osu, but this image depicts the corresponding mechanic in EBA. By reading the [wiki article on game states](https://osu.ppy.sh/wiki/en/Storyboard/Scripting/General_Rules#game-state), it appears that EBA used a slightly different system with 3 states, but I don't have an original copy of the game to verify whether this particular point would be correct.

ZeroPyrozen:
I've checked the Indonesian translation and it still has the same meaning as the changed part.

Walavouchey:
> I've checked the Indonesian translation and it still has the same meaning as the changed part.

Meaning... it was ambiguous?

ZeroPyrozen:
> Meaning... it was ambiguous?

My bad I was reading SBPassing image instead of SBFailing

PR #6872, opened 2022-02-10
Remove misleading line from the osu!mania hitsound RC

Kominaru:
The line in question was added by one of the mergers in https://github.com/ppy/osu-wiki/pull/6128, not being in the original PR. Since it's misleading and contradicts other parts of the RC (it would forbid mappers to have proper volume control for the still required hitnormal support), it's necessary to remove it (this was already agreed during discussion the after original PR was merged and was confirmed again recently).

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -181,7 +181,7 @@ This category contains explicit allowance statements of concepts and rules that 
 - **A beatmap may only contain one song file used by all difficulties.** Multiple song files within a single beatmap are unsupported and result in unexpected behaviour with preview times, metadata, etc.
 - **A song's audio file and hitsound files must be of reasonable quality.** Try to find the highest quality source file available rather than ripping a file from a streaming video website. Songs should be normalised to their original release volumes and should not be encoded to a bit rate higher than their original files.
 - **Beatmaps must be hitsounded.** Hitnormals give feedback to the player, and additions (whistles, claps, and finishes) accent the most important parts of the music.
-  - **osu!mania beatmaps do not require hitsound additions.** This is to allow for easier approachability to osu!mania mappers of different upbringings. It is still highly recommended to add hitsounds to improve the feel of your beatmaps. In cases where hitsounds are not used, *no* additions may be placed, and *no* sample control points should be used (ie. to adjust volume).
+  - **osu!mania beatmaps do not require hitsound additions.** This is to allow for easier approachability to osu!mania mappers of different upbringings. It is still highly recommended to add hitsounds to improve the feel of your beatmaps.
 - **All clicked parts of objects must have at least one hitsound which both...**
   - **...has a clear impact, whose peak is delayed no more than 5 milliseconds.** `normal-hitfinish.wav` from the default skin is exempt from this.
   - **...uses the `.wav` or `.ogg` file format.** `.mp3` should not be used here as it is inherently delayed.


Walavouchey:
> this was already agreed during discussion the after original PR was merged and was confirmed again recently

mind linking those discussions?

Kominaru:
> mind linking those discussions?
There you have
https://discord.com/channels/188630481301012481/542532524388384784/897675926966509619
https://discord.com/channels/90072389919997952/299846395031060480/935239667425624094

peppy:
I can't understand the sentence which is being removed. So if it is intended to remain, then it needs to be reworded. Specifically: "hitsounds" and "additions" are synonymous, so it doesn't make sense to forbid using "additions" when "hitsounds" aren't used.

If this is specifically about making volume adjustments when hitsounds aren't being used, then it should be reworded to focus on only that case.

Walavouchey:
> I may be wrong but Im pretty sure you had been the one to add this (or have it added),

peppy added this sentence in https://github.com/ppy/osu-wiki/commit/55226cf7f86890652bc5ef33367c6fc920170419 (after merging https://github.com/ppy/osu-wiki/pull/6128) and posted an explanation [here](https://discord.com/channels/188630481301012481/542532524388384784/897674449669730324)

Kominaru:
To sum up the situation, it is important to clarify that 
> Specifically: "hitsounds" and "additions" are synonymous

is not true, as it is clearly stated throughout the Ranking Criteria, and specifically right before the proposed line removal, in the general rule: 

> Beatmaps must be hitsounded. **Hitnormals** give feedback to the player, and **additions** (whistles, claps, and finishes) accent the most important parts of the music.
 · osu!mania beatmaps do not require hitsound **additions**. (...)
 
It is optional the use of **additions**, not of hitnormal or hitsounds of a whole (that is, hitnormal support wth proper volume is still mandatory in order to provide active feedback for players). This is the intended meaning of the original PR's wording, **and also the wording that was used in the rule that existed prior**. `Hitsounds = hitnormal + additions`, as stated just above the rule in question. 
 
 The added line has three problems/contradictions and adds no value which makes it necessary to remove it:

-  `In cases where hitsounds are not used` **Hitsounds** are always going to be present in beatmaps as at least a custom **hitnormal** is still always required.
- `no additions may be placed` Redundant with the wording of the rule
- `and no sample control points should be used (ie. to adjust volume).` These are still necessary and mandatory (if needed depending on the song) to control the volume of the required hitnormal support.

If absolutely required, we can replace (not reword) the line in question with something verbose/explanatory like `Custom hitnormal support is still required, with correct volume adjustment throughout the song`, but then again, it shouldn't be needed as it is already stated in the general rule and the previous rule didn't need it and never caused confusion despite being more complex/having more cases.

peppy:
The volume rule should remain in that case, but the addition thing is weird and makes no sense (even with the above explaining of it "not being synonymous" which i disagree with).

Kominaru:
As we discussed in Discord when you added this line, removing it doesn't allow mappers to "avoid having audible feedback" or "make hitnormal silent", it actually makes them have proper volume control, which is something we're enforcing already as this line poses a contradiction with the rest of the RC. Proper volume control is covered by the rule right next to this one in the RC, which reads as follows:

```
Hitsounds must be audible. Their purpose is to provide feedback, so hitsounds with extremely low volume or samples that blend with a song's samples are unacceptable.
```
because hitsounds is NOT 1:1 synonymous with additions (it never has been, its literally stated in the RC that hitsounds = hitnormal + additions). 

No other wording is needed because the only necessary thing is to remove the line

peppy:
Makes sense. My only remaining concern is that 100% of mappers are going to add one sample control point at 5% volume for every map stating "better experience". That's why I added the clause in the first place. Can we add coverage somewhere (if not already) requiring at least one sample control point have volume >80% or something like that? It will allow mappers to adjust volume for different sections of the beatmap without applying uniform changes that leave the beatmap in a "pre-customised" state.

The terminology is weird IMO, might want to consider changing that.

"Hitsounding" is the art of designing hitsounds around the map. Obviously it's impossible to "turn off" hitsounds, so there's no requirement of calling "adding additions" anything other than "hitsounding", which is more understood by the community no?

Walavouchey:
For some context, the terminology of "additions" referring to whistles, finishes and claps likely comes from years of being accustomed to how the stable client and surrounding documentation have called them.

The editor calls them additions (likely since the beginning):

screenshot

The file format specification calls them additions (has since the wiki used wikimedia):

20220214T080555-9194618473

The hitsound wiki article (among related ones) refers to them as additions ([original written by pishifat 2 years ago](https://github.com/ppy/osu-wiki/commit/39a0f913a484c2f119e03436c074d152b2e70659#diff-528d6705a183cf89c6c83821916f7e478bfeb34f55beafbd5151e1f9e10a7432)):

20220214T080911-5443879861

Having a term that specifically refers to whistles, claps and finishes is very useful imo, since unlike hitnormals, they have buttons in the editor that let you toggle them for each hit object. They quite simply act as "additions" on top of the "hitnormal" (two concurrent sounds).

For the ranking criteria (and elsewhere), we could of course remove any mention of "hitsound additions" and use "whistles, claps, and finishes", which to me sounds needlessly cumbersome. Unless someone has a suggestion for an alternative name, hitsound addition is as good of a term as is.

Kominaru:
> My only remaining concern is that 100% of mappers are going to add one sample control point at 5% volume for every map stating "better experience".
Mappers have never been allowed to do this before, not even when additions became optional in songs above a certain length for mania. We already have this rule in place:

> Hitsounds must be audible. Their purpose is to provide feedback, so hitsounds with extremely low volume or samples that blend with a song's samples are unacceptable.

Which ensures hitsounds always have just the right volume. Enforcing a certain volume (i.e. >80%) doesn't work because different songs require lower or higher hitsound volume depending on the song's own volume, intensity, percussion, etc, and mappers and BNs fit the hitsound volume to be audible in that context without overpowering the music, which is crucial.

peppy:
As an idea of a path forward: in lazer we are going to add replaygain support, at which point hitsounds will generally be required to be set to 100% for new beatmaps (the game will manage the relative audio level). But this is still a bit far away.

peppy (approved):

PR #6873, opened 2022-02-10
Clarifying AIMod unsnap rules in General RC

AJT127:
Seeking to clarify unsnap rules regarding false positives where AIMod says something less than 2ms from the tick is unsnapped, following long-term discussion in #nat and confusion amongst BNs/users as to how to handle these cases.

## Self-check

(https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)


Naxesss:
For context the current rule relies on AiMod pointing out >= 2 ms unsnaps. However, for slider ends AiMod also erroneously points out > 1 ms. Modern tools do not produce the same false positive, so it's quite obvious for BNs when it happens, [leading to confusion of whether these edge cases are rankable](https://osu.ppy.sh/beatmapsets/1217044/discussion/2644766/general#/2889466).

It's been generally agreed that ignoring the false positive should be allowed, hence this.

peppy:
Wonder if this is something we can hotfix on stable? Is it always slider ends? if you can provide an example beatmap I'll look into it (seems like a better alternative to having this wording change).

AJT127:
can also happen with reverse arrows e.g. https://osu.ppy.sh/beatmapsets/1217044#osu/2644766 Insane diff 00:43:155 (7) - I'm guessing that's also kind of a package deal with the sliderend also flagging as unsnapped though

peppy:
Does that imply that the lazer implementation is also wrong? Fixing it there is probably the first point of interest, if so..

Naxesss:
The lazer implementation is correct and equivalent to the one in Mapset Verifier. "Negligible" implies that the issue is so small or unimportant as to be not worth considering in the vast majority of cases (in this case not required by the RC), hence hidden by default. So unless you check the negligible checkbox in lazer (as I did in the screenshot), it will purposefully not detect < 2 ms unsnaps.

AiMod is the only one that has the same severity in these cases as for >= 2 ms unsnaps.

peppy:
I've pushed a change to cutting edge with changes that seem to be correct. Could people please test this and confirm?

AJT127:
it still seems to erroneously flag https://i.imgur.com/BdvXRT4.png as unsnapped on the backstabber map

peppy:
> it still seems to erroneously flag https://i.imgur.com/BdvXRT4.png as unsnapped on the backstabber map, but the other warnings are gone now

Thanks, I've fixed this one as well. Please give it another shot.

Naxesss:
Seems to work properly now.

peppy:
Are there any remaining issues or can we close this now?

Naxesss:
AiMod being iffy with unsnaps was the only reason for this PR, so now that that's fixed we can close this.

wiki/Ranking_Criteria/en.md
```diff
@@ -182,7 +182,7 @@ This category contains explicit allowance statements of concepts and rules that
...
 - **A beatmap's first [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) cannot be used to toggle [kiai](/wiki/Gameplay/Kiai_time).** Doing this will cause the kiai to flash before objects appear. An [inherited point](/wiki/Client/Beatmap_editor/Timing#inherited-timing-point) in the same position as the first uninherited point must be used to toggle kiai instead.
-- **Objects must be snapped to timeline ticks according to [AiMod](/wiki/Client/Beatmap_editor/AiMod).** Objects in a musical section requiring unsupported beat snap divisors (e.g. 1/11) can either:
+- **Hit objects must be snapped within less than 2 ms of any timeline tick.** [AiMod](/wiki/Client/Beatmap_editor/AiMod) will report these issues, as well as rare false positives. Objects in a musical section requiring unsupported beat snap divisors (e.g. 1/11) can either:
```
NoffyNoffeh, line 185:
```suggestion
- **Hit objects must be snapped within less than 2 ms of any timeline tick.** [AiMod](/wiki/Client/Beatmap_editor/AiMod) will report these issues, as well as rare false positives. False positives mainly occur on slider ends and reverses, and should be verified manually or with other tools. Objects in a musical section requiring unsupported beat snap divisors (e.g. 1/11) can either:
```

Naxesss: This seems like a roundabout way of suggesting:

```suggestion
- **Hit objects must be snapped within less than 2 ms of any timeline tick.** Use [MV](https://github.com/Naxesss/MapsetVerifier) or the lazer client to detect these issues accurately. Objects in a musical section requiring unsupported beat snap divisors (e.g. 1/11) can either:
```

Which is more concise and skips the hassle with AiMod entirely. AiMod may be fine for detecting the majority of these cases, but with it being unmaintained and outdated, I don't think it's something we should promote in the RC at all.

Walavouchey: ^ I agree that this would work better

Walavouchey (approved):

peppy (approved):

PR #6890, opened 2022-02-16
Add "Tools" section to the Ranking Criteria

Naxesss:
The purpose of this section is to inform people of where to find and things to understand about using modern tools that aid in checking the Ranking Criteria. These have been integral to the ranking process for a long time now, and has seen use by not just Beatmap Nominators, but also modders and mappers who recommend  them to each other ([one](https://osu.ppy.sh/beatmapsets/1674456/discussion/3426226/general#/2896002), [two](https://osu.ppy.sh/beatmapsets/1676024/discussion/-/generalAll#/2874983), [three](https://osu.ppy.sh/beatmapsets/1593558/discussion/-/generalAll#/2823464/7809264), [four](https://osu.ppy.sh/beatmapsets/1642077/discussion/-/generalAll#/2864607), [five](https://osu.ppy.sh/beatmapsets/1664196/discussion/-/generalAll#/2820754/7677896), etc), so I figured this should probably exist here.

- [x] outdate translations


Ephemeralis (approved):
Looks good to me - paging @peppy in just to make sure he's okay with wording though

It's actually kind of astonishing that this wasn't added when its use is practically ubiquitous among mappers/modders at this point.

Walavouchey (approved):
seems good to me as well

Naxesss:
lazer is missing many checks that mv has currently, and the ones that lazer has also exist in mv, hence neither instead nor alongside respectively - that will eventually change ofc once I've fully ported everything

peppy (approved):
seems fine for the time being

PR #6893, opened 2022-02-17
Update community meeting date

Walavouchey:
This also brings all translations up to date

For Traditional Chinese, I've simply copied existing meeting notes (and double-checked with machine translation) so there should be no error

## Self-check


peppy (approved):
Thanks 🙏

PR #6898, opened 2022-02-18
Update medal count in Medals

chromb:
## Self-check


peppy (approved):
Matches databased numbers.

Joehuu (approved):

PR #6900, opened 2022-02-18
Updated 'The Team' wiki listings and Staff Log

Blooshing:
## Self-check

- [x] copy changes to translations


peppy:
Feels like this should be automated listings rather than manually updated each time..

Walavouchey:
if only they were automated pages like the user group listings... (e.g. like https://osu.ppy.sh/groups/28)

Walavouchey:
there's an issue tracking this at https://github.com/ppy/osu-web/issues/6216

cl8n:
`osu-wiki translate-groups` from <https://github.com/cl8n/osu-wiki-bin> can do the translations at least, which is the most tedious part. it's why `meta/group-info` exists in this repo. all the past people who regularly updated group pages used this, and the wiki pages should probably include a comment saying you don't have to do this by hand...

as for osu-web progress, Staff Log is close to having a replacement, I just need to put together a frontend and do a final pass over everything. most of the code is already in master.
the reason group listings are duplicated to osu-wiki is bc some extra info has nowhere to go on osu-web: areas of focus, other fluent languages, etc

Walavouchey (approved):

PR #6906, opened 2022-02-20
Update community meetings page with latest meeting

Walavouchey:
wording copied from previous commits. still need agenda translations

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [ ] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)
  - [x] @DoryanR @Biokax (fr)
  - [x] @ZeroPyrozen (id)
  - [ ] @SiderealArt (zh-tw)
  - [x] @respektive (de)

PR #6912, opened 2022-02-21
Add entry for latest community meeting

peppy:
Not sure how this one should be listed, but best to get it visible on the page for those that are expecting to be able to access it from here.

### DIFF
--- a/wiki/Community/osu!_community_meetings/de.md
+++ b/wiki/Community/osu!_community_meetings/de.md
@@ -1,3 +1,8 @@
+---
+outdated: true
+outdated_since: 8032d6f5965191467a19538d5cf535c3c4c0df81
+---
+
 # osu! Community Meetings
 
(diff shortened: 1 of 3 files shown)


peppy:
@Ephemeralis not sure if you plan on summarising the meeting into some form of "meeting notes", but highlighting you in case that's going to happen. Not sure how important it is.

Joehuu:
There's already a PR adding this: https://github.com/ppy/osu-wiki/pull/6906

ZeroPyrozen:
There is on-going PR on #6906

peppy:
The title of that PR meant that even after checking the PR I did not see it. Have you considered naming the PR using a standard english prose rather than including an underscored filename? I think it reads better.

Also the language prefixes are pretty ugly, labels would work much better but I guess they are there for people that can't adjust labels. For cases where the PR is updating every language it's not really necessary to include `[AA, AB, AAC, AD, AD, AF,AF, AG]` though, IMO.

Issue #6207, opened 2021-10-09
Change tournament mappool list style 


Bazzadwarf:
Currently, tournament articles use double bulleted lists for mappools. [MWC 4K 2021](https://osu.ppy.sh/wiki/en/Tournaments/MWC/2021_4K) used numbered lists for the maps in its mappool, and the feedback seemed to be positive as it better correlates to how players refer to map picks in practice, mod abbreviation + number `NM5`, `DT3`, `FM1`, etc.

Although I don't have any preference, it would be a worthwhile change if this better reflects how mappools are represented and used.

Relevant discussion: https://github.com/ppy/osu-wiki/pull/6206#discussion_r725505257

# Progress

Articles and other resources that need updating

- [X] Official tournaments https://github.com/ppy/osu-wiki/pull/6495
- [X] osu! (Community) https://github.com/ppy/osu-wiki/pull/6573
- [X] Taiko (Community) https://github.com/ppy/osu-wiki/pull/6573
- [X] Catch (Community) https://github.com/ppy/osu-wiki/pull/6573
- [X] Mania (Community) https://github.com/ppy/osu-wiki/pull/6573
- [X] Multimode (Community) https://github.com/ppy/osu-wiki/pull/6516
- [X] Tournament article template https://github.com/ppy/osu-wiki/pull/6496

# Current
```
- NoMod
  - [LiSA - Confidence Driver (Skyflame) \[Assurance\]](https://osu.ppy.sh/beatmapsets/1597778#osu/3263082)
  - [IRON ATTACK! - INEXPERIENCE (DeviousPanda) \[EXTRA STAGE\]](https://osu.ppy.sh/beatmapsets/1597785#osu/3263098)
  - [Mio Yamazaki - Byoushin Zenkai Girl (Deca) \[Collab Extra\]](https://osu.ppy.sh/beatmapsets/1597860#osu/3263248)
  - [Camellia - Heartless being (Mir) \[Expert\]](https://osu.ppy.sh/beatmapsets/1597801#osu/3263131)
- Hidden
  - [BUTAOTOME - Sharekoube (Moph) \[Extreme\]](https://osu.ppy.sh/beatmapsets/1597784#osu/3263097)
  - [ZUN - Primodrial Beat \~ Pristine Beat (Halfslashed) \[Extra Stage\]](https://osu.ppy.sh/beatmapsets/1597824#osu/3263174)
- HardRock
[...]

PR #6920, opened 2022-02-23
Conform mod names according to ASC across the wiki

Walavouchey:
bit of a chonker pr

The [terminology section of the asc](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Writing#terminology) mandates that mods must be spelt a specific way (like in the game), and I can't help but agree.

I've replaced all occurrences by regex using this script below, and manually restored wrong changes while scrolling through the diff (links, filenames, etc.).

<details><summary>script</summary>

```py
import os
import regex

def file_iterator(roots: list):
    for item in roots:
        if os.path.isdir(item):
            for root, _, filenames in os.walk(item):
                for f in filenames:
                    filepath = os.path.join(root, f)
                    yield filepath
        elif os.path.isfile(item):
            yield item

replacements = [
    (regex.compile(r"No ?Fail", regex.IGNORECASE), "No Fail"),
    (regex.compile(r"Half ?Time", regex.IGNORECASE), "Half Time"),
    (regex.compile(r"Hard ?Rock", regex.IGNORECASE), "Hard Rock"),
    (regex.compile(r"Sudden ?Death", regex.IGNORECASE), "Sudden Death"),
    (regex.compile(r"Double ?Time", regex.IGNORECASE), "Double Time"),
    (regex.compile(r"Fade ?In", regex.IGNORECASE), "Fade In"),
    (regex.compile(r"Target ?Practice", regex.IGNORECASE), "Target Practice"),
    (regex.compile(r"Spun ?Out", regex.IGNORECASE), "Spun Out"),
    (regex.compile(r"Touch ?Device", regex.IGNORECASE), "Touch Device"),
    (regex.compile(r"Score ?V2", regex.IGNORECASE), "ScoreV2"),
    (regex.compile(r"Night ?core", regex.IGNORECASE), "Nightcore")
]

for file in file_iterator("."):
    if not file.endswith(".md"):
        continue
    handle = open(file, 'r', encoding='utf8')
    content = handle.read()
    for replacement in replacements:
        content = replacement[0].sub(replacement[1], content)
    handle.close()
    handle = open(file, 'w', encoding='utf8')
    handle.write(content)
    handle.close()
```

</details>

### Remarks

- ScoreV2 is a mod, as well as the scoring system, but the in-game multiplayer scoring mode on stable says "Score v2". I think that looks ridiculous, and probably some tournament article writers as well (they've often used "Score V2"), but I've changed it to match the in-game format.
  - cf8f1b9326e6c6784de06dbb15e43476ab6e3239 was done semi-automatically with some intellectual guesswork for the translations. You can skim it to understand the logic used (again, ScoreV2 for scoring system and "Score v2" for multiplayer setting). It also reverts incorrect changes to links from the first commit.
- I've skimmed through the whole diff and reverted any unintended changes and `grep`'d for potential mistakes. If we want to be more rigorous, I could programmatically check this in a similar way to how I reviewed the pulls for the [mappool list style change](https://github.com/ppy/osu-wiki/issues/6207), by doing checks on the `git diff` output to see if changes were made in mappools, profile links, etc.

More search-and-replaces like this probably to follow, for other terms the asc lists (probably not as involved as this, which was mostly tournament articles).

This can't quite be a ci check, since some of this depends on context (Fade In for example shouldn't replaces "fade in" or "fade into" in writing, or "Hard Rock" with "hardrock" (genre)), but the effort required to programmatically replace stuff like this once in a while isn't that much (well it was a just a few hours for me), and contributors and reviewers will hopefully be more aware over time after I start pointing it out from now on.

~~For tournament article writers, you may want to consider this a PSA to update any automatic tooling you use to generate mappools.~~

`SKIP_OUTDATED_CHECK`


Joehuu:
I thought the tournament formatting uses upper camel case.

ASC:
> When writing out game modifiers for tournament articles, they must instead use upper camel case (omit the space and keep the letter casing as seen above).

As for score v2, there's the string from [web](https://github.com/ppy/osu-web/blob/9da408eafdfccd683fd5cceb0e40c10d68812c09/resources/lang/en/beatmaps.php#L322) (`Score V2`) that's used for the multiplayer pages (e.g. https://osu.ppy.sh/community/matches/94510355). Not sure if we should use that instead of the wiki one.

Walavouchey:
> I thought the tournament formatting uses upper camel case.
>
> ASC:
>
> When writing out game modifiers for tournament articles, they must instead use upper camel case (omit the space and keep the letter casing as seen above).

Hmm, must've missed that part. I'm not entirely sure why that is the case though, but I can revert either the mappool parts or all mentions of mods in tournament articles (which to me would seem a bit off).

> As for score v2, there's the string from [web](https://github.com/ppy/osu-web/blob/9da408eafdfccd683fd5cceb0e40c10d68812c09/resources/lang/en/beatmaps.php#L322) (`Score V2`) that's used for the multiplayer pages (e.g. https://osu.ppy.sh/community/matches/94510355). Not sure if we should use that instead of the asc one.

Ok that is just confusing. Shouldn't there be some kind of standard as to which format to use?

Will just block this for now. Does @peppy have an opinion?

peppy:
Just to confirm, is this over whether the case should be 'v' or 'V'? Or also about the space?

Walavouchey:
Both

`Score v2` is the thing you see under the win condition dropdown in-game. `ScoreV2` is the mod name in-game, `ScoreV2` (along with `ScoreV1`) is the scoring method. Web seems to use `Score V2` (not consistent with how it is displayed in-game).

Should either be `Score V2` (implies `Score V1`) or `ScoreV1` across the board imo.

peppy:
You mention in your opening post that "Score v2" looks ridiculous – are you referring to the lowercase there or the space? If you think it's okay with the space, then let's go with the capitalised version. Rather than making huge changes to the wiki I'm happy to update stable to make it capital, if that is something you see as required. Else go with osu-web has, or propose a change.

Walavouchey:
Ah, yes the lowecase v looks off, since that's the only place it's written like that.

Going with osu-web contradicts the asc a bit though, since web refers to the mod as `Score V2` while stable refers to it as `ScoreV2`. Haven't checked lazer yet.

peppy:
Lazer doesn't (and won't) have a score v2 mod. I'd say we should go with the version with a space and update the mod name if anything.

Walavouchey:
Traced the asc tournament exception to https://github.com/ppy/osu-wiki/commit/c0729908df87cc45a3ba4c62417e2b64189f2a03

Joehuu:
> Traced the asc tournament exception to [c072990](https://github.com/ppy/osu-wiki/commit/c0729908df87cc45a3ba4c62417e2b64189f2a03), although i can't find the review it came from in the corresponding pr. do you know the reason why this was added @Joehuu?

The review that stirred that change was in https://canary.discord.com/channels/188630481301012481/218677502141399041/446298212937105409:

Screen Shot 2022-02-23 at 1 29 49 PM

but was added with no conservation, as I think that was a common format for tournaments at that time? It was originally worded as an option, but it changed to a "must" because of maybe consistency reasons?

I think the reason for that format of the first place is because stable formats it like that. When I did that review, I wanted it to be consistent with web and lazer which now adds that space.

The rule can be removed as there was no real reason other than that was a common format at that time.

Walavouchey:
I see. I'll also remove the asc rule in a separate pr then, and move forward with the replacing here

Walavouchey:
closing this in favour of superseding it later

PR #6924, opened 2022-02-23
Add guideline regarding long note length to o!m RC

Kominaru:
This guideline proposal is the product of the initial RC Proposal in the osu! forum during the past November: https://osu.ppy.sh/community/forums/topics/1463447. This appeared in response to widespread community complaints about the extremely short long notes in a qualified map, which were unreasonable to hit and release on time, even on simple break sections. 

The object of this proposal is to act as a baseline to indicate the minimum length warranted for long notes in osu!mania, following overwhelming community and BN/NAT support. 

We have determined that the presence of a long note release timing, together with possible limitations due to the hardware/software used, makes very short LNs (shorter than ~20-25ms) unreasonable to be hit and released with good accuracy even if they appear isolated, which makes them unsuitable for the ranked section.


Ephemeralis:
"Extremely short long notes" made me chuckle then wonder if that is even a remotely correct term to accurately describe what is actually going on.

Kominaru:
> "Extremely short long notes" made me chuckle then wonder if that is even a remotely correct term to accurately describe what is actually going on.
@Ephemeralis 
Now that you say that ye, the term "extremely" sounds stupidly funny lol
What would be better, "excessively"? Having trouble finding anything better to be fair

Walavouchey:
> Now that you say that ye, the term "extremely" sounds stupidly funny lol

slight whoosh, but "extremely *short long* notes" is an oxymoron

could alternatively use "hold note" for the whole mania rc instead

Kominaru:
> could alternatively use "hold note" for the whole mania rc instead

That's out of the question to be fair, no one uses that term and everyone uses long note. `short long notes` shouldn't be much of an issue though.

Kominaru:
Perhaps something more on the line with how length of LNs is described in the difficulty specific RC could be used. i.e. we could do:

> **Long notes should be held for at least 1/12 of a beat**. Faster release timings are unreasonable to play with full accuracy due to software or hardware limitations.

Walavouchey:
I'm not against either wording. They both sound alright to me

Kominaru:
Changed wording to 

> **Long notes should be held for at least 1/12 of a beat.** Faster release timings are unreasonable to play with full accuracy.

after eph's feedback. Should be ready to merge now

Ephemeralis (approved):

PR #6925, opened 2022-02-24
Replace winner badge image for osb! Storyboarding Contest #1

ZeroPyrozen:
## Self-check


Walavouchey:
[TheDuckMask's profile](https://osu.ppy.sh/users/7405768)

image

Walavouchey (approved):

Walavouchey:
Was there something @Ephemeralis needed to check here? It looks correct - the current image must've been an old revision or something.

peppy:
It's fine if you're confident, just seems weird that badges are changed after the fact. If it matches the one on the profile it should be fine.

Walavouchey:
My brain must've somehow thought I was looking at the same picture when reviewing, my bad.

PR #6929, opened 2022-02-25
Update Skinning Contest #2 Wiki Page

RockRoller01:
Really hope i didnt make a mistake with the flags.. would be great if someone could check

Currently is missing the podium image, can not be merged before this is added. Also needs to wait until newspost is out

## Self-check


RockRoller01:
I have accidentally included the newspost commits... closing and reopening another PR without them

PR #6942, opened 2022-02-28
Merge Muzukashii break proposal

Capuosu:
It has been discussed for a while now to change the wording of the rest moment guideline for specifically Muzukashii difficulties in taiko. This is to make it more clear to everyone what's actually accepted as a substitute, since a 3/2 break can sometimes feel fairly unintuitive.

The discussion can be seen here https://osu.ppy.sh/community/forums/topics/1325267?n=89

## Self-check


wiki/Ranking_Criteria/osu!taiko/en.md
```diff
@@ -138,7 +138,7 @@ If a Futsuu difficulty is required and used as the *lowest difficulty* of a beat
 - **1/4 patterns should not be longer than five notes.** Anything longer is likely to be too straining for intermediate players.
 - **Main snapping should consist of mostly 1/2, 1/1 or slower rhythms.** Occasionally using 1/4 rhythms is acceptable.
 - **There should be at least 1/2 distance between a spinner and its preceding note.** This is to ensure they do not overlap substantially and assure readability.
-- **At least 1 rest moment that is 3/2 or longer should be inserted after 16/1 to 20/1 of continuous mapping.** Using rest moments less frequently is acceptable if either the pace of the music makes rest moments counter-intuitive or if the continuously mapped part is overall more forgiving to the player.
+- **At least 1 rest moment that is 3/2 or longer should be inserted after 16/1 to 20/1 of continuous mapping.** Using **at least 3 consecutive rest moments that are 1/1 is ->an acceptable substitute<-** if either the pace of the music makes rest moments counter-intuitive or if the continuously mapped part is overall more forgiving to the player.
```
Walavouchey, line 141:
```suggestion
- **At least 1 rest moment that is 3/2 or longer should be inserted after 16/1 to 20/1 of continuous mapping.** Using at least 3 consecutive rest moments that are 1/1 is an acceptable substitute if either the pace of the music makes rest moments counter-intuitive or if the continuously mapped part is overall more forgiving to the player.
```

Walavouchey (changes requested):

Walavouchey:
Bolding a second time is not necessary, since it is part of the clarification/exception and not the main guideline.

wiki/Ranking_Criteria/osu!taiko/en.md
Capuosu: Hope I'm doing this right, already removed the arrows and will also get rid of the bolding

peppy (approved):
Seems okay to me, but both of these are pretty hard to mentally parse. Not sure how mappers and modders are able to handle this.

Walavouchey:
I'm sure mappers and modders have a better idea on the terminology - they came up with the suggestion after all

Walavouchey (approved):

Ephemeralis:
There has got to be a way to convey this information in a better format, but I suppose that is a far larger task than this single PR. Will merge for now.

Ephemeralis (approved):

PR #6959, opened 2022-03-05
Remove mention of the next community meeting date and other clean-up

Walavouchey:
updating live to the next date is no longer necessary since the introduction of the events feature on discord, which is faster to update (apart from twitter)

## Self-check

this pr was done entirely on a phone using the github interface. 6/10 wouldn't do again


peppy (approved):
Seems okay

PR #6969, opened 2022-03-06
Fix formatting in mappool link in `COT22`

Walavouchey:
small thing i noticed while checking https://github.com/ppy/osu-wiki/pull/6968 (error not from that pr)

image


TicClick (approved):
\*squints\*

PR #7016, opened 2022-03-15
Start to clean up `Glossary`

Gamingzone123:
moving articles out of glossary and into proper locations

not sure where to put `Touhosu!` and `Trigger_distance`

- [ ] add redirects in `redirect.yaml`


Walavouchey:
`Touhosu!` should probably be deleted

`Trigger_distance` should be in `Beatmapping/`

Walavouchey:
have you looked at https://github.com/ppy/osu-wiki/issues/6080? the first four commits here aren't in line with the task list. they probably need a bit more discussion first if it's a point of contention

closing for now

PR #7019, opened 2022-03-15
Increase cheating base restriction duration to 6 months

Ephemeralis:
These changes reflect new internal policy as of today (2022-03-16).

The team has noticed for a while now high re-offense rates in people getting their first and second restrictions, and have collectively agreed that the trivially short durations on current offenses are contributing to the problem. As it stands, a three month timeout is an offensively short punishment that is viewed with considerable dismay by the people who are doing the right thing.

Capping cooldowns at only 6 months for even the most egregious repeat offenders has only facilitated their return to game with relative abandon so long as they are capable of producing a convincing appeal in most cases. Accordingly, they have been lengthened to follow a pattern of 6-12-24 for the first three restrictions instead of the previous 3-6-6. It is our hope that harder punishments on this front will make people less willing to entertain the risk of having to sit out from the game for upwards of two years at a time.

**Importantly, this only applies to cheating restrictions.** Multi-accounting and other community offenses still have a 3 month minimum term and will continue to follow the 3-6-6 pattern for multiple offenses.

For the curious, existing offenses before this date will be grandfathered in under old terms.

I've added some wording about cooldowns and a new section on repeat offenses, plus a few missing entries from existing reasons/cooldowns.


peppy:
Is there going to be a further explanation as to what triggered this change somewhere else? A minimum requirement for making changes like this should be rationalising them in the pull request description.

Ephemeralis:
I've updated the PR description with an explanation/summary of the discussion surrounding these changes,

peppy (dismissed):

wiki/Help_centre/Account_restrictions/en.md
```diff
@@ -103,13 +111,15 @@ Please understand that sending an email to this address only guarantees a second
...
 | Excessive multi-accounting | 3 months | Appeal terms apply, 1 year tournament ban |
 | Account sharing | 3 months | Appeal terms apply, 1 year tournament ban, and partial rollback |
-| Cheating | 3 months | Appeal terms apply, 1 year tournament ban, and full rollback |
+| Cheating | 6 months | Appeal terms apply, 1 year tournament ban, and full rollback |
 | Excessive misconduct | Variable | Appeal terms apply. Cooldown will depend on history. |
-| Cheating/sharing for a tournament | 6 months | Appeal terms apply, appropriate rollback, permanent tournament ban |
-| Subsequent offences | 6 months | Appeal terms apply, appropriate rollback, tournament ban 1 year per offence |
+| Cheating/sharing for a tournament | 12 months | Appeal terms apply, appropriate rollback, permanent tournament ban |
+| Subsequent offences (multi-accounting, account sharing, community conduct) | 6 months | Appeal terms apply, rollback (where appropriate), tournament ban 1 year per offence |
+| Subsequent offences (cheating) | 12 months, up to 24 | Appeal terms apply, rollback, tournament ban 1 year per offence |
```
Walavouchey, line 124: Just to make sure I understand, repeatedly cheating and returning would give you a maximum of 24 months of cooldown, but any other offense can have the cooldown escalate indefinitely?

Ephemeralis: Subsequent offences cap at 6 months for non-cheating offenses, as described. For cheating offenses, they double from 6 to 12 to 24 and then everything after that is 24.

Walavouchey: Well what's this "minimum" in this table for if it doesn't apply to most of this table? Subsequent offences for non-cheating offences would always result 6 months of cooldown (meaning it's not really a minimum). Where does it apply? Is it possible to get more than 3 months of cooldown for (the first offence of) account sharing?

Ephemeralis: I've removed the mention of minimum from the table header and also re-ordered it to be smallest to largest in terms of applied cooldown.

wiki/Help_centre/Account_restrictions/en.md
```diff
@@ -81,11 +81,19 @@ Some offences will lead to your account being rolled back upon unrestriction. Th
...
 - **Your history is too severe.** In cases of extreme offenders, or players who get restricted again after they have already been unrestricted on a successful appeal, we might simply state that your history is too severe to rejoin the game at the current time. Should this apply to you, we recommend waiting **a long time** before appealing again, and making sure you do not break any further rules in that time.
 
+## What happens if this isn't my first time appealing?
+
+In most cases, each restriction on your account doubles the length of the next restriction and the cooldown associated with it.
+
+For example, if you are restricted for cheating, return on appeal, then cheat again, you will be restricted for a minimum of 12 months (double the six months base) before we will review any kind of appeal from you. If you return after this and cheat once again, you will be restricted for a minimum of 24 months.
```
Walavouchey, line 93: since it says "minimum" here too, is 24 not the cap?

wiki/Help_centre/Account_restrictions/en.md
```diff
@@ -109,7 +117,9 @@ Please understand that sending an email to this address only guarantees a second
...
 | Excessive multi-accounting | 3 months | Appeal terms apply, 1 year tournament ban |
 | Account sharing | 3 months | Appeal terms apply, 1 year tournament ban, and partial rollback |
-| Cheating | 3 months | Appeal terms apply, 1 year tournament ban, and full rollback |
+| Subsequent offences (multi-accounting, account sharing, community conduct) | 6 months | Appeal terms apply, rollback (where appropriate), tournament ban 1 year per offence |
+| Cheating | 6 months | Appeal terms apply, 1 year tournament ban, and full rollback |
 | Excessive misconduct | Variable | Appeal terms apply. Cooldown will depend on history. |
-| Cheating/sharing for a tournament | 6 months | Appeal terms apply, appropriate rollback, permanent tournament ban |
-| Subsequent offences | 6 months | Appeal terms apply, appropriate rollback, tournament ban 1 year per offence |
+| Cheating/sharing for a tournament | 12 months | Appeal terms apply, appropriate rollback, permanent tournament ban |
+| Subsequent offences (cheating) | 12 months | Appeal terms apply, rollback, tournament ban 1 year per offence |
```
Walavouchey, line 124: if it can go to 24, it should be mentioned

```suggestion
| Subsequent offences (cheating) | 12 to 24 months | Appeal terms apply, rollback, tournament ban 1 year per offence |
```

wiki/Help_centre/Account_restrictions/en.md
Ephemeralis: min terminology is used here so as not to infer that it's a "you're restricted for 24 months then you get to come back" when it's "you're restricted for 24 months THEN you can apply to remove that restriction", the latter causing restriction to last longer than 24 months

Walavouchey: oh right, not cooldowns, nvm

Walavouchey (approved):

peppy (approved):

PR #7022, opened 2022-03-15
A-s-c

### LINKED ISSUE #6973: Use the ASC for the language they are written in
## Problem

The content parity rule says that all languages should have the same content. This also applies to the ASC. Considering this, that means that every language's ASC describe what formats the English wiki needs to use. However, every language has its own rules and that should be written down somewhere, so that the translators know how to write articles for the wiki. Otherwise, the wiki will be heavily inconsistent and reviewers will have a hard time redirecting translators into the right direction.

To visualise the discrepancies between different ASC, I've searched through English and German and wrote down the differences that, in my opinion, should be included (or not) in the ASC for German:
- There is no equivalent to British/American English in German
- Specific terms, like osu! wiki and hit circle, are written as osu!-Wiki and Hit-Circle respectively[1]
- Date formats are different
- The acronyms `e.g.` and `i.e.` aren't used in German
- Em dashes aren't used as much in German (instead just `-` is used most of the times)

My main point is that it doesn't make sense to me to write down how the English wiki should look like in several languages. Furthermore, English contributors will probably look at the English ASC anyway.

## Proposal

My suggestion is to use the ASC for the language they are written in. This requires the ASC to be excluded at least partly from the content parity rule. A problem that might arise here, is that several contributors for a language need to create language specific ASC. Currently, we don't have enough contributors in every language, so some languages might be left behind. I understand that this is an argument against this suggestion.

Another idea might be to create a space next to the ASC, where contributors can write down language specific rules. For example, a subfolder called `Language specific rules`. I don't want to determine a specific structure here, this is just a general idea.

I don't have a problem if things stay like they are right now, but I would like to start a discussion about this, hence why I open this issue.

## Changes regarding the German ASC

If my first proposal isn't accepted, then the following tasks need to be done:

- [ ] Rewrite ASC to also include the English parts (rules that didn't apply to German were removed originally)
- [ ] Add header stating that the ASC are mainly for English (look at the [French ASC](https://osu.ppy.sh/wiki/fr/Article_styling_criteria/Writing) for an example)
- [ ] Create a space for the original ASC (before the rewrite), where it can be maintained better. I can imagine one of these options:
    - separate GitHub repo
    - separate Discord server
    - osu! forums

[1] Not really a valid point here, because I more or less set this rule myself. Though, I'm still including it because I've seen this writing style elsewhere (not just German).

TicClick:
resolves #6973

## Self-check


wiki/Article_styling_criteria/Writing/en.md
```diff
@@ -5,6 +5,18 @@
...
+
+Translations are subject to strict content parity with their English article, in the sense that they must have the same message, regardless of grammar and syntax. Any changes to the translations' meanings must be accompanied by equivalent changes to the English article.
+
+There are some cases where the content is allowed to differ:
+
+- Articles originally written in a language other than English (in this case, English should act as the translation)
+- Translations and explanations of English words that are common terms in the osu! community
+- External links
+- Tags
+- Information specific to a subcommunity or a certain language (such as [Article styling criteria/Writing](/wiki/Article_styling_criteria/Writing))
```
Walavouchey, line 18:
```suggestion
- Information specific to a subcommunity or a certain language (such as translations of this page)
```

Walavouchey (commented):
may also want to add something like "For language-specific writing standards, see the translations of this page" at the top (where translations would similarly refer to the english version)

Walavouchey (approved):

PR #7032, opened 2022-03-17
Update `Project Loved Team`

huuishuu:
Fixing a lot of errors and missing listings:
- Noffy was incorrectly marked as a retired member (despite being removed from the Project Loved group on website, she is still active as far as I am aware and has not yet left the team). She was also categorized as an osu! captain for...some reason.
- The "Retired members" section was missing several people & categories altogether (metadata, moderators).
- Fixing renames of users & made sure to reorder all usernames alphabetically (renames: Nikolayio -> Baal Sphere, PuigdemontLoL -> RandomeLoL)
- Translations have not been updated.

All updates to this page have been based off of the Project Loved website's [The team page](https://loved.sh/team).

## Self-check

### DIFF
--- a/wiki/People/The_Team/Project_Loved_Team/en.md
+++ b/wiki/People/The_Team/Project_Loved_Team/en.md
@@ -18,8 +18,8 @@ The Project Loved Team mostly consists of *captains* for each [game mode](/wiki/
 
 - ![][flag_US] [DigitalHypno](https://osu.ppy.sh/users/4384207)
 - ![][flag_GB] [Doomsday](https://osu.ppy.sh/users/18983)
-- ![][flag_US] [Wixonater](https://osu.ppy.sh/users/10083084)
 - ![][flag_US] [UberFazz](https://osu.ppy.sh/users/8646059)
+- ![][flag_US] [Wixonater](https://osu.ppy.sh/users/10083084)
 
 ### osu!taiko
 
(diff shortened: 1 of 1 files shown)


cl8n:
if this page's user list is going to be a lagging copy of https://loved.sh/team, then I don't see why it shouldn't be a link to that instead

huuishuu:
> if this page's user list is going to be a lagging copy of https://loved.sh/team, then I don't see why it shouldn't be a link to that instead

Wouldn't that be against wiki guidelines though? Pretty sure the official wiki shouldn't be linking users off-site to view the team listing. I'm not saying I disagree with your statements, I'm just pointing out what I've been told in the past.

Walavouchey:
it's the same with the listings of the bng, nat, gmt, etc. sure, they could all just point to the group listings on the website (which they already do 🤔) once the they can store and display more metadata about members of groups

not sure what this guideline is about, but i don't have an opinion on it

cl8n:
well none of the groups' user lists can be replaced by their /g/ pages because some data isn't shown there like you said. but for loved group, there is a public and reputable-enough page that shows the same stuff wiki would show. not a big deal to me either, just seems annoying to have regular copy-paste checkups on articles that don't need it

huuishuu:
Gonna close this PR since it looks like #7033 is more likely to be accepted.

PR #7035, opened 2022-03-18
Clean up `Client/File formats`

TicClick:
- simplified stuff, removed some odd wordings
- inclined to remove the winrar/7zip screenshots as well (up for discussion)

## Self-check


Walavouchey (changes requested):
I also don't think the screenshots are really necessary, and they aren't even embedded into the page

Walavouchey (approved):

PR #7041, opened 2022-03-19
Correct date in news post filename

### LINKED ISSUE #7040: Inappropriate date
https://github.com/ppy/osu-wiki/blob/master/news/2020-31-12-mappers-report-fall.md
This file name is 31/12 when it should be 12/31.

Walavouchey:
closes #7040

The date in the filename was in an incorrect format (the post itself is correct)

I don't think the website cares about the filename, so in theory nothing wonky should happen


peppy:
This *will* change the URL, but it's probably okay to do so for an article like this?

peppy (approved):

PR #7051, opened 2022-03-20
Switch headings *General* and *Mouse* in `Client\Keyboard_shortcuts`

The-Last-Cookie:
Moves the heading *Mouse* below the heading *General*, so *General* is the first heading in this section. This order seems more plausible to me.

## Self-check


Walavouchey:
mind doing the same for the translations?

The-Last-Cookie:
No problem. Can do this in a few minutes.

The-Last-Cookie:
Headings were moved in the non-outdated translations accourdingly.

Walavouchey (approved):

PR #7056, opened 2022-03-20
Update `TWC 2022` match results table

LeoFLT:
Fixes match results table links (they were pointing to MWC 7K lobbies instead of TWC).

## Self-check


Walavouchey (approved):
how even

PR #7059, opened 2022-03-21
Outdate id.md

AJT127:
## Self-check

- [ x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)


TicClick:
any particular reason for this? also, you meed to wrap it in three dashes: https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#front-matter

TicClick:
ok looks like it was supposed to be in https://github.com/ppy/osu-wiki/pull/6873. closing

PR #7062, opened 2022-03-21
Make audio requirements more lenient in the Ranking Criteria

Naxesss:
This addresses a latent issue introduced when dropping the 128 kbps minimum, which we did to prevent the need to up-encode for songs with no available source >= 128 kbps. To discourage people from using too low quality as a result of that change, we added "try to find the highest quality audio".

Since then, we have seen groups forcing mappers to change from 128 kbps to 192 kbps audio files quoting that rule, effectively raising the bar to 192 kbps, which is not supposed to be necessary. The benefit of this is sometimes negligible, and sometimes even detrimental when the suggested audio is badly encoded.

To fix this, we simply made 128 kbps a threshold after which further improvement is up to the mapper, as long the current audio sounds ok. We also cleaned up the structure of the rules a bit, making them easier to read than before.

Proposal thread: https://osu.ppy.sh/community/forums/topics/1540270

- [x] sanity check from fellow nat
- [x] waiting for https://github.com/ppy/osu-wiki/pull/6873, as it will likely handle outdating sooner than this PR


wiki/Ranking_Criteria/en.md
```diff
@@ -191,9 +191,12 @@ This category contains explicit allowance statements of concepts and rules that
 
 ### Rules
 
-- **A beatmap's audio file must use the `.mp3` or `.ogg` file format and have an average bit rate no greater than 192kbps.**
+- **The audio file of a beatmap must...**
+  - **...use the .mp3 or .ogg file format.**
```
Walavouchey, line 195:
```suggestion
  - **...use the `.mp3` or `.ogg` file format.**
```

Walavouchey (changes requested):

peppy (approved):

Walavouchey (approved):

peppy (approved):

PR #7057, opened 2022-03-21
Clean up `osu!supporter`

TicClick:
- article is now focused on the supporter tag as entity, not the process of its purchase
- updated some images and removed unnecessary ones

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

## TODO

- [x] expand the list of perks

PR #7080, opened 2022-03-24
Remove infoboxes in the osu!supporter article

Walavouchey:
follow-up to https://github.com/ppy/osu-wiki/pull/7057. removing because of alignment issues. reported at https://github.com/ppy/osu-web/issues/8743

note that if [this](https://discord.com/channels/188630481301012481/188630652340404224/956355142138073098) also applies to the wiki then maybe the correct step here is to live with the issue until it is fixed (and also combine the first two images into one infobox)

image

### DIFF
--- a/wiki/osu!supporter/en.md
+++ b/wiki/osu!supporter/en.md
@@ -16,13 +16,9 @@ The osu!supporter tag may be obtained from the [osu!supporter store page](https:
 
 ### Heart badge
 
-::: Infobox
 ![](img/userpage.png?1 "Profile with osu!supporter")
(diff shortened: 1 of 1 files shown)


cl8n (approved):

cl8n (changes requested):
just kidding will be fixed soon

peppy:
Let's apply this if it isn't fixed by the end of the week. Would rather fix the underlying issue.

notbakaneko:
No need to remove them https://github.com/ppy/osu-web/pull/8745

PR #7091, opened 2022-03-26
Fix the freshness check to make it consider composite language codes

TicClick:
## Self-check

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -147,7 +147,7 @@ jobs:
 
           MISSED_TRANSLATIONS=$( echo "$ORIGINAL_ARTICLES" | tr \\n \\0 | xargs -0 dirname | sort -u | {
             while read DIRECTORY; do
-              for ARTICLE_NAME in $( ls "$DIRECTORY"/??.md | grep -v en.md || true ); do
+              for ARTICLE_NAME in $( ls "$DIRECTORY"/(??|??-??).md | grep -v en.md || true ); do
                 if ! [[ "$TRANSLATIONS" =~ "$ARTICLE_NAME" ]]; then
                   grep -LzP -e "$MASK" "$ARTICLE_NAME" || true  # print only names of non-matching files, with multiline match (considering \n)
                 fi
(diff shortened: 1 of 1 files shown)


Walavouchey:
doesn't work https://github.com/Walavouchey/osu-wiki/runs/5705650526?check_suite_focus=true

.github/workflows/continuous-integration.yml
```diff
@@ -147,7 +147,7 @@ jobs:
 
           MISSED_TRANSLATIONS=$( echo "$ORIGINAL_ARTICLES" | tr \\n \\0 | xargs -0 dirname | sort -u | {
             while read DIRECTORY; do
-              for ARTICLE_NAME in $( ls "$DIRECTORY"/??.md | grep -v en.md || true ); do
+              for ARTICLE_NAME in $( ls "$DIRECTORY"/(??|??-??).md | grep -v en.md || true ); do
```
Walavouchey, line 150: this would work

image

```suggestion
              for ARTICLE_NAME in $( ls "$DIRECTORY"/??.md "$DIRECTORY"/??-??.md | grep -v en.md || true ); do
```

TicClick: it fails when there aren't any `??-??` files. I'll attempt something else later

TicClick:
@Walavouchey took another approach, seems to be working: https://github.com/TicClick/osu-wiki/runs/5710414992?check_suite_focus=true

.github/workflows/continuous-integration.yml
```diff
@@ -147,7 +147,10 @@ jobs:
 
           MISSED_TRANSLATIONS=$( echo "$ORIGINAL_ARTICLES" | tr \\n \\0 | xargs -0 dirname | sort -u | {
             while read DIRECTORY; do
-              for ARTICLE_NAME in $( ls "$DIRECTORY"/??.md | grep -v en.md || true ); do
+              for ARTICLE_NAME in "$DIRECTORY"/*.md; do
+                if ! [[ "$ARTICLE_NAME" =~ \/(..|..-..)\.md ]] || [[ "$ARTICLE_NAME" =~ \/en\.md ]] ; then
```
Walavouchey, line 151:
```suggestion
                if ! [[ "$ARTICLE_NAME" =~ \/(..|..-..)\.md ]] || [[ "$ARTICLE_NAME" =~ \/en\.md ]]; then
```

Walavouchey (approved):
lgtm

PR #7097, opened 2022-03-27
[DE] Update `Tournaments/OWC/2014`

respektive:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #7099, opened 2022-03-27
Fix formatting in OWC articles

respektive:
Noticed these errors in https://github.com/ppy/osu-wiki/pull/7097

I put the winning conditions under another heading, which is what the 2020/21 OWC articles do and removed the weird gaps in the stage instructions. Left the outdated articles untouched as they're outdated anyway.

## Self-check


wiki/Tournaments/OWC/2014/en.md
```diff
@@ -444,39 +444,30 @@ Sunday, 9 November 2014:
...
-    - In Semi-finals and Finals, you need to win 6 maps to win a match. (Best-of-11)
+| Stage | Match ID |
+| :-- | :-- |
+| Round of 16 | A, B, C, D, E, F, G, H |
+| Quarterfinals | I, J, K, L & R, S, T, U |
+| Semifinals | M, N & V, W, X, Y, Z, AA |
+| Finals week 1 | O & AB, AC, AD, AE |
+| Finals week 2 | P, Q |
+
+#### Winning conditions
```
Walavouchey, line 466: you can interrupt a list with a table (without an indent), so you don't need another heading

Walavouchey (changes requested):
as above

Walavouchey:
see [the oSEAT5 article](https://github.com/ppy/osu-wiki/blob/master/wiki/Tournaments/oSEAT/5/en.md?plain=1#L870L889) for an example

wiki/Tournaments/OWC/2014/en.md
respektive: Oh, I didn't know that. Will adjust accordingly.

respektive: reverted the heading

wiki/Tournaments/OWC/2014/en.md
```diff
@@ -444,39 +444,29 @@ Sunday, 9 November 2014:
...
+| Round of 16 | A, B, C, D, E, F, G, H |
+| Quarterfinals | I, J, K, L & R, S, T, U |
+| Semifinals | M, N & V, W, X, Y, Z, AA |
+| Finals week 1 | O & AB, AC, AD, AE |
+| Finals week 2 | P, Q |
+
+12. **Win conditions:**
+   - In Group Stage, you need to win 4 maps to win a match. (Best-of-7)
+   - In the Round of 16 and the Quarter-finals, you need to win 5 maps to win a match. (Best-of-9)
+   - In Semi-finals and Finals, you need to win 6 maps to win a match. (Best-of-11)
```
Walavouchey, line 469:
```suggestion
    - In Group Stage, you need to win 4 maps to win a match. (Best-of-7)
    - In the Round of 16 and the Quarter-finals, you need to win 5 maps to win a match. (Best-of-9)
    - In Semi-finals and Finals, you need to win 6 maps to win a match. (Best-of-11)
```

Walavouchey (approved):

PR #7147, opened 2022-04-08
Add clause about unsolicited PMs in `Bot account`

Trigonoculus:
Per internal discussion, this PR clarifies that bots should not send unsolicited private messages.

## Self-check

### DIFF
--- a/wiki/Bot_account/en.md
+++ b/wiki/Bot_account/en.md
@@ -20,7 +20,7 @@ The main difference between personal accounts and bot accounts is in the rate li
 - Personal accounts can send 10 messages every 5 seconds
 - Bot accounts can send 300 messages every 60 seconds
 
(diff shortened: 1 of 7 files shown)


Walavouchey:
note that some bots, like the [mappersguild](https://osu.ppy.sh/users/23648635) bot (not sure of any other examples), do send private messages for informative or administrative purposes

peppy:
> note that some bots, like the [mappersguild](https://osu.ppy.sh/users/23648635) bot (not sure of any other examples), do send private messages for informative or administrative purposes (like [bn evaluation feedback](https://osu.ppy.sh/wiki/en/People/The_Team/Nomination_Assessment_Team/Evaluations))

I'd count that as mostly-solicited. Still best to add this clause as a general rule and have people ask for exceptions as required.

peppy (approved):

PR #7171, opened 2022-04-13
Re-enable code ownership for the `Help centre` articles

TicClick:
found in https://github.com/ppy/osu-wiki/pull/7170, broken in https://github.com/ppy/osu-wiki/pull/6723. if anything, only Ephemeral has edited these since then, so no unattended edits were made


peppy (approved):

peppy:
Nice find.

PR #7175, opened 2022-04-13
[DE] Add `Brand_identity_guidelines` and `Client/Interface/Cookie`; update `Gameplay`, `People` and `Help_centre`

The-Last-Cookie:
Notice for reviewers: I've also applied some styling changes in `Help_centre`, because I think that it improves readability.

## Self-check


wiki/People/de.md
```diff
@@ -1,13 +1,10 @@
----
-outdated: true
-outdated_since: 7880a47c4e5df0bb9470cdd8edc48daa42e6f277
----
-
 # Menschen
```
respektive, line 1: I'm thinking maybe this should be translated as `Personen` or `Leute`? I think both work better than `Menschen`. :thinking:

The-Last-Cookie: Yeah, good point. I also thought about this myself. The main page uses `Personen`, so I'm going to stick to this. `Leute` sounds a tiny bit too casual in my opinion.

```suggestion
# Personen
```

wiki/People/de.md
```diff
@@ -1,13 +1,10 @@
----
-outdated: true
-outdated_since: 7880a47c4e5df0bb9470cdd8edc48daa42e6f277
----
-
-# Menschen
+# Personen
 
 Indexseite für Artikel über Menschen oder Gruppen.
```
respektive, line 3:
```suggestion
Indexseite für Artikel über Personen oder Gruppen.
```
just noticed this should be changed as well then whoops

The-Last-Cookie, line 3: Actually this should be changed too, shouldn't it?

respektive (dismissed):

The-Last-Cookie:
TODO: I need to add in the (unexpected) changes from #7194 . In particular, this line here, above `website` in the table:
```markdown
| [Tournament bans](/wiki/Help_centre/Tournament_bans) | All about tournament bans: overview, common reasons, and cooldowns |
```

respektive (approved):

Walavouchey (approved):

PR #7177, opened 2022-04-13
[DE] Mark `Client/Beatmap_editor/Menu` as needing native review

The-Last-Cookie:
See TODO which has been added.

### DIFF
--- a/wiki/Client/Beatmap_editor/Menu/de.md
+++ b/wiki/Client/Beatmap_editor/Menu/de.md
@@ -2,7 +2,7 @@
 no_native_review: true
 ---
 
-<!-- TODO: This article should use the informal form. Furthermore, there are also some untranslated sentences and grammar mistakes. -->
+<!-- TODO: This article should use the singular form of "you" (`Du` instead of `Sie`), unless the English version gets rewritten to have no pronouns. Furthermore, there are also some untranslated sentences and grammar mistakes. -->
 
 # Menü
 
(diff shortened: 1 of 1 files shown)


wiki/Client/Beatmap_editor/Menu/de.md
```diff
@@ -1,3 +1,9 @@
+---
+no_native_review: true
+---
+
+<!-- TODO: This article should use the informal form. Furthermore, there are also some untranslated sentences and grammar mistakes. -->
```
Walavouchey, line 5: ... informal? the english version seems to use "you" a lot, but is otherwise not "informal" (and shouldn't be)

respektive: in German we have both a formal and informal version for you (`Sie` vs. `Du`). Generally the German articles use `Du`, which is a more personal and friendly form, while `Sie`, which this article is using, is a polite and formal form used in like letters and such.

I suggest the TODO should specify it's about the informal form of "you".

Walavouchey: let's reword this a bit so no one else gets confused

```suggestion
<!-- TODO: This article should use the singular form of "you" (`Du` instead of `Sie`), unless the English version gets rewritten to have no pronouns. Furthermore, there are also some untranslated sentences and grammar mistakes. -->
```

Walavouchey (approved):

PR #7183, opened 2022-04-14
Remove formatting from section headings

TicClick:
https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#sections:

> Section headings must not skip a heading level (i.e. do not go from a level 2 heading to a level 4 heading) and must not contain formatting or links.


TicClick:
@Nivalyx @hikikuman please say if you are okay with the change as-is, as most of the articles on the list are Indonesian and this may be unintended

peppy (approved):

Walavouchey (approved):

PR #7189, opened 2022-04-15
April update of the team and staff log

Blooshing:
## Self-check

`SKIP_OUTDATED_CHECK` for GMT stuff, which only has some clean-up


wiki/People/The_Team/Global_Moderation_Team/de.md
```diff
@@ -61,10 +61,10 @@ Die [Seite für das globale Moderationsteam](https://osu.ppy.sh/groups/4) listet
 | ![][flag_HU] [[ Another ]](https://osu.ppy.sh/users/3416573) | Ungarisch | Chat-Moderation |
 | ![][flag_CH] [\[ryuu\]](https://osu.ppy.sh/users/5698467) | Russisch | Chat-Moderation |
 | ![][flag_US] [abraker](https://osu.ppy.sh/users/4635891) |  | Forum-Moderation |
-| ![][flag_CA] [Azer](https://osu.ppy.sh/users/2155578) |  | Verwaltung von Turnieren |
+| ![][flag_CA] [Azer](https://osu.ppy.sh/users/2155578) |  | Verwaltung von Wettbewerben |
```
The-Last-Cookie, line 64: I just noticed that the `group-info` file in the meta folder has an incorrect value. `Turnier` and `Wettbewerb` have a large difference in usage on osu! related topics, as can be [seen in the German ASC](https://osu.ppy.sh/wiki/de/Article_styling_criteria/Writing#%C3%BCbersetzungen-f%C3%BCr-h%C3%A4ufige-begriffe). In this section are common terms and their correct translation listed. `Turnier` is used only ***and only*** for tournaments. In the contrary, `Wettbewerb` is used only for contests.

I would very much appreciate it if the changed files are fixed in this PR. Otherwise, I can do it in a separate one. I will open a PR for the `group-info` file later or tomorrow.

wiki/People/The_Team/Beatmap_Nominators/en.md
```diff
@@ -190,49 +178,53 @@ The tables listed below indicate the game mode(s) in which each Beatmap Nominato
...
+| ![][flag_US] [Yogurtt](https://osu.ppy.sh/users/2649717) |  |
 
 #### osu!taiko
 
 | Name | Additional languages |
 | :-- | :-- |
-| ![][flag_JP] [Chromoxx](https://osu.ppy.sh/users/1881639) | Japanese |
+| ![][flag_TR] [Burak](https://osu.ppy.sh/users/13213075) | Turkish |
+| ![][flag_KR] [Konpaku Sariel](https://osu.ppy.sh/users/533502) | Korean |
 | ![][flag_FI] [pupu5](https://osu.ppy.sh/users/7140680) | Some Japanese, some German |
```
Hiviexd, line 198: pupu5 isn't a BN, they can safely be removed from the list

.osu-wiki-bin.json
```diff
@@ -0,0 +1,4 @@
+{
+  "osuApiKey": "",
```
Ephemeralis, line 2: i don't think you meant to include this

Ephemeralis (dismissed):

wiki/People/The_Team/Global_Moderation_Team/de.md
Walavouchey: i've fixed this for GMT here in [ebaac52](https://github.com/ppy/osu-wiki/pull/7189/commits/ebaac52d6e16df6f95747635444d0581fc89c190), but do pr the change to `group-info`

Walavouchey (approved):

wiki/People/The_Team/Global_Moderation_Team/de.md
The-Last-Cookie: Thanks, will do 👍

PR #7195, opened 2022-04-17
[EN] Update on `osu!supporter`

ImMock:
It seems like the profile cover no longer is replaced when the osu!supporter tag expires.


peppy:
Hmm, it never was.. not sure where this came from.

peppy (approved):

PR #7222, opened 2022-04-23
Add a shell script for running CI checks locally

### LINKED ISSUE #7155: Make ci easy to run locally
The best solution is probably just a shell script in the repository root that runs the relevant ci steps

Walavouchey:
Closes #7155 

The in-lined checks from the workflow file are split into their own scripts.

Run with `bash run-ci.sh` from the repository root. I've checked that the github workflow still works with a test pr at https://github.com/Walavouchey/osu-wiki/pull/21.

## Notes:

- The scripts check for errors in changed files in the working branch (committed changes), just like in regular ci runs.
  - In the case of `run_yamllint.py`, all files are checked, but that can easily be changed by using its `--target` parameter.
  - Retrieving changed files requires being in a feature branch with the way I've done it here, but I suppose that is to be expected.
  - May add a way to run the checks for all files as well later.
- The scripts now take a pair of commit hashes as parameters (first and last in the branch), which is supplied by `run-ci.sh`, but they work the same even when passed a shallow clone commit from the workflow (and I have to thank @bdach for the good explanation at https://github.com/ppy/osu-wiki/pull/3712 which helped me understand how this works).
  - `check_outdated_tags.sh` optionally takes a third argument for the first commit in the branch or pr, depending on whether the script is run locally or in the workflow. The first commit in the branch can't be retrieved the same way as when done locally because of the shallow clone.
- Error reporting (the colouring for `Error:`, `Warning:`, etc.) in some of the scripts has been done using [workflow commands](https://docs.github.com/en/enterprise-server@3.2/actions/using-workflows/workflow-commands-for-github-actions), which I've replaced with ansi escape codes in the individual scripts.
- May also add a `.ps1` script for windows users later.


TicClick:
from the first glance it looks alright, but I think it's incomplete without a short README on how to set up the tooling for a local run (`nvm` and `npm i`, `python`'s virtualenv) -- just a few lines of "get this, run `that --with-some-flags" is enough

peppy:
```sh
osu-wiki on  local-ci via  v18.0.0
❯ ./run-ci.sh
-e
--- Binary file size check ---

grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
\e[0;90mNotice:\e[m No excessively large files detected
-e
--- Run remark ---

xargs: illegal option -- d
usage: xargs [-0opt] [-E eofstr] [-I replstr [-R replacements] [-S replsize]]
             [-J replstr] [-L number] [-n number [-x]] [-P maxprocs]
             [-s size] [utility [argument ...]]
-e
--- Run yamllint ---

./run-ci.sh: line 12: python: command not found
-e
--- Broken wikilink check ---

./run-ci.sh: line 15: python: command not found
-e
--- Outdated tag check ---

grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
        [-e pattern] [-f file] [--binary-files=value] [--color=when]
        [--context[=num]] [--directories=action] [--label] [--line-buffered]
        [--null] [pattern] [file ...]
grep: invalid option -- P
```

frenzibyte:
Just to note on the `grep -P` and `xargs -d` issues in that log, that occurs due to the ones shipped in macOS (i.e. FreeBSD) being old enough to not support that. Installing latest one via homebrew and linking it fixes it (`brew install grep`, `brew install findutils`). 

The `python` one I'm not entirely sure about, but it was also not found for me and turns out I have Python 3 installed, which its name is `python3` instead of `python` I suppose? alias/link can work.

No clue about the rest, this is as far as I've got.
```
╰─λ bash ./run-ci.sh                                                                                                                                           0 (2.126s)

--- Binary file size check ---

Notice: No excessively large files detected

--- Run remark ---

npm ERR! could not determine executable to run

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/salman/.npm/_logs/2022-04-25T07_45_13_945Z-debug-0.log

--- Run yamllint ---

Traceback (most recent call last):
  File "/Users/salman/Projects/osu-wiki/.github/scripts/ci/run_yamllint.py", line 6, in <module>
    import yamllint.cli
ModuleNotFoundError: No module named 'yamllint'

--- Broken wikilink check ---

Notice: No articles to check.

--- Outdated tag check ---

Notice: No original articles are edited, exiting
```
```
[I] ╭─salman at Salmans-MBP in /Users/salman/Projects/osu-wiki (local-ci ✚7)
╰─λ cat /Users/salman/.npm/_logs/2022-04-25T07_45_13_945Z-debug-0.log                                                                                          0 (0.790s)
0 verbose cli /opt/homebrew/Cellar/node/18.0.0/bin/node /opt/homebrew/lib/node_modules/npm/bin/npm-cli.js
1 info using npm@8.6.0
2 info using node@v18.0.0
3 timing npm:load:whichnode Completed in 0ms
4 timing config:load:defaults Completed in 1ms
5 timing config:load:file:/opt/homebrew/lib/node_modules/npm/npmrc Completed in 1ms
6 timing config:load:builtin Completed in 1ms
7 timing config:load:cli Completed in 1ms
8 timing config:load:env Completed in 0ms
9 timing config:load:file:/Users/salman/Projects/osu-wiki/.npmrc Completed in 0ms
10 timing config:load:project Completed in 2ms
11 timing config:load:file:/Users/salman/.npmrc Completed in 0ms
12 timing config:load:user Completed in 0ms
13 timing config:load:file:/opt/homebrew/etc/npmrc Completed in 0ms
14 timing config:load:global Completed in 0ms
15 timing config:load:validate Completed in 0ms
16 timing config:load:credentials Completed in 1ms
17 timing config:load:setEnvs Completed in 0ms
18 timing config:load Completed in 6ms
19 timing npm:load:configload Completed in 6ms
20 timing npm:load:mkdirpcache Completed in 0ms
21 timing npm:load:mkdirplogs Completed in 1ms
22 verbose title npm exec remark -qf --no-stdout --silently-ignore --report=vfile-reporter-position --color
23 verbose argv "exec" "--" "remark" "-qf" "--no-stdout" "--silently-ignore" "--report=vfile-reporter-position" "--color"
24 timing npm:load:setTitle Completed in 2ms
25 timing config:load:flatten Completed in 1ms
26 timing npm:load:display Completed in 2ms
27 verbose logfile logs-max:10 dir:/Users/salman/.npm/_logs
28 verbose logfile /Users/salman/.npm/_logs/2022-04-25T07_45_13_945Z-debug-0.log
29 timing npm:load:logFile Completed in 2ms
30 timing npm:load:timers Completed in 0ms
31 timing npm:load:configScope Completed in 0ms
32 timing npm:load Completed in 13ms
33 silly logfile done cleaning log files
34 http fetch GET 200 https://registry.npmjs.org/remark 472ms (cache revalidated)
35 timing command:exec Completed in 490ms
36 verbose stack Error: could not determine executable to run
36 verbose stack     at getBinFromManifest (/opt/homebrew/lib/node_modules/npm/node_modules/libnpmexec/lib/get-bin-from-manifest.js:17:23)
36 verbose stack     at exec (/opt/homebrew/lib/node_modules/npm/node_modules/libnpmexec/lib/index.js:114:15)
36 verbose stack     at async module.exports (/opt/homebrew/lib/node_modules/npm/lib/cli.js:78:5)
37 verbose pkgid remark@14.0.2
38 verbose cwd /Users/salman/Projects/osu-wiki
39 verbose Darwin 21.4.0
40 verbose node v18.0.0
41 verbose npm  v8.6.0
42 error could not determine executable to run
43 verbose exit 1
44 timing npm Completed in 530ms
45 verbose code 1
46 error A complete log of this run can be found in:
46 error     /Users/salman/.npm/_logs/2022-04-25T07_45_13_945Z-debug-0.log
```

peppy:
My point was that these shouldn't need to be worked around from my end. The grep usage can be changed to work well with the default bundled `grep` on macOS and `python3` can be checked for and used if present (I didn't go as far as testing whether it works, or whether the scripts use a 2.x variant).

.github/scripts/ci/check_outdated_tags.sh
```diff
@@ -0,0 +1,64 @@
+MASK='---\n(.*\n)*?((outdated: |outdated_since: ).+\n)(.*\n)*?---'
+PULL_REQUEST_TAG='SKIP_OUTDATED_CHECK'
+
+FIRST_COMMIT_HASH=$1
+LAST_COMMIT_HASH=$2
+FIRST_PR_COMMIT_HASH=$3
+
+# https://en.wikipedia.org/wiki/ANSI_escape_code#3-bit_and_4-bit
+function echo_red () { echo -e "\e[0;31m$1\e[m"; }
```
TicClick, line 9: `printf` could be a fitting replacement: 
image

.github/scripts/ci/check_outdated_tags.sh
```diff
@@ -0,0 +1,64 @@
...
+ORIGINAL_ARTICLES=$( diff_files 'wiki/**/en.md' )
+if [[ -z "$ORIGINAL_ARTICLES" ]]; then
+  echo "$( echo_grey 'Notice:' ) No original articles are edited, exiting"
+  exit 0
+fi
+
+# obtain directories of the modified en.md files, then list translations in them which:
+# - are not modified in the PR (assuming that the author took care of them)
+# - don't have the outdated/outdated_since markers
+MISSED_TRANSLATIONS=$( echo "$ORIGINAL_ARTICLES" | tr \\n \\0 | xargs -0 dirname | sort -u | {
```
TicClick, line 40: you can reuse the `tr \n \0` trick to work around `xargs` not understanding `-d` in other scripts (which will become `xargs -0`)

.github/scripts/ci/check_outdated_tags.sh
```diff
@@ -0,0 +1,64 @@
...
+  while read DIRECTORY; do
+    for ARTICLE_NAME in "$DIRECTORY"/*.md; do
+      if ! [[ "$ARTICLE_NAME" =~ \/(..|..-..)\.md ]] || [[ "$ARTICLE_NAME" =~ \/en\.md ]]; then
+        continue
+      fi
+      if ! [[ "$TRANSLATIONS" =~ "$ARTICLE_NAME" ]]; then
+        grep -LzP -e "$MASK" "$ARTICLE_NAME" || true  # print only names of non-matching files, with multiline match (considering \n)
```
TicClick, line 47: it is true that `grep -E` should work on macOS. feel free to use any ugly solution here as I plan to rewrite this check in anything but shell later

run-ci.sh
```diff
@@ -0,0 +1,20 @@
+FIRST_COMMIT_HASH=$( git log master..$( git branch --show-current ) --pretty=oneline | tail -1 | awk '{ print $1 }' )
```
TicClick, line 1:
```suggestion
FIRST_COMMIT_HASH=$( git log master..$( git branch --show-current ) -1 --pretty=format:%H --reverse )
```

run-ci.sh
```diff
@@ -0,0 +1,20 @@
...
+ARTICLES=$( git diff --diff-filter=d --name-only ${FIRST_COMMIT_HASH}^ ${LAST_COMMIT_HASH} "wiki/**/*.md" "news/*.md" )
+
+echo -e "\n--- Binary file size check ---\n"
+bash .github/scripts/ci/inspect_binary_file_sizes.sh ${FIRST_COMMIT_HASH} ${LAST_COMMIT_HASH}
+
+echo -e "\n--- Run remark ---\n"
+bash .github/scripts/ci/run_remark.sh ${FIRST_COMMIT_HASH} ${LAST_COMMIT_HASH}
+
+echo -e "\n--- Run yamllint ---\n"
+python .github/scripts/ci/run_yamllint.py --config .yamllint.yaml
```
TicClick, line 12: I think we need a small piece of code which will bootstrap the environment if it doesn't exist, like

```sh
pip3 install venv
python3 -m venv ./.venv
source ./.venv/bin/activate
pip install -r requirements.txt
```

(and also do not care about 2.x since it's expired)

run-ci.sh
```diff
@@ -0,0 +1,20 @@
+FIRST_COMMIT_HASH=$( git log master..$( git branch --show-current ) --pretty=oneline | tail -1 | awk '{ print $1 }' )
+LAST_COMMIT_HASH=$( git rev-parse HEAD )
+ARTICLES=$( git diff --diff-filter=d --name-only ${FIRST_COMMIT_HASH}^ ${LAST_COMMIT_HASH} "wiki/**/*.md" "news/*.md" )
+
+echo -e "\n--- Binary file size check ---\n"
+bash .github/scripts/ci/inspect_binary_file_sizes.sh ${FIRST_COMMIT_HASH} ${LAST_COMMIT_HASH}
```
TicClick, line 6: since it's not GitHub-only now, we could move the scripts to their own top-level directory and blacklist it in [`.gitattributes`](https://github.com/ppy/osu-wiki/blob/master/.gitattributes)

run-ci.sh
```diff
@@ -0,0 +1,20 @@
+FIRST_COMMIT_HASH=$( git log master..$( git branch --show-current ) --pretty=oneline | tail -1 | awk '{ print $1 }' )
+LAST_COMMIT_HASH=$( git rev-parse HEAD )
+ARTICLES=$( git diff --diff-filter=d --name-only ${FIRST_COMMIT_HASH}^ ${LAST_COMMIT_HASH} "wiki/**/*.md" "news/*.md" )
+
+echo -e "\n--- Binary file size check ---\n"
```
TicClick, line 5: nit: this, being the very first announcement, produces an unnecessary empty line at the beginning

TicClick (changes requested):

.github/scripts/ci/check_outdated_tags.sh
TicClick: (and, since some shells provide it as a builtin as well as `echo`, it makes sense to use `/usr/bin/printf`)

Walavouchey: i'd rather put `#!/bin/bash` at the top of the file than deal with that

TicClick: that wouldn't work under macOS as well due to `bash` being old
```sh
bash-3.2$ env echo "\e[0;31mred\e[m"
\e[0;31mred\e[m
bash-3.2$ echo "\e[0;31mred\e[m"
\e[0;31mred\e[m
```

(`echo` is v 1.18 2005/01/10 here)

run-ci.sh
Walavouchey: are you implying this should be in-lined in the script?

TicClick: I just noticed that we have at least one undeclared dependency (`yamllint`, see frenzibyte@'s comment) and decided to add a note here

as for inlining.. likely not, because we probably need the same for node

run-ci.sh
Walavouchey: we now have a funny problem of `printf "--- Binary file size check ---\n\n"` causing a syntax error. adding the beginning newline fixes it

run-ci.sh
Walavouchey: I've added a section in the contributing file. Very open to suggestions here

CONTRIBUTING.md
```diff
@@ -21,3 +21,32 @@ Luckily, if you're not willing to get too involved with the technicalities of Gi
...
+
+## Testing
+
+All pull requests are automatically run through a series of continuous integration checks (CI) in order to catch various common errors.
+
+For people familiar with the command line, these CI checks can also be run locally.
+
+### Requirements
+
+- Python 3.X
```
peppy, line 33:
```suggestion
- Python 3.x
```

CONTRIBUTING.md
```diff
@@ -21,3 +21,32 @@ Luckily, if you're not willing to get too involved with the technicalities of Gi
...
+- Node.js 16
+
+### Setup
+
+```sh
+pip3 install venv
+python3 -m venv ./.venv
+source ./.venv/bin/activate
```
peppy, line 41: Is this actually required? It doesn't look well maintained and fails on non-bash shells:

```log
DEPRECATION: Configuring installation scheme with distutils config files is deprecated and will no longer work in the near fu
ture. If you are using a Homebrew or Linuxbrew Python, please see discussion at https://github.com/Homebrew/homebrew-core/iss
ues/76621
ERROR: Could not find a version that satisfies the requirement venv (from versions: none)                                    ERROR: No matching distribution found for venv                                                                               ./.venv/bin/activate (line 40): Unsupported use of '='. In fish, please use 'set VIRTUAL_ENV "/Users/dean/Projects/osu-wiki/.
venv"'.                                                                                                                      from sourcing file ./.venv/bin/activate                                                                                      source: Error while reading file './.venv/bin/activate'
```

If we *are* using this then it needs a `.gitignore` rule.

CONTRIBUTING.md
```diff
@@ -21,3 +21,32 @@ Luckily, if you're not willing to get too involved with the technicalities of Gi
...
+pip install -r requirements.txt
```
peppy, line 42:
```suggestion
pip3 install -r requirements.txt
```

Although I argue the value of that txt file. Why not just include it inline?

peppy (changes requested):

CONTRIBUTING.md
Walavouchey: I'll wait for @TicClick's opinion on this one

TicClick: I forgot that `venv` is a built-in module (unlike `virtualenv`), so the first one is my mistake.

as for the scripts, running `activate` and the likes is a matter of convenience if you don't want to specify the path to the environment's interpreter every time. it's still available under `./.venv/bin/python3`, and if this is intended to be a copy-pastable command, we may as well just use the direct path instead of a shell's discovery mechanisms.

(also, it has scripts for non-Bash shells, `activate.csh` and `activate.fish`)

CONTRIBUTING.md
TicClick: we can inline it as it only consists of a single requirement for now, but I'd prefer reusing it here and in the GitHub workflow for unification purposes and as it may potentially grow like node's `package.json` (hopefully not to that extent though)

.github/workflows/continuous-integration.yml
```diff
@@ -87,7 +59,7 @@ jobs:
       - name: setup and run yamllint on .yaml and .md files
         run: |
           pip install yamllint
```
TicClick, line 61: I think as we only use it in Python scripts we should put it into `scripts/`
```suggestion
          pip install -r scripts/requirements.txt
```

requirements.txt
```diff
@@ -0,0 +1 @@
+yamllint
```
TicClick, line 1: let's pin the version while we're at it
```suggestion
yamllint==1.26.3
```

run-ci.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+FIRST_COMMIT_HASH=$( git log master..$( git branch --show-current ) --pretty=format:%H | tail -1 )
+LAST_COMMIT_HASH=$( git rev-parse HEAD )
+ARTICLES=$( git diff --diff-filter=d --name-only ${FIRST_COMMIT_HASH}^ ${LAST_COMMIT_HASH} "wiki/**/*.md" "news/*.md" )
+
```
TicClick, line 6: may want to add something like that, following peppy's suggestion:
```bash

python="python3"
if test -x "./.venv/bin/python3"; then
  python="./.venv/bin/python3"
fi

...

"$python" scripts/ci/run_yamllint.py --blah --blahblah
```

TicClick (changes requested):

CONTRIBUTING.md
```diff
@@ -21,3 +21,32 @@ Luckily, if you're not willing to get too involved with the technicalities of Gi
...
+## Testing
+
+All pull requests are automatically run through a series of continuous integration checks (CI) in order to catch various common errors.
+
+For people familiar with the command line, these CI checks can also be run locally.
+
+### Requirements
+
+- Python 3.x
+- Node.js 16
```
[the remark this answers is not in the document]
peppy: Taking it to the next level would be to just use docker, avoiding all of the venv and dependency nonsense. Can be done as a future effort if that's easier though.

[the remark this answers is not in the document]
TicClick: docker is overkill while separate environments work just fine (on the other hand we wouldn't probably even need that if all of our CI was in Node.js)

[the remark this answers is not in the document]
peppy: It's not overkill if it makes running 10x easier on platforms like windows. But sure.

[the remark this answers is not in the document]
TicClick: this really depends on how much we want to support bare Windows instead of recommending people use subsystems like Cygwin or WSL (which is notoriously easy to set up). if we want to simplify things as much as possible, then Docker may be an eventual good solution

run-ci.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+python="python3"
+if test -x "./.venv/bin/python3"; then
+  python="./.venv/bin/python3"
+fi
+
+FIRST_COMMIT_HASH=$( git log master..$( git branch --show-current ) --pretty=format:%H | tail -1 )
+LAST_COMMIT_HASH=$( git rev-parse HEAD )
+ARTICLES=$( git diff --diff-filter=d --name-only ${FIRST_COMMIT_HASH}^ ${LAST_COMMIT_HASH} "wiki/**/*.md" "news/*.md" )
```
TicClick, line 10: I've just noticed that it only takes committed changes into account due to `${FIRST_COMMIT_HASH}^` part. if I don't commit anything, the diff will be empty (and if I stage files, it will also be empty when requested without `--cached`). this may and will produce misleading results

TicClick (changes requested):
this is the only thing that needs attention by now, once it's fixed I think we're good (until other issues for non-standard setups arise)

CONTRIBUTING.md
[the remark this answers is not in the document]
peppy: I personally have an aversion to running things like node or python on my main development system due to security concerns, and will only run it within docker or a VM.

peppy:
Working well enough for me now 👍 .

TicClick:
a note that with the amount of workarounds for macOS (`stat -f %z` instead of `stat --format %s` was the last drop) making a minimal Docker image may be a much better idea than I initially thought

TicClick:
currently, the CI script would run ok locally on Linux but fail on macOS due to [different `stat` flags](https://github.com/ppy/osu-wiki/pull/7222/commits/2dcf3e837b986b83b141acd278c72f07b436d60c#diff-dbf405161c6ceff849d25da59742ef4f8271322ddc730de4a4acfe78abe5472eR21-R22). I've made a draft Dockerfile which installs dependencies in a container and then mounts the osu! wiki repository into `/wiki`: https://github.com/TicClick/osu-wiki/commit/7bf2ea50a8bbe5f309ff2083a1d2587faea8e6e7

The dockerfile needs further work, because 1) I use Docker extremely rarely, and 2) it installs dependencies from `package.json` into `/home/osu-wiki/node_modules`, and `npx` looks for binaries in the current directory, so unless we add a `./node_modules -> /home/osu-wiki/node_modules` symlink or make `run-ci.sh` unusable for local use without Docker (by changing `npx remark` to `/home/osu-wiki/node_modules/.bin/remark`, for example), it would break.

peppy:
The docker implementation looks pretty good as a starting point, but let's leave that as a follow-up.

To fix the file size check, how about the `wc` solution [mentioned here](https://unix.stackexchange.com/a/321502)? In testing it looks to work across platforms (make sure to use the version that removes the prefixing space).

CONTRIBUTING.md
```diff
@@ -21,3 +21,30 @@ Luckily, if you're not willing to get too involved with the technicalities of Gi
...
+
+```sh
+python3 -m venv ./.venv
+.venv/bin/pip3 install -r scripts/requirements.txt
+npm install
+```
+
+### Running
+
+From the repository root on your feature branch:
```
TicClick, line 46:
```suggestion
Using `git add`, commit or stage files you want to have checked, then run from the repository root on your feature branch:
```

TicClick (changes requested):

run-ci.sh
```diff
@@ -1,11 +1,20 @@
 #!/bin/bash
 
+# This runs CI checks for committed and unstaged files, but not untracked files.
+# For newly added files, stage or commit them before running these checks.
+
+CURRENT_BRANCH=$( git branch --show-current )
+if test ${CURRENT_BRANCH} == 'master'; then
```
TicClick, line 7:
```suggestion
if test ${CURRENT_BRANCH} = 'master'; then
```

TicClick (approved):

adamwillisMastery (commented):
Changes been reviewed.

PR #7230, opened 2022-04-23
Update `osu!trivium Quiz 2021` VOD links

Bazzadwarf:
The VOD links that were present on the page led to a content not found page (I assume the video was the original twitch broadcast and not the video highlight) so the IDs have been updated to match. Also added a link to Mangomizer's YouTube reupload just in case for whatever reason the twitch video gets lost again.


Walavouchey:
I'm sure they were *VOD* links when I put them there, but they must've been deleted and reinstated or something along the way

Walavouchey (approved):

PR #7236, opened 2022-04-24
[EN/ID] Update `Countries that participated in osu! tournaments` with `TWC 2022` results

Bazzadwarf:
No major table reordering for this one thankfully. I copied the changes over to the ID translation as it is only updating two tables, a fluent speaker still might want to double-check just in case. 

## Self-check


mousesalt (approved):
ID looks good 👍

Walavouchey (approved):

PR #7288, opened 2022-05-06
Add reference- and in-article link checking to the wikilink checker

Walavouchey:
The wikilink checker is updated with changes from https://github.com/Walavouchey/osu-wiki-tools/commit/76f2b53ad7569455a69eba665f461c372899ab3f to https://github.com/Walavouchey/osu-wiki-tools/commit/cdbfa960d9ef301a1c8158bf28e20696fd66290f to make some more progress towards https://github.com/ppy/osu-wiki/issues/6233. The full commit history along with tests can be viewed in that [repo](https://github.com/Walavouchey/osu-wiki-tools).

Huge thanks to @TicClick for pushing the project along!

## Summary

The two main features of reference- and in-article link checking require additional parsing logic to handle ignoring comments and code blocks, and getting the correct identifier from headings and custom identifier tags. For in-article links, the articles they point to are parsed for their identifiers in order to check if the #link-fragment is correct, and if not, all possible identifiers are printed as suggestions.

By default, all links from outdated articles are ignored. The plan is to eventually turn on link checking for all files in CI runs when the remaining errors have been fixed, so this is to prevent contributors from having to edit already outdated articles.

Currently on `master` this this runs for about 40s with `--outdated` and about 32s without on my Windows machine (when running on all articles). There are 152 remaining errors without counting outdated articles and 253 errors in total.

## Example output

The first two are real examples and the last three are conjured up.

<details><summary>`MissingReferenceError`:</summary>

image

The declaration has a missing space after the colon, which I suppose is fine by remark standards (remark catches undefined references in general), but this should be fine to leave as an error: `[flag_QA]:/wiki/shared/flag/QA.gif "Qatar"`
</details>

<details><summary>`MissingIdentifierError`:</summary>

image
</details>

<details><summary>`BrokenRedirectError`:</summary>

image
</details>

<details><summary>`MalformedLinkError`:</summary>

image

This is also caught by remark, but we throw an error if the conditions for it are met (see [`link_checker.py`](https://github.com/ppy/osu-wiki/pull/7288/files#diff-78f4e39f2c31ef72b87d4588346529cc36f1f42fdec296bee8c38459c3101b05R33-R35)) to be on the safe side, just in case there is a case where it's not caught by remark.
</details>

<details><summary>`LinkNotFoundError`:</summary>

image
</details>

## Testing

[Test branch](https://github.com/Walavouchey/osu-wiki/pull/22) | [Error output](https://github.com/Walavouchey/osu-wiki/runs/6324858416?check_suite_focus=true) | [Pass output](https://github.com/Walavouchey/osu-wiki/runs/6325152584?check_suite_focus=true)

<details><summary>Image of error output</summary>

image
</details>

<details><summary>Image of pass output</summary>

image
</details>


TicClick (approved):

cl8n:
I think you should use https://github.com/Walavouchey/osu-wiki-tools as a package rather than committing this much code to osu-wiki. python equivalent of https://github.com/ppy/osu-wiki/blob/28f915c30bf056b1fa91275513695f615aecbe31/package.json#L23-L26 , if that's a thing, or publish to a registry

TicClick:
@Walavouchey how's it going?

Walavouchey:
superseded by above pr

PR #7317, opened 2022-05-11
Update medal count

Blooshing:
## Self-check


Blooshing:
was asked by <chromb> if I could update the the medal count in the wiki.

Blooshing:
This should be good. The translations that are outdated weren't touched.

PR #7324, opened 2022-05-12
Add `vote` tag to content voting process

RockRoller01:
looked for this page with the search and was surprised to not see it pop up under "vote"

## Self-check


TicClick:
translation tags are a combination of the original article's tags and their translations, but you can omit them as they aren't that important -- just add the same tag to every translation that's still up to date

Walavouchey (approved):

PR #7366, opened 2022-05-23
Update `Countries that participated in osu! tournaments` for `CWC 2022`

Bazzadwarf:
Split the commits to make this slightly easier to review. I copied the changes over to the FR and ID translation as it is updating a table, a fluent speaker still might want to double-check just in case.

## Self-check


Biokax (dismissed):
fr.md is good

Bazzadwarf:
> you may want to sort the indonesian one as well

Applied the same sorting format to the ID translation as the FR article has, the GitHub files changed tab diff is pretty awful so you probably want to check the commit itself for checking over.

Walavouchey:
looking through the commits, it seems ok at a glance. going to get this in and make a script for updating the translations before next time rolls around, because checking this every time is becoming too tedious with the different sorting (although i suppose most of this is copy-pasted anyway)

Walavouchey (approved):

PR #7379, opened 2022-05-24
Add explanation for translated blockquotes to the ASC

### LINKED ISSUE #6862: Add explanation in the article styling criteria for translated quotes
Sometimes, it may occur that quotes need to be translated for the wiki. Letting the reader be able to understand what is being said in a quote is essential. However, content, meaning and emotions often get lost through the translation.

In my opinion, it should be explicitly highlighted when quotes are translated into English (or other languages). This way, the reader is made aware of that this is not the *original quote*. My proposal is to add this as a rule in the article styling criteria in the following or some other form:

````markdown
```markdown
> have fun good luck

—The-Last-Cookie (translated into English)
```
````

[@Walavouchey mentioned that I should open an issue](https://github.com/ppy/osu-wiki/pull/6839#discussion_r800088937) about this.

### PR DESCRIPTION
**Pull request by The-Last-Cookie**
Proposal to add the idea that is described in #6862 

Closes #6862 

## Self-check

### CODE REVIEW

**Comment by Walavouchey** [comment]
I don't think this is the best way to handle it. this whole blockquote thing should be split into

1. usage rules in asc/formatting (only quotations, not to be used as general indentation)
2. notes on citation style in asc/writing

**Comment by The-Last-Cookie** [comment]
Do you think it is enough to just move the text to `ASC/Writing`, maybe under `Logical quotations`? Is the text in itself fine?

File: wiki/Article_styling_criteria/Formatting/en.md [hunk 1]
```diff
@@ -794,7 +794,7 @@ The following is an example of what a table should look like:
 
 ## Blockquotes
 
-The blockquote is limited to quoting text from someone. It must not be used to format text otherwise. The name of the quoted person that follows the quote must be prepended by an em dash (`—`) as follows:
+The blockquote is limited to [quoting text someone or somewhere](/wiki/Article_styling_criteria/Writing#block-quotation). It must not be used to format text otherwise.
```
The-Last-Cookie, line 797:
```suggestion
The blockquote is limited to [quoting text from someone or somewhere](/wiki/Article_styling_criteria/Writing#block-quotation). It must not be used to format text otherwise.
```

Walavouchey:
i've pushed a change which should illustrate what kind of direction i intended here

multiline quotes [kinda break CI atm](https://github.com/ppy/osu-wiki/commit/eccdb192e0909257805ec39ae1e56ff3c64c2852), so i'd wait on that, but yes it probably would have an example in ASC/Formatting

wiki/Article_styling_criteria/Formatting/en.md
The-Last-Cookie: or
```suggestion
The blockquote is limited to [quoting someone or something](/wiki/Article_styling_criteria/Writing#block-quotation). It must not be used to format text otherwise.
```

TicClick:
looks good to me

Walavouchey (approved):

PR #7382, opened 2022-05-25
Update nomination limit for taiko BNs

Capuosu:
The past few days the nomination cap has been discussed and we came to the conclusion to update taiko only from 75 to 100 unique nominations because of the simplicity and activity of some nominators in relation to other game modes.

## Self-check


wiki/People/The_Team/Beatmap_Nominators/Rules/en.md
```diff
@@ -23,7 +23,10 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
   - **Falling below the 90 day requirements with more than 60 days worth of nominations will result in an activity warning.** 
   - When warned for activity, minimum activity requirements for your respective game mode(s) must be met over the course of one month. 
   - Failing to meet the required minimum when being warned for it prior may result in removal from the Beatmap Nominators.
-- **You cannot exceed 75 unique nominations over any 90 day period.** Those exceeding this amount will first be warned and then removed if they do not manage their activity accordingly.
+- Beatmap Nominators are expected to stay below a maximum number of nominations, with each mode having its own nomination cap over a 90 day period. Exceeding this amount will lead to a warning, followed by a removal if they fail to manage their activity accordingly.
+
+  - osu!, osu!catch, osu!mania: 75 unique nominations.
+  - osu!taiko: 100 unique nominations.
```
Walavouchey, line 29:
```suggestion
- Beatmap Nominators are expected to stay below a maximum number of nominations over any 90-day period, with each mode having its own nomination cap. Exceeding this amount will lead to a warning, followed by removal upon failure to properly manage activity.
  - osu!, osu!catch, osu!mania: 75 unique nominations.
  - osu!taiko: 100 unique nominations.
```

Ephemeralis:
I've asked @peppy to pull some stats on unique users engaging with osu!taiko maps at the moment since other preliminary statistics (as [seen here](https://puu.sh/J1T2I/a5eab618c7.png) seem to indicate that there may be a great excess of incoming osu!taiko maps entirely out of proportion with any people actually playing them, which this change would potentially worsen significantly.

Should probably hold off on merging this until that data is pulled and discussions around it have been had.

Hiviexd:
Hi, I'm one of the two taiko BNs on the activity limit, I personally don't think the problem about maps being ranked at a quick pace is that big of a concern since even if the playcount may be more thinly spread, that's not a guarantee that a significant amount of taiko players have an issue with that.

the limit feels like a huge demotivator and is just delaying the inevitable, because the maps that were modded will eventually get nominated. Is there really a difference between a map getting ranked today or getting ranked after 3-4 days? The cap makes workflow extremely limiting and adds a lot of unnecessary downtime for the BN and the mapper, this is the same issue that was presented when there was a limit on how you aren't allowed to re-nominate a set within 24h, which didn't turn out well as you know. 
this all makes it unfair to feel like you're being punished for doing well.

I can do a survey and send it to as many active taiko players as possible (through big taiko discord servers mainly) which asks the following questions:
- do you check for newly ranked maps regularly?
- do you feel like there are too many maps getting ranked right now?

this should give us a bit of an idea about the taiko community's perspective towards the amount of ranked maps we get nowadays.

Ephemeralis:
Good to go ahead with this change - after further discussion, nomination limit is not where the brakes would be applied in this instance anyway. I was mistaken in thinking it was!

Walavouchey (approved):

Walavouchey (approved):

Walavouchey:
> And a forewarning that the number of maps that will proceed from qualified to ranked _may drop for taiko_ in the near future. So keep that in mind while making this change?

if activity nomination activity were to drop for a significant amount of time, maybe we can revisit this.

PR #7395, opened 2022-05-27
Use different outdated tag for translations

### LINKED ISSUE #7383: Add new tag to distinguish outdated translations from outdated base article
> another idea is to have a separate `translation_outdated` tag, since an `outdated` in a translation doesn't distinguish whether the translation is outdated compared to the english version or if the english version is outdated (because the tag is carried over to translations in the latter case).
>
> this separate tag could then be used for [clayton.cc](https://clayton.cc/osu-wiki-status/en) to display if a translation is outdated, so that translators can more easily tell which articles need to be worked on

_Originally posted by @Walavouchey in https://github.com/ppy/osu-wiki/issues/2645#issuecomment-1137786144_

cl8n:
View diff of each commit rather than whole PR. Second commit is automated as described. Closes #7383

Before merge:

- [x] Have osu-web recognize these tags https://github.com/ppy/osu-web/pull/8963

After:

- Update any other tooling around osu-wiki
- Update currently open PRs
- Update ASC https://github.com/ppy/osu-wiki/pull/7397


DoryanR:
A real revolution for us, the translators, it will really help us to distinguish the outdated articles from the articles that follow the outdated of the original article.

Let's see what it will do on `clayton.cc`.

Walavouchey (approved):

PR #2156, opened 2019-04-04
Ranking Criteria/BN rule updates

pishifat:
an unreasonable amount of threads this time

https://osu.ppy.sh/community/forums/topics/847666: standardizing semi-common metadata phrases
~~https://osu.ppy.sh/community/forums/topics/832111: minor wording adjustment to match how users respond to guidelines~~ removed this one, see thread
https://osu.ppy.sh/community/forums/topics/868716: moved from general to osu! specific
https://osu.ppy.sh/community/forums/topics/860714: added rule about applause sfx filetype
https://osu.ppy.sh/community/forums/topics/877557: clarified how source can be used in unclear situations
https://osu.ppy.sh/community/forums/topics/837952 and https://osu.ppy.sh/community/forums/topics/853985: edited bg regulations
https://osu.ppy.sh/community/forums/topics/871492: "must" changed to "should"
https://osu.ppy.sh/community/forums/topics/872345: clarified multi-mode bn exception
https://osu.ppy.sh/community/forums/topics/744298: clear sliderpath clarification
https://osu.ppy.sh/community/forums/topics/857987: general guideline cleanup
https://osu.ppy.sh/community/forums/topics/857982: osu! guideline cleanup
https://osu.ppy.sh/community/forums/topics/888625: reverted part of veto rule
https://osu.ppy.sh/community/forums/topics/889929: handling whitespaces in tags
https://osu.ppy.sh/community/forums/topics/857306 ctb thing 1
https://osu.ppy.sh/community/forums/topics/844454 ctb thing 2
https://osu.ppy.sh/community/forums/topics/889616 difficulty naming

---

PR #1225, opened 2018-04-08
change sanity limits

Okorin:
outcome of: https://osu.ppy.sh/forum/t/718100

### Status

done

PR #7433, opened 2022-06-03
Update `Ranking Criteria`

pishifat:
applying a handful of mostly minor changes from the rc subforum. i'm planning to use footnotes to make the rest of the article easier to read once this is merged

changes:

- [VO metadata standardization](https://osu.ppy.sh/community/forums/topics/1344928) - current wording enforced incorrect metadata
- [Allow backgrounds used as storyboard images to use storyboard image dimensions](https://osu.ppy.sh/community/forums/topics/1435518) - avoids redundant images in the song folder. if there are technical problems with allowing large images as bgs in these cases, nobody would know besides devs(?)
- [Broaden source field usage](https://osu.ppy.sh/community/forums/topics/1498950) - original thread wanted unofficial tourney-specific songs to be prioritized over "osu!" in the source field, so this change influences that + other media sources in the same boat
- [Broaden tags requirements](https://osu.ppy.sh/community/forums/topics/1492863) - stupidly written words in song fields -> stupidly written words in ALL fields are required to be tagged
- [Clarify exceptions for extra timing points](https://osu.ppy.sh/community/forums/topics/1564613) - the current wording was confusing enough for someone to write a thread misinterpreting it
- [Capping millisecond window between notes](https://osu.ppy.sh/community/forums/topics/1563145) - as per the conflict on https://osu.ppy.sh/beatmapsets/1691083 / https://osu.ppy.sh/beatmapsets/1691083/discussion/-/generalAll#/3020552

---

- [x] NAT approval
- [x] peppy approval


wiki/Ranking_Criteria/en.md
```diff
@@ -294,3 +295,7 @@ This category contains explicit allowance statements of concepts and rules that
   - **Refrain from having multiple sprites active while not rendered.** Active sprites will still process commands regardless of whether they are visible or not. Should this be the case for longer periods of time, instantiate new sprites instead, for when visibility is regained.
   - **When using many commands of the same type on a sprite, try leaving at least 16 ms between their start times.** 60 commands per second is often more than enough for any sprite to make smooth transitions on an average setup. This is for the sake of reducing file size and loading times.
   - **Fade out sprites activated from triggers after usage.** Triggers will activate from their first possible command and stay active until the end of the difficulty, which is why fading these out when done is preferable.
+
+## Notes
+
+[^maximum-dimensions]:Backgrounds which are also used as storyboard images may follow [the storyboard images rule for maximum dimensions](/wiki/Ranking_Criteria#rules.7) instead.
```
cl8n, line 301:
```suggestion
[^maximum-dimensions]: Backgrounds which are also used as storyboard images may follow [the storyboard images rule for maximum dimensions](/wiki/Ranking_Criteria#rules.7) instead.
```

cl8n (changes requested):
needs the space

wiki/Ranking_Criteria/en.md
```diff
@@ -229,8 +230,8 @@ This category contains explicit allowance statements of concepts and rules that
 - **The following are requirements for background images:**
   - **Minimum width:** 160px
   - **Minimum height:** 120px
-  - **Maximum width:** 2560px
-  - **Maximum height:** 1440px
+  - **Maximum width:** 2560px[^maximum-dimensions]
+  - **Maximum height:** 1440px[^maximum-dimensions]
```
peppy, line 234: These dimensions are larger than I recall them being, kinda scary in terms of the technical requirements but I guess it's what has been in place. Also I don't know who made the storyboard rule about 17m pixels but that's pretty high too. Dangerously.

wiki/Ranking_Criteria/en.md
```diff
@@ -106,7 +107,7 @@ Important to understand before using:
 - **Commas, `vs.`, `feat.`, `CV:` and any other symbols linking or designating artists must include a trailing space.** Excluding commas, a leading space is also required if the marker is preceded by a word.
 - **Any form of `vs.`, `VS`, etc. must be written as `vs.` when used as an indicator for collaboration between two or more artists.** Alternative casing may be used to match the rest of the field.
 - **Any form of `feat.`, `feat`, `ft.`, etc. must be written as `feat.` when used as an indicator for an artist featured in the song.** Alternative casing may be used to match the rest of the field.
-- **When a fictional character is credited as the singer of a song, their credit is to be formatted in a `Character (CV: Voice Actor)` format.** For live action, credit the voice actor only.
+- **When a fictional character is credited as the singer of a song, their credit is to be formatted in a `Character (CV: Voice Actor)` or `Character (VO: Voice Actor)` format.** For live action, credit the voice actor only.
```
Walavouchey, line 110: this would be a good place to put a footnote that explains the difference between CV and CO

wiki/Ranking_Criteria/en.md
```diff
@@ -30,6 +30,7 @@ Important to understand before using:
 ### Rules
 
 - **No two hit objects can be placed on the same tick.** This includes hit circles and the durations of sliders and spinners. osu!mania difficulties are exempt from this.
+- **No two inputs can be less than 20 milliseconds apart.** osu!mania and osu!catch difficulties are exempt from this.
```
Naxesss, line 33: Are we sure we want this at 20 ms instead of the 10 ms enforced by AiMod/MV? The thread seems to suggest the change *needs* to happen despite community backlash, but after re-reading both the thread and internal discussions, I do not see the reason why.

As far as I understood this is a diff-calc related issue, which I have since confirmed is tracked and worked on by the performance team. Apart from that, we have found no issues with keeping the current sanity limit of 10 ms.

So I feel like there may be a misunderstanding here.

wiki/Ranking_Criteria/en.md
Walavouchey: bg image context: [pr](https://github.com/ppy/osu-wiki/pull/2156), [thread 1 (mostly about resolution)](https://osu.ppy.sh/community/forums/topics/837952?n=1), [thread 2 (mostly about file size)](https://osu.ppy.sh/community/forums/topics/853985?n=1)

people use high-fidelity monitors these days so i can definitely understand this one. i'm not sure why a static 1440p image would cause technical concern

sb image context: [pr](https://github.com/ppy/osu-wiki/pull/1225), [thread](https://osu.ppy.sh/community/forums/topics/718100?n=5)

apparently @smoogipoo suggested it, with 4096x4096 being a little under 17m

wiki/Ranking_Criteria/en.md
```diff
@@ -144,7 +145,7 @@ Important to understand before using:
 
 - **If the creator of the beatmap has remixed or covered the song, they are free to name it appropriately to signal that this song is a special version.** In this case the original songs should still be clearly indicated in the title or tags in order for players to be able to search for the original songs.
 - **In the case of compilations or remixes, the original song title(s) and artist(s) should be included in tags.** This is to ensure that players can find all beatmaps of one song by searching the same thing without getting vastly different results.
-- **Songs with metadata that contains ambiguous or hard to write unicode characters should add easily searchable variations or romanisations of these words to the beatmap's tags.**
+- **Any metadata that contains ambiguous or hard to write unicode characters should add easily searchable variations or romanisations of these words to the beatmap's tags.**
```
Walavouchey, line 148:
```suggestion
- **Any metadata that contains ambiguous or hard-to-write unicode characters should be accompanied by easily searchable variations or romanisations of these words in the beatmap's tags.**
```

wiki/Ranking_Criteria/en.md
smoogipoo: I have no recollection of suggesting this, fwiw. So you can discard anything I suggested and re-evaluate as if I didn't.

peppy: some of the forum threads mention wanting to get my opinion but either i am very forgetful (not usually the case) or that never happened 😅

wiki/Ranking_Criteria/en.md
peppy: Definitely should be understood before making a call.

Can you cc in the performance team that you have discussed this with, and link an issue/tracking thread?

wiki/Ranking_Criteria/en.md
Walavouchey: concerns aside, there are ranked storyboards with images above e.g. 10m pixels fwiw:

- https://osu.ppy.sh/beatmapsets/935708 (4000 x 4000 = 16000000)
- https://osu.ppy.sh/beatmapsets/1134425 (4096 x 2731 = 11186176)
- https://osu.ppy.sh/beatmapsets/1188324 (4500 x 2531 = 11389500)
- https://osu.ppy.sh/beatmapsets/1246880 (2804 x 4096 = 11485184)

there aren't many (probably any) cases where large images are necessary though, because the images can just be split into multiple sprites. [sakuraburst - 「skyshifter vip」](https://osu.ppy.sh/beatmapsets/701482#osu/1637011) for example uses three 1920x1080 px images to create a vertical panorama

peppy: Multiple smaller images is definitely the preference, to the point I'd consider changing this criterial to reduce the limit to 12m or lower...

wiki/Ranking_Criteria/en.md
Naxesss: I asked @apollo-dw on 2022-04-22 about getting an issue opened, who replied simply saying that it is already tracked and known in the PP dev server, and that he'd try getting a fix out in the next wave of changes. That is all I know.

apollo-dw: Yup, this is tracked in the PP server (there are no relevant github issues / other threads to my knowledge, if wanted). In my eyes, the SR inflation issue is going to be addressed as soon as the diff-calc refactor stuff happens that makes it much more feasible. I'd be hesitant to say that the diff-calc issue should be taken into account with deciding the amount of milliseconds here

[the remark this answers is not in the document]
Walavouchey: apparently this was [addressed in the thread](https://osu.ppy.sh/community/forums/topics/1563145?n=2) with the wording of "active input" (i.e. key down; as in active object, things you click on), but the wording here seems to miss that point. i'd prefer using "active object" here since it encompasses the same meaning, although there should probably be a stub for it first (or have the [osu!catch-related stub](https://osu.ppy.sh/wiki/en/Glossary/Active_object) expanded)

pishifat: i was under the impression that 10ms would be too little and 20ms was a solid minimum based on discussions that happened the day this issue came up. i'm not an authority on which should be chosen since this is a technical concern, but from a mapping side, 10ms wouldn't be a problem imo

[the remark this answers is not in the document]
pishifat: this refers to 2 keydowns as wala says. the idea is to allow sliderend->next object timeframes to be smaller than 20ms. that said, i've changed "inputs" -> "key presses" for now, and can do what wala suggests about article expansion in the future

wiki/Ranking_Criteria/en.md
```diff
@@ -177,7 +178,7 @@ This category contains explicit allowance statements of concepts and rules that
...
 - **[Uninherited timing points](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must be the same in every difficulty of a beatmap.** Each point must have the same [BPM](/wiki/Beatmapping/Beats_per_minute) and [offset](/wiki/Beatmapping/Offset) in each difficulty.
-- **There must not be extra [uninherited timing points](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) in any difficulty.** These may accidentally affect main-menu pulsing, add unwanted sounds to the [Nightcore mod](/wiki/Game_modifier/Nightcore), or cause timing to shift. Acceptable uses include:
+- **There must not be extra [uninherited timing points](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) in any difficulty.** These may accidentally affect main-menu pulsing, add unwanted sounds to the [Nightcore mod](/wiki/Game_modifier/Nightcore), or cause timing to shift. Exceptions to this rule include:
```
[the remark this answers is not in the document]
pishifat: hmm the original wording led to misinterpretations already, so i'd rather not keep that, but i'm open to other ideas. "Acceptable uses of extra uninherited timing points include:" is the intent, so maybe that? it's a mouthful though

wiki/Ranking_Criteria/en.md
Naxesss: As long as the framerate limited input is per-key it should be fine at 10 ms I think (ignoring diffcalc), since the idea is to press both buttons at the same time. If it queues them up so the next input comes at the next frame, it could be a problem. 

But yeah we need more details about how the frame limited input works to make an informed decision here I think.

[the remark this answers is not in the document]
peppy: I definitely prefer "key press" or "key down", since "active input" means nothing to us as developers, let alone users.

wiki/Ranking_Criteria/en.md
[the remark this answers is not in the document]
peppy: How about rewording the whole paragraph to read more like

"Uninherited timing points should only be used when required for timing purposes..."

Also I'm not sure uninherited is a word.

wiki/Ranking_Criteria/en.md
peppy: Which existing beatmaps have objects closer than 20ms? Would be good to test them out on both stable and lazer to see how things behave.

wiki/Ranking_Criteria/en.md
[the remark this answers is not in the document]
Walavouchey: Inherited and uninherited are terms mappers have used for ages

I suppose it would be:

```suggestion
- **[Uninherited timing points](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) should only be used when required for timing purposes.** Unnecessary timing points may accidentally affect main-menu pulsing, add unwanted sounds to the [Nightcore mod](/wiki/Game_modifier/Nightcore), or cause timing to shift. Other acceptable uses include:
```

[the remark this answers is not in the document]
peppy: Two things:

- The listed acceptable uses are all timing related, so I'd drop the "Other"
- Maybe change "accidentally" to "unintentionally"

wiki/Ranking_Criteria/en.md
pishifat: none ranked, but the reason this rule is being considered is https://osu.ppy.sh/beatmapsets/1691083#osu/3455732 (13ms)

peppy: After discussion with everyone that has any knowledge surrounding these issue, 10ms sounds fine, except for one special case: After a slider end, the next object needs to be >20ms.

How that is phrased is up for discussion.

pishifat: split into two rules, how does this sound?
`- **There must be at least 10 milliseconds between a hit circle and the next object.** osu!mania and osu!catch difficulties are exempt from this.`
`- **There must be at least 20 milliseconds between the end of a slider and the next object.** osu!mania and osu!catch difficulties are exempt from this.`

wiki/Ranking_Criteria/en.md
```diff
@@ -30,7 +30,8 @@ Important to understand before using:
 ### Rules
 
 - **No two hit objects can be placed on the same tick.** This includes hit circles and the durations of sliders and spinners. osu!mania difficulties are exempt from this.
-- **No two key presses can be less than 20 milliseconds apart.** osu!mania and osu!catch difficulties are exempt from this.
+- **There must be at least 10 milliseconds between a hit circle and the next object.** osu!mania and osu!catch difficulties are exempt from this.
```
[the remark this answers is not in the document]
pishifat: not sure simplifying would work too well. the map with circles 13ms apart doesn't snap them to supported timeline ticks (1/24), so it'd be unrankable by the simplification, or the logic that permits 1/24 would also permit circles <10ms apart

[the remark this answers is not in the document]
peppy: I'm fine if that's the case, but would defer to others for further opinions.

peppy (approved):
Looking okay from my end.

Walavouchey (approved):

peppy (approved):

PR #7442, opened 2022-06-04
Use wikilink checker externally

Walavouchey:
Supersedes https://github.com/ppy/osu-wiki/pull/7288, part of https://github.com/ppy/osu-wiki/issues/6233

[Walavouchey/osu-wiki-tools](https://github.com/Walavouchey/osu-wiki-tools) is put into a [pip package](https://pypi.org/project/osu-wiki-tools/) and imported when running CI, as [suggested](https://github.com/ppy/osu-wiki/pull/7288#issuecomment-1122044268) by @cl8n

Not much in the actual script used has changed since the other PR, apart from [updating it to use the new `outdated_translation` tag](https://github.com/Walavouchey/osu-wiki-tools/commit/dd668e3ea5e42d1f4ab1bfb4ac4e50cc00c07de6)

The commits for each version are tagged in the osu-wiki-tools repo, and further updates will simply require a version bump here

## Testing

[Test branch](https://github.com/Walavouchey/osu-wiki/pull/23) | [Error output](https://github.com/Walavouchey/osu-wiki/runs/6737733550?check_suite_focus=true) | [Pass output](https://github.com/Walavouchey/osu-wiki/runs/6737749573?check_suite_focus=true)


.github/workflows/continuous-integration.yml
```diff
@@ -73,8 +73,10 @@ jobs:
             exit 0
           fi
 
+          pip install osu-wiki-tools
+
```
TicClick, line 77: you have installed it on the line 63 already
```suggestion
          pip install osu-wiki-tools

```

.github/workflows/continuous-integration.yml
```diff
@@ -73,8 +73,10 @@ jobs:
...
           ARTICLES=$( git diff --diff-filter=d --name-only ${{ github.sha }}^ ${{ github.sha }} "wiki/**/*.md" "news/*.md" )
-          python scripts/ci/find_broken_wikilinks.py --target $ARTICLES
+          find-broken-wikilinks --target $ARTICLES
```
TicClick, line 79: it looks like all the scripts from the package are added to `PATH`. I feel like making a small CLI module that would proxy the arguments and run necessary scripts is better for clarity and usability. something like that (in addition to the way the scripts are invoked right now):

```python
wikitools-cli check-links --target ... --some-other-flag=...
```

please keep in mind that it's not a requirement for the PR to be merged, but something to do to the tools collection anyway, separately

TicClick (changes requested):

TicClick (approved):
I approve of the change, but again, I took part in this, so I'd like someone else to cast their vote (and maybe look at the code if they will)

peppy (approved):

PR #7449, opened 2022-06-04
Outdate very old mapping techniques articles

cl8n:
even the parts that aren't "wrong" are super old and don't reflect really anything about mapping today


TicClick:
> even the parts that aren't "wrong" are super old and don't reflect really anything about mapping today

could you include this into the articles as an inline comment for quick access, in case anyone takes it upon themselves to refresh the info?

cl8n:
hopefully that's good enough

TicClick (approved):
it is

PR #7471, opened 2022-06-08
Fix some of the broken wiki links

TicClick:
(didn't touch outdated articles)

`SKIP_WIKILINK_CHECK` | `SKIP_OUTDATED_CHECK`


TicClick:
some of the links in the articles I touched are still broken, as they point to outdated articles that don't have corresponding headings. forgot what was the tag for ignoring the check.

Walavouchey (approved):

PR #7526, opened 2022-06-18
Update Medal Count

Blooshing:
## Self-check


peppy:
I don't know how productive it is to have a dynamic number like this on the wiki...

PR #7376, opened 2022-05-24
Replace informal usage of hyphens with more appropriate formatting

The-Last-Cookie:
This replaces every hyphen (`-`) with an em dash (`—`) where applicable as according to the [ASC](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Writing#em-dash). For finding out every dash, I've used [the script in this PR](https://github.com/The-Last-Cookie/osu-wiki/pull/1) (this also contains the full diff the script produced). Date ranges and names were left unchanged. I'm not sure for some dash replacements in this PR, so this has to be reviewed on a line-to-line basis (especially number ranges, for example `7-10`).

`SKIP_OUTDATED_CHECK`

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #7530, opened 2022-06-18
Update dash usage rules in the ASC

Walavouchey:
Accompanying changes to the ASC along with https://github.com/ppy/osu-wiki/pull/7376

Feel free to suggest different wording or added clarification

I've intentionally used double quotes instead of code blocks because dashes and hyphens are tricky to tell apart in monospace fonts

- [x] May need to check how the translations will follow suit, if at all (checked off by clayton -- can be done at another time)


wiki/Article_styling_criteria/Writing/en.md
```diff
@@ -336,9 +336,9 @@ Contractions must not be used.
 
 The serial comma, also known as the Oxford or Harvard comma, must be used.
 
-### Em dash
+### Dashes
 
-Em dashes (`—`) must be used without spaces around them. Hyphens must not be used in their place.
+Refer to [Wikipedia's Manual of Style for dash usage](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style#Dashes). When punctuating a sentence however, use em dashes with spaces around them (" — "), not spaced en dashes (" – "). Hyphens ("-") must not be used in place of dashes.
```
TicClick, line 341: > Hyphens must not be used in place of dashes.

I hope "must" here means that they will be *eventually* converted to dashes, not that it *must* be done at the review stage, since in #7376 most reviewers agree it's not a tragedy if new hyphens pop up

Walavouchey: there's no consequence for not following the asc. if no one catches it, the same could be said for any "rule" here

TicClick: right

cl8n: in practice "must" in asc is like "if the reviewer(s) feel it's important enough to correct at this time" so I don't think there is much of a difference here

TicClick (approved):

peppy (approved):

PR #7532, opened 2022-06-19
[FR] Change "le/de osu!" to "l/d'osu!" and "du osu!" to "de l'osu!"

Walavouchey:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review) @Biokax @DoryanR 

based on [discussion in discord](https://discord.com/channels/188630481301012481/218677502141399041/972559065249112184)

```sh
for file in $( find -type f -name "fr.md")
do
    perl -i -pe "s/\b([ldLD])e ((?:\**)|\[)os([ub])/\1'\2os\3/g" $file
    perl -i -pe "s/\b([dD])u ((?:\**)|\[)os([ub])/\1e l'\2os\3/g" $file
done
git restore wiki/People/The_Team/Staff_Log/2009/fr.md
```

PR #7541, opened 2022-06-20
Use ICU/CLDR language display names for group page translations

cl8n:


wiki/People/The_Team/Beatmap_Nominators/fr.md
```diff
@@ -46,11 +46,11 @@ Les listes ci-dessous indiquent dans quel(s) mode(s) de jeu chaque Beatmap Nomin
 
 | Nom | Langues supplémentaires |
 | :-- | :-- |
-| ![][flag_PH] [-Aqua](https://osu.ppy.sh/users/7150015) | Philippin |
+| ![][flag_PH] [-Aqua](https://osu.ppy.sh/users/7150015) | Filipino |
```
DoryanR, line 49:
```suggestion
| ![][flag_PH] [-Aqua](https://osu.ppy.sh/users/7150015) | Philippin |
```

wiki/People/The_Team/Global_Moderation_Team/fr.md
```diff
@@ -65,74 +65,74 @@ La page du groupe [Global Moderation Team](https://osu.ppy.sh/groups/4) donne la
...
+| ![][flag_BR] [LeoFLT](https://osu.ppy.sh/users/3668779) | Portugais, espagnol | Modération du tchat, gestion des tournois |
 | ![][flag_DE] [Loctav](https://osu.ppy.sh/users/71366) | Allemand | Modération de la communauté modding/mapping |
 | ![][flag_TW] [Loneight](https://osu.ppy.sh/users/663131) | Chinois | Modération du tchat |
-| ![][flag_GB] [mangomizer](https://osu.ppy.sh/users/1893718) | Cantonais, Chinois | Gestion des tournois |
+| ![][flag_GB] [mangomizer](https://osu.ppy.sh/users/1893718) | Cantonais, chinois | Gestion des tournois |
 | ![][flag_BR] [Maot](https://osu.ppy.sh/users/3914271) | Portugais | Modération de la communauté modding/mapping |
 | ![][flag_NO] [MillhioreF](https://osu.ppy.sh/users/941094) |  | Développement d'osu! |
 | ![][flag_VN] [My Angel Chino](https://osu.ppy.sh/users/20547597) | Vietnamien | Modération du tchat |
-| ![][flag_PH] [Nathanael](https://osu.ppy.sh/users/2295078) | Philippin | Modération du tchat, modération du forum, support technique |
+| ![][flag_PH] [Nathanael](https://osu.ppy.sh/users/2295078) | Filipino | Modération du tchat, modération du forum, support technique |
```
DoryanR, line 99:
```suggestion
| ![][flag_PH] [Nathanael](https://osu.ppy.sh/users/2295078) | Philippin | Modération du tchat, modération du forum, support technique |
```

wiki/People/The_Team/Support_Team/fr.md
```diff
@@ -24,8 +24,8 @@ La [page de groupe de la Support Team](https://osu.ppy.sh/groups/22) énumère t
 | ![][flag_US] [Death](https://osu.ppy.sh/users/3242450) |  |
 | ![][flag_US] [Dntm8kmeeatu](https://osu.ppy.sh/users/5428812) |  |
 | ![][flag_NO] [MillhioreF](https://osu.ppy.sh/users/941094) |  |
-| ![][flag_PH] [Nathanael](https://osu.ppy.sh/users/2295078) | Philippin |
-| ![][flag_SE] [Saten](https://osu.ppy.sh/users/444506) | Suédois, Espagnol |
+| ![][flag_PH] [Nathanael](https://osu.ppy.sh/users/2295078) | Filipino |
```
DoryanR, line 27:
```suggestion
| ![][flag_PH] [Nathanael](https://osu.ppy.sh/users/2295078) | Philippin |
```

DoryanR (changes requested):
The only errors for the French language are listed below.

`Filipino`, I have the impression that it has not been translated, because it is English. When I check on Wikipedia, it is `Filipino`.

DoryanR:
Also, I have another problem (which has nothing to do with the changes here) concerning group info.

When the same person speaks more than one language, there is a `some <language>` addition, except that in French it doesn't fit all languages.

Let me explain: for example, if I a person speaks French, and a little Italian, in our side we will say `français, un peut de Italien`, except that it doesn't fit, because it sounds weird to say `un peu de Italien`. The real spelling should be `un peu d'Italien`. 
This problem occurs in many languages in the articles affected by this PR.

In view of what Walavouchey did on #7532, I wonder if it would be possible to do the same thing here, it would be more French.

DoryanR:
> I thought so too about "Filipino" but it actually is translated like that in most languages that use latin characters. not sure if there's some particular reason for that https://github.com/unicode-org/cldr/blob/main/common/main/fr.xml#L192

That's weird, that's not what I saw [on Wikipedia](https://fr.wikipedia.org/wiki/Philippin)

> what kind of rule is it to decide between `un peu de` and `un peu d'` ? is that something i can determine by program?

Hmm well, I don't think so, for example for Filipino it's `un peut de Philippin`, but for Arabic it's `un peut d'Arabe`.

Apart from I guess making a list of the different terms for each language (like `group-info`), I don't really see how you can do it. 
For me there are no real rules, I'll have to discuss it with the other translators to really know what it is.

Ganondorf-1:
Generally, you have to use `un peu d'` when it's a vowel (un peu d'espagnol, un peu d'italien...) and `un peu de` when it's a consonant (un peu de français, un peu de japonais...)

Also.. I'm not 100% sure about whether we should use "philippin" or "filipino", but Google and other websites keep trying to change "filipino" into "philippin" so I guess we can go with "philippin"

DoryanR:
In the list everything seems good to me, except for the Hungarian, where I would put rather 'un peut d'Hongrois'.

Walavouchey:
the french wikipedia article i linked uses "le/de hongrois" 🤔

meta/group-info/fr.yaml
```diff
@@ -32,39 +29,5 @@ nat:
...
-  sr: Serbe
-  sv: Suédois
-  te: Telugu
-  th: Thaïlandais
-  tr: Turc
-  ur: Urdu
-  vi: Vietnamien
-  yue: Cantonais
-  zh: Chinois
+  partial_vowel_prefix: un peu d'<language> # This key is specific to FR
```
Walavouchey, line 32: to match with `partial` :)

```suggestion
  partial_vowel: un peu d'<language> # This key is specific to FR
```

wiki/People/The_Team/osu!_Alumni/fr.md
```diff
@@ -106,7 +106,7 @@ La page [osu! Alumni group page](https://osu.ppy.sh/groups/16) donne la liste de
 | ![][flag_CA] [jonathanlfj](https://osu.ppy.sh/users/270377) | BAT |
 | ![][flag_AR] [juankristal](https://osu.ppy.sh/users/443656) | Gestion des tournois |
 | ![][flag_MX] [Kai](https://osu.ppy.sh/users/4537) | BAT |
-| ![][flag_CA] [karterfreak](https://osu.ppy.sh/users/1031958) | GMT, créateur d'osu!media |
+| ![][flag_CA] [karterfreak](https://osu.ppy.sh/users/1031958) | GMT, créateur de osu!media |
```
Walavouchey, line 109: this should be "créateur d'osu!media"

wiki/People/The_Team/Global_Moderation_Team/id.md
```diff
@@ -73,42 +73,42 @@ Halaman daftar [Global Moderation Team](https://osu.ppy.sh/groups/4). Sebagai ta
...
 | ![][flag_GB] [JBHyperion](https://osu.ppy.sh/users/4879508) |  | Moderasi chat, moderasi komunitas mapping/modding |
-| ![][flag_HK] [kanpakyin](https://osu.ppy.sh/users/394326) | Kanton, Mandarin, Jepang | Moderasi chat |
+| ![][flag_HK] [kanpakyin](https://osu.ppy.sh/users/394326) | Kanton, Tionghoa, Jepang | Moderasi chat |
 | ![][flag_DE] [Kite](https://osu.ppy.sh/users/134572) | Jerman | Moderasi komunitas mapping/modding |
 | ![][flag_RU] [Kobold84](https://osu.ppy.sh/users/3227533) | Rusia | Moderasi chat |
 | ![][flag_JP] [KSHR](https://osu.ppy.sh/users/409957) | Jepang | Moderasi chat |
 | ![][flag_RU] [Kudou Chitose](https://osu.ppy.sh/users/9936528) | Rusia | Moderasi chat |
 | ![][flag_RU] [Kyubey](https://osu.ppy.sh/users/2195646) | Rusia | Moderasi chat |
-| ![][flag_FI] [Laurakko](https://osu.ppy.sh/users/7253731) | Finlandia | Moderasi chat |
+| ![][flag_FI] [Laurakko](https://osu.ppy.sh/users/7253731) | Suomi | Moderasi chat |
```
Walavouchey, line 87: that's finnish in finnish 🤔

cl8n: and finnish in id according to cldr

meta/group-info/fr.yaml
cl8n: "prefix" is for "vowel prefix" cuz its when language starts with vowel

Walavouchey (approved):

PR #7187, opened 2022-04-15
Reorganise many articles according to #6080

Gamingzone123:
start to move stuff around towards #6080 

there is no Storyboarding\Specifications\Concepts\Coordinate_space for `osu!_pixel` to be linked to

waiting on:

- #7270 
- #7442

depends on:

- (separate) <s>the timing article being rewritten/amalgamated from `Timing_section` and Client\Beatmap_editor\Timing</s>
- (separate) <s>discussion of where to put offset articles</s>

I also don't get how:

- (separate) <s>the current `Touhosu` article can be merged into the `April_fools` article as they seem completely separate from one another</s>
- (Fixed) <s>CI is getting an issue with a pipe and I'm not sure how to fix this</s>

- [x] add redirects in `redirect.yaml`

PR #7547, opened 2022-06-21
Fix redundant redirects

cl8n:
there are many other problems with redirects but this is only fixing where redirects referred to other redirects, or an existing article. pls no expand scope


Walavouchey:
hmm seems like the timing and slider tick rate ones were leftovers from reverted moves in https://github.com/ppy/osu-wiki/pull/7187

PR #7549, opened 2022-06-21
[ID/FR] Update osu-wiki status link to their specific language

daud-ahn:
## Self-check


daud-ahn:
what is wrong with the task-list-completed check?

Walavouchey:
last time it took about an hour for https://github.com/ppy/osu-wiki/pull/7545

PR #7559, opened 2022-06-24
Fix style of a CI script

TurnipGuy30:
Removed unnecessary indentation from one line.


TicClick (approved):
pointless but harmless

PR #7594, opened 2022-07-03
Add Murumoo to 2021 Elite Nominator newspost

Ephemeralis:
Upon request of the NAT.


peppy:
Kinda weird adding after the fact but okay

peppy (approved):

news/2022-03-22-community-contributors-elite-nominators.md
```diff
@@ -112,6 +112,12 @@ A dedicated Beatmap Nominator since 2018, [**\_Stan**](https://osu.ppy.sh/users/
 
 Without a doubt, \_Stan is a major force in the 7K scene, and things would be very different without him!
 
+[![](/wiki/shared/news/2022-03-22-community-contributors-elite-nominators/Murumoo.jpg)](https://osu.ppy.sh/users/1653229)
```
frenzibyte, line 115: I think you got the wrong user ID there? 👀

cl8n:
seems both very pointless as an announcement, with practically no visibility, and very misleading for anyone looking back, with no indication that this one was amended later.

think it'd be better off as either its own news post or wait for the next one. or at least say it was added later, if those two options don't work for some reason

PR #7599, opened 2022-07-04
Disable CI workflow runs for the master branch

Walavouchey:
they're only redundant since all changes are run through them via pull requests anyway, and they pointlessly keep failing when they are skipped

[context](https://discord.com/channels/188630481301012481/218677502141399041/993418941454102619)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,9 +1,6 @@
 name: osu-wiki continuous integration
 
 on:
-  push:
-    branches:
-      - master
   pull_request:
     branches:
       - master


peppy:
Any specific reason for doing this? They seem to be completing quite fast so I'd probably just let it be unless there's a good reason.

Walavouchey:
not really other than failing whenever they are skipped, and although that can be fixed more surgically, me and clayton figured they might not be necessary at all. close if you will.

cl8n:
also, this workflow *doesn't* check that the whole repo is in a valid state at this commit, so it was kind of odd to run on master to begin with

peppy:
Yeah it is a bit weird that it doesn't check everything. Probably need to eventually have a separate workflow for that so we can see the overall status. Maybe.

TicClick:
would rather make a proper fix which I never really implemented

TicClick (approved):

cl8n:
will revisit CI on master when we're able to run checks on more than just changed files

cl8n (approved):

PR #7607, opened 2022-07-05
Update `Skinning/osu!` and `Skinning/History`

### LINKED ISSUE #7246: Remove hitcircleoverlay being animatable
This used to be a thing, but recently got changed to no longer be possible. Animated elements should also get added to skinning/history.
There is no entry for this in the changelog, but seems to have happened around a month ago based on comments. I have tested it myself and it no longer works. Maybe @peppy wants to verify that this got removed

RockRoller01:
Sets hitcircleoverlay, sliderstartcircleoverlay and sliderendcircleoverlay as non-animatable and adds their animatable sprites to skinning/history

closes #7246

## Self-check


TicClick (dismissed):

RockRoller01:
there actually is one mistake still, it says sliderstartcircleoverlay instead of sliderendcircleoverlay at once spot

TicClick (approved):
good catch

PR #7608, opened 2022-07-06
Add `MWC 4K 2022` wiki article

LeoFLT:
🎹 

News post to come at a later date

## Self-check


peppy:
Note that the non-@2x banner wasn't correctly sized so I fixed that up.

peppy (approved):

PR #7626, opened 2022-07-09
[EL] Add `group-info`

Blooshing:
Would like to wait on any type of merge, I have a person that is willing to review this, but if they don't get to it in an orderly fashion then `no_native_review: true` is there for such reason. 

## Self-check


meta/group-info/el.yaml
```diff
@@ -0,0 +1,35 @@
+---
+no_native_review: true
+---
```
cl8n, line 3:
```suggestion
no_native_review: true
```

cl8n (approved):

PR #7643, opened 2022-07-13
Update `MWC 4K 2022` staff listing, flag reference syntax

LeoFLT:
Updates the staff listing. Also updates flag references to the new `::{ flag="XX" }::` format.
## Self-check


peppy:
Did something change apart from the flag syntax? I can't seem to see it.

LeoFLT:
> Did something change apart from the flag syntax? I can't seem to see it.

Davvy has been removed from Mappool helpers at his request, his name was before guden on that table

peppy (approved):

PR #7682, opened 2022-07-22
Add final member of the osu!taiko pp committee

Lawtrohux:
This adds the final member of the osu!taiko pp committee, and changes the wording as active changes are now implemented.

`SKIP_OUTDATED_CHECK`


TPGPL (approved):
looks ok

Walavouchey:
the new table row can be copied to the translations while keeping the outdated marker if changing the one paragraph is too much trouble

TPGPL (approved):
lgtm

Walavouchey (approved):
outdated hash check failed for some reason, will investigate

https://github.com/ppy/osu-wiki/runs/7468467079?check_suite_focus=true

PR #7699, opened 2022-07-25
[JA] Add `Game_mode/osu!`

yuzupon1133:
## Self-check


peppy (changes requested):
Seems fine apart from one particle

peppy (approved):

PR #7719, opened 2022-07-29
Update `MWC 4K 2022` team and staff listings

LeoFLT:
All team rosters are finalized.

## Self-check


peppy (approved):

RushFTK:
just curiosity, is community stream might allowed to add in links part?
* **community stream** means the streams host by community member, major for localization, such as Chinese, Português , etc.   
* *only large enough community stream should be consider, depend on tournament team's choice. one stream max per language.*

PR #7730, opened 2022-07-30
Fix commit hash validation not working in the outdated checker

Walavouchey:
This reverts commit 179fe2eab37b7cad431f1a00b3695df63d94c5d9.

this broke the commit hash check in the outdated checker which uses git-show to check if the hash is valid

actions/checkout@v3 must be doing the checkout slightly differently than what we used to have, but i haven't looked into it that much

[test branch](https://github.com/Walavouchey/osu-wiki/pull/26)

noticed this in https://github.com/ppy/osu-wiki/pull/7682


cl8n:
#7731 instead

PR #7731, opened 2022-07-31
Checkout with full clone in CI

cl8n:
can't use `fetch --depth` (by itself) after all because one of Walavouchey's checks requires the full commit history to work. ~~in this update, pull in the entire object DB from the PR merge commit, but only get the blobs needed to populate work tree~~ now just normal clone


peppy:
Why not just keep it simple and do a normal clone at this point? We don't really care about performance.

cl8n:
performance seems fine anyway :tada:

peppy (approved):

PR #7744, opened 2022-08-01
Fix typos and remove flag syntax

Blooshing:
For some reason `::{ flag=PH }::` was not correctly working for the news post and thus led to the news post showing only `::{}::`, which is not favorable. Also, it was brought to my attention that the osu!catch score showcase was incorrectly marked as 4xmiss when the stable score was an FC. Retested it and it shows incorrect still based on the playback speed in lazer. I've made an issue, but until that is resolved there is no way for me to resolve the osu!catch score showcase. 

## Self-check

### DIFF
--- a/news/2022-08-01-osumonthly-1.md
+++ b/news/2022-08-01-osumonthly-1.md
@@ -62,7 +62,7 @@ Not content to allow osu! to vacuum up all the glory, the members of the osu!tai
 
 Showcasing some of the best drum bashers the community has to offer, EGTS has a storied history of being one of the most competitively and visually impressive contests around, sporting enough dons and kats to even make the Taikonator blush.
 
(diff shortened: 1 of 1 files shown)


peppy (approved):

news/2022-08-01-osumonthly-1.md
```diff
@@ -62,7 +62,7 @@ Not content to allow osu! to vacuum up all the glory, the members of the osu!tai
 
 Showcasing some of the best drum bashers the community has to offer, EGTS has a storied history of being one of the most competitively and visually impressive contests around, sporting enough dons and kats to even make the Taikonator blush.
 
-The GTS team's insanely talented designer ::{{ flag=PH }}:: [OsuMe65](https://osu.ppy.sh/users/852867), has also made a trailer for the tournament. Check it out and leave your thoughts down below!
```
TPGPL, line 65: isnt the syntax `::{ flag=PH }::` though?

PR #7768, opened 2022-08-04
Change outdated term "iDevice" to "iOS device" in osu!stream article 

rzkalqdr:
After further discussion on the #osu-wiki Discord channel, we figured it would be best to update the term "iDevice" into something more usual and clearer and agreed to use "iOS device(s)" instead.  

## Self-check

### DIFF
--- a/wiki/osu!stream/en.md
+++ b/wiki/osu!stream/en.md
@@ -33,7 +33,7 @@ Development for this version ended in 2020. For more information, read the [blog
   - If requested for Apple ID, key-in the Apple ID and the password.
 - Plug-in the iOS device to the device (that has osu!stream).
 - Transfer the osu!stream files using iTunes.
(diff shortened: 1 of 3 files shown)


wiki/osu!stream/en.md
```diff
@@ -12,13 +12,13 @@ Development for this version ended in 2020. For more information, read the [blog
 
 ## Installation
 
-![](img/Os-Devices.jpg "osu!stream in iDevice")
+![](img/Os-Devices.jpg "osu!stream in iOS devices")
 
-### iDevice's App Store method (Suggested)
+### iOS' App Store method (Suggested)
```
Walavouchey, line 17:
```suggestion
### iOS App Store method (suggested)
```

cl8n (approved):

PR #7785, opened 2022-08-05
[FR] Updated `osu!stream` wording to match the en version

riesha:
## Self-check

Replaced `iDevice` by `appareil iOS` to match the english version, i was thinking of making more changes as some anglicisms like `tels que des hold circles et des connected beats` sound weird to me but i assume they're intentional so i left them in.


TPGPL:
Hi, thanks for your contribution, although a PR applying those changes has already been opened (#7773).

If you want to make any further changes to this translation, you can review the existing PR or make a new one after it gets merged.

PR #7806, opened 2022-08-08
[ID] Update `Modding/Kudosu` and `Community/Mappers' Guild`

ZeroPyrozen:
## Note

Mappers' Guild article doesn't have any change in the Indonesian article since the English article only update the grammar.

## Self-check


Nivalyx (commented):
looks good but i noticed that you used both `postingan mod` and `*mod post*` in the pr

i think both words are fine tho, so it's up to you as to which one to use i suppose (as long as they're consistent ofc)

TPGPL (approved):

PR #7824, opened 2022-08-10
Escape unintentional span syntax

cl8n:


TicClick (approved):

Walavouchey:
it doesn't seem to be an issue.. ?

image

Walavouchey (approved):
nvm

image

cl8n:
some parser will accept `::\S.+::` as a span so it's just to be safe, for the contests line. the rest are actually broken on osu

PR #7821, opened 2022-08-10
[DE] Add `Performance points`

Mysterius79:
Complete but some translation might be wrong

PR #7844, opened 2022-08-14
[DE] Adjust wording rules regarding notes

FamousPig:
## Self-check

Adjustment of existing articles to follow shortly.


The-Last-Cookie:
Now that I'm thinking about it, does this conflict with Anmerkungen in the [footnotes](https://osu.ppy.sh/wiki/de/Article_styling_criteria/Formatting#anmerkungen)? I would like to hear other opinions on this.

FamousPig:
`Note:` -> `Anmerkung:`
`Notes and References` -> `Anmerkungen und Referenzen`

I don't see a problem here.

wiki/Article_styling_criteria/Formatting/de.md
```diff
@@ -271,7 +271,7 @@ Ein Hinweis sollte an eine angemessene Stelle in einem Abschnitt platziert werde
 *Warnung: {Warnung}.*
 ```
 
-- `Tipp` sollte für Fakten oder triviale Details genutzt werden.
+- `Anmerkung` sollte für Fakten oder triviale Details genutzt werden.
```
Pasi4K5, line 274: I'm not sure if it's a good idea to remove `Tipp` entirely. The reason I made the suggestion to  change `Tipp` to `Hinweis` (https://github.com/ppy/osu-wiki/pull/7821#discussion_r945193645) is because I see a `Tipp` as more of a piece of advice rather than a note. But `ppv1, die originale Implementierung für Performance Points, hatte auch ein Änderungsprotokoll, welches in seinem [Forumsbeitrag](https://osu.ppy.sh/community/forums/topics/92185) einsehbar ist.` is not advice. It's what I would call an `Anmerkung`, so I do think that `Anmerkung` fits better here, but there are still notes that are helpful and can be considered `Tipps`, like these ones for example:
https://github.com/ppy/osu-wiki/blob/9402695d2e9f6f5d2f46393bd8056b545f33f0de/wiki/Client/Interface/de.md?plain=1#L258
https://github.com/ppy/osu-wiki/blob/9402695d2e9f6f5d2f46393bd8056b545f33f0de/wiki/Article_styling_criteria/Formatting/de.md?plain=1#L745

So my suggestion would be to change the usage of `Tipp` instead of removing it and use `Anmerkung` for what `Tipp` was supposed to be used previously.

```suggestion
- `Tipp` sollte für hilfreiche Ratschläge genutzt werden.
- `Anmerkung` sollte für Fakten oder triviale Details genutzt werden.
```

FamousPig: Thing is, all of these are denoted as `Note:` in the English original, where they could be denoted as 'Tip:' just as well. With that in mind I am against adding additional denotations in the German translations specifically.

The-Last-Cookie: Both `Tipp` and `Hinweis` are hard to differentiate, as they point the reader to something very specific. I would call a `Anmerkung` a more general `Hinweis` that only gives (slightly essential) background information. So I agree with FamousPig, that we should stick to four notices.

In general, I am fine with this change.

The-Last-Cookie (approved):
> Note: -> Anmerkung:
> Notes and References -> Anmerkungen und Referenzen
>
> I don't see a problem here.

I thought them to be a bit confusing as they are so close to each other (especially articles that just use `Anmerkungen` for this section) but if you are fine with this then I am too.

cl8n (approved):

PR #7866, opened 2022-08-18
Update BN and NAT listings

Nivalyx:
By the request of [Kotoha](http://osu.ppy.sh/users/7823498) himself. 

I left bg.md untouched on purpose as the team listing on that language appears to be heavily outdated, as opposed to how the listing is on the other languages.


cl8n (approved):
jsyk these pages are updated partially automatic and it's usually easier to do it that way. see the [link in the markdown comment above these tables](https://github.com/ppy/osu-wiki/blob/3b0403a25f0325f4c8f29ce843c00789d94f6d4a/meta/group-info/TRANSLATING.md), and also [these instructions to keep staff log in sync](https://github.com/ppy/osu-wiki/blob/3b0403a25f0325f4c8f29ce843c00789d94f6d4a/wiki/People/The_Team/Staff_Log/2022/en.md?plain=1#L386-L398)

PR #7868, opened 2022-08-18
[DE] Add `Beatmap/Difficulty`

The-Last-Cookie:
## Self-check


wiki/Beatmap/Difficulty/de.md
```diff
@@ -0,0 +1,82 @@
...
+In einigen Fällen wie beispielsweise [AIMod](/wiki/Client/Beatmap_editor/AiMod) werden Schwierigkeitsgrade ausgehend von der Sternebewertung in 6 Klassen wie folgt eingeteilt:
+
+- ![](/wiki/shared/diff/easy-o.png?20211215) Easy: 0.0★ bis 1.99★
+- ![](/wiki/shared/diff/normal-o.png?20211215) Normal: 2.0★ bis 2.69★
+- ![](/wiki/shared/diff/hard-o.png?20211215) Hard: 2.7★ bis 3.99★
+- ![](/wiki/shared/diff/insane-o.png?20211215) Insane: 4.0★ bis 5.29★
+- ![](/wiki/shared/diff/expert-o.png?20211215) Expert: 5.3★ bis 6.49★
+- ![](/wiki/shared/diff/expertplus-o.png?20211215) Expert+: 6.5★ und darüber
+
+Beachte, dass die beabsichtigten Schwierigkeitslevel der meisten Beatmaps zwar mit den angegebenen Farben oder Klassen übereinstimmen, einige jedoch aufgrund bestimmter Faktoren erheblich davon abweichen können. Deshalb bieten die [Ranking-Kriterien](/wiki/Ranking_Criteria) entschiedenere Spielregeln, um die tatsächliche Schwierigkeit eines Levels unabhängig der Sternebewertung zu bestimmen.
```
FamousPig, line 82: Here you are using `Level` as a stand-in for `difficulty` when the original is talking about `difficulty level and spread`.
Aside from dropping the idea of `difficulty spread` altogether, using `Level` here would probably be fine, though it technically goes against the [ASC](https://osu.ppy.sh/wiki/de/Article_styling_criteria/Writing#terminologie), but I still want to confirm that is actually what you were going for.
If not, my suggestion would be as follows:

```suggestion
Beachte, dass die beabsichtigten Schwierigkeitslevel der meisten Beatmaps zwar mit den angegebenen Farben oder Klassen übereinstimmen, einige jedoch aufgrund bestimmter Faktoren erheblich davon abweichen können. Deshalb bieten die [Ranking-Kriterien](/wiki/Ranking_Criteria) entschiedenere Spielregeln, um den tatsächlichen Schwierigkeitsgrad, sowie die Verteilung der Schwierigkeitsgrade, einer Beatmap unabhängig von der Sternebewertung zu bestimmen.
```

FamousPig (changes requested):

wiki/Beatmap/Difficulty/de.md
The-Last-Cookie: Although I understood the meaning of `spread` in this case, I couldn't find a good translation I was content with myself. So, thanks for your suggestion.

> it technically goes against the ASC

I mean, the ASC is not hammered in stone and this (only in the German ASC existing and relatively) new section (refering to `translations of common terms`) can be adjusted any time. But I do agree with you that we should stick to `Schwierigkeitsgrad (/-stufe)` altogether and don't make it more complex than it needs to be. Again, as mentioned above, word variety was a reason why I used `Level` here.

wiki/Beatmap/Difficulty/de.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Schwierigkeitsstufen und Sternebewertung
+
+Seit dem 27. Juli 2021 ordnet die osu!-Webseite einer Schwierigkeitsstufe abhängig von ihrer Sternebewertung nach dem folgenden Spektrum automatisch eine Farbe zu:
+
+![Farbspektrum der Schwierigkeitsbewertung von osu!](/wiki/shared/star-rating/spectrum.png)
+
+In einigen Fällen wie beispielsweise [AIMod](/wiki/Client/Beatmap_editor/AiMod) werden Schwierigkeitsgrade ausgehend von der Sternebewertung in 6 Klassen wie folgt eingeteilt:
+
+- ![](/wiki/shared/diff/easy-o.png?20211215) Easy: 0.0★ bis 1.99★
```
The-Last-Cookie, line 75: A minor nitpick I just noticed: star rating information could be written down with commas instead of full stops. osu! web does this too, and using commas in floats is more common for Europeans.

This also occurs in other articles (namely subarticles of `Medals`), so this change is out-of-scope here.

FamousPig (approved):

cl8n (approved):

PR #7873, opened 2022-08-19
Remove wav files from "The Sound Design of osu!(lazer)" folder

0x84f:
This fix brings:
~~Name fix (date change) of the folder where wiki assets are held is also incorrect~~
Removal of .wav files present within wiki asset folder which have been moved onto assets.ppy.sh 

## Self-check

- [X] ~~*(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)~~


cl8n (changes requested):
name of news post file shouldn't change if it's already published. that part is used for the link

0x84f:
> name of news post file shouldn't change if it's already published. that part is used for the link

Do you reckon the other changes should be done then or do we leave it as is, considering its in production?

cl8n:
the other changes are fine

cl8n (approved):

PR #7875, opened 2022-08-20
[ES] Add `Circle_size`

Ekkusu:
## Self-check


Ekkusu:
now i noticed a mistake but i'm not sure.
In the part of tags, should I also translate them or leave them like that? Well, I would suppose yes, although I can't find that regulation.... (not sure)

TPGPL:
English tags shouldn't be removed, but you can add the translations for them as separate tags (it's not mandatory though)

Ekkusu:
Thank you~, I already fixed it, also I changed a mistranslated term

TicClick (approved):

PR #7927, opened 2022-08-30
[NL] Update `Beatmapping` page

TomG16:
Complete revamp of the Dutch/NL beatmapping page to look like the EN page which is way clearer and gives more information/guidance.

Current page is short and is outdated.

trigger ci 
## Self-check


TomG16:
Unsure what's causing it to fail. Looking at #L42 I only see necessary lines which should not be removed.

TicClick:
> Unsure what's causing it to fail. Looking at #L42 I only see necessary lines which should not be removed.

[`e1122f0`](https://github.com/ppy/osu-wiki/pull/7927/commits/e1122f04b52151d408f14b0a848f678a5590fd58) was the issue

also, to clarify, translations need to follow the original articles with respect to content — is that the case here, or not (can't tell only by the "look like" part)?

TomG16:
> > Unsure what's causing it to fail. Looking at #L42 I only see necessary lines which should not be removed.
> 
> [`e1122f0`](https://github.com/ppy/osu-wiki/pull/7927/commits/e1122f04b52151d408f14b0a848f678a5590fd58) was the issue
> 
> also, to clarify, translations need to follow the original articles with respect to content — is that the case here, or not (can't tell only by the "look like" part)?

Yes that is the case here. it follows the original article with of course some very minor grammatical difference to better fit with the Dutch language.

TPGPL (approved):

PR #7958, opened 2022-09-06
Refine ambiguous slider rule in ranking criteria 

apollo-dw:
See https://osu.ppy.sh/community/forums/topics/1519425

## Self-check


wiki/Ranking_Criteria/osu!/en.md
```diff
@@ -187,7 +187,7 @@ If a Normal difficulty is required and used as the *lowest difficulty* of a beat
 #### Rules
 
 - **Every slider must have a clear and visible path of movement to follow from start to end.** Sliders that overlap themselves without straightforward slider borders and sliders whose individual sections are unreadable cannot be used. A slider's end position must be clear under the assumption that a player has a skin which makes slider end circles fully transparent. 
-  - If a slider's follow circle covers the whole slider path, the movement of its slider ball may be ambiguous.
+  - If a slider's follow circle can be active (i.e. the slider can be held) over its duration without any player movement away from the slider head, the movement of its slider ball may be ambiguous.
```
peppy, line 190: I still don't really understand this. Like I both don't understand how this would be visually, nor whether this line is saying that this is an exception, or a reason explaining the line before or something else.

peppy: That reads *so* much better.

peppy (dismissed):

wiki/Ranking_Criteria/osu!/en.md
apollo-dw: @peppy Wording is good with me 👀

peppy:
I guess this needs OUTDATED_CHECK updates if anyone wants to grab that.

Walavouchey (approved):

PR #7983, opened 2022-09-10
[FR] Update `Contribution_guide/GitHub_Desktop`

Ganondorf-1:
Looks like a link was changed in the English version of the article because it didn't work anymore.
The link that was used in the French translation still works, so there's no reason to change it

## Self-check


agatemosu:
https://support.microsoft.com/fr-fr/windows/extensions-de-nom-de-fichier-courantes-dans-windows-da4a4430-8e76-89c5-59f7-1cdbbc75cb01
Could you put this link to match the version in English

TPGPL:
to be honest, i would prefer the translation to use the official Microsoft article instead of some weird-looking website

this way it would also match the English article better

Ganondorf-1:
Good point, and thanks for providing the link

TPGPL (approved):

PR #7988, opened 2022-09-11
Add `Changes to osu!mania Star Rating & Performance Points` news post 

molneya:
I'll need some ideas for the introduction and potentially also the conclusion. An image showing the [accuracy calculation](https://github.com/ppy/osu/blob/aa823161f18b957d65e4bab024d10bce92e6885e/osu.Game.Rulesets.Mania/Difficulty/ManiaPerformanceCalculator.cs#L76) is needed unless there is some other way of showing the formula. I've also checked with @Kominaru that I'm not forgetting any changes. 

## Self-check


news/2022-08-30-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,58 @@
...
+This primarily affects maps with heavy usage of chords (especially in higher key modes), for example:
+
+- [Cardboard Box - The Limit Does Not Exist [[7K] Infinity]](https://osu.ppy.sh/beatmapsets/379758#mania/831204) 7.45\* --> 7.71\*
+- [Manabu Namiki - Tenshi (True Final Boss) [[4K] MIPHATHER]](https://osu.ppy.sh/beatmapsets/1503789#mania/3080978) 5.36\* --> 5.44\*
+
+### Using accuracy instead of score for PP
+
+Score has long since been the measure of skill a player has on the map. But with [osu!(lazer)](https://github.com/ppy/osu) making changes to the scoring system, it no longer makes sense to base performance on this metric.
+
+With [a new change](https://github.com/ppy/osu/pull/18749), pp will now be based on the accuracy the player gets on the map using this formula:
```
0x84f, line 38: consistency 

```suggestion
With a [new change](https://github.com/ppy/osu/pull/18749), pp will now be based on the accuracy the player gets on the map using this formula:
```

0x84f (changes requested):
other than this, is there any specific reason why every sentence has it's own row? seems a bit unnecessary for some things

news/2022-08-30-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,58 @@
...
+
+Compared to the old system, this buffs very high-accuracy scores while nerfing lower-accuracy ones. However, some scores that previously received no pp will find that they now do reward some pp.
+
+### Other changes
+
+As a consequence of accuracy now determining pp, this led to some other changes:
+
+- **No Fail rewards 0.75x of No Mod pp**. Previously, due to No Fail's mod multiplier limiting score to below 500,000 you couldn't gain any pp with it.
+- **"Accuracy pp" has been removed**. It would previously be rewarded to scores that had a score above 960,000. This has been accounted for in the new accuracy pp formula.
+- **Star rating to pp curve formula has been simplified**. Only minor differences here.
```
Walavouchey, line 52: this doesn't say much at all.. what change does this refer to? maybe there's something more clarifying we can put here

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,62 @@
...
+- [Cardboard Box - The Limit Does Not Exist [[7K] Infinity]](https://osu.ppy.sh/beatmapsets/379758#mania/831204): 7.45\* --> 7.71\*
+- [Manabu Namiki - Tenshi (True Final Boss) [[4K] MIPHATHER]](https://osu.ppy.sh/beatmapsets/1503789#mania/3080978): 5.36\* --> 5.44\*
+
+### Using accuracy instead of score for performance points
+
+Overall score has historically been used as the measure of skill a player demonstrates on a given map. With the new direction [osu!(lazer)](https://github.com/ppy/osu) is taking, it no longer makes sense to base performance entirely on this metric.
+
+With a [new change](https://github.com/ppy/osu/pull/18749), awarded pp will now be based on the accuracy the player gets on the map using this formula:
+
+put accuracy formula image here
```
Ephemeralis, line 44: still missing an image for this

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,62 @@
...
+
+At 80% accuracy and below, the player will receive no pp for their score. This scales linearly to 100% accuracy where the player receives the maximum possible pp for their play.
+
+Compared to the old system, this buffs very high-accuracy scores while nerfing lower-accuracy ones. However, some scores that previously received no pp will find that they now do reward some pp.
```
Ephemeralis, line 48: any examples cited for this? seems like something worth mentioning

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Other changes
+
+As a consequence of accuracy now determining pp, this has led to some other changes:
+
+- **No Fail now rewards 0.75x of No Mod pp.** Previously, due to the multiplier for this mod limiting score to below 500,000, it was impossible to gain any pp with it. This is no longer the case.
+- **"Accuracy pp" has been removed.** Scores above 960,000 previously awarded a bonus to overall pp gained in a bonus termed as "accuracy pp". This has been removed but accounted for in the new accuracy pp formula described above.
+- **Star rating to pp curve formula has been simplified.** These changes are only minor.
```
Ephemeralis, line 56: is this curve formula thing important to mention at all? will the end user see any actual change from this? best to provide an example if they would like the other changes

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,62 @@
+---
+layout: post
+title: "Changes to osu!mania Star Rating & Performance Points"
+date: 2022-10-04 12:00:00 +0000
+---
+
+Big changes are arriving in the osu!mania scene as difficulty calculation receives its first set of updates in a long time! Read on for more info!
```
Ephemeralis, line 7: i've reworded this based on the original entry line, but are these changes really *big*? it's basically a set of bug fixes with LN and chord calculations plus a pivot from score into accuracy if I'm reading it right

Ephemeralis (changes requested):

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
MChecaH: The biggest of changes is the Score -> Accuracy change.

Even if they aren't *big* per se, the community sees it differently given how scarce these are to begin with. Should be fine.

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Long note difficulty changes
+
+A [new change](https://github.com/ppy/osu/pull/17913) in long note calculation fixes issues with fast releases giving disproportionately more star rating than they should.
+
+Before this change, all overlapping long notes would receive a fixed bonus. After this change, this bonus now scales proportionally with how long ago the last long note release was.
+
+This change also fixes an oversight with long notes in the rightmost column of a map taking precedence in overlap calculation.
+
+An [interactive demonstration](https://www.desmos.com/calculator/7eboanfzne) by [**Evening**](https://osu.ppy.sh/users/2193881) helps visualize how this change affects the long note bonus if you're interested.
```
MChecaH, line 25: Mentioned in the dev server that for those seeking for extra information, https://eve-ning.github.io/#/blog/004/ could be linked as it goes much more in depth on how the algorithm has changed.

Helpful annex given that it's the __only__ place where this information is compiled.

Small tangent however: this solution isn't fool proof as the release timings were accounted as if they were initial press timers, which I'm unsure whether it be good to mention or just fix moving on.

peppy:
Just a heads up to anyone that is involved with the osu!mania SR/PP changes: the deployment process is waiting on this news post being in good shape.

It has outstanding reviews and missing content.

cc/ @Ephemeralis @molneya and anyone else in the osu!mania community that is able to show interest.

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
Kominaru: It's hard to represent a particular example of this, specially because the display acc is not the accuracy formula being used to compute pp now. If we still want to use a very simplified, it could be sensible to reword it to say something very roughly along the lines of 

> Compared to the old system, this will reward plays with high accuracy and lower score (e.g. 96% acc, 780k score), compared to plays with low accuracy but higher score (e.g. 94% acc, 830k score).

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
Kominaru: The change is minimal in terms of effect (users will not notice this), but I still worry that a change like this should be documented in a somewhat accesible place. I don't think we have any place to achieve this rather than the newspost

Would a wording in the lines of 

> **Star rating to pp curve formula has been simplified.** This change doesn't have noticeable effects in the pp rewarded to plays, and is only intended to make the calculations more efficient.

work better?

Kominaru: That being said, I think that if we do include it, we should describe *how* it was changed (with the formulas themselves) rather than just mentioning the change and not going further

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
Kominaru: Would this be adequate (in terms of formatting)?

Figure_1

I used the matplotlib style template, but I can't seem to get the right font to be used. If someone wants to reproduce it with the proper font (Inter) or make tweaks, I attach a gist to the script to generate these:

https://gist.github.com/Kominaru/931e5e44ec983d8ce5a87a447f13a6d0

Kominaru:
> Just a heads up to anyone that is involved with the osu!mania SR/PP changes: the deployment process is waiting on this news post being in good shape.
> 
> It has outstanding reviews and missing content.
> 
> cc/ @Ephemeralis @molneya and anyone else in the osu!mania community that is able to show interest.

Made some comments on the open review points

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
molneya: I think removing it is best. It's not intended to have any drastic effect on pp calculation and seems to be causing more confusion than its worth. Explaining the difference between 2 very similar formulas won't help much either.

news/2022-08-30-changes-to-osu-mania-sr-and-pp.md
molneya: https://github.com/ppy/osu-wiki/pull/7988#discussion_r989977288

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
Walavouchey: i'm pretty sure this point is covered with the graph kominaru posted. the segmented part has been simplified to a linear formula, which i believe is what this is referring to

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
molneya: added a link to the blog post

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
Kominaru: > i'm pretty sure this point is covered with the graph kominaru posted. the segmented part has been simplified to a linear formula, which i believe is what this is referring to

It's not the same formula: the graph in the post above covers the `acc -> % of maximum pp awarded` curve. This point here corresponds to the `sr -> maximum pp awarded` curve instad

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
molneya: I've added a slightly reworded version of what Kominaru said. Not sure if a map example is exactly relevant here? I'm not too high level of a mania player anymore so I'll need some suggestions on if that's necessary.

MChecaH: The map this is affected by heavily depends on a player's Accuracy so it'd probably be misleading to cite one specific example.

news/2022-10-09-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Long note difficulty changes
+
+A [new change](https://github.com/ppy/osu/pull/17913) in long note calculation fixes issues with fast releases giving disproportionately more star rating than they should.
+
+Before this change, all overlapping long notes would receive a fixed bonus. After this change, this bonus now scales proportionally with how long ago the last long note release was.
+
+This change also fixes an oversight with long notes in the rightmost column of a map taking precedence in overlap calculation.
+
+An [interactive demonstration](https://www.desmos.com/calculator/7eboanfzne) by [**Evening**](https://osu.ppy.sh/users/2193881) helps visualize how this change affects the long note bonus if you're interested. Those seeking yet more information may also be interested in this
```
tybug, line 25: > Those seeking yet more information may also be interested in this

hanging sentence? interested in what?

Ephemeralis: holdover from previous review, fixing

molneya: More than that was changed that wasn't supposed to. Fixed now

news/2022-10-04-changes-to-osu-mania-sr-and-pp.md
molneya: Remade image with correct font

molneya:
I've added the last of the images so this should be good to go now

news/2022-10-09-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Long note difficulty changes
+
+A [new change](https://github.com/ppy/osu/pull/17913) in long note calculation fixes issues with fast releases giving disproportionately more star rating than they should.
+
+Before this change, all overlapping long notes would receive a fixed bonus. After this change, this bonus now scales proportionally with how long ago the last long note release was.
+
+This change also fixes an oversight with long notes in the rightmost column of a map taking precedence in overlap calculation.
+
+An in-depth [blog post](https://eve-ning.github.io/#/blog/004/) and [interactive demonstration](https://www.desmos.com/calculator/7eboanfzne) by [**Evening**](https://osu.ppy.sh/users/2193881) helps visualize how this change affects the long note bonus if you're interested.
```
Walavouchey, line 25:
```suggestion
An in-depth [blog post](https://eve-ning.github.io/#/blog/004/) and [interactive demonstration](https://www.desmos.com/calculator/7eboanfzne) by [**Evening**](https://osu.ppy.sh/users/2193881) may help visualise how this change affects the long note bonus if you're interested.
```

news/2022-10-09-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,61 @@
...
+This primarily affects maps with heavy usage of chords (especially in higher key modes), for example:
+
+- [Cardboard Box - The Limit Does Not Exist [[7K] Infinity]](https://osu.ppy.sh/beatmapsets/379758#mania/831204): 7.45\* --> 7.71\*
+- [Manabu Namiki - Tenshi (True Final Boss) [[4K] MIPHATHER]](https://osu.ppy.sh/beatmapsets/1503789#mania/3080978): 5.36\* --> 5.44\*
+
+### Using accuracy instead of score for performance points
+
+Overall score has historically been used as the measure of skill a player demonstrates on a given map. With the new direction [osu!(lazer)](https://github.com/ppy/osu) is taking, it no longer makes sense to base performance entirely on this metric.
+
+With a [new change](https://github.com/ppy/osu/pull/18749), awarded pp will now be based on the accuracy the player gets on the map using this formula:
```
Walavouchey, line 42:
```suggestion
With a [new change](https://github.com/ppy/osu/pull/18749), the awarded pp will now be based on the accuracy the player gets on the map using this formula:
```

news/2022-10-09-changes-to-osu-mania-sr-and-pp.md
```diff
@@ -0,0 +1,61 @@
...
+
+![](/wiki/shared/news/2022-10-09-changes-to-osu-mania-sr-and-pp/accuracy-equation.png)
+
+At 80% accuracy and below, the player will receive no pp for their score. This scales linearly to 100% accuracy where the player receives the maximum possible pp for their play. Compared to the old system, this rewards high-accuracy plays with low score (e.g. 96% accuracy, 780k score) compared to lower-accuracy plays with high score (e.g. 94% accuracy, 830k score). However, some scores that previously received no pp will find that they now do reward some pp. The differences are visualised below, calculated using OD 8:
```
Walavouchey, line 46:
```suggestion
This formula is almost exactly the same as how the actual accuracy you'd see on the results screen is computed. With this "accuracy" that's used in difficulty calculations, the only change is that MAX judgements are weighted more. See the [wiki article on accuracy](/wiki/Gameplay/Accuracy#osu!mania) if you'd like to compare.

Additionally, 80% accuracy and below, the player will receive no pp for their score. This scales linearly to 100% accuracy where the player receives the maximum possible pp for their play. Compared to the old system, this rewards high-accuracy plays with low score (e.g. 96% accuracy, 780k score) compared to lower-accuracy plays with high score (e.g. 94% accuracy, 830k score). However, some scores that previously received no pp will find that they now do reward some pp. The differences are visualised below, calculated using OD 8:
```

Walavouchey:
otherwise ok 👍

news/2022-10-09-changes-to-osu-mania-sr-and-pp.md
molneya: Not sure about this one. It's similar, but a huge difference is that the MAX judgement actually contributes to accuracy with this formula, so it might be misleading. The lead in to the second paragraph also sounds weird.

Walavouchey: it *is* the exact same formula, except that MAX judgements have a weight of 320 instead of 300. i get your point though so i've edited the suggestion

molneya: That reads much better.

Walavouchey (dismissed):

peppy (approved):

PR #7998, opened 2022-09-14
Update medal count in `Medals`

MegaMixYT:
There are 6 more medals after recent update, so medals count is now 280.

## Self-check


Pisapou:
Could you update this in all other md files so they don’t get outdated just for a number which don’t need to be translated ?

MegaMixYT:
> Could you update this in all other not outdated md files so they don’t get outdated just for a number which don’t need to be translated ?

I can, but I'm not sure how should I do that

Walavouchey (approved):

PR #8022, opened 2022-09-18
[DE] Update `Storyboard/Scripting/osu_File_Toggles`

The-Last-Cookie:
This article was in need of an update. Its structure was slightly different from the English article and it also used the formal form of address. Furthermore, two phrases were missing. So, here is an update for this article.

## Self-check


wiki/Storyboard/Scripting/osu!_File_Toggles/de.md
```diff
@@ -1,11 +1,11 @@
-# Storyboard Scripting - Schalter
+# Storyboard-Einstellungen in .osu-Dateien
```
FamousPig, line 1: Is there a precedent of this title format? I don't know if I agree with translating `scripting` as `Einstellungen`.

FamousPig (approved):

wiki/Storyboard/Scripting/osu!_File_Toggles/de.md
The-Last-Cookie: So, the English title is `Storyboard .osu file toggles` and if I understand it correctly, this article lists entries that can be added to `.osu` or `.osb` files to *configure* stuff related to storyboards.

Originally, this article translated `toggles` with `Schalter`, but I think `Einstellungen` fits better in this context.

As a suggestion, I could agree to use `Einstellungen beim Storyboard-Scripting in .osu-Dateien` if that is what you are going for.

TPGPL (approved):

PR #7925, opened 2022-08-30
Add `Changes to osu! Star Rating & Performance Points` news post

apollo-dw:
Some good maintenance is probably required on this (like adding the banner and making sure links are correct). It's been proof read by the other committee members to check for correctness.

Also I'll come up with a good signoff line shortly, I just didn't want to let it collect dust while I think 😅 

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #8058, opened 2022-09-26
Move performance points/star rating news post banners to one folder

Walavouchey:
prereq for https://github.com/ppy/osu-wiki/pull/7925, https://github.com/ppy/osu-wiki/pull/7915, and https://github.com/ppy/osu-wiki/pull/7988


Walavouchey:
alternatively, make a new banner because the posts also relate to pp and not just star rating (which this banner doesn't convey that well i think). up to someone else to decide

peppy:
Would be cool if we had one banner per ruleset too.

Walavouchey:
@arflyte

peppy (approved):

Issue #7799, opened 2022-08-07
Brand identity guidelines should mention how to correctly write "osu!"


LittleEndu:
How to correctly write "osu!" already exists at [https://osu.ppy.sh/wiki/en/Article_styling_criteria/Writing#osu!](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Writing#osu!) so maybe just move that to https://osu.ppy.sh/wiki/en/Brand_identity_guidelines and link to *Brand identity guidelines* from *Article styling criteria > Writing*

Issue #7439, opened 2022-06-04
Brand Identity Guidelines Design Kit shows incorrect "Yellow (-1)" color hex code


hongaaronc:
Wasn't quite sure where to file this issue. I figured the wiki is the closest thing, but the actual issue is with a PDF asset hosted on Drive, with a link to it in the wiki.

This is referring to the asset linked [here](https://github.com/ppy/osu-wiki/blob/master/wiki/Brand_identity_guidelines/en.md?plain=1#L5).

In Page 9 of the document, the color code for "Yellow (-1)" is incorrect. It looks like an inadvertant mistake from copy+pasting, since it's the same hex code as "Light Blue (-1)".

![BrandIdentityError](https://user-images.githubusercontent.com/5133530/171980551-01be9a1b-0228-488a-aa36-b44ebaf5270f.png)

TicClick:
@arflyte please take a look (the doc belongs to you: https://github.com/ppy/osu-web/issues/4996#issuecomment-534363999)

Coppertine:
It was referred whe  it was originally made with the response being that the swatch itself is not used any more
https://github.com/ppy/osu-web/issues/4996#issuecomment-554184420

peppy:
The whole guidelines needs an update, since most has moved to a new colour scheme and figma resources.

arflyte:
Yes, we no longer use it. Here are the new palettes. https://www.figma.com/file/VIkXMYNPMtQem2RJg9k2iQ/Asset%2FColours?node-id=0%3A1

TicClick:
it seems we can't just insert a link to Figma due to licensing https://github.com/ppy/osu-wiki/pull/8061#discussion_r985267217. wonder if manually exporting assets featured in the design kit from Figma and packing them in a new archive would work

PR #8061, opened 2022-09-27
Update `Brand identity guidelines` with osu! spelling rules

TicClick:
closes https://github.com/ppy/osu-wiki/issues/7799
## Self-check


wiki/Brand_identity_guidelines/en.md
```diff
@@ -2,39 +2,69 @@
 
 The brand identity guidelines is a set of standards for branding osu!-related projects. All use does not require approval.
 
-[Download the design kit here.](https://drive.google.com/file/d/1TmUot5nu49p71icz4u3G68njLAQOeQrG/view?usp=sharing) The linked design kit contains some outdated assets. Also note that the included PDF is outdated and this page should be used instead.
+## Assets
+
+*Note: Some parts of the design are still work in progress, unless stated otherwise.*\
+*See also: [osu-resources — Assets used by osu!](https://github.com/ppy/osu-resources)*
+
+The [osu! Figma library](https://www.figma.com/file/6m10GiGEncVFWmgOoSyakH/osu!-Figma-Library) contains design assets such as the osu! logo, mod icons, colour palettes, and many more. To obtain them, log into Figma, mark objects for download using rectangular selection, and click the `Export` tab in the top right corner.
```
Walavouchey, line 10:
```suggestion
The [osu! Figma library](https://www.figma.com/file/6m10GiGEncVFWmgOoSyakH/osu!-Figma-Library) contains design assets such as the osu! logo, mod icons, colour palettes, and many more. To obtain them, log into Figma, mark objects for download using rectangular selection, and click the `Export` tab in the top-right corner.
```

wiki/Brand_identity_guidelines/en.md
```diff
@@ -2,39 +2,69 @@
...
+
+The osu! website design is licensed under AGPL version 3 or later. Please see [the licence file](https://github.com/ppy/osu-web/blob/master/LICENCE) for more information.
```
Walavouchey, line 12: the figma contains both lazer and web designs, but i don't think the figma stuff has any licence specified from what i can tell

wiki/Brand_identity_guidelines/en.md
```diff
@@ -2,39 +2,69 @@
...
+:::
 
 This is the full colour version of the osu! cookie logo. The subtle drop shadow is part of the full colour logo.
 
 Please use the logo as provided by the design kit without any modification.
 
 ### Clear space area
 
-![](img/clear-space-area.png)
+::: Infobox
```
Walavouchey, line 57: the paragraph below is so short that the infobox will look kinda wonky

wiki/Brand_identity_guidelines/en.md
TicClick: from what I was able to find, most, if not all lazer assets are used on the web and are therefore under AGPL, too. the remaining part, hosted in ppy/osu-resources, is *mostly* [CC-BY-NC 4.0](https://github.com/ppy/osu-resources/blob/master/LICENCE.md), save for fonts

wiki/Brand_identity_guidelines/en.md
TicClick: I suspect that but hope it will not, otherwise I need to remove the two infoboxes above as well to not make this specific image look out of the place

wiki/Brand_identity_guidelines/en.md
peppy: No licence means copyright by default. Which is where we will leave things for figma. They are designs in progress and not to be considered CC or otherwise.

TicClick: this raises a few questions:
- what are the usage rules for designs of the osu! cookie, mod and modE icons, and other things featured previously in the design kit?
- same, but for WIP? 
- can Figma be used as a design source at all in this case?

my understanding is that for (1) and (2) it's "use with attribution, we don't care for as long as you mention the copyright and don't make profit off it", but that is not "fair use" either

peppy: > * what are the usage rules for designs of the osu! cookie, mod and modE icons, and other things featured previously in the design kit and considered ready for use?

Everything in the asset kit is free for all use. They are assets provided to be used in designs or external applications. Still need to understand trademarks, which are a separate thing altogether.

> * same question, but for WIP assets — can they be used for personal things and small projects clearly not affiliated with osu!?

No

> * basically, can Figma be used as a design source at all in this case, and if yes, what should I add as a word of caution?

Figma is public only so the designs can be implemented in osu! projects. Any other use should be avoided.

TicClick: alright, what I am trying to solve is the design kit being mostly out of date (https://github.com/ppy/osu-wiki/issues/7439). if linking Figma for color/asset stuff isn't the way to go, then what is? perhaps a curated archive of some things exported from Figma, then?

peppy: Yeah, it should be updated. But i'm not sure that will ever happen.

TicClick: alright I'm rolling this chunk back as it is more sophisticated than it looked like from the start

wiki/Brand_identity_guidelines/en.md
TicClick: removed

TicClick:
I have removed the Figma link addition and reduced the PR to the osu! spelling rules

peppy (approved):

Walavouchey (approved):

Issue #5546, opened 2021-05-22
`hitcircleoverlay`'s documentation regarding `sliderstartcircleoverlay`/`sliderendcircleoverlay` usage is ambiguous


frenzibyte:
![CleanShot 2021-05-23 at 00 08 37](https://user-images.githubusercontent.com/22781491/119240823-0843a100-bb5b-11eb-8960-b037e84fd9bc.png)

On testing, `hitcircleoverlay` is never used for sliders unless the circle's texture itself (`sliderstartcircle`/`sliderendcircle`) isn't skinned, in which case `hitcircle` and `hitcircleoverlay` will be used for that circle instead.

frenzibyte:
This have had hit me once again and I feel like the explanation above isn't quite understandable so I'm gonna reword it here:

The current documentation mentions "the `hitcircleoverlay` is also used for `sliderstartcircleoverlay`/`sliderendcircleoverlay` if not skinned", but from testing through this with osu!(stable) after stumbling around this:
 - if a `sliderstartcircle`/`sliderendcircle` exists but no `sliderstartcircleoverlay`/`sliderendcircleoverlay`, then no overlay will be shown, rather than using `hitcircleoverlay` which is what the documentation seems to indicate.
 - if no `sliderstartcircle`/`sliderendcircle` exists, then `hitcircle` will be used in place of them and `hitcircleoverlay` will be used as the overlay in that case.

peppy:
Probably worth submitting a PR yourself given that no one else has taken this on.

PR #8077, opened 2022-09-28
Update `Skinning/osu!` with `hitcircleoverlay.png` visibility rules

TicClick:
resolves https://github.com/ppy/osu-wiki/issues/5546. one recorded issue less, many more hidden ones to go.

## Self-check


cl8n (approved):

frenzibyte (commented):
Reads much better than before. :+1:

PR #8043, opened 2022-09-23
[DE] Add `Gameplay/Game modifier/Flashlight`

The-Last-Cookie:
## Notes

There are three distinct terms that sound pretty similar. My proposal is to translate them as follows:
- `score multiplier` - `Score-Multiplikator` (changes when activating/deactivating difficulty changing mods)
- `combo (score multiplier)` - `Punktemultiplikator` (Combo value itself)
- `combo multiplier effect` - `Multiplikatoreffekt` (see #8029)

This will need changes in the following articles (if approved, that is):
- [x] Adjust `Client/Interface` and `Gameplay/Game_modifier`
    - #8079 

### Other notes

- not sure on `hitting area` --> `Trefferbereich`

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #8079, opened 2022-09-29
[DE] Adjust translation for score multiplier

The-Last-Cookie:
Like proposed in #8043, this adjusts the translations in the wiki accordingly.


TicClick (approved):
update the branch if/when it's approved

FamousPig (approved):

PR #8100, opened 2022-10-02
Fix preview text not showing in osu!monthly September news post

Walavouchey:
image

two of the videos also have issues, not sure what's up with that yet

image

image


Walavouchey:
superseded by https://github.com/ppy/osu-wiki/pull/8102

PR #8117, opened 2022-10-03
[ES] Update `People/peppy`

agatemosu:
## Self-check


TicClick (approved):
looks small enough to skip the native review part

PR #8134, opened 2022-10-05
Fix flags in contest page

Hiviexd:
wish that we could preview new flags in vscode just like the old ones :sob:


Walavouchey:
if you're adding links to the headings, why not all of them?

TPGPL (approved):

PR #8150, opened 2022-10-06
[DE] Add game modifiers `10K`, `Auto` and `Autopilot`

The-Last-Cookie:
After the related terminology was adjusted and approved in #8043 and #8079, now it is time to translate the missing game modifier articles. This PR is part 1 of the coming PR series.

## Translation notes

- ~~`hold bar` is meant to be `hold note`, thus this should be translated with `Hold-Note`~~
    - [x] requires #8159, awaits negotiation on the correct terminology

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #8159, opened 2022-10-08
Fix terminology in `Gameplay/Game modifier/Auto`

The-Last-Cookie:
Walavouchey [brought to my attention](https://github.com/ppy/osu-wiki/pull/8150#issuecomment-1270490644) that there are no *hold bars*, only *hold notes*.

- [x] outdate translations

## Self-check


Walavouchey:
blocked because apparently [the meaning of this line isn't so clear](https://discord.com/channels/188630481301012481/218677502141399041/1028269653048430672). this needs further testing for what cases can cause auto to miss

TicClick:
I tried doing some testing with 1/16th-snapped sliders on adjacent rows, but wasn't able to find anything combo-breaking. asked Quenlla to find an example of such a map, but if anyone else gets it first, please do make a post

TicClick:
I asked Maxus:

> [8:05 PM]Maxus: From what i know:
> 
> 1. If regular note overlap with another regular note.
> 2. If note overlap at the beginning of LN press.
> 3. If note overlap at the LN body.
> 4. If note overlap at the LN tails.
> 
> That should be the most usual case.
> [8:06 PM] TicClick: is all of that while the objects are on the same column?
> [8:07 PM] Maxus: yeah, they must be on same column, and must be on exact same timepoint.
> [8:09 PM] Maxus: hold bar should have same definition as hold notes actually. It's more of old word that already got replaced and not used anymore nowadays.
> [8:10 PM] TicClick: so the edit is correct, then?
> [8:10 PM] TicClick: because I thought for quite a while that that means the judgement line itself
> [8:10 PM] TicClick: (another person suggested that it might be the column)
> [8:11 PM] TicClick: I guess I can also change `but may miss densely-stacked notes` to `but may miss objects appearing simultaneously on the same column`
> [8:12 PM] TicClick: because just placing a row of non-overlapping 1/16th notes won't break anything
> [8:12 PM] Maxus: I think it's better to simplify by saying same column and same timestamp.

Walavouchey (dismissed):

TicClick (approved):

TicClick:
for clarity: https://discord.com/channels/188630481301012481/218677502141399041/1052322628632846387

> [9:34 PM] Protastic101: Ok, so auto will potentially miss if there are notes within +-1 ms or each other, but will not miss if they have at least 2ms between them
> [9:40 PM] Protastic101: correction: Auto will miss if there are notes within 1ms of each other in the same column
> [9:45 PM] Protastic101: which is basically the same thing as having two objects placed in the same column at the same time
> [10:20 PM] Antalf: auto will fail if there is a good amount of p-notes (and all of their variations) present in a map, this only applies for mania and there is one specific p-note that will completely fail the map regardless if it was played 99% accurately even if just one note was present in the map (called a fake note)
> [10:21 PM] Antalf: forgot to a explain: a p-note is when the ending timestamp of a long note is set earlier than the start of said long note
> [10:34 PM] Protastic101: ^ To make such a note, you'd need to manually edit the .osu, and there's a whole host of other potential side effects with doing so, so most players will likely not encounter such an issue

PR #8165, opened 2022-10-08
Update peppy's Twitter avatar in `Twitter`

tybug:
twitter provided this image in 400x400, but I've downscaled to match the existing 100x100 images, which seemed like an intentional size choice. Let me know if 400x400 would be preferred.

## Self-check


Walavouchey:
i don't think the resolution affects the appearance in the table, so i believe it's ok to put 400x400 in there. the file size is small anyway

tybug:
I've updated to the 400x400 version

Walavouchey (approved):

PR #8216, opened 2022-10-16
Cache `node_modules` in Github Actions to win 40–50s

TicClick:
this speeds the check up by ~40%

- without cache (`install remark`): https://user-images.githubusercontent.com/4686101/196014127-4f1e770f-bb5a-4744-9afc-57a2d5860636.png
- with cache (`install remark`): https://user-images.githubusercontent.com/4686101/196014090-f5709e1e-5804-4ddf-8667-53afa21051e4.png

once this is merged, we need to run the action from `master` once in a while to populate the cache, [or it won't be hit](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -35,7 +35,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: node_modules
-          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
+          key: node-modules-${{ hashFiles('package.json', 'package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
 
       - name: install remark
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -29,7 +29,15 @@ jobs:
           cache: npm
           node-version: 16
 
+      - name: load node_modules from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package-lock.json') }}
+
       - name: install remark
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
 
       - name: run remark on changed files
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -24,6 +24,7 @@ jobs:
         run: bash scripts/ci/inspect_binary_file_sizes.sh ${{ github.sha }} ${{ github.sha }}
 
       - name: set up Node.js
+        id: setup-node
         uses: actions/setup-node@v3
         with:
           cache: npm
@@ -34,7 +35,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: node_modules
-          key: node-modules-${{ hashFiles('package-lock.json') }}
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
 
       - name: install remark
         if: steps.cache-node-modules.outputs.cache-hit != 'true'


TicClick:
a small note on how this works:
- `load node_modules from cache` step tests cache for `key` and extracts cached directories listed in `with.path` on success
- there's an extra step `Post load node_modules from cache` added automatically at the end of the workflow (if multiple cache actions are used, these steps are added in reverse order), which creates the cache entry or overwrites it
- the workflow has access to the default branch's cache (`master` for us) and its base branch's cache if it has one.

different feature braches have different caches, so to make use of this, we need to run the workflow with caching step from `master` at least once. if we don't run it periodically, then hope that [at least once a week](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy) someone makes a new PR or updates an existing one:

> GitHub **will remove any cache entries that have not been accessed in over 7 days**. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB.
>
> If you exceed the limit, GitHub will save the new cache but will begin evicting caches until the total size is less than the repository limit.

.github/workflows/continuous-integration.yml
```diff
@@ -24,12 +24,21 @@ jobs:
...
 
+      - name: load node_modules from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
+
       - name: install remark
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
```
peppy, line 41: Is it not quite safe to leave this in place? Is there much of a saving from skipping this command run?

TicClick: it's basically the gist of the whole PR. not skipping it results in `npm ci` installing it all from scratch (see timings from https://github.com/TicClick/osu-wiki/pull/11)

by the way, why is it not safe? reading cache is scoped to the feature branch and its base for read, and writing to the branch itself, so you can't affect someone else's PR by doing changes in yours (unless I misunderstand your question)

peppy: for instance, in the case the cache is in a bad state (ie. a "bpm download failure" gets cached somehow) there is no post step to verify and recover

TicClick: `npm ci` is the only thing we use that writes to `node_modules`. if it fails, the whole workflow fails earlier than cache creation takes place (it's one of the last steps).

I can look into verifying the cache on hit but I think we're still good

peppy: Can try and see how it plays out. I just know how much of a PITA cache can be if you cache wrongly.

TicClick: done some digging:
- apparently `npm c(lean)i(nstall)` does its own integrity checks
  - removing `node_modules` is also the first thing it does, even with `--dry-run`
- running `npm install --no-audit --dry-run` validates nothing, and `npm ci --no-audit` provides only a marginal gain
- you can't quickly verify integrity of all installed packages
- at the same time, GitLab CI themselves mention `node_modules` as one of folders potentially subject to caching in their workflows: https://docs.gitlab.com/ee/ci/caching/#cache-nodejs-dependencies

I am convinced that if for whatever reason the cache will end in a bad state, linter will also fail and we'll take note of that. as an extra precaution against it being created already malformed, I can add `hashFiles(package.json)`

peppy (approved):

.github/workflows/continuous-integration.yml
cl8n: I used `npm ci` instead of `npm i` originally since the idea was to cache `~/.npm`, which is npm's internal cache for package downloads, then rebuild node_modules every run. `npm i` can be used if you *want* to keep previous node_modules

the difference between caching .npm and node_modules for a one-project environment like this comes down to whether it's faster to unpack from a smaller cache (.npm) or use a larger cache (node_modules). haven't tested that myself, I just used the github built-in option (.npm)

.github/workflows/continuous-integration.yml
```diff
@@ -24,12 +24,21 @@ jobs:
...
         with:
           cache: npm
           node-version: 16
 
+      - name: load node_modules from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package.json', 'package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
```
cl8n, line 38: don't see how package.json being part of this key changes anything

.github/workflows/continuous-integration.yml
```diff
@@ -24,12 +24,21 @@ jobs:
         run: bash scripts/ci/inspect_binary_file_sizes.sh ${{ github.sha }} ${{ github.sha }}
 
       - name: set up Node.js
+        id: setup-node
         uses: actions/setup-node@v3
         with:
           cache: npm
```
cl8n, line 30: if node_modules cache works fine then this can be removed
(wasn't working anyway but I'm still not sure why...)

.github/workflows/continuous-integration.yml
```diff
@@ -24,12 +24,21 @@ jobs:
...
+      - name: load node_modules from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: node_modules
+          key: node-modules-${{ hashFiles('package.json', 'package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
+
       - name: install remark
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
```
cl8n, line 42: should test `npm i` with no condition, if you didn't already. does feel kinda weird to skip over all npm checks and have later steps fail with some arcane error on bad cache

cl8n (commented):
late review :tada:

.github/workflows/continuous-integration.yml
TicClick: impressively, it changes nothing, will remove

Issue #4282, opened 2020-10-02
Skinning Criteria for sliderset somewhat vague


BProK:
https://osu.ppy.sh/help/wiki/Ranking_Criteria/Skin_Set_List

Spinner set has both basic and new spinner style implemented into one, which actually messes with BN applications,
Seperating the lists as "Basic spinner set" and "New spinner set", then adding any files that are used by both (e.g. spinner-approachcircle.png) to both lists would help clarify this. There is no good reason to have it be vague, as it (right now) acts as two sets under one chart?

Even the header says "If you include elements labeled as "basic spinner style", you must not include elements labeled as the "new spinner style"."
It's pretty clear that these are intended to be separate sets.

PR #8221, opened 2022-10-16
Separate osu spinner set table

TheMoMan:
Resolves https://github.com/ppy/osu-wiki/issues/4282


wiki/Ranking_Criteria/Skin_Set_List/en.md
```diff
@@ -95,28 +97,48 @@ Please note that the recommended sizes for all elements in this set is handled v
 
 ### Spinner set
 
-All elements excluding spinner-osu.png and spinner-rpm.png are gameplay relevant. If you include elements labelled as "basic spinner style" in the notes, you **must not** include elements labelled as the "new spinner style".
+osu! has two spinner styles - the old style, and the new style. Skins can only use one spinner style. Elements that are exclusive to a specific style **cannot** be mixed with elements of the other style.
+
+This is because the old spinner style will take priority over the new one if both styles are present. The new spinner style will only work if the user-specific skin does not use the old spinner style and supports skin versions 2 or higher, or if the default skin is forced in the beatmap.
```
TicClick, line 102: https://user-images.githubusercontent.com/4686101/196163509-c6d8af58-4b5c-4da4-87e7-3b06f5a13ddc.png not that my suggestion is ideal, but it's a way forward
```suggestion
osu! has two spinner styles, only one of which can be used in a skin at the same time. Using exclusive elements from different styles is **not allowed**: if both styles are present, the old one will take priority over the new one.

The new spinner style will only work if the user-specific skin does not use the old spinner style and supports skin versions 2 or higher, or if the default skin is forced in the beatmap.
```

TicClick (changes requested):
(RockRoller said that the spinner style sets are correct)

wiki/Ranking_Criteria/Skin_Set_List/en.md
TheMoMan: Do agree current wording is kinda uh, awkward. Still trying to think of a good way to do it.

Walavouchey: maybe like this

```suggestion
osu! has two spinner styles, only one of which can be used in a skin at the same time. Using exclusive elements from both of them is **not allowed**: if elements from both are present, the old style will take priority.

The new spinner style will only work if the user-specific skin does not use the old one and supports skin versions 2 or higher, or if the default skin is forced in the beatmap.
```

TicClick (approved):

PR #8223, opened 2022-10-17
Add new PP/SR news post banners

Ephemeralis:
Designs courtesy of kaetwo. These should be used going forward, instead of the old (mode-agnostic) star rating image.


Walavouchey (changes requested):
very against applying these retroactively. the posts have already been long published

Ephemeralis:
The catch one is going to use them, so it seems strange not to use the others.

peppy:
Should only be for new posts.

Walavouchey:
if it feels awkward to use them only for the catch post so suddenly, then they could be put on hold until the next wave of changes

Ephemeralis:
May as well only use them for the next batch, then.

Walavouchey:
i suggest renaming to `pp-sr-{osu,taiko,catch,mania}.jpg`

- "banner" is redundant because of folder name
- "ctb" is an old term and "tko" uhhh

to my ears pp/sr sounds better than sr/pp and i believe the news post titles follow that order already

Ephemeralis:
Swapped 'ctb' for 'fruits' and followed the naming schema above.

peppy:
Is there a reason for "use non-minified assets"? The filesizes seem a touch exorbitant.

Ephemeralis:
jpeg-recompress smallfry was introducing massive banding and artifacting on these banners, looks like jpegmini into imageoptim seems to be much better.

peppy (approved):

PR #8232, opened 2022-10-19
[FR] Update link to `Little Drummer Girl` for `osu! originals`

Coppertine:
## Self-check

(As this is a non visual change, reviewing from a native french speaker is not required.)


Coppertine:
Using them as seperate PR's to prevent cross committing across languages.

Walavouchey:
?

updating the page along with translations in one pr is preferred, especially when the change isn't a wording change and doesn't require review by translators. having multiple PRs is just unnecessary in this case, and would be even more so with 10 different languages

PR #8239, opened 2022-10-20
[ID] Clean up `BanchoBot`

Nivalyx:
Also added a flag reference to Echo ( http://osu.ppy.sh/users/431 )'s profile in the article across all applicable translations.


TicClick:
> the flag couldn't be added on pages that are still based on the *very* old version of the article such as it.md

the check skips outdated articles automatically. if they don't match the original anymore and aren't marked with `outdated_translation: true`, this is worth fixing instead of silencing the tool

Nivalyx:
> the check skips outdated articles automatically. if they don't match the original anymore and aren't marked with `outdated_translation: true`, this is worth fixing instead of silencing the tool

ah, okay

wowcake (dismissed):
good to go 👍

TicClick (dismissed):

TicClick (approved):

PR #8244, opened 2022-10-21
Run all checks in parallel for early feedback

TicClick:
see this PR for an example. did this mainly because I got irritated a few times trying to fix a check and failing subsequent ones. now, status area shows where specific issues lie.

- full repository checkout takes 30-60s, which is the only bottleneck now that `node_modules` are cached (#8216).
- checkouts are running simultaneously in n=5 containers but don't seem to overly compete for network.

I also tried caching a repository but it was NOT worth it:
- packing ~1.3 Gb into a cache archive was taking a while (maybe around the same time that was spent on checkout)
- in dependent steps, extracting the cache was done at a higher speed (~70-150 Mb/s)
- additionally, extracted repository was lacking important information such as remote refs, the merge commit itself (on which our checks are run), and I have eventually realized that I was trying to replicate `actions/checkout` but in a poor way
- moreover, finding a good cache key is also a challenge

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -74,7 +74,7 @@ jobs:
         with:
           cache: pip
           cache-dependency-path: scripts/requirements.txt
-          python-version: 3.x
+          python-version: '3.10'
 
       - name: Install Python packages
         run: pip install -r scripts/requirements.txt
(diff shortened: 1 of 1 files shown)


TicClick:
surprisingly, it seems to perform much worse than it did during my tests in https://github.com/TicClick/osu-wiki/pull/9. not sure we should sacrifice speed **that** much -- perhaps I'll find another way of delivering feedback

PR #8263, opened 2022-10-28
Make custom hitnormals obligatory in the osu!mania RC

Kominaru:
The goal of this PR is to add a simple clarification to the osu!mania exception of the "mandatory hitsounds" rule in General RC. 

While the current wording does not obligate mappers to add one or more custom hitnormal samples overriding the default hitnormal, it has for years beenan unwritten rule and standard *de facto* to override the default hitnormal in all osu!mania ranked sets (in fact, everyone had been interpreting the current ruling as having this meaning). This is done in order to improve the audible feedback of the map for those users who prefer to enable hitsounds. 

Generally this is done to use less "punchy" hitnormals instead of the default ones, which usually become obnoxious to the player due to the high object density and concurrency in osu!mania beatmaps. Hitnormals used by mappers as overrides of course still provide good audible feedback (this is mandated by RC), but they are less overwhelming and provide a better experience for those players who choose to play with hitsounds.

Thus, current RC reads: 

> **osu!mania beatmaps do not require hitsound additions.** This is to allow for easier approachability to osu!mania mappers of different upbringings. It is still highly recommended to use hitsound additions to improve the feel of your beatmaps.

and after discussing among the osu!mania BN and NAT, we propose the following clarification, also simplifying the rule a bit:

> **osu!mania beatmaps do not require hitsound additions, but a custom hitnormal sample must override the default hitnormal.** It is still recommended to use hitsound additions to improve the audible feedback of your beatmaps.

Of course, any proposal for an alternative or more clear wording is welcomed

Cheers

### DIFF
--- a/wiki/Ranking_Criteria/en.md
+++ b/wiki/Ranking_Criteria/en.md
@@ -200,7 +200,7 @@ This category contains explicit allowance statements of concepts and rules that 
   - **...not be encoded upwards from a lower bitrate.**
 - **A beatmap may only contain one song file used by all difficulties.** Multiple song files within a single beatmap are unsupported and result in unexpected behaviour with preview times, metadata, etc.
 - **Beatmaps must be [hitsounded](/wiki/Beatmapping/Hitsound).** Hitnormals give feedback to the player, and additions (whistles, claps, and finishes) accent the most important parts of the music.[^normal-vs-addition]
-  - **osu!mania beatmaps do not require hitsound additions.**[^normal-vs-addition] This is to allow for easier approachability to osu!mania mappers of different upbringings. It is still highly recommended to use hitsound additions to improve the feel of your beatmaps.
+  - **osu!mania beatmaps do not require hitsound additions, but a custom hitnormal sample must override the default hitnormal.**[^normal-vs-addition] It is still recommended to use hitsound additions to improve the audible feedback of your beatmaps.
 - **All clicked parts of objects must have at least one hitsound which both...**
   - **...has a clear impact, whose peak is delayed no more than 5 milliseconds.** `normal-hitfinish.wav` from the default skin is exempt from this.
   - **...uses the `.wav` or `.ogg` file format.** `.mp3` should not be used here as it is inherently delayed.


cl8n:
well for one this is a rule change and not a "clarification", the line being edited had nothing to do with hitnormals and it doesn't imply anything about them

> in fact, everyone had been interpreting the current ruling as having this meaning

seems wrong for same reason. what part of these rules makes it sound like you need a custom hitnormal

wording seems ok but I think these would be better as two separate rules, since the custom hitnormal requirement is not related to using or not using additions

Kominaru:
> well for one this is a rule change and not a "clarification", the line being edited had nothing to do with hitnormals and it doesn't imply anything about them

I understand it's a bit hard to contextualize, but in osu!mania the "hitsounds are mandatory" rule has always been read as meaning "you must hitsound your maps, using appropiate custom samples" by all the community (this includes mappers, modders, and BNs). The second part of this interpretation has always been followed without being questioned since it's agreed as being adequate. We recently learnt from another mode's NAT that the "custom sample" part isn't intended to be mandatory with the current wording of the general rule. So in the end, for osu!mania this is indeed a "clarification" or "correction" of the rule to reflect the years-long community's interpretation.  If needed, we can pull some data on what percentage of the ranked sets use a custom hitnormal since the current wording is in place. 

 
> wording seems ok but I think these would be better as two separate rules, since the custom hitnormal requirement is not related to using or not using additions

I agree with this, we just tried to fit the clarification in the current wording, but it becomes a bit convoluted. Perhaps it's more adequate to use the same method as on the o!m spread rules, and use the general RC to write

> - **osu!mania beatmaps are bound by a different set of hitsound requirements, defined in the [osu!mania ranking criteria]**

And instead detail these two points more cleanly in the mode-specific RC, such as: 

> - **osu!mania beatmaps do not require hitsound additions**, but using them is recommended to improve the audible feedback of your beatmaps.
> - **A custom hitnormal sample must at all times override the default hitnormal**. Using the default hitnormal can become obnoxious to players due to the game-mode's high object density.

Walavouchey:
> So in the end, for osu!mania this is indeed a "clarification" or "correction" of the rule to reflect the years-long community's interpretation. If needed, we can pull some data on what percentage of the ranked sets use a custom hitnormal since the current wording is in place.

this is a *rule change* precisely because the new wording mandates a thing that wasn't mandated before. the community simply misinterpreted the rule to specify more than it actually does.

yes, it's a technicality, but it means that this pr has to go through the same process as any other rc change (i.e. can't be merged immediately like wording fixes). context is irrelevant in that regard.

Kominaru:
Going through the whole RC Proposal process for something which is a standard de facto (and has been for years, due to everyone misreading a rule) feels like really unneeded bureaucracy that delays something simple. That being said, I understand that this discussion is going to be more effective and less frustrating if we just go through the mere formality of making a RC proposal in forums and coming back here afterwards. 

Thanks for your feedback
Cheers

peppy:
Seems super weird. I think we should be improving the default rather than forcing all beatmaps to not use it.

Walavouchey:
custom beatmap skin samples can be turned off in the settings, so this wouldn't really force people to not use the defaults (and their skin would also need to use the default samples if they wanted them)

peppy:
I meant the beatmappers, not the players. Unless I'm misunderstanding the change.

Walavouchey:
looking at the proposal thread, it seems that the main problem is that the normal-hitnormal is too loud and obnoxious with mania's high object density, and doesn't fit most songs people use. common sense among mappers/bns is to always override the normal samples so people who play with hitsounds and map sounds enabled (considered a very niche playstyle in mania, because many disable hitsounds completely) don't have to be "bombarded" with normal-hitnormals in case their skin doesn't provide one.

> rather than forcing all beatmaps to not use it.

for a summary of other suggestions/concerns with this proposal see [this post](https://osu.ppy.sh/community/forums/topics/1667260?n=23), one of them being making this a guideline instead, which should address your concern. let's see how the discussion pans out.

 > I think we should be improving the default

since stable is feature-locked and most people use stable, this is assumed not to be an option.

peppy:
> since stable is feature-locked and most people use stable, this is assumed not to be an option.

improvements could of course be made in lazer? dunno what you're getting at here.

Walavouchey:
yes, but that's not relevant for the ranking criteria atm. (also i'm only summarising what other people have said in the forum thread)

Kominaru:
The wording as of now seems to be have muted into the following (forum poll would need a rerun in a new proposal thread because it has deviated too much from the original proposal): 

In general RC, fold into osu!mania RC to match existing mania-specific exceptions of General RC: 

> - Beatmaps must be hitsounded (...)
>   - **osu!mania beatmaps are bound by a different set of hitsound requirements defined in the [osu!mania ranking criteria](https://osu.ppy.sh/wiki/en/Ranking_Criteria/osu%21mania)**

and the osu!mania RC would include a two new rules and a new guideline:

> - **osu!mania beatmaps do not require hitsound additions**, but using them is recommended to improve the audible feedback of your beatmaps.
> - If the normal-sampleset is chosen and no additions are present in the map, the default normal hit-normal must be overriden by a custom sample that is not the default normal-hitnormal.
> - Frequent use of the default normal-hitnormal sample is discouraged. Using this sample too frequently can become obnoxious to players due to the game mode's high object density and concurrent objects.

This seems to be the majoritary opinion as of now

Kominaru:
Bear in mind I'm acting solely as a messenger here, I personally don't agree with the current majoritary wording, it is really cluttery. I think something way simpler like a single guideline like:

> **Frequent use of the default normal-hitnormal sample is discouraged.** Using this sample too frequently can become obnoxious to players due to the game mode's high object density and concurrent objects.

works perfectly without all the proposed additions vs no additions separation.

Protastic101:
As of https://osu.ppy.sh/community/forums/topics/1667260?n=40, I think this PR can be closed as we can instead work to make the lazer samples not as intrusive and have that backported into stable.

Walavouchey:
feel free to reopen or make a new pr if discussion heads that way

PR #2743, opened 2019-10-29
Update difficulty naming schemes from other games

jakads:
- Added "Outer Oni" to osu!taiko difficulty names
- Added "IIDX" after beatmania to prevent confusion
- Added "Leggendaria" difficulty to beatmania IIDX
- Fixed "Pop 'n Music" to "Pop'n Music"
- Fixed "DJMax" to "DJMAX"
- Added various difficulty names differing by SOUND VOLTEX version
- Added Arcaea
- Added Lanota
- Added SEGA's rhythm games (maimai, CHUNITHM, Ongeki)
- Added Rayark's rhythm games (Cytus, Deemo, VOEZ)

As rhythm games are getting more variable, so should the wiki be.

PR #8268, opened 2022-10-29
Update `Ranking criteria/Difficulty naming` with Expert+ difficulty name of the CHUNITHM series

YumeMuzi:
## Self-check

Reference: https://chunithm.sega.jp/music/ultima/
The chunithm official website recorded all of the ultima difficulties. 

On the other hand, WORLD'S END diffname is in the same situation as the Utage/enkai (宴) (https://maimai.sega.jp/maimai_finale/song/enkai/) in maimai, which were not meet official standards.


RushFTK (dismissed):
Really need localization?

TicClick:
> Really need localization?

don't think so -- difficulty names are not localized

on another note, @YumeMuzi do you know where the `WORLD'S END` suggestion originally comes from? can it be used as an alternative to `ULTIMA` perhaps? cc @Jakads who added that in #2743

YumeMuzi:
> > Really need localization?
> 
> don't think so -- difficulty names are not localized
> 
> on another note, @YumeMuzi do you know where the `WORLD'S END` suggestion originally comes from? can it be used as an alternative to `ULTIMA` perhaps? cc @Jakads who added that in #2743

I found SEGA also recorded `WORLD'S END` list (https://chunithm.sega.jp/music/we/) which means it's an officially recognized difficulty, therefore it could be an alternative to `ULTIMA`.

but imo `ULTIMA` will be slightly lower than `WORLD'S END` cuz `ULTIMA` has its difficulty index number (mostly below 15), `WORLD'S END` is using the number of stars (just like the old version of osu, mostly below 5 stars), which beyond of difficulty index number system.

TicClick (approved):

PR #8276, opened 2022-10-31
[ID] Update/Cleanup `osu! terms of service`

Nivalyx:
An update to the Indonesian translation of the `osu! terms of service` article, which is actually slightly outdated compared to the current en.md page but is not tagged w/ the `outdated` marker for some reason (。。*)

Also did some cleanup/proofreading as well while I'm on it.


wowcake (approved):
everything is good to go as far i've checked 👍

TicClick (approved):

PR #8284, opened 2022-11-02
Use correct profile flags and links in some articles

TicClick:
found automatically, verified manually. only obvious mistakes are fixed (typos, copypasta). a link or two are removed for cases where it's clear it is wrong, but the user is restricted and correct link can't be obtained

I decided to keep a lot of other mismatches, as people may have had different flags in the past (and I have no way of knowing that)

- `SKIP_OUTDATED_CHECK` since Chinese translation of the ranking criteria council article doesn't have flags


peppy:
> I decided to keep a lot of other mismatches, as people may have had different flags in the past (and I have no way of knowing that)

Can you explain which cases you updated and which you didn't? It's a bit unclear as it stands.

TicClick:
what was updated:
- cases when a flag is visually similar (say, BG vs RU) and I can tell that instantly, or cases where a user from ex-USSR (UK, BY) changed the flag to that of their homeland instead of Russian one
- clear typos (IR -> IE, UK *(that is Ukraine, correct)* -> GB)
- clear copypaste (two users with identical profile links but different names -- obviously one of the links should point somewhere else)
- users I am familiar with (say, Damnae is from France -- why would he have a German flag all of a sudden?)

what is kept:
- everything else where flags are different and I don't know if they were changed or not (example: yf_bmp -- CN vs HK)
- users who I know have changed flags (example: Lanturn -- UK vs CA)

Walavouchey:
> wrong profile links. if a user is restricted, but the link leads to someone else's profile instead, I removed the link (think there was only one or two such issues)

could also be a typo in the name?

TicClick:
> > wrong profile links. if a user is restricted, but the link leads to someone else's profile instead, I removed the link (think there was only one or two such issues)
> 
> could also be a typo in the name?

true for one case -- see my commit. for the other tournament, I verified via the thread that the user who applied for participation was restricted

peppy (approved):
quick check looks okay

PR #8298, opened 2022-11-05
[ID] Update `Modding and mapping timeline`

Nivalyx:
Also did some cleanup to the existing Indonesian translation as well (so that it could be made more compact) as usual.


TicClick (approved):

PR #8308, opened 2022-11-08
Add Afrikaans language to `group-info`

hikikuman:
## Self-check

Need to add Afrikaans for Zuid-Afrikaan before updating bn/staff listing.


cl8n:
do this at the same time / same PR as updating staff log so that you get feedback if it's working (program will either tell you or it'll be obvious that it's not translated, don't remember which)

PR #8310, opened 2022-11-09
Fix incorrect description in `Game modifier/Double Time`

tematikkp:
Doesn't affect the CS.

## Self-check


TicClick:
thanks for catching this. I made the change to the translations as well, copying existing wording for Korean, and fixed the `Half Time` article in the same way.

TicClick (approved):

PR #8311, opened 2022-11-09
Same change as en.md (diff sama seperti en.md)

tematikkp:
## Self-check


Walavouchey:
move these changes to https://github.com/ppy/osu-wiki/pull/8310. the other languages need to be updated as well.

PR #8317, opened 2022-11-10
Fix Dusk-'s flag in `Nomination_Assessment_Team`

Hiviexd:
the UK instead of GB blunder strikes again


Walavouchey:
superseded by https://github.com/ppy/osu-wiki/pull/8318

PR #8326, opened 2022-11-11
Fix grammar and styling inconsistency in `Game_modifier/Mirror`

The-Last-Cookie:
## Self-check


agatemosu:
The translations are not misspelled

agatemosu:
Instead of marking them as outdated, you could copy the translation from another mod.

fr: À propos
id: Tentang
ru: Краткая информация

TicClick (approved):

PR #8370, opened 2022-11-21
RC metadata change: Change the requirement of romanised title and romanised artist.

Protastic101:
RC proposal thread found [here](https://osu.ppy.sh/community/forums/topics/1670642?n=1). Currently, if a beatmap A is ranked or loved, then beatmap B of the same song should follow the first ranked/loved beatmap's title and artist fields. There's no exception to this unless beatmap A's metadata breaks other rules or is completely wrong. This change makes it so that if an artist has given a translation or romanisation, future beatmaps are able to respect this even if prior beatmaps used a different translation or romanisation.

## Self-check

- [ ] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [ ] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #8380, opened 2022-11-23
Revert RC metadata change #8370

Protastic101:
Due a miscommunication, the metadata RC proposal [here](https://osu.ppy.sh/community/forums/topics/1670642?n=14) had not been updated to properly address the (relatively few) community concerns. More discussion has been taking place as a result of this change, so the thread has been re-opened to allow people to properly voice concerns and modifications to the proposal. In the meantime, because there is still discussion, I am requesting the file be reverted to its original pre-metadata change. Apologies for the inconvenience.

## Self-check


peppy:
but why

Walavouchey:
i messaged the person who made the [latest post](https://osu.ppy.sh/community/forums/topics/1670642?n=14), waiting for confirmation on whether people think this was an oversight

Walavouchey:
there were unaddressed concerns in the thread, and it was mistaken as having reached a conclusion by thread op and pr op

forcing official translations/romanisations (what this change did) leaves no room for edge cases where self-romanisation may be preferred, e.g. with given name–surname order, or otherwise conforming to established metadata standards defined in the rc

further discussion is needed for what priority there should be between "official romanisation vs official translation vs correct romanisation"

Walavouchey (approved):

PR #8399, opened 2022-11-28
Add `Results - Mappers' Choice Awards 2021` news post

Hiviexd:
scheduled for today (28th)


Ephemeralis:
Using inspect element swaps as a rudimentary holdover, Wala's concerns seem to be well founded - the text *is* quite small.



Given this is the bulk of the article, it's probably not going to be a good experience for mobile readers at all. This being said, the background/avatar segments are the largest and I figure most people will identify the maps based on this, it might work. Whether it'll work well or not is another question.

Walavouchey:
if the images are to be redesigned, i suggest making them a bit taller and utilising the horisontal space better

if not, then always using one row of text could work as well

Hiviexd:
thoughts on this redesign?
image

Walavouchey:
zooming in to fit the images to the width of my phone screen, it's a *lot* better and everything is legible. here is a comparison with all widths normalised, including project loved banners since that's one of the current uses of a bunch of images in news posts

comparison

i do have some points of consideration but they may be non-issues. i'd be fine with this as-is too.

- with the pink overlay taking up half of the image, there may not be enough horizontal room for the map bg to be recognisable, although it may still be ok with manual cropping for each bg
  - many song titles will bleed over anyway, so moving the overlay a bit to the left could resolve this
  - lowering the image height slightly would give a higher aspect ratio and thus more horizontal space for the bg
- with a bit more than triple the previous height (some to account for breathing room in the design), the post will definitely be taller than any other news post, but maybe that's fine
- i suspect you're intending to have one image for each section (with no way of linking each beatmap), since your demo wouldn't work otherwise because stacked images have sizeable spacing around them

  image

- i recommend keeping the rounded corners

Hiviexd:
kaetwo's response

> counter points: 
-the map bg has sufficient space to be recognizable, it only takes up a third of the image, weve done worse in other corsace tournaments. also i dont want text to overlay the image more than it needs to, itll look too busy and probably more difficult to read especially with his previous point of also being more mobile focused
-not much point to lowering the image height when youre dealing with 272 images
-rounded corners, sure and also isn't it possible to mosaic linked images with no gaps like a forum post?
-because corsace isn't an officially affiliated with osu, we try not to include "osu" in anything
we are better off not using images if it becomes too much of an issue

Walavouchey:
> isn't it possible to mosaic linked images with no gaps like a forum post?

confirmed this with clayton, it will display correctly with no gaps but we'll need to add backslashes after each image (will do this in a sec). so all is fine

> because corsace isn't an officially affiliated with osu, we try not to include "osu" in anything

even if it appears on a news post? like wiki articles, the content that appears in news posts is held to a certain standard and quality - hence the review process - which includes using official terms where they're preferred.

> we are better off not using images if it becomes too much of an issue

i encourage you do! it's a very cool idea and i wish people continue to push boundaries like this

> all other points

yeah ok

Hiviexd:
updated with new designs, also ran them thru an optimizer

Walavouchey (approved):

Ephemeralis (approved):

PR #8385, opened 2022-11-25
Update `People/The Team/Beatmap Nominators` and `Staff log/2022`

aceticke:
Add new Taiko BNs and move Lumenite- to full
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #8409, opened 2022-11-29
[ES] Update ` People/The_Team/Beatmap_Nominators`

agatemosu:
## Self-check


TicClick (commented):
please apply https://github.com/ppy/osu-wiki/pull/8385 as well

TicClick:
note: while `Holandés` seems to be a more widespread option than `Neerlandés` (and two Spanish-speaking GMTs agreed on that), I'd like to point out that it's [an official language name](https://es.wikipedia.org/wiki/Idioma_neerland%C3%A9s) which IMO shouldn't be overriden unless totally necessary. besides, it will be overwritten every time someone runs https://github.com/cl8n/osu-wiki-bin/ which updates group translations automatically, based on a standardized set of languages and other locale-specific information

cl8n:
might not be relevant here but if you want to override the language name in my program, you can still do that. see `fr`'s settings as example

TicClick (approved):

PR #8412, opened 2022-11-29
Allow clipped audio in `Song_Content_Rules`

qia-1:
https://osu.ppy.sh/community/forums/topics/1671335 

Some songs use clipping as a sound design technique. This change allows audio files of songs with intentional clipping, as long as they aren't too loud (in which case the file's volume can be lowered). Also note that the rule against clipping has been inconsistently applied so far, with some maps (e.g https://osu.ppy.sh/beatmapsets/559928) being allowed to be ranked despite heavy use of clipping, but others being removed.

## Self-check


peppy:
I don't know, it could be argued that clipping implies "extremely and unbearably loud" but alas.

To confirm, is this for songs which were *produced* with clipping? As in it's not going to allow users to take a song and amplify it beyond limits, right?

TicClick:
@qia-1 I find this change redundant. the clause you are arguing for is already in [the ranking criteria, audio guidelines section](https://osu.ppy.sh/wiki/en/Ranking_Criteria#guidelines.3), while song content rules disallow tracks like https://www.youtube.com/watch?v=MUdLrQLbjQ0

> The audio file and hitsound files of a beatmap should not feature any audible and unwarranted sound distortions, like clipping, muffling, or crackling **that is clearly not intended by the artist or part of the song's identity**

qia-1:
> To confirm, is this for songs which were _produced_ with clipping? As in it's not going to allow users to take a song and amplify it beyond limits, right?

Yes. If someone amplifies a song like that it would still be considered unbearably loud. 

> I find this change redundant.

It's been enforced again legitimate songs, like 1103 mentions in the thread.

Maybe a different wording which specifies that only unintended clipping is disallowed would alleviate you two's concerns.

qia-1:
Tried to update the PR, it should now change the file to include
`- Extremely and unbearably loud
- Contains audible clipping unintended by the artist`

wiki/Rules/Song_Content_Rules/en.md
```diff
@@ -4,7 +4,8 @@ With a wide variety of music available, most tracks will be fine for use in osu!
 
 - Heavily political
 - Promoting racial tension or division
-- Extremely and unbearably loud or excessively clipped
+- Extremely and unbearably loud
+- Contains audible clipping unintended by the artist
```
peppy, line 8:
```suggestion
- Contains audible clipping or distortion unintended by the artist
```

peppy (approved):

PR #8429, opened 2022-12-05
Add new RGTS 2022 original track in `osu! originals`

Walavouchey:
## Self-check

- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review) @Pisapou @Ganondorf-1 (dunno if mappool section should be translated)


Pisapou:
except for `Wildcard` that I would probably translate into `Joker` , everything seems ok. I still prefer to wait for Ganon's approval to confirm you to change Wildcard translation

Ganondorf-1:
I don't think mappool brackets/sections should be translated at all; we have never translated "tiebreaker" or anything like that because doing so would be confusing (and pointless since even French tournaments use the English terms)

Pisapou (approved):
So everything is fine coucerning french translation

TicClick (approved):

PR #8455, opened 2022-12-11
[DE] Add `Legal/Privacy`

The-Last-Cookie:
## Self-check


wiki/Help_centre/Account/de.md
```diff
@@ -218,7 +218,7 @@ Das bedeutet aber nicht, dass du einen neuen Account erstellen kannst! Mehrere A
 
 ### Kann ich meinen Account löschen? {#account-deletion}
 
-*Für mehr Informationen über die Löschung von Accounts, siehe: [osu!-Datenschutzrichtlinie § Deine Rechte und Kontrolle](/wiki/Legal/Privacy#your-rights-and-control)<!--TODO: change link as soon as German article is available-->.*\
+*Für mehr Informationen über die Löschung von Accounts, siehe: [osu!-Datenschutzerklärung § Ihre Rechte und Kontrolle](/wiki/Legal/Privacy#ihre-rechte-und-kontrolle).*\
```
The-Last-Cookie, line 221: I'm not fond of having both formal and informal in here, but let's go with it for now. It's not set in stone, so it could be changed back if the majority of the German community is bothered by it.

Walavouchey: it's the title of an article and section, why would it be wrong

The-Last-Cookie: Don't mind it too much, merging that is fine. If I were to link an external article here which uses the formal form, it would practically be the same. It's not as bad as I made it out to be.

TicClick (approved):

PR #8462, opened 2022-12-13
Update `Community/Projects`

minusQuantumNeko:
Updated Osekai links and description

## Self-check

### DIFF
--- a/wiki/Community/Projects/fr.md
+++ b/wiki/Community/Projects/fr.md
@@ -1,5 +1,6 @@
 ---
-outdated: true
+outdated_since: f31678789896bc4d36e44abb8d89925fb095dfb8
+outdated_translation: true
 ---
 
 # Projets
--- a/wiki/Community/Projects/fr.md
+++ b/wiki/Community/Projects/fr.md
@@ -1,4 +1,5 @@
 ---
+outdated: true
 outdated_since: f31678789896bc4d36e44abb8d89925fb095dfb8
 outdated_translation: true
 ---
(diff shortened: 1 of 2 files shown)


wiki/Community/Projects/fr.md
```diff
@@ -1,5 +1,6 @@
 ---
-outdated: true
```
TicClick, line 2: this alone should be enough for the outdated article checker to keep quiet. will take a look at what's going on under the hood

wiki/Community/Projects/fr.md
```diff
@@ -1,5 +1,6 @@
...
+outdated_since: f31678789896bc4d36e44abb8d89925fb095dfb8
+outdated_translation: true
```
TicClick, line 3: you are changing the meaning from "we DON'T know since when the translation is outdated" to "we DO know since when the translation is outdated", ignoring many other potential changes that need to be backported

I get that you added this in attempt to comply with the CI check step, but it is still wrong

TicClick (commented):
the wording change is alright, the front matter one isn't -- will check what's going on

wiki/Community/Projects/fr.md
minusQuantumNeko: > if the translation was already outdated before, then we just don't put in `outdated_since`
> 
> (https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#outdated-articles)

so should I just remove `outdated_since` and that's it?

Walavouchey: no, add back `outdated: true`

Walavouchey (approved):

PR #8465, opened 2022-12-13
Fix wrong date in `Beatmap Spotlights Winter 2023`

agatemosu:
## Self-check

You can't travel to the past like that.


TicClick (approved):
unfortunately so

PR #8471, opened 2022-12-14
Add description of osu!(lazer)'s file storage system

bdach:
- [x] Depends on https://github.com/ppy/osu/pull/21628

The intention for this article is to link it up with https://github.com/ppy/osu-wiki/pull/8467 so that it can be read from within the game client.

Content has been mostly pulled from [here](https://github.com/ppy/osu/wiki/User-file-storage) and [here](https://gist.github.com/bdach/faf66e74f4b4307d867cbe6e3b44ef60), and lightly retouched to become a somewhat-coherent article and to use a neutral register.

Curious if the hard link description makes sense to readers. There may be too much content there, please advise if you think there's too much info there and some needs to be snipped.

## Self-check


wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,52 @@
...
+
+At the time of writing, the correct procedure to manually modify skins or beatmaps outside of the facilities available in-game is to export the item, update it, and then reimport it again with the desired changes.
+
+## Migration from osu!(stable)
+
+### Via hard links
+
+On most Windows systems, osu!(lazer) will be able to import data from the stable version of the game without having to create a second copy of the data on disk. This is possible thanks to an operating system feature called *hard links*.
+
+A hard link is conceptually similar to a *shortcut* (also known as a *symbolic link*), in that it is a method that allows a user to have the same file available from multiple different places on their filesystem. However, while shortcuts are just plain files that point to a different file (and therefore require additional space), hard links work one level deeper, at the level of the filesystem itself.
```
peppy, line 37: shortcuts and symbolic links on windows are a different beast, but might be fine to gloss over this for the user's sake here.

bdach: I'll just snip the mention for simplicity.

peppy:
I'd probably want to get some user feedback on this. The coverage of hard links is detailed, but maybe too much for the average user.

I think the important points to convey are:
- No extra disk space will be used
- But if you check via windows explorer it will look like each install is using space
- Making external changes to files might break things (optional, as discussed maybe we should just gloss over this until it becomes an issue to keep it simple..)

bdach:
I've attempted to simplify the explanation of hard links and cut down on the redundancy of the content. See if it feels any better now.

wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,47 @@
+# File storage in osu!(lazer)
+
+By default, osu!(lazer) stores beatmaps, skins, and score replay files in the following directories:
+
+- `%appdata%/osu/files` (Windows),
+- `~/.local/share/osu/files` (Linux & macOS),
+- `Android/data/sh.ppy.osulazer/files/files` (Android).
+
+On desktop platforms, the file store can also be moved in its entirety to another location, using the `Change folder location...` button in client settings.
```
Walavouchey, line 9:
```suggestion
On desktop platforms, the file store can also be moved in its entirety to another location, using the `Change folder location...` button in the client settings.
```

wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,47 @@
...
+
+osu!(stable)'s storage structure, where data was directly accessible to users as standard files, resulted in a lot of teething issues due to having to explicitly handle unwanted modifications of key files. To avoid this, osu!(lazer) employs a more stringent method of file storage. As such, there is no equivalent of the `Songs` and `Skins` folders. All files that are imported to lazer are stored under filenames that reflect their [SHA-256 hashes](https://en.wikipedia.org/wiki/SHA-2). Mappings to these files are held inside a client database.
+
+For example, a file with the SHA-256 hash of
+
+```
+1a47929b6056d34d25a95eeb2012395ceed66af6f40cc37c898a08482d6325d2
+```
+
+shall be stored under the path of
```
Walavouchey, line 21:
```suggestion
would be stored under the path of
```

wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,47 @@
...
+
+```
+files/1/1a/1a47929b6056d34d25a95eeb2012395ceed66af6f40cc37c898a08482d6325d2
+```
+
+This allows for saving storage space by preventing duplicates of the same file from being persisted to disk, and prevents users (or other applications) from easily tampering with files that should not be tampered with. All inconveniences compared to stable that arise from this choice have been addressed by either allowing exports or adding new in-game management features, or will be addressed in due time by new features.
+
+At the time of writing, the correct procedure to manually modify skins or beatmaps outside of the facilities available in-game is to export the item, update it, and then reimport it again with the desired changes.
```
Walavouchey, line 29:
```suggestion
For now, the correct procedure to manually modify skins or beatmaps outside of the facilities available in-game is to export the item, update it, and then reimport it again with the desired changes.
```

wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,47 @@
...
+
+A hard link is conceptually similar to a *shortcut* in that it is a method that allows a user to have the same file available from multiple different places on their filesystem. However, while shortcuts are just plain files that point to a different file (and therefore require additional space), hard links work one level deeper, at the level of the filesystem itself.
+
+In simple terms, two hard-linked files are just two different names for the same piece of physical space on the hard drive. The same data is available through either of those names. This means that once the migration completes using hard links, both the `Songs` folder in stable, and the `files` folder in lazer will contain files which will point to shared underlying data on the disk. Deleting files from one installation will not affect the other. Updating or editing files in one installation will also not affect the other if osu! is used to make changes, rather than external applications.
```
Walavouchey, line 39: i think this will get the point across a bit more concisely

```suggestion
In simple terms, two hard-linked files are just two different names for the same piece of physical space on the hard drive. This means that once the migration completes using hard links, both the `Songs` folder in stable, and the `files` folder in lazer will contain files which will point to shared underlying data on the disk. Deleting files from one installation will not affect the other. Updating or editing files in one installation will also not affect the other, as long as osu! is used to make the changes.
```

you say "Updating or editing files in one installation will also not affect the other" but that kind of implies that changes to a beatmap in stable will not be reflected in lazer or vice versa

could reword to

```suggestion
In simple terms, two hard-linked files are just two different names for the same piece of physical space on the hard drive. This means that once the migration completes using hard links, both the `Songs` folder in stable, and the `files` folder in lazer will contain files which will point to shared underlying data on the disk. Deleting files from one installation will not affect the other. Updating or editing files in one installation will seamlessly be reflected on the other, as long as osu! is used to make the changes.
```

wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,47 @@
...
+
+Note that if disk usage is checked using the `Properties` window in Explorer or any other similar method, it will appear that both the `Songs` directory and the `files` directory are consuming disk space, potentially misleading users to think that the data is stored twice. This is an artifact of file size accounting, and the actual underlying data is **not** duplicated. This can be verified by comparing the total amount of free space on the entire disk before and after the migration.
+
+As hard links are a filesystem-level feature, it is required that the hard-linked files reside on a common drive, and that the drive is formatted using NTFS.
+
+### Via file copy
+
+On other operating systems and filesystems where hard links are not available (or the functionality to create them has not been implemented yet), the migration from stable to lazer works by copying all files across from the old installation to the new. Both installs are therefore fully separated, but also consume twice the disk space.
```
Walavouchey, line 47: idk maybe remind the reader of the selling point of the new system

```suggestion
On other operating systems and filesystems where hard links are not available (or the functionality to create them has not been implemented yet), the migration from stable to lazer works by copying all files across from the old installation to the new. Both installs are therefore fully separated, but also consume (less than, due to osu!(lazer)'s storage optimisations) twice the disk space.
```

Walavouchey (changes requested):

wiki/Client/Release_stream/Lazer/File_storage/en.md
bdach: not sure about changing to a subjunctive `would`? like, this *is* going to happen, that's what the game *does*, that's a fact and there's no need to conditionalize this i don't think. but i don't care enough to argue

wiki/Client/Release_stream/Lazer/File_storage/en.md
bdach: > you say "Updating or editing files in one installation will also not affect the other" but that kind of implies that changes to a beatmap in stable will not be reflected in lazer or vice versa

it doesn't, so i'm not sure what point is being made here - the second suggestion seems to just be wrong

wiki/Client/Release_stream/Lazer/File_storage/en.md
bdach: i think this is just being overly pedantic and muddying the waters if i'm honest

wiki/Client/Release_stream/Lazer/File_storage/en.md
Walavouchey: if the second suggestion is just factually wrong then that's that, seems like i had a misunderstanding. just consider the first one.

but there seems to be a bit of detail glossed over then, because if the data the files point to is the same, you'd intuitively expect changes on one end be reflected on the other, no? i.e. a beatmap imported from stable using hard links would make lazer's files... point to the same data as the beatmap files on stable. idk but if editing one really doesn't affect the other (as if they're using separate data) then that is just contradictory and leaves move questions

bdach: any change made using either osu! client is deleting one of the hard links and creating a file anew. which is why the other client will not be affected. this is intentional because if you think about how lazer's file store works (each file is stored under its own SHA) then having a change come in from stable *without* breaking the hard link will corrupt lazer's file storage structure.

i'm not sure how to explain this without creating complications in understanding. i'm not sure too many users who aren't technical will care?

wiki/Client/Release_stream/Lazer/File_storage/en.md
Walavouchey: my argument is that *if* such a file exists, then it *would* be stored at so-and-so path. we're talking about a hypothetical file and not a real one

what you wrote makes sense too, but the problem with "shall" is that it's a bit too archaic for the wiki. you could use "will" instead.

bdach: meh it's fine

wiki/Client/Release_stream/Lazer/File_storage/en.md
Walavouchey: yes, being able to read well and concisely is important for the wiki, but so is correctness and clarity. i'm not for oversimplifying to the point of being imprecise, or potentially causing misunderstanding or conflicting information.

would this work for you?

```suggestion
On other operating systems and filesystems where hard links are not available (or the functionality to create them has not been implemented yet), the migration from stable to lazer works by copying all files across from the old installation to the new. Both installs are therefore fully separated, but also consume up to twice the disk space.
```

bdach: i could be fine with that last suggestion, sure

wiki/Client/Release_stream/Lazer/File_storage/en.md
Walavouchey: thank you, that makes sense. i don't think conveying this should be too hard

```suggestion
In simple terms, two hard-linked files are just two different names for the same piece of physical space on the hard drive. This means that once the migration completes using hard links, both the `Songs` folder in stable, and the `files` folder in lazer will contain files which will point to shared underlying data on the disk. Deleting files from one installation will not affect the other. Updating or editing files causes the files to become separate, and will therefore also not affect the other installation, as long as osu! is used to make the changes.
```

maybe i could've come to this perhaps simple conclusion myself if i thought about it harder, but i still think it would be better to spell it out

Walavouchey (dismissed):

wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,47 @@
+# File storage in osu!(lazer)
+
+By default, osu!(lazer) stores beatmaps, skins, and score replay files in the following directories:
+
+- `%appdata%/osu/files` (Windows),
+- `~/.local/share/osu/files` (Linux & macOS),
```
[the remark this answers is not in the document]
bdach: i forgor :skull:

Should be better now i think? I'll update the pinned discussion osu!-side if yes because that still has the old one.

bdach:
I've taken the liberty of adding back-links from the help centre upgrade guide to this new article in [7bd3de4](https://github.com/ppy/osu-wiki/pull/8471/commits/7bd3de4fcd071596869de8f612d9e3f35632be92).

peppy (approved):

wiki/Client/Release_stream/Lazer/File_storage/en.md
```diff
@@ -0,0 +1,48 @@
+# File storage in osu!(lazer)
+
+By default, osu!(lazer) stores beatmaps, skins, and score replay files in the following directories:
+
+- `%appdata%/osu/files` (Windows),
```
Walavouchey, line 5: just noticed that this is incorrect, it's `%appdata%/roaming/osu/files` iirc

bdach: `%appdata%` already leads to `C:\Users\$USERNAME\AppData\Roaming`. As opposed to `%localappdata%` which leads to `C:\Users\$USERNAME\AppData\Local`. See https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables (`CSIDL_APPDATA` vs `CSIDL_LOCAL_APPDATA`).

PR #8494, opened 2022-12-20
[ES] Update `group-info`

agatemosu:
## Self-check


meta/group-info/es.yaml
```diff
@@ -1,34 +1,29 @@
...
   roles:
-    academy: fundador de osu!academy
+    academy: fundador de la osu!academy
     admin: administración
     chat: moderador del chat
     dev: osu!dev
     media: creador de osu!media
     support: equipo de soporte
-    tournaments: gestor de torneos
+    tournaments: gestión de torneos
```
juankristal, line 10: Because this section is about roles I wouldn't change `gestor` to `gestión` as the title of the role is correct as is. For the areas the update is correct.

agatemosu: aaaaa

agatemosu: would it be the same in admin?

agatemosu: would it be the same in admin?

juankristal: Yes. Nice catch.

juankristal (approved):

PR #8497, opened 2022-12-21
Link Datawalke's osu! profile in "The Team"

cl8n:
here's a page where all the names & usernames are linked together https://github.com/Datawalke


Walavouchey (approved):
seems to check out, [post history and all](https://osu.ppy.sh/community/forums/topics/54)

PR #8514, opened 2022-12-25
[ZH] Add `SEV_Rating`

YumeMuzi:
## Self-check


TicClick (approved):

PR #8525, opened 2022-12-27
Update BN/GMT/NAT listings

Blooshing:
## Self-check


Hiviexd:
would it be possible to include Kotoha's namechage in this PR?

https://osu.ppy.sh/users/7823498

Walavouchey:
i guess this was [already logged](https://github.com/ppy/osu-wiki/blob/master/wiki/People/The_Team/Staff_Log/2022/en.md?plain=1#L597) but not updated on this page?

Blooshing:
Guess I'll just turn this into a staff update

Walavouchey:
this can get merged since it's already in the staff log. i was waiting for if you wanted to include name changes here

Walavouchey:
in general it's better to do full staff updates, otherwise the pages will get weirdly out of sync

wiki/People/The_Team/Beatmap_Nominators/zh.md
```diff
@@ -100,7 +100,7 @@ BN 分为两类：正式 (Full) 成员和见习 (prob) 成员。
 | ::{ flag=US }:: [Yogurtt](https://osu.ppy.sh/users/2649717) |  |
 | ::{ flag=CN }:: [Yugu](https://osu.ppy.sh/users/3161834) | 汉语，日语 |
 | ::{ flag=IT }:: [Zekk](https://osu.ppy.sh/users/9704802) | 意大利语 |
-| ::{ flag=SE }:: [Zer0-](https://osu.ppy.sh/users/4260033) | 瑞典语 |
+| ::{ flag=SE }:: [Kudosu](https://osu.ppy.sh/users/11038155) | 瑞典语 |
```
juankristal, line 103: This applies to all languages but shouldnt this be the FR flag? Does he know Swedish or should be French?

Pretty sure Zer0- is still a BN as well, so he should not be removed.

juankristal: Also sort them alphabetically.

juankristal (changes requested):
Some flags are wrong and a member was removed despite still being in the team

wiki/People/The_Team/Beatmap_Nominators/zh.md
Blooshing: 🤷 tbh the only reason why it was changed is because of ticclick's tool saying that the name was changed. 

I shoulda double checked but for now I won't use the tool.

TicClick: the issue was with wrong identifiers someone copied and pasted, not with the names

TicClick:
Firika's removal is correct, since they formally had a dual role (NAT + GMT)

TicClick (approved):

PR #8528, opened 2022-12-28
Add AESS1 & 2 to the list of tournaments

AetherianFI:
Added Aetherian's Summer Singles 1 and 2 to the list of tournaments.

## Self-check


TicClick (approved):
funny that there's no article for AESS run no. 1

PR #8558, opened 2023-01-02
Fix o!m RC General Hard Guideline Only Appearing in 4k/7k Specific Guideline

Protastic101:
All translations except JP and RU were updated by just figuring out which two lines were the exact same and cutting then pasting the 4K line into the general guidelines and removing the excess line in the 7k specific.

## Self-check

`SKIP_WIKILINK_CHECK`


juankristal (commented):
Pretty sure for jp.md:

```5つ以上のノーツで構成されるアンカーは避けてください。 プレイヤーは読みやすいですが、アンカーが長いと大変です。```

Would be the translation that has to be moved up. Once that is sorted I'll merge this as it otherwise seems fine

The Russian translation seems to be outdated anyways and does not contain this guideline to begin with.

juankristal (approved):

PR #8575, opened 2023-01-06
Fix outdated translation incorrectly being in tag

nanaya:
Why does it keep happening


Walavouchey:
you'll be delighted that work is being done to fix this https://github.com/Walavouchey/osu-wiki-tools/pull/12

Walavouchey (approved):

PR #8576, opened 2023-01-06
Add a note about lazer scores not being permanent

peppy:
Want to make this very prominent so people are not confused when we begin to reset scores.


wiki/Help_centre/Upgrading_to_lazer/en.md
```diff
@@ -11,6 +11,8 @@ The eventual goal is that this version will be released as an update which will
 
 "lazer" is a codename and will eventually be dropped as it becomes the primary release of the game. The rest of this document will refer to osu!(lazer) as "lazer" and osu!(stable) as "stable" for simplicity.
 
+*Notice: We are still figuring out game balance and mechanics. For now, **scores set on lazer should not be considered permanent**.*
```
Walavouchey, line 14: notices are supposed to be right below the heading, not at the end of a section

peppy: What's the rational for this? Are they going to be displayed differently in the future?

Walavouchey: this is by convention, similar to [hatnotes](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#hatnote), mimicking how they're displayed on wikipedia:

image

they play a similar role, and are sometimes stacked together. here's an example of a hatnote and a notice together:

image

Walavouchey: they're not the same thing as these kinds of boxes if you were thinking of that

image

Walavouchey (approved):

Ephemeralis (approved):

PR #8630, opened 2023-01-16
[VI] Update `BanchoBot`

TranT3011:
review for -lt but 3 months later because i was busy
also i kinda forgot how to use github so i accidentally merged my master branch lol

## Self-check


TicClick:
from a first glance, it doesn't look correct, as `!roll` works both with numbers and arbitrary strings: https://github.com/ppy/osu-wiki/blob/master/wiki/BanchoBot/en.md?plain=1#L62-L78. what are you trying to do, exactly?

TranT3011:
> from a first glance, it doesn't look correct, as `!roll` works both with numbers and arbitrary strings: https://github.com/ppy/osu-wiki/blob/master/wiki/BanchoBot/en.md?plain=1#L62-L78. what are you trying to do, exactly?

i dont really understand what "arbitrary strings" is
but to my undstanding "argument" is like the result of a equation and it just sounds weird to what we were taught at school because pretty much number works the same way

TicClick:
> `!roll <argument>/<number>`

this line means that `!roll` works both with numbers and with whatever you feed it:

```
15:51:25 <TicClick> !roll 12345
15:51:28 <BanchoBot> TicClick rolls 9859 point(s)
15:51:34 <TicClick> !roll anything you want
15:51:36 <BanchoBot> TicClick rolls 52 point(s)
15:51:38 <TicClick> !roll and even more
15:51:42 <BanchoBot> TicClick rolls 17 point(s)
```

it does not mean number division or anything you suggest about math, so you don't need to remove the `<argument>` part -- that is an incorrect edit

TranT3011:
> > `!roll <argument>/<number>`
> 
> this line means that `!roll` works both with numbers and with whatever you feed it:
> 
> ```
> 15:51:25 <TicClick> !roll 12345
> 15:51:28 <BanchoBot> TicClick rolls 9859 point(s)
> 15:51:34 <TicClick> !roll anything you want
> 15:51:36 <BanchoBot> TicClick rolls 52 point(s)
> 15:51:38 <TicClick> !roll and even more
> 15:51:42 <BanchoBot> TicClick rolls 17 point(s)
> ```
> 
> it does not mean number division or anything you suggest about math, so you don't need to remove the `<argument>` part -- that is an incorrect edit

oh ok ima change it to sth else

TicClick:
if that affects only the translation, sure. if it's something that changes the meaning of the article in general, regardless of the language, you need to edit the original article, `en.md`, instead

TranT3011:
> if that affects only the translation, sure. if it's something that changes the meaning of the article in general, regardless of the language, you need to edit the original article, `en.md`, instead

hi finally reply because i was lazy
no im not changing the meaning of the article

TicClick (approved):

TicClick:
seems to be small enough to skip the native review step

PR #8634, opened 2023-01-17
Mark `History of osu!/2009` a stub

tematikkp:
## Self-check


wiki/History_of_osu!/2009/en.md
```diff
@@ -1,3 +1,8 @@
+---
+outdated: true
+needs_cleanup: true
```
Walavouchey, line 3: tbh i think this is more appropriate

```suggestion
stub: true
```

Walavouchey (dismissed):
the translations need to have the same tags too

TicClick:
I went ahead and applied Walavouchey's suggestion

TicClick (approved):

PR #8710, opened 2023-01-28
Update `osu!mania Modding and Mapping Hub Server` article

mousesalt:

replaced with a new verification system.
copied from other servers' articles (same as FR and DE articles). :wink:


TicClick:
uhh, could you point where you copied the translations from? having a bit of a hard time finding the sources..

mousesalt:
> uhh, could you point where you copied the translations from? having a bit of a hard time finding the sources..

both from taiko hub server article in this sub folder.

TicClick (approved):

PR #8752, opened 2023-02-02
Adjust supporter message for first-time givers

The-Last-Cookie:
Companion to [osu-web#9820](https://github.com/ppy/osu-web/pull/9820)

- [x] This message will be visible on the next (localisation) roll out of osu-web. I'm opening this beforehand.

The maintainers may decide when to merge this.

~~Intentionally not outdating the translations as of now, in case contributors of the translations want to give their input to this PR. It is such a small change.~~

Some translations will need to be outdated anyway, so might as well do it now. This PR is ready for merge from my side.

## Self-check

### DIFF
--- a/wiki/osu!supporter/en.md
+++ b/wiki/osu!supporter/en.md
@@ -143,8 +143,8 @@ osu!supporter may also be gifted to another player by entering their username in
 
 After the transaction is completed, there will be a new event in the `Recent Activity` section in the profile of both the giver and the recipient:
 
-- `{username} has supported osu! - thanks for your generosity!` if it is the first time the user gets osu!supporter or makes a gift.
-- `{username} has once again chosen to support osu! - thanks for your generosity!` if the user has been an osu!supporter or has made a gift before.
+- `{username} has supported osu! - thanks for your generosity!` if it is the first time the user gets osu!supporter or gifts it to someone else.
+- `{username} has once again chosen to support osu! - thanks for your generosity!` if the user has been an osu!supporter or gifted it before.
 - `{username} has received the gift of osu!supporter!` in case the tag was gifted.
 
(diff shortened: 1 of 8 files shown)


wiki/osu!supporter/en.md
```diff
@@ -143,8 +143,8 @@ osu!supporter may also be gifted to another player by entering their username in
 
 After the transaction is completed, there will be a new event in the `Recent Activity` section in the profile of both the giver and the recipient:
 
-- `{username} has become an osu! supporter - thanks for your generosity!` if it is the first time the user gets osu!supporter.
-- `{username} has once again chosen to support osu! - thanks for your generosity!` if the user has been an osu!supporter before.
+- `{username} has supported osu! - thanks for your generosity!` if it is the first time the user gets osu!supporter or makes a gift.
+- `{username} has once again chosen to support osu! - thanks for your generosity!` if the user has been an osu!supporter or has made a gift before.
```
Walavouchey, line 147:
```suggestion
- `{username} has supported osu! - thanks for your generosity!` if it is the first time the user gets osu!supporter or gifts it to someone else.
- `{username} has once again chosen to support osu! - thanks for your generosity!` if the user has been an osu!supporter or gifted it before.
```

Walavouchey:
these strings are localised via crowdin, so updating needs to be done in two places. perhaps a comment can be left while outdating the translations, something like "the osu!web strings on on line # have been updated as of #osu-web-pr. to update this page, also check if the corresponding strings on crowdin need to be updated."

The-Last-Cookie:
For the record, the dutch translation seems to have a different structure. I have left an extra comment in the respective file.

Walavouchey (approved):

peppy (approved):

PR #8756, opened 2023-02-03
Rephrase rule in Twin Trials articles

pishifat:
a few people asked, so it's worth rewording this

### DIFF
--- a/news/2023/2023-02-01-twin-trials-contest-beatmapping-phase.md
+++ b/news/2023/2023-02-01-twin-trials-contest-beatmapping-phase.md
@@ -58,7 +58,8 @@ These rules only apply to the **beatmapping phase**. The storyboarding phase wil
   - Teams must have 2 members. No more, no less.
 - **Submissions must not be publicly revealed until after the judging phase ends.**
 - **Submissions must not be Qualified until after the community vote ends.**
-- **A user cannot create more than one submission per game mode.**
+- **Each team cannot create more than one submission per game mode.**
+- **A user cannot participate in more than one team per game mode.**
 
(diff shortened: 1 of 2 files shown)


news/2023/2023-02-01-twin-trials-contest-beatmapping-phase.md
```diff
@@ -58,7 +58,7 @@ These rules only apply to the **beatmapping phase**. The storyboarding phase wil
   - Teams must have 2 members. No more, no less.
 - **Submissions must not be publicly revealed until after the judging phase ends.**
 - **Submissions must not be Qualified until after the community vote ends.**
-- **Each team cannot create more than one submission.**
+- **A user cannot create more than one submission per game mode.**
```
Walavouchey, line 61: huh? is this implying that both members of the team can submit a mapset each for the same game mode? don't you get like 8 max submissions out of this with one team, two per game mode?

Walavouchey: what case is this actually trying to cover

pishifat: if someone does that, they would have created 2 submissions per game mode
i think it's ok to let people submit to different modes though (since they act as different contests with different competitors and judges etc)

Walavouchey: that's assuming the two users collaborated on at least one submission, in which case yeah at least one user made two submissions. it's this team --> user rewording that makes no sense

other than that it's ok

```suggestion
- **Each team cannot create more than one submission per game mode.**
```

pishifat: i've added a new line to clarify. i was trying to minimize the rule into one but i guess it's confusing

Walavouchey: and to clarify https://osu.ppy.sh/comments/2606379

> can a team map multiple gamemodes in a submission?

this is not allowed in any case, all 5 diffs in the same submission are of the same mode, right?

pishifat: that's unrelated

Walavouchey (approved):

peppy (approved):

PR #8245, opened 2022-10-22
Rewrite judgement articles

Walavouchey:
part of https://github.com/ppy/osu-wiki/issues/8189

### stuff to check

- [x] Opinions on whether we want to use the new judgement names throughout articles. I thought it seemed ok, even though technically they only appear in lazer.
  - [x] ~~Blocked on finalised judgement names?~~ another day
- [x] ~~The mania Perfect, Meh, and Miss windows differ from the [OD article](https://osu.ppy.sh/wiki/en/Beatmap/Overall_difficulty#osu!mania) as i've taken all of them from lazer. Will need to confirm whether this is an actual gameplay difference between lazer and stable.~~
  - Turns out the OD article was just wrong and put the Miss window in the place of the Meh window (and left out miss windows)
- [x] Confirm hit window/hit error rounding shenanigans
  - The hit windows are truncated after accounting for OD, while the hit error uses rounded time derived from the audio engine.
  - DT and HT make the clock tick faster and slower respectively, which scales the hit error and hit windows, which may affect how much the hit windows can vary. Haven't thought it through thoroughly.
- [x] Confirm spinner spin requirements

### remarks

- ~~The mania hold note table was created by empirical testing (I have no idea what the logic there is, but it's different from lazer I believe)~~
  - [x] Potentially redo table due to [1.5x-as-lenient release windows](https://github.com/ppy/osu/blob/60c8ef3fe5fe9adcb204814d975b7ed6c832fc8c/osu.Game.Rulesets.Mania/Objects/Drawables/DrawableHoldNoteTail.cs#L18-L23)
     - redone using stable source
- no idea what to do about the catch article, it doesn't seem to have the concept of "judgements" in the same way anyway

### future effort

- separate article for judgement names in other rhythm games
- fix mentions of judgements across the wiki
- move slightly out-of-scope stuff like score and accuracy-related information out of the judgement articles
- fill index page with a more general gameplay-related description about judgements ([judgement line, approach circle](https://github.com/ppy/osu-wiki/pull/8245#issuecomment-1288261369))

PR #8791, opened 2023-02-11
Add lazer feature comparison article

Walavouchey:
with the upcoming lazer mechanics panel i want this at least pr'd, but this may be mergeable sort of as is with some adjustments to the mess that are the long bullet point lists. either way, some eyes on this would be appreciated, especially suggestions for what's missing.

- expanding the non-gameplay difference sections into fuller descriptions and images is probably the way forward (slowly, eventually.). if there's a good way to tabulate it then that'd probably be a good short-term solution too.
- recording and editing gifs takes time and the file sizes get quite large, so it might not be worth considering for every little thing as i quite quickly came to realise
- some of the information in the gameplay difference tables are educated guesses, and i've marked those i'm more unsure of with question marks. dev opinions may or may not differ significantly though so that should be checked
- in the future, a good idea might be to include github issues/discussions in the gameplay differences info tables


peppy:
@Walavouchey I would see value in getting this live ASAP then improving as we go, rather than leaving in a draft PR state.

Walavouchey:
@peppy please check if the tables in the gameplay section are accurate.

i'll see if i can get this to an acceptable state today

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+
+#### The health system is different
+
+This is mostly due to changes to the judgement system. For example, Geki and Katu judgements do not exist in lazer, which affected health on stable.
+
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | No? |
```
peppy, line 106: yes but only to simplify things.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+
+#### The Easy mod no longer pauses gameplay when recovering from failure
+
+<!-- paragraph? -->
+
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | No |
+| Classic mod revert support | No |
+| Intentionally changed | No |
```
peppy, line 95: no, just not implemented yet i guess

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+
+#### The Flashlight gradient does not match stable
+
+<!-- comparison images for relevant game modes  -->
+
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | No |
+| Classic mod revert support | No |
+| Intentionally changed | No |
```
peppy, line 84: this was recently revisited, at least for osu! mode and should match pretttttttttttty close now

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | Yes |
+| Needs further consideration | Yes? |
+
+#### Hitsound volume cannot be adjusted in between notes
+
+In stable, it is possible to change the volume of ghost taps in osu!taiko and osu!mania, or create gradual osu!taiko swell volume changes. In lazer, there is only per-note volume, making that impossible.
```
peppy, line 131: This seems like a bug, and I'm not sure it deserves to be listed here. Probably better off making an issue if one doesn't exist (I have a strong feeling one does already though).

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | No |
+| Classic mod revert support | Yes |
+| Intentionally changed | Yes |
+| Needs further consideration | No |
+
+#### The spinner spin speed cap of 477 RPM has been removed
+
+Instead of a speed cap, spinners now have a score cap. The reason for this is inconsistency in how score is awarded, because the algorithm is dependent on frame rate.
```
peppy, line 241: That's not really the reason. It was just a simple implementation we haven't done another pass on yet. Calculations on lazer are done in a way that frame rate should not be an issue, at very least. But to make spinners feel good it may factor back in, we'll see.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+
+Some adventurous beatmaps exploit glitches in the stable client that allowed for very weird slider mechanics. These range from zero-length sliders acting as invisible circles, to cross-screen stretched and squished sliders.
+
+More discussion and consideration will be needed for how much of Aspire beatmaps will be compatible going forward. For example, invisible circles might become a properly supported feature in the future.
+
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | Yes |
```
peppy, line 262: I'd argue "No" is the answer here. Like we didn't intentionally break them. We just intentionally haven't spent effort on fixing them yet. But your call.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+| Needs further consideration | Yes |
+
+### osu!taiko
+
+#### The playfield aspect ratio is limited
+
+The osu!taiko playfield is stretched based on window size, making the apparent AR of gameplay change in line with window size. Originally this was limited completely as it affects playability but switched to only limiting the maximum aspect ratio due to community push-back for having it "as it was on stable".
```
peppy, line 269: Not sure this still applies

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | No |
+| Needs further consideration | Yes |
+
+#### Drumrolls do not prevent mashing
+
+In stable, drumrolls could not be hit too quickly or too slowly. This restriction has been lifted, just like in ScoreV2.
```
peppy, line 291: Are you sure? Looking at the implementation that does definitely not look to be the case. Each tick in the drumroll has a judgement window.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | No? |
+| Needs further consideration | Yes |
+
+### osu!catch
+
+<!-- TODO: are there any differences... ? -->
```
peppy, line 311: probably

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | Yes |
+| Needs further consideration | Yes |
+
+#### Hold notes may sometimes have missing ticks
+
+This is a bug that has not yet been fixed.
```
peppy, line 343: Again, worth reporting as a bug rather than listing here?

I've never heard of this one FWIW.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | No |
+| Needs further consideration | Yes |
+
+#### SV is limited to between 0.01 and 10
+
+Some beatmaps with SV gimmicks like teleports or stops will not look as intended, but are otherwise playable.
```
peppy, line 354: I'm pretty sure it's limited the same way on stable, and we are matching it. It'd be worth searching for issues regarding this and providing example beatmaps which don't work.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+There are [designs](https://www.figma.com/file/ytnnne2TH8Z956Jxhiypqq/Beatmap-Editor-2) for an upcoming built-in storyboard editor similar to the one in stable, but no work on this has started yet.
+
+  - The trigger system is not implemented
+  - Beatmap backgrounds do not have the correct stretch mode for 4:3 storyboards.
+
+## Some cosmetic features are missing
+
+These are features that are planned to come back at some point, but have not due to low priority.
+
+- Beatmap and replay comments do not exist
```
peppy, line 421: I'm not sure these will be returning.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,438 @@
...
+- Cursor ripples
+- Non-skinnable components
+  - Gameplay leaderboard
+  - Cutscene skip button
+  - Section fail/pass
+  - Results screen
+  - Song select
+  - Pause menu
+
+## Technical differences
```
peppy, line 433: There's probably many more we can add here, but for the future I guess.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: then i guess it would be "intentionally changed: no, needs further consideration: yes"

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: indeed: https://github.com/ppy/osu/discussions/17247. i got most of these issues from either open issues, discussions or while researching stable code. i do need to add links later as mentioned in op

> In stable, you can change the volume of "ghost taps", or create fluent swell volume changes. In Lazer, there's only per-note volume, making that impossible.

perhaps it should be split off to a separate issue

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: hmm this was something i vaguely recalled hearing from some dev discussion, but can't find it anywhere. either way, it is true that frame rate affects spinning on stable, but yes i can remove it if it wasn't the actual reasoning.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: whether temporary or not, the interpretation i'm going with is that this wasn't unimplemented by oversight, so that would be a "yes".

"needs further consideration: yes" accounts for what you mean, no?

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: this is copied from https://github.com/ppy/osu/wiki/Gameplay-differences-from-osu!stable#osutaiko-gamefield-aspect-ratio-is-limited and i believe this reflects how it stands currently.

i do know about https://github.com/ppy/osu/pull/22581 which may change things

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: yes, drumroll ticks have a hit window in both stable and lazer, but stable scorev1 has an additional layer that blocks input if you're hitting [more than twice as fast as ticks appear or hitting slower than every 5th tick](https://github.com/ppy/osu-wiki/pull/8245/files#diff-b82a37959e54dc656a70a58c3417fa40c6bced969b74229307f3fcc785425150R29) ([relevant stable code](https://github.com/peppy/osu-stable-reference/blob/1531237b63392e82c003c712faa028406073aa8f/osu!/GameplayElements/HitObjects/Taiko/SliderTaiko.cs#L362-L396))

drumroll ticks don't have miss windows unlike other hit objects, so in lazer (and stable scorev2) you're able to mash through the drumroll as long as some of your keypresses land on the ticks' hit windows

there is an open issue about this at https://github.com/ppy/osu/issues/5694

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: i heard this from you lol <https://discord.com/channels/188630481301012481/188630652340404224/1032979827596210237>

can agree removing this one because i can't find an issue or discussion about such a thing.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: maybe i should reword this, because this specifically refers to teleports and stops (which maybe use extreme bpm values instead, haven't checked). scroll speed is definitely being limited in some way.

there are plenty of discussions that link example beatmaps:

https://github.com/ppy/osu/discussions/15638
https://github.com/ppy/osu/discussions/14835#discussioncomment-1372777
https://github.com/ppy/osu/discussions/15016

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: existing issue: https://github.com/ppy/osu/issues/4796

it will still be listed since it's a difference

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: Yeah, I think so.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: I've tested and indeed it doesn't seem to work. I'll take a look into this today and open an issue if it's not a simple fix.

peppy: https://github.com/ppy/osu/pull/22659 :)

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: As in I would say "Intentionally changed: no" here. Because saying "no we're not focusing on that yet" doesn't mean we intentionally made a change. Using that argument, basically everything in this document would be a "yes" to this answer (because it's not looked into yet).

Supporting "aspire" beatmaps is one of the most time intensive tasks, which requires actively investigating sometimes megabyte large files using non-standard mechanics which we have never seen before, reading through thousands of lines of stable code and trying to understand how they work. Then taking that and deciding how we can support that in a completely different code base, if at all.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: With the current implementation, it doesn't need to be listed there or here anymore. For all standard aspect ratios (read: all, since stable was limited to standard aspect ratios) it behaves the same as stable.

The PR you linked fixes lazer now being *too* flexible, allowing a range that stable broke at (and lazer also breaks at right now – the goal of that PR is to reduce visual breakage).

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: Should be reworded to "niconico-style scrolling comments", since there are also beatmap page comments (which are already implemented in the game).

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: how close is close enough i wonder

https://user-images.githubusercontent.com/36758269/219944801-f4aa5916-ff73-4712-9793-77e43b051282.mp4

Walavouchey: taiko

| stable | lazer |
| :-: | :-: |
| screenshot703 | osu_2023-02-19_12-31-20 |

https://user-images.githubusercontent.com/36758269/219945689-6cd4015f-0cce-4749-a034-e11a3f145cd8.mp4

flashlight size seems to match at least

https://user-images.githubusercontent.com/36758269/219945508-56dd8b85-5bcc-4358-9003-bd338de2df7d.mp4

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: i've opted for a comment that links to the above pr for now

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: reworded

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: reworded heading

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,450 @@
...
+
+#### Drumrolls do not prevent mashing
+
+In stable, drumrolls could not be hit too quickly or too slowly. This restriction has been lifted, just like in ScoreV2.
+
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | Yes? |
```
Walavouchey, line 285: still unsure about this one (https://github.com/ppy/osu-wiki/pull/8791#discussion_r1106846327 didn't really give much of an indication)

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,450 @@
...
+#### The PERFECT judgement hit window scales with OD
+
+This used to be a constant ±16 ms regardless of overall difficulty.
+
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | Yes |
+| Needs further consideration | No? |
```
Walavouchey, line 362: and this

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
```diff
@@ -0,0 +1,450 @@
...
+#### Converts no longer have different hit windows
+
+Beatmaps converted from the osu! game mode into osu!mania had a set of hit windows that were not affected by OD.
+
+|  |  |
+| :-- | :-: |
+| Breaks backwards compatibility | Yes |
+| Classic mod revert support | No |
+| Intentionally changed | Yes? |
+| Needs further consideration | No? |
```
Walavouchey, line 373: and these

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: the absolute state of lazer mania flashlight
| stable | lazer |
| :-: | :-: |
| stable-mania-flashlight | lazer-mania-flashlight |

Walavouchey: catch seems to match pretty much perfectly


https://user-images.githubusercontent.com/36758269/219947705-265a1133-5df3-4352-8d20-8a663e67cd42.mp4

Walavouchey:
i've resolved the conversations where i've applied your suggestions directly

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
Walavouchey: i've split up this section into taiko and mania in https://github.com/ppy/osu-wiki/pull/8791/commits/e57c1a213e8ac9a9cfc50432331e8c2522856aa6

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: It's merged so please remove this section now.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: Dunno, whatever you think.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: Not sure. I think leaving the question mark is probably fine for now.

wiki/Client/Release_stream/Lazer/Feature_comparison/en.md
peppy: I have no fucking idea. This sounds like a stable oversight from day one.

peppy:
I think this is in a good initial state. Great job on this effort 👍

peppy (approved):

PR #8821, opened 2023-02-18
Update `MWC7K 2023` with tournament results

LeoFLT:
Still missing a few assets (will be added at a later date). Translation checks shouldn't be necessary as country names were taken from the same tables that were edited.

# Self-check


TPGPL (commented):
i think L57 about donating to the prize pool can be removed

TPGPL (approved):

PR #8838, opened 2023-02-22
[DE] Adjust rule of pronoun usage

The-Last-Cookie:
The first version of the `perspective` section in the ASC was rather inaccurate and more or less a fitting translation of the English original.

After the addition of a rule for formal pronouns in #8435, there came the discussion up on discord that the German ASC doesn't correctly define what pronouns should be used (and when).

This PR is a proposal to solve this inconsistency.

## Self-check


respektive (approved):
I don't have a strong opinion on pronoun usage for the wiki so sure, let's go with this

TicClick (approved):

PR #8866, opened 2023-03-01
Fix formatting error in `Featured Artist showcase beatmaps`

TheRealStevie:
## Self-check


Walavouchey (approved):
ty for catching this!

PR #8911, opened 2023-03-08
Fix wrong flag in `Beatmap Spotlight Curators`

aceticke:
Ukraine flag was set to UK instead of the correct UA, so I fixed it for now.

## Self-check


aceticke:
I gotta push the translations sec

Walavouchey (approved):

PR #8922, opened 2023-03-10
[FR] Add outdated tags to `Users with unique titles`

agatemosu:
## Self-check


TicClick (approved):

TicClick:
thanks for catching this

PR #8926, opened 2023-03-10
Remove the `Beatmapping/Health drain` stub in favour of `Gameplay/Health`

TicClick:
part of https://github.com/ppy/osu-wiki/issues/6080 -- the main article provides far better coverage of drain and health, so the stub is unnecessary

## Self-check


Walavouchey (commented):
links should be replaced, not deleted, right?

Walavouchey (dismissed):

Walavouchey (approved):

Issue #7254, opened 2022-04-27
Make a top-level `Offset` category and group existing offset articles under it


Walavouchey:
Split off from https://github.com/ppy/osu-wiki/issues/6080

[Discussion](https://discord.com/channels/188630481301012481/218677502141399041/923229492498010263) [on](https://discord.com/channels/188630481301012481/218677502141399041/964863600655212604) [Discord](https://discord.com/channels/188630481301012481/218677502141399041/968775283526234142) has come to the conclusion that the offset articles would on their own reside in separate categories, but me and TicClick have agreed to instead group them together as subarticles under an `Offset` directory.

This index would ideally provide an overview of what offsets are, what they do and what types of offsets exist in osu!, essentially replacing [`Disambiguation/Offset`](https://osu.ppy.sh/wiki/en/Disambiguation/Offset). Note that it would encompass all types of offsets (the timing kind, not spatial), with the basic concept of keeping things in sync in gameplay:

- Offsets that shift everything relative to the audio track (these three do the exact same thing)
  - [Local offset](https://osu.ppy.sh/wiki/en/Glossary/Local_song_offset)
  - [Universal offset](https://osu.ppy.sh/wiki/en/Client/Options/Universal_offset)
  - [Online offset](https://osu.ppy.sh/wiki/en/Beatmap/Online_offset)
- [Beatmap offset](https://osu.ppy.sh/wiki/en/Beatmapping/Offset) (the "time between the start of the audio file and the song's first downbeat", i.e. the first red line) (not to be confused with beatmap offset in lazer, which is local offset renamed. i've also heard this being called "editor offset")
- [Lead-in](https://osu.ppy.sh/wiki/en/Glossary/Lead-in) (the `AudioLeadIn` setting in a `.osu` file, which delays the start of the whole beatmap)
  - Perhaps a mention about storyboard intros, which effectively do the same thing
- Video offset (set in the `.osu file`, the delay from the audio track's start time to before a video starts)
  - This doesn't have an article, but it's mentioned in [`Disambiguation/Offset`](https://osu.ppy.sh/wiki/en/Disambiguation/Offset) (and it doesn't really need one, but can be mentioned)

General clean-up of the offset articles is also part of this (adding images and using more precise wording)

PR #8954, opened 2023-03-14
Move offset articles to the top-level category

TicClick:
part of https://github.com/ppy/osu-wiki/issues/7254, continuation of https://github.com/ppy/osu-wiki/pull/8921

didn't move the timing points and the audio lead-in articles as I think they will benefit from coupling with their present categories

## Self-check


Walavouchey (commented):
glossary can be deleted

`Glossary/img/*` can be moved to `meta/unused`, if not deleted as well

Walavouchey (approved):

PR #8959, opened 2023-03-14
Miscellaneous BN page fixes

aceticke:
## Self-check


TicClick (approved):
(20 commits compressed into 1)

PR #8984, opened 2023-03-19
Add "Doomsday - Length of Location" to the list of `osu! originals`

Nivalyx:
was made aware that [Doomsday](http://osu.ppy.sh/users/18983) (yes, that Doomsday) had been releasing a couple of originals in which he had also mapped himself so yea


Walavouchey:
since these were uploaded on other platforms before osu, we'll need to confirm whether these songs were specifically made for those maps

for example, kurokotei's galaxy collapse isn't listed because even though they made a map of their song, the song itself wasn't made specifically for osu

Nivalyx:
yup

i doublechecked this with Doomsday himself and apparently "Length of Location" was the only one that premiered as an osu! beatmap

image

TicClick (approved):

PR #8994, opened 2023-03-21
Use standard video embed size in `Project Loved: March 2023` news post

Walavouchey:
to match other (recent) featured artist news posts

the intent is to have this be the case in all posts from now on (also in osu!monthly posts, etc.). some use 100%, some use 95%, some use specific pixel values and it's very inconsistent

question is if this should be done retroactively for older posts


pishifat:
> question is if this should be done retroactively for older posts

if this happens, only apply it to 16:9 videos. a few fa videos are fairly tall (like https://osu.ppy.sh/home/news/2021-10-30-new-featured-artist-garlagan) so they'd cover too much screen space at high %s

Ephemeralis (approved):

PR #9003, opened 2023-03-23
Update skinning directory structure mentioned in `osu!tourney/Skinning`

zSmoxKe:
It seems this has changed, I fixed it.

## Self-check


Walavouchey:
this has been in the article since forever. was it changed at some point?

zSmoxKe:
> > this has been in the article since forever. was it changed at some point?
> 
> It looks like so... If you do a new tourney installation and you follow the steps in the wiki, it does not work. You have to put your skin in `<your-osu-tourney-installation>/Skins/User`.

Okay now I think I got it, gameplay skins need to be placed in `<your-osu-tourney-installation>/Skins/User`. while the backgrounds of the client must be placed into another folder called "tournament" inside the "User" folder. so the skinning directory structure was good, but there was an inaccuracy in the sentence after where it says "The skin elements can be placed in this folder and support `.jpg` and `.png` file extensions." It's not skin element but the backgrounds images of the client instead.

PR #9023, opened 2023-03-25
Add `Music theory/Layer`

TicClick:
## Self-check


wiki/Music_theory/Layer/en.md
```diff
@@ -0,0 +1,29 @@
+---
+tags:
+  - musical layer
+  - sound layer
+---
+
+# Layer
+
+A **musical layer** is a part of the song which contains features with a common aspect between them. A significant part of [mapping](/wiki/Beatmapping) in every [game mode](/wiki/Game_mode) revolves around following different kinds of layers (percussion, vocals, sound patterns) consistently, and switching between them for the purpose of emphasis and contrast.
```
Walavouchey, line 9:
```suggestion
A **musical layer** is a part of the song which contains features with a common aspect between them. A significant part of [beatmapping](/wiki/Beatmapping) in every [game mode](/wiki/Game_mode) revolves around consistently following different kinds of layers (percussion, vocals, and sound patterns), and switching between them for the purpose of emphasis and contrast.
```

wiki/Music_theory/Layer/en.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Description
+
+Typically, a single layer in a musical piece represents a group of instruments of the same family or [range](https://en.wikipedia.org/wiki/Range_(music)), or — alternatively — voices of the [same type](https://en.wikipedia.org/wiki/Voice_type). Since layers are abstract, they can be limited to just a single instrument or kind of sound, depending on the context they are discussed in. Other criteria for layer subdivision include [rhythm](https://en.wikipedia.org/wiki/Rhythm) patterns, [melody](https://en.wikipedia.org/wiki/Polyphony), [harmony](https://en.wikipedia.org/wiki/Harmony), or [pitch](https://en.wikipedia.org/wiki/Pitch_(music)).
+
+## Mapping
+
+In osu! as a game, song layers are represented through different [patterns](/wiki/Beatmap/Pattern) and sections of the beatmap. While picking a layer to follow is usually up to a beatmap author's discretion, the choices are usually the most obvious and recognisable ones:
```
Walavouchey, line 17: i'm pretty sure it's more like the rhythm choice depends on which layer is followed

the patterns may or may not relate to layers or specific instruments, but the logic of when to use which pattern can depend on way more things

wiki/Music_theory/Layer/en.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Percussion, such as drums, hi-hats, or other similar instruments providing the [beat pattern](/wiki/Music_theory/Beat)
+- Instruments forming the main melody pattern(s)
+- Vocal party
```
Walavouchey, line 21: "beat pattern" is a bit of a nebulous connection

as mentioned, the rhythms of these is taken into consideration and the one followed directly corresponds to where to place hit objects on the timeline, as well as which hit object to use

wiki/Music_theory/Layer/en.md
```diff
@@ -0,0 +1,29 @@
...
+
+### osu!mania
+
+Due to the unique nature of [osu!mania](/wiki/Game_mode/osu!mania), the most prominent layers of a song are often represented by separate columns on the playfield.
```
Walavouchey, line 25: i have a hard time imagining this is true. mania is very pattern-centric, and i don't think i've ever seen a map with this kind of gimmick (granted, i'm still a beginner at mania)

wiki/Music_theory/Layer/en.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Improvisation
```
Walavouchey, line 27: super big thing in taiko, but in osu! at least it'd be more likely to call this "overmapping"

```suggestion
### Improvisation / overmapping
```

wiki/Music_theory/Layer/en.md
TicClick: I meant the sounds which form the tempo (1/1 beats and smaller notes that surround them). I can change "beat pattern" to something more direct to clarify that.

wiki/Music_theory/Layer/en.md
TicClick: yes. I wanted to explain that the musical layers translate to the hit objects (well, isn't that evident), but that seemed to be too simple

wiki/Music_theory/Layer/en.md
TicClick: provided I have a total of 6 impressive unranked plays, I should probably retract the section as a whole -- it may not apply to the majority of maps.

wiki/Music_theory/Layer/en.md
TicClick: I'd rather insert a reference to the overmapping article instead, don't like slashes in the section headings

wiki/Music_theory/Layer/en.md
Walavouchey: "rhythm" is the more direct word here, and i don't think i understand your explanation, since the "beat" is just a repeating pulse

you're not mapping to a "tempo" (or in other words mapping to the "speed of the song", whatever that would mean), but the various *rhythms* formed by the percussion, other instruments, vocals, etc.

it's the terminology that i'm fundamentally concerned about

TicClick: fair enough, removed

wiki/Music_theory/Layer/en.md
```diff
@@ -0,0 +1,25 @@
...
+  - musical layer
+  - sound layer
+---
+
+# Layer
+
+A **musical layer** is a part of the song which contains features with a common aspect between them. A significant part of [beatmapping](/wiki/Beatmapping) in every [game mode](/wiki/Game_mode) revolves around consistently following different kinds of layers (such as percussion, vocals, and sound patterns), and switching between them for the purpose of emphasis and contrast.
+
+## Description
+
```
Walavouchey, line 12: correct me if i'm wrong, but this kind of section isn't really a thing in other articles

```suggestion
```

TicClick: in the music theory? yes, it isn't. my intention was to make a very brier description of what a layer is, then expand on every side or that in separate sections. think wikipedia, a well-formed article has a leading section with a short overview, followed by a detailed breakdown. it usually works better than large pieces of text without any subdivision, especially in large articles, so I wanted to give it a go for smaller topics as well

TicClick: on a second thought, the page the way I wrote it looks better without this heading since the second paragraph reads smoothly enough as a continuation of the first one

Walavouchey (approved):

PR #8867, opened 2023-03-01
Change some links to directly go to the osu! weppage

TheRealStevie:
So starters (like me 😅) won't get confused navigating through.
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [ ] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #9031, opened 2023-03-26
CI: exclude non-articles from style checks, and fix minor visual nuisances

TicClick:
unblocks https://github.com/ppy/osu-wiki/pull/8867

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -64,8 +64,8 @@ jobs:
       - name: run remark on changed files
         run: |
           readarray -t REMARK_TARGET_FILES <<< $(
-            # ppy/osu-wiki#8867 -- disable style checks for regular Markdown files, such as README.md
-            echo "${{ env.INTERESTING_FILES }}" | grep -e .md$ | grep -E -v '^[[:upper:]]+\.md$' || true
+            # ppy/osu-wiki#8867 -- disable style checks for non-article Markdown files, such as README.md
+            echo "${{ env.INTERESTING_FILES }}" | grep -e .md$ | grep -E -v '^[A-Z-]+\.md$' || true
           )
           bash scripts/ci/run_remark.sh --target "${REMARK_TARGET_FILES[@]}"
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -111,8 +111,7 @@ jobs:
           for (( i=1; i<=MAX_ATTEMPTS; i++ )); do
             # get the first commit of the branch associated with the PR; GitHub's ubuntu-latest has curl/jq: https://github.com/actions/virtual-environments
             API_RESPONSE=$(
-              curl -v \
-                --header 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -63,7 +63,10 @@ jobs:
 
       - name: run remark on changed files
         run: |
-          readarray -t REMARK_TARGET_FILES <<< $( echo "${{ env.INTERESTING_FILES }}" | grep -e .md$ || true )
+          readarray -t REMARK_TARGET_FILES <<< $(
+            # ppy/osu-wiki#8867 -- disable style checks for regular Markdown files, such as README.md
+            echo "${{ env.INTERESTING_FILES }}" | grep -e .md$ | grep -E -v '^[[:upper:]]+\.md$' || true
```
Walavouchey, line 68: what kind of wack-ass syntax is this? wouldn't something like this work?

```suggestion
            echo "${{ env.INTERESTING_FILES }}" | grep -e .md$ | grep -E -v '^[A-Z-]+\.md$' || true
```

.github/workflows/continuous-integration.yml
```diff
@@ -63,7 +63,10 @@ jobs:
 
       - name: run remark on changed files
         run: |
-          readarray -t REMARK_TARGET_FILES <<< $( echo "${{ env.INTERESTING_FILES }}" | grep -e .md$ || true )
+          readarray -t REMARK_TARGET_FILES <<< $(
+            # ppy/osu-wiki#8867 -- disable style checks for regular Markdown files, such as README.md
```
Walavouchey, line 67:
```suggestion
            # ppy/osu-wiki#8867 -- disable style checks for non-article Markdown files, such as README.md
```

run-checks.sh
```diff
@@ -57,7 +57,8 @@ function main() {
     exit 0
   fi
 
-  interesting_articles=$( echo "${interesting_files}" | grep -e ^wiki/ -e ^news/ | grep .md$ )
+  # ppy/osu-wiki#8867 -- disable style checks for regular Markdown files, such as README.md
```
Walavouchey, line 60:
```suggestion
  # ppy/osu-wiki#8867 -- disable style checks for non-article Markdown files, such as README.md
```

.github/workflows/continuous-integration.yml
TicClick: https://www.gnu.org/software/grep/manual/html_node/Character-Classes-and-Bracket-Expressions.html not PCRE, but it works (: can change for readability if you insist

```
/home/user % grep --version
grep (GNU grep) 3.7
--- snip ---
/home/user % echo -en 'a\nb\nDONALD-DUCK\nen.md\nCONTRIBUTING.md' | grep -Ev '^[[:upper:]]+\.md$'
a
b
DONALD-DUCK
en.md
```

Walavouchey: as is it wouldn't exclude [`TEMPALTE-SERIES.md`](https://github.com/ppy/osu-wiki/blob/master/wiki/Tournaments/TEMPLATE-SERIES.md) i'm pretty sure, so may want to add the dash

PR #9045, opened 2023-03-27
[ES] Update `Main Page`

agatemosu:
## Self-check


TicClick (approved):
most if not all of the titles are taken from the respective pages, and it's mostly about adding new links, so I think a native review is not required (simple enough), similarly to https://github.com/ppy/osu-wiki/pull/9044

PR #7807, opened 2022-08-08
[ES] Add `Community/How you can help!`

Raphiel-R:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] ~~*(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)~~
SKIP_WIKILINK_CHECK

PR #9052, opened 2023-03-28
[ES] Retranslate `How you can help!`

agatemosu:
## Self-check


TicClick (approved):

TicClick:
(not adding `no_native_review: true` as the translation was done by other person in https://github.com/ppy/osu-wiki/pull/7807)

Issue #4871, opened 2021-01-29
Better documentation of osu! history


MilkyIQ:
There's way too many holes in the "[History of osu!](https://osu.ppy.sh/wiki/en/History_of_osu%21/2018)" pages as well as basically no information pre-2011 and post-2018. Things like the history of the website design, dev blogs or updates during beta releases, as well as major milestones in player achievements (e.g., breaking the 1000pp barrier) throughout the years are pretty much nowhere to be found there. There's also been no updates to the [history of the wiki](https://osu.ppy.sh/wiki/en/History_of_osu%21/osu%21_wiki) post-2018. On top of this, the articles are very unorganized and need *heavy* cleanup in order to be serviceable. 

Thus, I propose that the History of osu! articles be updated and cleaned up with the following tasks:

- [ ] Cleanup or rewrite previously-existing articles to fit with current ASC guidelines
- [ ] Remove any frivorlous or unneccessary content
- [ ] Create articles for the years 2019, 2020, and 2021 (2021 bc of the pp update)
- [ ] Create articles for the beta years (2007 - 2008 iirc)
- [ ] Reference facts and events with sources where appropraite (links in HTML comments)
- [ ] Incorporate a "List of April Fools jokes" section or article into the main page as per #117

peppy:
I would love to see this section get a lot more coverage. I think it's important to preserve all the things that happened along the way. Let me know if I can be of help in providing any historical data or exports.

MilkyIQ:
So far I've been following the convention that gameplay records (e.g., pp scores, leaderboard changes, etc.) are not to be mentioned on the general History of osu! articles. I've been doing this for just the fact that they're subjective, a major PITA to keep track of, and because they probably deserve their own article because of all the little nuances (either in or out of the wiki), much like osu!lazer.

But as I continue to keep reading articles that mention "Cookiezi did x! Rafis did y!", I'm wondering whether we should mention the 1000pp milestone. It's a milestone that is very monumental and is backed by years of history. Mentioning it would be akin to the way we mention lazer milestones (like when it first went open source, when it (eventually) "replaces" stable, and the like. 

[...]

PR #9055, opened 2023-03-28
Add `History of osu!/2022`

The-Last-Cookie:
Here, I have a proposal for the 2022 history article.

This text is in no way complete. Things I mainly included are new featured artists, new lazer mods, presentations by core team members, contests and 15th anniversary related stuff.

What I'm not sure about are the osu! world cups and the april fool's joke. The osu! world cups for example were included in some of the newer articles (2017–2018), but not in all.

Part of #4871

The update for the main article will come in a follow-up.

## Tasks

- [x] Add entry about r/place

## Self-check


Walavouchey:
- where's r/place? it was one of the biggest events of yesteryear
- i don't think new featured artists should be listed in these history pages like this
- the roundtable should definitely be listed, but i'm not entirely sure about the world cup for the same reason as above
  - if there were some general stats for the year in an infobox table at the top then maybe it could go there, smth like "new Featured artists: 100, osu! World Cup winner: United States, new beatmaps: 100,000, new Ranked beatmaps: 10,000", but that's an idea for another time
- writing is pretty good at a quick glance

The-Last-Cookie:
> where's r/place?

At first glance, I thought about it to be more of a Reddit than an osu! thing and I was not sure if it should be included. Nonetheless, I fully agree with you that this event is definitely worth to mention.

For what it's worth, the `2017` article does not mention r/place.

> featured artists

The older history articles do mention the new artists specifically.

My other suggestion is to remove them entirely. Listing only a few of them (or making entire tables about join dates) doesn't make sense in my opinion. If one wants to find out more about featured artist announcements, they should be referred to the respective news post.

TicClick (changes requested):
some typos and missing links that I have shelved for a week already, didn't do a full-scale review yet

The-Last-Cookie:
To give an update about this article, I've began collecting sources about r/place that can be viewed [in this gist](https://gist.github.com/The-Last-Cookie/04fbcaa687a36bcb170b9a4d66c63fd0).

Furthermore, I will continue to extract the information out of the sources (it's just a bullet point list as of now) and maybe also cleanup the formatting, so that it *might* be used for a standalone article later on.

After this is done, I will begin writing a small entry of r/place for this article.

Walavouchey:
while the info and sources you've collected will be useful, a mention in this history article definitely doesn't need to be very elaborate.

The-Last-Cookie:
I've added a basic overview for r/place describing the event and showing two artworks from the pixel canvas.

wiki/History_of_osu!/2022/en.md
```diff
@@ -0,0 +1,206 @@
+# History of osu! 2022
+
+## January
+
+More artists joined the [Featured Artist](/wiki/People/Featured_Artists) program of osu! in the new year, namely [NIWASHI](https://osu.ppy.sh/beatmaps/artists/228), [DGK](https://osu.ppy.sh/beatmaps/artists/229), [Haywyre](https://osu.ppy.sh/beatmaps/artists/230), [satella](https://osu.ppy.sh/beatmaps/artists/231), [Vansire](https://osu.ppy.sh/beatmaps/artists/232), [Boom Kitty](https://osu.ppy.sh/beatmaps/artists/233), [Annabel](https://osu.ppy.sh/beatmaps/artists/234), and [Maduk](https://osu.ppy.sh/beatmaps/artists/235).
```
Walavouchey, line 5:
```suggestion
More artists joined the [Featured Artist](/wiki/People/Featured_Artists) program in the new year, namely [NIWASHI](https://osu.ppy.sh/beatmaps/artists/228), [DGK](https://osu.ppy.sh/beatmaps/artists/229), [Haywyre](https://osu.ppy.sh/beatmaps/artists/230), [satella](https://osu.ppy.sh/beatmaps/artists/231), [Vansire](https://osu.ppy.sh/beatmaps/artists/232), [Boom Kitty](https://osu.ppy.sh/beatmaps/artists/233), [Annabel](https://osu.ppy.sh/beatmaps/artists/234), and [Maduk](https://osu.ppy.sh/beatmaps/artists/235).
```

Walavouchey (commented):
some general feedback:

- this is a history article, so past tense is expected, but you seem to switch between past and present randomly
- many of the changes i've made mainly make the language more terse and less narrative, trimming away some embellishments from the explanations.

i'll commit the changes myself, but those are the main points i want you to take home if you want to read over the changes yourself.

Walavouchey (approved):

TicClick (approved):

PR #9076, opened 2023-04-04
Document new supported locales and rename `tl` translations to `fil`

Walavouchey:
- [x] make sure ci doesn't explode due to the 3-letter language code https://github.com/ppy/osu-wiki/pull/9497

today i learned that tagalog is basically the same language as filipino. since tagalog isn't listed in [osu-web](https://github.com/ppy/osu-web/blob/master/app/Libraries/LocaleMeta.php) but filipino is, the 3 translations can be renamed instead of deleted


Nivalyx:
*(( make sure ci doesn't explode ))*

TicClick (approved):

Walavouchey:
referring to this here https://github.com/Walavouchey/osu-wiki-tools/blob/master/wikitools/file_utils.py#L44-L45

TicClick:
oh well

PR #9098, opened 2023-04-06
[ES] Clean up `Tournaments/MWC/2014`

UnknownOsu:
## Self-check


wiki/Tournaments/MWC/2014/es.md
```diff
@@ -1,16 +1,16 @@
 ---
-outdated_translation: true
+outdated_translation: false
```
Walavouchey, line 2: it's still outdated, because it is not up to date with the english version

Walavouchey (commented):
https://github.com/ppy/osu-wiki/pull/9053 does a lot of bulk changes, one of them being that it uses "beatmaps" instead of "mapa", contrary to this pr. which should it be?

cc/ @agatemosu

agatemosu:
> https://github.com/ppy/osu-wiki/pull/9053 does a lot of bulk changes, one of them being that it uses "beatmaps" instead of "mapa", contrary to this pr. which should it be?
> 
> cc/ @agatemosu 

I made the changes because "beatmap" does not have a translation into Spanish and the wiki sometimes uses "beatmap" and other times "map". I don't know if in Spanish I should only put "map".

TicClick (approved):
merging (no native review for one week or longer). no `no_native_review: true` since the translation is outdated anyway and this PR is a small cleanup

PR #9119, opened 2023-04-08
[ES] Update `Reporting bad behaviour`

UnknownOsu:
## Self-check

I changed some photos because I couldn't find the user in the report photos in the profile, so I put my profile instead. I did the same with the other photos.

The [report beatmap picture](https://github.com/ppy/osu-wiki/blob/master/wiki/Reporting_bad_behaviour/img/report-beatmap.png) I left it unchanged because the [report beatmap button](https://i.imgur.com/DpqQrMt.png) is currently missing. So I left it as it was.

Tell me if any picture is wrong, I will change it as soon as possible.


agatemosu (commented):
It's really necessary?

UnknownOsu:
Why not? If there is the possibility of putting it in Spanish, I think it would be better for the reader.

TicClick:
it's been slightly more than a month now, is this ready to be merged?

UnknownOsu:
According to me, yes, unless there is something missing.

TicClick (approved):

PR #9129, opened 2023-04-10
[ES] Update `Beatmap discussion`

UnknownOsu:
## Self-check

Change of images to the current interface (where applicable) and translated language.

I do not recommend changing "mapa(s)" to "beatmap(s)", the words would not make sense with the images.

Also, I don't understand why "beatmap(s)" cannot be translated to Spanish. A great majority use the word "mapa(s)", it even appears on the game website; I think it should not be left untranslated.


TicClick (dismissed):

TicClick (approved):

PR #9154, opened 2023-04-14
Add Corne2Plum3 to the list of Global Moderation Team

tematikkp:
## Self-check


Walavouchey:
people listings are updated in batches, in tandem with the [staff log](https://github.com/ppy/osu-wiki/blob/master/wiki/People/Staff_log/2023/en.md?plain=1#L188-L200) in order to keep things in sync, and one is already planned along with other updates to the gmt article

please excuse the inconvenience. this should probably be indicated better to reduce wasted effort.

PR #9158, opened 2023-04-15
Expand `Tournaments`

FazlyMR:
Adds `Indonesian Taiko Beginner Showdown 2` into the `Asian-Oceanian Taiko Showdown` community tournament series.

This is to also to reflect a change I made to a still-open (as of time of writing) PR on a major listing update for the Indonesian translation of `Tournaments`.

Also, as of time of writing, folder `AOTS/ITBS_2` does not yet exist but added as the link to said new tournament as a prediction based on naming patterns of past tournaments within that series.

SKIP_WIKILINK_CHECK

## Self-check


Walavouchey:
> Also, as of time of writing, folder `AOTS/ITBS_2` does not yet exist but added as the link to said new tournament as a prediction based on naming patterns of past tournaments within that series.

please add it either as part of or after the pr that adds that tournament article

PR #9255, opened 2023-04-29
Add `Jinjin - Midnight Musing` to osu! originals

Coppertine:
## Self-check

Charter thought only FA's can get the spot.. 🤔


Coppertine:
Above PR does not modify section this song is added into.

Walavouchey (approved):

PR #9196, opened 2023-04-21
Update links to `osu!_Discord_Server` in `Main Page` 

Sitoria:
Fix `osu!dev Discord Server` to `osu! Discord Server` cuz of merging [the wiki page](https://github.com/ppy/osu-wiki/pull/9176) recently

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #9176, opened 2023-04-18
Update and rename `osu!dev Discord server` to `osu! Discord server`

Nivalyx:
related to http://github.com/ppy/osu-wiki/issues/9127.

(was intended to be a quick fix at first, but as the whole article is quite lagging behind i decided to expand it to include what's still missing as of now)

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #9310, opened 2023-05-07
Change instances of `osu!dev` to `osu!`, fix broken links and add missing redirect

The-Last-Cookie:
You could call this a continuation of #9196 and #9176.

Many articles contain the link `/wiki/Community/osu!dev_Discord_server`. The `redirect.yaml` file awaits a `/wiki/osu!dev_Discord_server` for redirecting to the new link `/wiki/Community/osu!_Discord_server`. Thus, these articles have an incorrect link, making the link's function cease to exist.

I've manually searched for articles containing the above link and replaced `osu!dev` with `osu!`, so that the link works again.

I'm not sure if the redirection in `redirect.yaml` is still needed, hence I'm adding this task:

- [x] ~~Remove link to `/wiki/Community/osu!dev_Discord_server` in `redirect.yaml`~~

Please tell me if this is not wanted, and instead the link in `redirect.yaml` should be changed.

## Self-check

`SKIP_OUTDATED_CHECK`


Walavouchey (approved):

Walavouchey:
> I'm not sure if the redirection in `redirect.yaml` is still needed, hence I'm adding this task:
> 
> Please tell me if this is not wanted, and instead the link in `redirect.yaml` should be changed.

redirects are placed whenever an article changes location, such that the old link still works. in principle this is not just to account for incorrect links on the wiki, but also links to the wiki from other places.

PR #9333, opened 2023-05-08
Fix mania mapping guide claiming 5K maps can use special style

bdach:
So I was reviewing https://github.com/ppy/osu/pull/23429 today. The gist of that pull is that lazer had wrong copy saying that special style was applicable to 5K maps (while - _to the best of my understanding_ - it actually is applicable to 6K maps, or 5K+1). Generally not being very knowledgeable about mania, I figured I'd crosscheck wiki just to be sure, and lo and behold, this guide also says 5K can use special style? Confusion ensues.

I ended up checking source and I am pretty sure that is false, _to the best of my understanding_.

To be sure I'm not misreading, I also ran stable. Didn't change much in my view of things, it's still 6K that can be special, _as far as I understand_.

So here I go seeing if the wiki will accept this as a fix as well.

---

I know enough French to also attempt to fix that version as well. I can't read German, but the German version [seems to be saying something about 7K and 9K being special](https://osu.ppy.sh/wiki/de/Guides/osu!mania_Mapping_Guide#advanced) which is just grand, but that translation is outdated and I can't write the language so I'm not touching that. The Indonesian translation is also outdated and also has no mention of special style, so I'm not touching that either.

## Self-check

### DIFF
--- a/wiki/Guides/osu!mania_Mapping_Guide/en.md
+++ b/wiki/Guides/osu!mania_Mapping_Guide/en.md
@@ -94,7 +94,7 @@ Alright! Let's drag the `.mp3` of the song we want to map into osu! and a new ma
 | :-- | :-- |
 | Stacking: | This point has no effect for osu!mania so no need to change anything here. |
 | Allowed Modes: | With this point, you change the mode of the editor to the one you want to use for your map. This guide is about osu!mania so we use "osu!mania" of course. Should you choose "All", your editor will be set to osu!. **Changing this option while in an existing osu!mania map will overwrite the map.** |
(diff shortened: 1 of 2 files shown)


The-Last-Cookie:
For your information, the German version needs a rewrite from the ground up, as there are grammar and styling issues all over the place (apart from the article being outdated). So, it's save to say this can be skipped.

Walavouchey:
special styles are entirely a skin.ini setting, which in stable is editable through the mania keyboard layout wizard (only for 6k and 8k)

the game doesn't prevent you from ticking the box in the beatmap setup tab even if the key mode is wrong, nor does it actually do anything with the value. having it displayed in song select or affect the skin layout has been a [requested](https://osu.ppy.sh/community/forums/topics/210488) [feature](https://osu.ppy.sh/community/forums/topics/304999) though

Walavouchey (approved):

PR #9280, opened 2023-05-03
Clean up `Menu`

Sunnyboy971:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #9409, opened 2023-05-18
Further cleanup of `Client/Beatmap_editor/Menu`

The-Last-Cookie:
PR #9280 updated the tables listing the various menu buttons and their functionalities.

Looking at the beatmap editor however, you can notice that there were more changes to be done.

The French translation already has images that are up to date.

I will put the German update into a separate PR.

## Self-check


TicClick (approved):

TicClick:
I've changed the PR's description slightly because everything you put after `---` followed by a blank line will make its way into the website's changelog

PR #9427, opened 2023-05-21
Change custom identifier syntax from `{#` to `{id=`

The-Last-Cookie:
[According to the ASC](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#custom-identifiers), the old syntax with `#` is not used anymore and instead `id=` should now be utilised.

The old syntax was not rendered properly if special characters like `!` were present.

Identifiers are displayed in plain text on the website when the old syntax is used

## Notes

- There was a missing identifier in `Help_centre/Client/zh` (I added it)
- I also updated outdated translations (if there were identifiers present, that is)
- Some french articles still have french identifier. Maybe we should change these to English.

## Self-check


peppy:
Was this done using a script or regex? If so can you include that?

Walavouchey:
@sylphi it should be fine to update it here. if you've made the same changes, then there are no conflicts, and if there are any, then they're probably not hard to resolve.

The-Last-Cookie:
> Was this done using a script or regex? If so can you include that?

For the record, I used a [self-written python script](https://github.com/The-Last-Cookie/python-utilities/blob/master/osu-wiki-search.py) to search for every file containing `{#`. Following that, I used search and replace in each file to change these to `{id=`.

The-Last-Cookie:
@sylphi @Walavouchey already mentioned that after this PR has been merged, you can merge the master branch into the one of your PR (`git pull origin master`, `git checkout asc_formatting` and `git merge master`). If the change is the same (and as far as I can see that is indeed the case), then there will be no conflicts. Otherwise, solving the arising problem should be no problem.

Ephemeralis (approved):

PR #9440, opened 2023-05-23
Set a proper flag for osu!team in `Developers` and `Global Moderation Team` articles

Hiviexd:
https://osu.ppy.sh/users/4341397

image


Walavouchey (approved):
i wonder when this was changed

PR #9443, opened 2023-05-24
[UK] Update `BBCode`

xKuroiUsagix:
Fully updated Ukrainian translation for BBCode wiki page.  Added images for the Ukrainian editor and font-size selector. All tags descriptions were re-translated, since previous translation was outdated and hard to understand. 

## Self-check


TicClick (approved):

PR #9450, opened 2023-05-27
[UK] Add `Rules`

xKuroiUsagix:
## Self-check


wiki/Rules/uk.md
```diff
@@ -0,0 +1,63 @@
...
+5. **Обирайте ваші підписи та аватари з розумом.** Якщо вони анімовані, вони не повинні блимати надто яскраво. Зображення також повинні бути достатньо малого розміру для користувачів з повільнішим інтернет з'єднанням, в ідеалі менше половини мегабайту (500 кілобайт), якщо можливо.
+6. **Не публікуйте рекламу, реферальні посилання, або інший набридливий або настирливий контент.** Такі речі як партнерські програми, або реферальні посилання є веселими для людини, яка це публікує та набридливими для всіх інших. Контент, який автоматично завантажується чи автовідтворюється, підлягає під це правило також.
+7. **Посилання на застосунки або програми сумнівного характеру заборонені.** Простіше кажучи, якщо існують найменші сумніви щодо легальності певного застосунку чи програми, їх не бажано публікувати.
+
+## Правила публікації бітмап
+
+*Зауважте: Ці правила стосуються **всіх опублікованих мап**, а не тільки рейтингових. Щоб ознайомитися детальніше з правилами, які стосуються рейтингових мап, відвідайте сторінку [Рейтингові Критерії](/wiki/Ranking_Criteria)*
```
TicClick, line 41: quick note (same applies to similar lines): at least in Russian you don't need to capitalize regular nouns, which both of the words in "ranking criteria" are -- the title, being a regular part of the sentence, should use sentence case. I imagine it would be the same in Ukrainian

xKuroiUsagix: That's a good point, but to be honest there is no such rule for both Russian and Ukrainian about how to write things like this. My thoughts were next: "Ranking Criteria it's basically the name of the page, so in Ukrainian we use capital letters the same way as in English.".

I personally think it's question about consistency. In Russian translation they use names the same way everywhere and it works fine. I believe there is nothing wrong with translating titles as I do, since it's more about personal preferences. The only thing which is important here it's how consistent translation is.

If you insist on changing these titles, so they look the same way as Russian's - I am fine with that. Since both options are acceptable for me.

TicClick: lemme expand a bit on that -- IMO two things are mixed up here:

1) if an article's name is a regular phrase or a word (not limited to nouns), AND the article is referred to in a regular sentence (casually mentioned as it would have been without a link), then you should use sentence case like you usually would -- see `частина свiту` from the lead section of https://uk.wikipedia.org/wiki/%D0%84%D0%B2%D1%80%D0%BE%D0%BF%D0%B0
2) if an article's name is a proper noun or an acronym (see: OWC), or has special case (see: peppy), or the sentence itself has a fixed format, where the title must be used as-is (see: an example from https://osu.ppy.sh/wiki/ru/Article_styling_criteria/Formatting#%D0%BF%D1%80%D0%B5%D0%B4%D0%BC%D0%B5%D1%82-%D1%81%D1%82%D0%B0%D1%82%D1%8C%D0%B8), then you should use the exact title

RC isn't really special in this regard -- "ranking criteria" is "a lot of ranking-related regulations" and should therefore be used in sentence case when it's a casual mention -- likewise with https://osu.ppy.sh/wiki/VCC, which is a list of regulations from another area and not A Special Name For A Feature Or A Document.

in fact, both articles' own titles should be in sentence case so as to not mislead -- it's something the maintainers have only done for a subset of the osu! wiki corpus to get rid of the stylistic plague which Title Case Naturally Is.

xKuroiUsagix: Thanks for explanation.

TicClick (approved):

PR #9457, opened 2023-05-28
[ID] Update/Retranslate `BBCode`

Nivalyx:
took some liberties every here and there in this one (so that the translation can be better understood by native readers) but then again i've been doing that for a long while lol

hopefully everything's good enough (ó﹏ò｡)


TicClick (approved):

PR #9493, opened 2023-06-04
Update `CWC 2023` rules, schedules and links

LeoFLT:
Adds Semifinals mappool, schedule. Updates the Quarterfinals results with the results from today.

Updates the match rules to account for a replay scenario that was not previously considered:

6. In the case of a team starting a map with less than the optimal number of players on their roster due to an unforeseen circumstance (e.g. a technical issue or unavoidable player absence), the team in question will be granted the right to fill their roster before play commencemences. The remainder of the original roster must stay the same, that is, only player additions are allowed – teams must not swap the players already present in the lobby.


wiki/Tournaments/CWC/2023/en.md
```diff
@@ -467,7 +506,8 @@ The final standings for the Qualifier stage can be found in the following [sprea
...
      - For a ready timer: the referee will issue the `!mp start 10` command, regardless of how many players from each team are present in the lobby, using `!mp kick` on any extra players for each team, starting from the top (i.e. the first valid player combination for each team will be forced to play the pick). The results for such games are to be taken as is.
      - Repeat offenders may receive further sanctions from the tournament managers.
-6. Each team will receive one "tactical timeout" of 2 minutes, to be used as extra time to pick, ban, or ready for a beatmap. *The tactical timeout is optional.*
+6. In the case of a team starting a map with less than the optimal number of players on their roster due to an unforeseen circumstance (e.g. a technical issue or unavoidable player absence), the team in question will be granted the right to fill their roster before play commencemences. The remainder of the original roster must stay the same, that is, only player additions are allowed – teams must not swap the players already present in the lobby.
```
Walavouchey, line 509: tautology

```suggestion
6. In the case of a team starting a map with less than the optimal number of players on their roster due to an unforeseen circumstance (e.g. a technical issue or unavoidable player absence), the team in question will be granted the right to fill their roster before play commences. The remainder of the original roster must stay the same, i.e. only player additions are allowed.
```

Walavouchey (commented):
cool vod links

Walavouchey (approved):

PR #9509, opened 2023-06-07
[DE] Add `Client/Program files`

The-Last-Cookie:
## Self-check


wiki/Client/Program_files/de.md
```diff
@@ -0,0 +1,110 @@
+# osu!-Programmdateien
+
+*Siehe auch: [osu!-Dateiformate](/wiki/Client/File_formats)*
+
+![Die Dateistruktur des Installationsordners von osu! auf Windows und macOS](img/file_structure.jpg "Die Dateistruktur des Installationsordners von osu! auf Windows und macOS")
+
+Die **osu!-Programmdateien** sind eine Reihe von Dateien, die für die Ausführung von osu! erforderlich sind und die verschiedenen Aktivitäten des Benutzers speichern.
```
The-Last-Cookie, line 7: I didn't want to use `verfolgen/aufzeichnen` in the traditional sense of `track`, because it makes it seem like osu! is some data kraken company, which is not true at all.

wiki/Client/Program_files/de.md
```diff
@@ -0,0 +1,110 @@
...
+
+### .exe (Anwendung)
+
+Das Kernstück. Klicke darauf, um es zu starten (gilt nur für Windows). Die .exe-Dateien können ohne Bedenken geöffnet werden, vorausgesetzt, es wurde das osu!-Installationsprogramm von der offiziellen Webseite verwendet.
+
+osu!.exe (Starte osu!)
+
+### .dll (Anwendungserweiterung)
+
+Diese .dll-Dateien sind die Komponenten und Dependencies von osu!.
```
The-Last-Cookie, line 110: I wanted to find an alternative for `Dependency` at first, and I think that some might use `Abhängigkeit`. Though, Dependency is still pretty common, so I'm going with this.

respektive (approved):
looks good 👍

Walavouchey (approved):

PR #9607, opened 2023-06-30
Fix `OWC 2022`, `MWC7K 2023` staff listings

LeoFLT:
Tables were meant to be updated on MWC 7K but ended up on the wrong article (I blame the vscode open editors tab)

## Self-check


Walavouchey:
so no mapper list for owc 2022?

LeoFLT:
> so no mapper list for owc 2022?

Not for the time being at least, we have about 88 maps to go through (some of which are collabs). Gotta make sure nobody is missing

Walavouchey (approved):

PR #9618, opened 2023-07-03
Remove `Additional languages` column and add an `Assigned category` to the NAT bot listing

aceticke:
## Self-check

[Requested on Discord](https://kanonluver727.s-ul.eu/SMQ6kqUv)

Indonesian takes a different approach to translating "official team presence" opting to use "osu!team", thus I did not include it as it would be misleading given the NAT bot only relays messages from the NAT not the osu!team.


Hiviexd:
honestly at this point i'd just have the user column only, assigned category feels redundant here

if consistency is a concern then sure keep it

aceticke:
> honestly at this point i'd just have the user column only, assigned category feels redundant here
> 
> if consistency is a concern then sure keep it

cc @Walavouchey for this one i believe

wiki/People/Nomination_Assessment_Team/en.md
```diff
@@ -146,9 +146,9 @@ The [Nomination Assessment Team group page](https://osu.ppy.sh/groups/7) lists a
 
 ### Bot accounts
 
-| Name | Additional languages | Assigned category |
+| Name | Assigned category |
 | :-- | :-- | :-- |
```
Walavouchey, line 150: need to remove a column in the alignment row too, would be broken as is

Walavouchey:
> honestly at this point i'd just have the user column only, assigned category feels redundant here

at that point you could just remove the table and refer to it with one paragraph at the end. that may give more room to explain what it's used for

aceticke:
Closing, handling with a paragraph instead as Walavouchey mentioned

PR #9620, opened 2023-07-03
Fix case of an article link and remove spaces in `Legal` articles

agatemosu:
## Self-check

`SKIP_OUTDATED_CHECK`


wiki/Legal/Privacy/zh.md
```diff
@@ -4,8 +4,6 @@ legal: true
 
 # osu! 隐私政策
 
-**上一次翻译更新：2023 年 2 月 3 日**
-
```
CloneWith, line 8: From my point of view, these two lines should be kept. We have been using them for a long time to indicate if the translation is up to date. And when the original file is updated we will change this too.

It seems that this is file-specific and no other translations have done this 
 before, so it's better to mention it here.

agatemosu: I don't think it is necessary to put the latest edition of the translation because it already puts the version date of the file below

CloneWith: Resolved, thanks! 😄

peppy (approved):

PR #9658, opened 2023-07-11
Miscellaneous clean-up in a few articles

agatemosu:
The website says [direct messages](https://osu.ppy.sh/community/chat), but the wiki articles say "private messages" and this is the only article that puts DM, but it was recently updated. Should DM be changed to PM in the other articles?

## Self-check


Walavouchey:
whether it's "direct", "private" or "chat" messages doesn't really matter i think

agatemosu:
> whether it's "direct", "private" or "chat" messages doesn't really matter i think

yes, but it is so that the same words are used

Walavouchey (approved):

PR #9677, opened 2023-07-14
Fix incorrect date listing on `Centurions`

Nivalyx:
As mentioned and requested by one of the Centurion mappers (Andrea) via PM :

image


Nivalyx:
For the record as to why the current listed date (`2009-01-11) is wrong :

The article currently refers to [this map](http://osu.ppy.sh/s/11975) as Andrea's earliest Ranked beatmap, which is famous for having its actual Ranked date bugged on the website as shown below.

I've cross-checked this with Andrea himself and I can confirm that Andrea's earliest Ranked beatmap is actually [this one](http://osu.ppy.sh/s/5396) from February 2009.

image

Joehuu (approved):

PR #9685, opened 2023-07-15
Removed unnecessary path in `Upgrading_to_lazer`

Sclafus:
## Self-check

### DIFF
--- a/wiki/Help_centre/Upgrading_to_lazer/en.md
+++ b/wiki/Help_centre/Upgrading_to_lazer/en.md
@@ -10,7 +10,7 @@ The eventual goal is that this version will be released as an update which will 
 
 ## Feature comparison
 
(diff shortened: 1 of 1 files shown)


The-Last-Cookie:
The ASC state the following in [`Sub-article links`](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#sub-article-links):

```md
Wiki links that point to a sub-article should include the parent article's folder name in its link text. See the following example:

*See also: [Beatmap Editor/Design](/wiki/Client/Beatmap_editor/Design)*
```

Ask a maintainer about this but I think it is indented.

Sclafus:
I agree @The-Last-Cookie, but in practice this has never been enforced. Here [Bancho_(server)](https://github.com/ppy/osu-wiki/blob/master/wiki/Bancho_(server)/en.md) as a practical example.
This was my reasoning behind this PR, plus it looks very out of place [in the website](https://osu.ppy.sh/wiki/en/Help_centre/Upgrading_to_lazer#feature-comparison) imho.

Walavouchey (approved):

agatemosu (approved):
I was thinking the same

peppy (approved):

PR #9730, opened 2023-07-18
[DE] Remove outdated translation tag in `Do you really want to ask peppy?`

The-Last-Cookie:
Unfortunately, this was not removed in #9656 

See [here](https://github.com/ppy/osu-wiki/commit/f11d9c05d92e6bde2a45dcc573612853d9ecd418#diff-fd671f3c93d1fe25330defa8bc45a25304c21f14f400c485a41641dcad6c2c4aR3).

## Self-check


TicClick (approved):
good catch, I'll try being more attentive next time

PR #9741, opened 2023-07-21
Move Muzukashii pattern length guideline to rules in osu!taiko ranking criteria

Hiviexd:
discussion thread: https://osu.ppy.sh/community/forums/topics/1791571?n=1

german translation was handled by a german nat


cl8n (approved):

TicClick (approved):
think it's been around for long enough to warrant a merge

PR #9767, opened 2023-07-25
Update `Help Centre/Account` with new flag change policy

Hiviexd:
according to initial rules section in https://github.com/ppy/osu-web/issues/10382

### DIFF
--- a/wiki/Help_centre/Account/en.md
+++ b/wiki/Help_centre/Account/en.md
@@ -11,6 +11,7 @@ tags:
   - device
   - username
   - deletion
+  - flagchange
 ---
 
 # Account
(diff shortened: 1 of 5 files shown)


wiki/Help_centre/Account/en.md
```diff
@@ -180,13 +181,13 @@ Your profile will show all your previous purchased name changes, but we only sel
 
 ### The flag/country on my account profile is wrong! {id=wrong-flag}
 
-**Unfortunately, we cannot change this for you.**
-
-The country associated with your osu! profile is based on the geographical location of the connection you used to register the account and is permanently assigned during account creation.
+**Yes, but only under certain circumstances.**
```
TicClick, line 184: this needs to be an answer to the heading, which contains a question different from the one you are answering

Hiviexd: oops didn't pay attention

cl8n (dismissed):
also might make sense to write the rules in second-person (e.g. "you" instead of "a user") since that's what a lot of the rest of this article does.

fine to me either way

wiki/Help_centre/Account/en.md
```diff
@@ -180,13 +183,13 @@ Your profile will show all your previous purchased name changes, but we only sel
...
 
-The country associated with your osu! profile is based on the geographical location of the connection you used to register the account and is permanently assigned during account creation.
+A user is eligible for a flag change when:
 
-In the past, we offered the ability for users to change their account country by providing proof of residence documents, but after countless instances of abuse by unscrupulous players, we were left with no option but to stop offering such services.
-
-Please understand that we cannot adjust this under any circumstances, regardless of people you've heard of or might know who have had it done before.
+- There is at least 6 months of login data since the creation of the account, or since the last flag change
+- At least 4 of the 6 months have logins from the same country
+- The user is not registered for any ongoing tournaments
```
peppy, line 192: I really didn't write these on the wiki for them to be plastered everywhere. Simplify it and don't include the actual conditions, they may/will change and will not be announced.

peppy (changes requested):

wiki/Help_centre/Account/en.md
Hiviexd: does something like this work?
`If you have been consistently logging in from a different country over a long period of time, you will be prompted to change your account country in the [account settings page](https://osu.ppy.sh/home/account/edit).`

peppy: `If you have been consistently connecting from a country that differs from your account's country over a long period,...` sounds better

Walavouchey (approved):

peppy (approved):

PR #9811, opened 2023-08-01
Update `SEA Summer Suiji Showdown 3`'s badge design

Nivalyx:
as requested by the winners and the tournament organizers

image


Walavouchey:
you'll need to add a query string to the image urls on the relevant page (e.g. `?1`), otherwise it'll use the old cached image

cl8n (approved):

PR #9820, opened 2023-08-02
Update all Twitter links and references to X

githacki:
## Self-check

This PR replaces all Twitter links with the new X redirects, and all Twitter references in english files with "X (formerly Twitter)"


NiceAesth:
is there really a point to any of this, especially the links since they redirect to twitter anyway? just an annoyance to be redirected twice

Redo7:
Can we not stroke billionaire ego and validate their shitty decisions? Thanks.

TicClick:
+1, feels unnecessary, especially with all the turmoil surrounding the rename, and the fact that "X" is way less recognizable

githacki:
> +1, feels unnecessary, especially with all the turmoil surrounding the rename, and the fact that "X" is way less recognizable

which is why i added "(formerly Twitter)" in english files

Walavouchey:
apart from what has been said above, if this were to happen we would also not remove twitter from front matter tags.

let's reconsider at another point, maybe once "x" is common parlance or when osu!web switches over.

0xallie:
As much as I hate Elon Musk (especially for his transphobia and platforming nazis), I think it makes sense to change it to "X (formerly Twitter)" just to avoid any confusion, including for future users who didn't really know Twitter (though Twitter is extremely well known so that is probably gonna be a rare thing until the next generation, if they don't reverse course until then). However it is probably best to wait until they've fully switched over to x.com and got rid of the Twitter branding in more places.

Ephemeralis:
For the sake of clarity and readability, I can't really see us referring to it as anything other than Twitter for likely the remaining length of the platform's history, or until such a point that 'X' takes over as common vernacular for the service (highly unlikely).

Issue #1563, opened 2018-07-31
FAQ needs a rewrite


Joehuu:
https://osu.ppy.sh/help/wiki/FAQ

I probably won't have a question about the game modes let alone the origin of the game. It was also copied and pasted from the game modes article. There are other stuff ("Common issues and questions" section on FAQ?) which I won't say, but it definitely needs a update.

Gamingzone123:
works towards: https://github.com/ppy/osu-wiki/issues/962

PR #9845, opened 2023-08-06
Fix minor grammatical error in `FAQ`

CrazyCSIW6:
## Self-check

- TicClick: `SKIP_OUTDATED_CHECK`


TicClick (approved):
while the edit is correct, the whole article has long been scheduled for a rewrite, so don't worry too much about it: https://github.com/ppy/osu-wiki/issues/1563

PR #9846, opened 2023-08-06
Use tables for bit flag information in `.osu (file format)`

ekgame:
## Self-check

The sections about bit flags were confusing, especially to people unfamiliar with binary.
- Previously, the information was 0-indexed, but now it's 1-indexed - it makes more sense to say "1st bit" than "0th bit".
- The two sections for type information were joined into the same table - it makes sense to have the bits in order, even if their meaning falls out of order.
- The object types table is raw HTML to make use of `rowspan` for colour hax information. I'm not sure how this is handled when it's converted to HTML on the website. I hope it's fine, because Markdown does not support `colspan`/`rowspan` (as far as I know).


Walavouchey:
the wiki doesn't support html

ekgame:
Any suggestions for how to handle the rowspan? 
image

bdach:
> Previously, the information was 0-indexed, but now it's 1-indexed - it makes more sense to say "1st bit" than "0th bit".

I disagree with this. Zero-indexing bits is basically done everywhere where bits are used in computing. Every single [packet diagram](https://www.wireshark.org/docs/wsug_html_chunked/ChUsePacketDiagramPaneSection.html) I've seen is zero-indexed. Zero indexing also directly translates into the bit shifts you need to perform to get the mask (zeroth bit is `1 << 0`, first bit is `1 << 1`...)

Providing values for the bit flags is likely helpful, but reindexing the bits isn't, to me at least.

> Any suggestions for how to handle the rowspan with Markdown?

Do the "color hax" bits need to go into that table at all? Can't normal flowing text underneath the table be used to convey that information?

Walavouchey:
the page is intended for developers, where 0-indexing is the norm. the "decimal value" column also isn't necessary

again, html isn't supported, so you'd have to put the information on the same row

agatemosu:
Another option is to put "Same as above"

ekgame:
The reason I decided to try an update this part of the format documentation was because of a conversation I had where I had to explain that "3" actually means the 4th bit, which is actually 8 in the .osu file. I thought the leap of 3 (in the docs) => 4 => 8 (in the actual file) was a bit confusing for somebody less familiar familiar with bit flags. Even I remember myself looking at it back in the day confused about this part, opting to make test beatmaps to understand how it actually works. Even now, when I wanted to refresh my memory, I had to look at a [parsing implementation](https://github.com/nojhamster/osu-parser/blob/master/index.js#L168) when it finally made sense.

The reason I want to keep the "decimal value" column is to make it easier to look at a type value in the file and do the math of what the value actually means. Even when you have a pretty good sense for binary, it's unnecessary mental work to convert 0-based index to a decimal value.

I'm not married to the idea of 1-based bit index, so I reverted it back to 0-based. I also placed the "colour hax" information below the table as an explanation for the table entries.

image

Walavouchey:
> The reason I want to keep the "decimal value" column is to make it easier to look at a type value in the file and do the math of what the value actually means. Even when you have a pretty good sense for binary, it's unnecessary mental work to convert 0-based index to a decimal value.

it's "unnecessary mental work" because the decimal value is not relevant for anything, and for the individual colour hax bits, utterly meaningless

Walavouchey (approved):

PR #9865, opened 2023-08-08
Outdate `Beatmap discussion`

aceticke:
Markdown has been added to the discussion so should be updated to reflect this, furthermore, inaccuracies such as "Tags display important roles below a users name. Only roles relevant for the beatmap discussion are shown." need to be addressed. Images could also do with updating to the newer designs.

## Self-check


agatemosu:
I don't think it's necessary to update the images, the only thing that changed was the font. And there is a pending redesign

TicClick:
a lot of things could be done better, images included, so it would benefit from a full rewrite rather than speficic edits

TicClick (approved):

PR #9893, opened 2023-08-14
Revert "Fix typo in `Skinning/skin.ini`"

heipizhu4:
@ChickenGoldSS 
After some discussion, we've found that we have made a mistake here. The default of combo prefix is indeed score, not combo. I apologize for the mistake I have made.


ChickenGoldSS (approved):
sorry,sad...

peppy:
Yep, this is not a typo.

peppy (approved):

PR #9904, opened 2023-08-15
Update `osu!supporter` with new friend limits and other clean-up

Walavouchey:


wiki/osu!supporter/en.md
```diff
@@ -127,28 +129,30 @@ The remaining osu!supporter duration, as well as total amount of contributions,
...
+The recipient is not notified of who gave them the tag, but when gifting osu!supporter, an optional message can be attached to be sent alongside the notification email.
 
-### Conclusion
+### Confirmation
 
 After the transaction is completed, there will be a new event in the `Recent Activity` section in the profile of both the giver and the recipient:
 
 - `{username} has supported osu! - thanks for your generosity!` if it is the first time the user gets osu!supporter or gifts it to someone else.
 - `{username} has once again chosen to support osu! - thanks for your generosity!` if the user has been an osu!supporter or gifted it before.
 - `{username} has received the gift of osu!supporter!` in case the tag was gifted.
```
aceticke, line 152:
```suggestion
- `{username} has received the gift of osu!supporter!` in the case where the tag was gifted.
```

could be nicer wording

wiki/osu!supporter/en.md
```diff
@@ -127,28 +129,30 @@ The remaining osu!supporter duration, as well as total amount of contributions,
...
+:::
+
+osu!supporter may also be gifted to another player by entering their username below the user card, or by clicking `Gift supporter tag` on their profile page. More people can be added by going back to the store page or someone else's profile and repeating the process.
+
+The recipient is not notified of who gave them the tag, but when gifting osu!supporter, an optional message can be attached to be sent alongside the notification email.
 
-### Conclusion
+### Confirmation
 
 After the transaction is completed, there will be a new event in the `Recent Activity` section in the profile of both the giver and the recipient:
```
aceticke, line 148: users can also choose to hide this from their own activity in the case where they are the one who purchased it, worth a mention?  
image

aceticke: maybe something like:

```suggestion
After the transaction is completed, there will be a new event in the `Recent Activity` section in the profile of both the giver and the recipient:[^hide-event]

[^hide-event]: The giver can choose to hide this event from their `Recent Activity` on their profile when at checkout.
```

wiki/osu!supporter/en.md
Walavouchey: not sure about "nicer" but it's more words when fewer suffice

PR #9946, opened 2023-08-20
[ES] Add `User title`

agatemosu:
## Self-check


TicClick:
(my god, it never merged itself)

PR #9948, opened 2023-08-20
Update `MWC 4K 2023` for the RO16 stage

LeoFLT:
Adds all Round of 16-related information (matchups/schedules/map pool); adds Round of 32 results.

## Self-check


Walavouchey (approved):
very cool addition of vod links

PR #9967, opened 2023-08-23
[EN/DE] Add interview about osu!lazer's sound design to `History of osu!/2022`

The-Last-Cookie:
I decided to include the translation because the change is so small, though it would be not a problem to move this to a separate branch if necessary.

## Self-check


respektive (approved):
:ok_hand:

peppy:
Probably okay, but I dunno where we draw the line between "history" and "listing every news post". @Walavouchey thoughts on this?

Walavouchey:
i also disagree with adding individual news posts like this. i'm not sure what makes this history-worthy - would some youtube videos or tweets have similar value?

PR #9973, opened 2023-08-24
Update game mode icons

Hiviexd:
old icons are very outdated by current standards, so i replaced them with ones that the game currently use + match the diff icons style in `/wiki/shared/diff`

preview:

image


Walavouchey:
they're so lifeless <.<

image

image

Hiviexd:
idk what to tell u the old ones are truly outdated.

here's an idea, how about i color the new ones based on each mode's official color? aka osu! would be pink, taiko green, catch blue, mania purple

Walavouchey:
certainly better, but they're rather dark

image

Walavouchey (approved):
will wait for more eyes on this

TicClick:
the colours look good to me (dunno where they come from but will leave that up to you)

Redo7:
Wouldn't say the current icons are outdated by any means. They have coherent visuals aligning with the overall theme of the site very well. I can't tell if swapping them to just the icon itself provides any meaningful change, but it does seem it follows the trend of overly simplifying things for no apparent reason. Even when given some identity by coloring them, it still seems like a downgrade.

cl8n:
can the color go in the background instead? I think the white icons look better since they match with other inline text (which is where this icons are primarily used in the wiki)

and yeah agree with everything @Redo7 said^

Redo7:
That's what I was thinking too, but not like their current state is in need of any change imo

Hiviexd:
personally i would really prefer to be consistent with the style that's already being used by the website, alongside the existing diff icons used by the wiki here https://github.com/ppy/osu-wiki/tree/master/wiki/shared/diff, hence why i made this PR

Walavouchey:
for the colour scheme, refer to the main mascots and the world cup motifs

@cl8n the problem may just be that the colours are too dark and stand out less against the background. could be worth testing, although imo they're good as they are (@Tanza3D @arflyte ?)

@Redo7 for context, the current game mode icons use a design from back when the website looked like this:



the difficulty icons were later updated when the website design changed, but the game mode icons lagged behind. this pr intends to make them align more with the current design

i suppose you're right that the design has been simplified. on the old website, the icons looked like this:

image

cl8n:
I just thinking you're reading kinda far into this to call it "outdated". it's literally same icon just with a backing color unless I'm missing something else. it has good reason to not match exactly with some other usages on osu-web here because it's not used in the same contexts (icon for text like said before), or even for the same reasons(the color denotes difficulty on the website)

I can justify to myself why this change isn't "that bad" or w/e but I just see no need for the downgrade when there is no actual issue to begin with.

peppy:
No idea where these colours are from, but we've never used them anywhere.

peppy:
@arflyte please advise

Walavouchey:
they can't have `@2x` versions because that would make them bigger in size

peppy:
> they can't have `@2x` versions because that would make them visually bigger in size

I don't get this. That's not how it works.

arflyte:
I'm confused with the request. You just wanna add colours to the ruleset icons? If so, just add a set of coloured icons, not replacing them.

Walavouchey:
the intention here is very much not to have separate icons, but to update the current ones

regardless of whether this goes through, can you confirm what the game mode colours are? (used in the world cup banners, mascot designs, some badges etc)

peppy:
I think it's best to open an issue to discuss this if we want to change the design. This isn't in a state it's going to get merged.

PR #9988, opened 2023-08-30
Add a tag to `Visual content considerations` article

Hiviexd:
Visual content considerations are commonly abbreviated as VCC so this should help with searchability a lot (I hate typing the full page name in search lol)

`SKIP_OUTDATED_CHECK` because it doesn't make sense to add an english abbreviation to other languages

### DIFF
--- a/wiki/Rules/Visual_content_considerations/en.md
+++ b/wiki/Rules/Visual_content_considerations/en.md
@@ -1,3 +1,8 @@
+---
+tags:
+  - vcc
+---
+
 # Visual content considerations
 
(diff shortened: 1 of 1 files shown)


peppy (approved):

Walavouchey:
note that you can also type https://osu.ppy.sh/wiki/vcc into the url bar and you'll get redirected to the correct article (see [redirect.yaml](https://github.com/ppy/osu-wiki/blob/master/wiki/redirect.yaml))

> `SKIP_OUTDATED_CHECK` because it doesn't make sense to add an english abbreviation to other languages

translations are required to have all english tags as well as any applicable translations of them ([asc](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#tagging-articles))

PR #10011, opened 2023-09-04
Update `People/Completionists`

Walavouchey:
other than the date being wrong, it turns out the query used for checking EEEEEEEEEEEEEEE's completion count [didn't account for approved beatmaps](https://discord.com/channels/188630481301012481/218677502141399041/1148291035466838096) (which are in the ranked section nowadays)

the numbers for the other completionists are sourced from osu!alternative anyway, which is a database (or rather, a discord bot that provides an interface for a database) sourced from data.ppy.sh dumps, which people have been using to track completion progress


cl8n (approved):
(assuming that the new date and diff count is accurate. I didn't read that convo or check it myself)

PR #10015, opened 2023-09-07
Update Hivie's role and add tag in `osu! team`

Hiviexd:
this one kinda slipped thru the cracks


cl8n (approved):
even working on loved i cant tell what would differentiate a "manager" from anyone else cuz most people there contribute to management in some way. but if you just wanted to mention this on your entry for the team page, i guess it's okay :thinking:

PR #10055, opened 2023-09-15
Fix notbakaneko's country flag

cl8n:
gives me errors every time while checking against online info


peppy:
Not sure how that happened but fixed now.

PR #10067, opened 2023-09-20
Replace the `task-list-completed` GitHub app with a GitHub action

TicClick:
unfortunately this seems necessary due to https://github.com/stilliard/github-task-list-completed/issues/25 -- the check coming from the `task-list-completed` app occasionally becomes stale, only displaying the "Waiting for status to be reported" placeholder, and whether or not you can break its torpor is pure luck (it's probably something on GitHub's end, but I personally can't care anymore)

https://github.com/Shopify/task-list-checker is the new action. keep in mind that it only considers the PR's description, while [`task-list-completed`](https://github.com/marketplace/task-list-completed) reads all comments and tests them for checkboxes

@peppy please:

- [ ] uninstall https://github.com/marketplace/task-list-completed from the repo
- [ ] change `GITHUB_TOKEN` access to `Read and write permissions` in the repo settings → actions

### DIFF
--- a/.github/workflows/task-list-checker.yml
+++ b/.github/workflows/task-list-checker.yml
@@ -1,0 +1,20 @@
+name: GitHub Task List Checker
+on:
+  pull_request:
+    branches:
+      - master
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
+jobs:
+  task-list-checker:
+    runs-on: ubuntu-latest
+    permissions:
+      statuses: write  # required for https://docs.github.com/rest/commits/statuses#create-a-commit-status
+    steps:
+      - name: Check for incomplete task list items
+        uses: Shopify/task-list-checker@main
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/task-list-checker.yml
+++ b/.github/workflows/task-list-checker.yml
@@ -8,11 +8,11 @@ on:
       - reopened
       - synchronize
       - edited
+permissions:
+  statuses: write
 jobs:
   task-list-checker:
     runs-on: ubuntu-latest
-    permissions:
-      statuses: write  # required for https://docs.github.com/rest/commits/statuses#create-a-commit-status
     steps:
       - name: Check for incomplete task list items
         uses: Shopify/task-list-checker@main
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -10,6 +10,8 @@ on:
       - synchronize
       - edited
 
+permissions: {}
+
 jobs:
   ci:
     name: changed files
--- a/.github/workflows/manual-sync.yml
+++ b/.github/workflows/manual-sync.yml
@@ -14,6 +14,9 @@ on:
         required: true
         default: "false"
 
+permissions:
+  contents: write
+
 jobs:
   sync:
     name: Execute Sync Task
--- a/.github/workflows/welcome-new-contributors.yml
+++ b/.github/workflows/welcome-new-contributors.yml
@@ -4,6 +4,9 @@ on:
   pull_request_target:
     types: [opened]
 
+permissions:
+  pull-requests: write
+
 jobs:
   welcome-new-contributor:
     runs-on: ubuntu-latest


TicClick:
marking as a draft because it doesn't seem to work lol https://github.com/ppy/osu-wiki/actions/runs/6250016020/job/16967940347?pr=10067 neither on empty items nor on complete ones

peppy:
What's the write access for in this case?

Also we may want to propagate this change to other repos, like lazer.

cl8n:
the permission is for writing commit statuses

.github/workflows/task-list-checker.yml
```diff
@@ -0,0 +1,20 @@
...
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
+jobs:
+  task-list-checker:
+    runs-on: ubuntu-latest
+    permissions:
+      statuses: write  # required for https://docs.github.com/rest/commits/statuses#create-a-commit-status
```
cl8n, line 15: permissions should be restricted as much as possible for all other workflows as well

also this can be defined at workflow scope which may be more convenient https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

cl8n (changes requested):

.github/workflows/task-list-checker.yml
cl8n: also I don't think your comment is correct. this is good practice of course but shouldn't be required with a permissive token https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

TicClick: the comment is a leftover from past attempts to make it work

TicClick:
tested on my repository that our actions still work with limited permissions:
- [`welcome-new-contributor`](https://github.com/TicClick/osu-wiki/actions/runs/6263287802/job/17007340247?pr=25#logs)
- [syncing a feature branch to `master`](https://github.com/TicClick/osu-wiki/actions/runs/6263237261/job/17007169373)
- [wiki checks](https://github.com/TicClick/osu-wiki/actions/runs/6263287749/job/17007339888?pr=25#logs) (was able to read a list of commits)

---

[`task-list-checker`](https://github.com/TicClick/osu-wiki/actions/runs/6263287734/job/17007339876?pr=25) also works 

<details>
<summary>here's what the token was required for</summary>

image

</details>

cl8n (approved):

TicClick:
JFYI there's an alternative solution to the permissions issue:

1. make a classic[^1] [personal token](https://github.com/settings/tokens) with a `repo:status` access scope and unlimited expiry date
2. put it into `ppy/osu-wiki`'s secrets and change `secrets.GITHUB_SECRET` in this PR to `secrets.{new secret name}`

the downside is that the new action will fail in pull requests made to forks, unless the forks' owners generate their own tokens
(also that the token may be eventually revoked and then stuff breaks, but that's easily noticeable)

[^1]: fine-grained tokens can't be issued for use longer than a year

sr229:
> JFYI there's an alternative solution to the permissions issue:
> 
> 1. make a classic[1](#user-content-fn-1-940863b3886305f0cdd032cbf8d0b20e) [personal token](https://github.com/settings/tokens) with a `repo:status` access scope and unlimited expiry date
> 2. put it into `ppy/osu-wiki`'s secrets and change `secrets.GITHUB_SECRET` in this PR to `secrets.{new secret name}`
> 
> the downside is that the new action will fail in pull requests made to forks, unless the forks' owners generate their own tokens (also that the token may be eventually revoked and then stuff breaks, but that's easily noticeable)
> 
> ## Footnotes
> 1. fine-grained tokens can't be issued for use longer than a year [↩](#user-content-fnref-1-940863b3886305f0cdd032cbf8d0b20e)

Just to add to this, these tokens are ephemeral by default which reduces the likelihood of a supply chain attack, so I believe its due diligence to audit the actions we use and precisely tag the version we want to prevent any kind of misuse.

TicClick:
good point -- what I suggested as a replacement for task list handling isn't tagged..

TicClick:
(hopefully permanently) closing as per https://discord.com/channels/188630481301012481/218677502141399041/1157343244322078770

> [5:48 PM] peppy: hmm
> [5:49 PM] peppy: i may have fixed the todo cheker app not working
> [5:49 PM] peppy: please lemme know if it works better

PR #10072, opened 2023-09-21
Update `MWC 4K 2023` Grand Finals schedule

LeoFLT:
Also updates a few of the Grand Finals picks' mappers.

## Self-check


cl8n:
I thought that map format always has the set host in parenthesis just to disambiguate what set you're talking about

is there already precedent for writing the map's actual creators in there?

LeoFLT:
> I thought that map format always has the set host in parenthesis just to disambiguate what set you're talking about
> 
> is there already precedent for writing the map's actual creators in there?
> 
> e: yeah all the mappools below this just use set host from what I can tell so I don't think you should change this by itself.

It was changed since in my eyes crediting every mapper that worked on the songs for this round was more important than following unspoken standards, as map sets are already hyperlinked everywhere they appear in.

The intent was to have every set have the proper mappers listed, not just the set host (as some mappers come from other games to help without having an osu! account). I will revert the change for now though since it's fairly inconsequencial.

PR #10077, opened 2023-09-21
CI: only warn about commit fetching failures when they really happen

TicClick:


Walavouchey (approved):

Walavouchey:
tested in https://github.com/Walavouchey/osu-wiki/pull/31 ([action](https://github.com/Walavouchey/osu-wiki/actions/runs/6264037365/job/17009804099?pr=31)). this also prevents it from running curl 3 times when it has already succeeded

PR #10111, opened 2023-09-27
Update "User" section for `skin.ini`

Walavouchey:
i don't think this part should be in the skin.ini article since it has nothing to do with skin.ini files but idk where to put it without restructuring the skinning articles

[reported on discord](https://discord.com/channels/188630481301012481/188630652340404224/1156491646041595945)

### DIFF
--- a/wiki/Skinning/skin.ini/en.md
+++ b/wiki/Skinning/skin.ini/en.md
@@ -109,13 +109,7 @@ If your skin folder does not contain a `skin.ini` file, it will default to this 
 
 ### User
 
-**Not a version, but always force [latest](#latest)**
-
-- **Never** use this method when distribute skins!
-- The skin folder **must** be named `User`.
-- Use this method if you only want to change a few things (e.g. cursor or numbers, etc).
-- Does not require the `skin.ini` file.
-- This folder will always force `Version: latest`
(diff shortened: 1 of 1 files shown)


wiki/Skinning/skin.ini/en.md
```diff
@@ -109,13 +109,9 @@ If your skin folder does not contain a `skin.ini` file, it will default to this
 
 ### User
 
-**Not a version, but always force [latest](#latest)**
+This is not version, but a special skin folder that is created when the default skin is modified, e.g. by dragging in a custom menu background into the game window (with [osu!supporter](/wiki/osu!supporter#extra-skinnable-elements)) or when attempting to modify osu!mania keybinds for the default skin from the options.
```
peppy, line 112:
```suggestion
This is not a version, but a special skin folder that is created when the default skin is modified, e.g. by dragging in a custom menu background into the game window (with [osu!supporter](/wiki/osu!supporter#extra-skinnable-elements)) or when attempting to modify osu!mania keybinds for the default skin from the options.
```

pointed out by twitch chat..

peppy (changes requested):

peppy (approved):

PR #10115, opened 2023-09-28
[DE] Add `Beatmap/Genre and language`

The-Last-Cookie:
The two genres `hip hop` and `rock` link Wikipedia articles in their description. I had a quarrel on how to proceed with those while translating. Originally, I saw these options:

1. translate literally and link to english article
2. translate literally and link to german article
3. translate in free form

The third option is very likely not wanted and should only be used as a backup.

While both Wikipedia articles[^wiki] essentially say the same, they do it with very different words (see the third option) and I'm not sure if the second option makes sense for something which is meant as a citation. Despite this, I still decided to go for this option, because it is probably the most user friendly one.

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

[^wiki]: I'm refering to the German and English articles for each genre.

PR #10117, opened 2023-09-28
Fix citation in `Beatmap/Genre and language`

The-Last-Cookie:
Noticed this in #10115 

Looking at the wikipedia article, it uses `syncopated` instead of `unsyncopated` now.

## Self-check


Walavouchey (approved):
shouldn't these be footnotes?

peppy (approved):

PR #10127, opened 2023-10-01
Update outdated trivia point in `Profile badge`

mmattbtw:
## Self-check


chromb:
wouldn't this be better to update rather than nuke entirely?

mmattbtw:
> wouldn't this be better to update rather than nuke entirely?

agreed; "Badges can link to other web pages for more context, like the tournament's forum page or wiki entry."

Ephemeralis (approved):

PR #10138, opened 2023-10-06
Allow key modes of 12-18K in osu!mania RC

maxuskazuma:
RC proposal can be found here: https://osu.ppy.sh/community/forums/topics/1787329?n=1

To give a bit of background, we already working with this roughly 3 months, from asking peppy's permission via email and double asking permission via sit-in and also internal discussion with testing it in lazer client as well, which we got the permission from peppy to qualify 12-18K map as part of the test to ensure nothing breaks before making any PR for changing keymodes rule. 12-18K already been test qualified and ranked now, and there is no issue found from them which already been clarified by some people that already play the map.


wiki/Ranking_criteria/osu!mania/en.md
```diff
@@ -58,7 +58,7 @@ Overall rules and guidelines apply to every kind of osu!mania difficulty. Rhythm
 - **Two notes cannot be placed in a single column at the same timestamp.**
 - **No column can be left empty.** If you want to use fewer keys, change the number of keys in the Song Setup.
 - **There must not be more than 6 notes pressed at any given moment in Insane or lower difficulties.** Using more than 6 notes must also follow a reasonable spread to the next lower difficulty. This rule does not apply to ends of long notes, as they are released, not pressed.
-- **Beatmaps can only use 4 to 10 keys.** Anything else is not supported for the Ranked section.
+- **Beatmaps can only use 4–10, 12, 14, 16, or 18 keys.** Anything else is not supported for the Ranked section.
```
cl8n, line 61:
```suggestion
- **Beatmaps can only use 4–10, 12, 14, 16, or 18 keys.**
```
seems unnecessary to me, this should be obvious because it's the ranking criteria.

wiki/Ranking_criteria/osu!mania/en.md
```diff
@@ -58,7 +58,7 @@ Overall rules and guidelines apply to every kind of osu!mania difficulty. Rhythm
...
 - **If multiple key modes are present in a single beatmap, the key mode must be denoted in all difficulty names. Otherwise, the key mode must not be denoted.**
 - **Beatmaps using the special N+1 style of mapping must be denoted as such under the `Advanced` tab of song setup in the editor and "N+1" must be added to the tags. For beatmaps containing multiple key modes, difficulties using this style must be denoted using N+1 notation (e.g. using 7K+1 instead of 8K on difficulty names).**
```
cl8n, line 63: I don't feel strongly about this either way, but you might want to reword this to be more generic than "N+1" now that you could rank DP special styles

cl8n (changes requested):
nice work, this is a really cool change to see

wiki/Ranking_criteria/osu!mania/en.md
Walavouchey: i'm not sure what that is or if/where it's mentioned in the rc

can you be a bit more concrete? are you referring to another proposal?

aryoadhi: N+1 purpose was to differentiate two distinct map styles sharing same key count in 4-10K. Mostly it refers to differentiating 7K+1 and 8K, due to both are quite popular modes to map on these keycount, but also covers anything else like 5K+1 and 6K and novel styles like 4K+1. https://osu.ppy.sh/community/forums/topics/1694238

Due to how broad the playstyles offered by 12K+, the ruling about key count in diff names can be enforced for 12K+, which also extends to mentioning the playstyle it offered. N+1 should remain there if there is only one special key (on left column) present in the map. 4-10K descriptor should remain unmentioned unless multiple key modes present.

Walavouchey: as an update people seem to be working on an updated wording to this rule and a new [supplementary article](https://discord.com/channels/188630481301012481/218677502141399041/1161131392701309038) that this rule can link to

wiki/Ranking_criteria/osu!mania/en.md
```diff
@@ -19,8 +19,13 @@ This set of **osu!mania ranking criteria** lays out [rules and guidelines](/wiki
...
+- **Column**: Vertical slices of the [playfield](/wiki/Game_mode/osu!mania#playfield) where notes are placed. Note that [editor timestamps](/wiki/Modding/Timestamp) refer to columns by their [zero-indexed](https://en.wikipedia.org/wiki/Zero-based_numbering "\"Zero-based numbering\" on Wikipedia") position starting from the left.
 - **Key mode**: The number of columns present in a difficulty.
+- **Playstyle**: A mapper's intended style of [key bindings](/wiki/Game_mode/osu!mania#controls) or hand positions used to play a difficulty.
+- **Standard playstyle**: The most common type of playstyle for key counts 10 or below, where no column is of particular significance, and the player is expected to control each key with a unique finger.
+  - Examples of standard playstyles: 4K, 7K, 10K
+  - Examples of non-standard playstyles: 7K+1, 10K2S, 14K DP
+- **N+1 playstyle**: A specific type of playstyle where the leftmost column is mapped independently from the rest of the columns, which otherwise form a standard playstyle. Playstyles may also be considered "N+2" if they add a second independently-mapped column.
```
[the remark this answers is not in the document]
cl8n: from the perspective of the mapper it always goes on left, then the player chooses if it goes on the left or right. this definition and the rule that refers to it are only really here to help explain conventional usage of the beatmap setting, which is also only used to mark a special column on the left. it sounds weird because it is weird...

cl8n:
also should've mentioned earlier but I'm keeping my block on this pr cuz I'm waiting on Maxus to check my updates. those were put together mostly by me, Konomi, and Ryu Sei in the 10k+ discord server

wiki/Ranking_criteria/osu!mania/en.md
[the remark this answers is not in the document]
aryoadhi: Since beginning, osu!mania stable editor always treats leftmost column as special column on N+1 maps, despite honoring mapper's skin preference for coloring purposes.

wiki/Ranking_criteria/osu!mania/en.md
```diff
@@ -19,8 +19,13 @@ This set of **osu!mania ranking criteria** lays out [rules and guidelines](/wiki
 ### Common terms
 
 - **BPM**: Tempo of a song measured in beats per minute.
-- **Column**: Vertical lines on the playfield where notes are placed. The editor uses `0|1|2|3|4|5|6|7|8|9` for timestamp labels.
+- **Column**: Vertical lanes of the [playfield](/wiki/Game_mode/osu!mania#playfield) where notes are placed. Note that [editor timestamps](/wiki/Modding/Timestamp) refer to columns by their zero-indexed position starting from the left.
 - **Key mode**: The number of columns present in a difficulty.
+- **Playstyle**: A mapper's intended style of [key bindings](/wiki/Game_mode/osu!mania#controls) or hand positions used to play a difficulty.
+- **Standard playstyle**: The most common type of playstyle for key counts 10 or below, where no column is of particular significance, and the player is expected to control each key with a unique finger.
```
MChecaH, line 25: `Key modes` and `Standard playstyles` would end up meaning the same thing. Not technically on paper, but with how key modes are understood by the community, it would have that overlap.

Ideally, `Key modes` by definition should already encompass the definition of `Standard playstyles`.

wiki/Ranking_criteria/osu!mania/en.md
```diff
@@ -58,13 +63,14 @@ Overall rules and guidelines apply to every kind of osu!mania difficulty. Rhythm
...
 - **There must not be more than 6 notes pressed at any given moment in Insane or lower difficulties.** Using more than 6 notes must also follow a reasonable spread to the next lower difficulty. This rule does not apply to ends of long notes, as they are released, not pressed.
-- **Beatmaps can only use 4 to 10 keys.** Anything else is not supported for the Ranked section.
-- **If multiple key modes are present in a single beatmap, the key mode must be denoted in all difficulty names. Otherwise, the key mode must not be denoted.**
-- **Beatmaps using the special N+1 style of mapping must be denoted as such under the `Advanced` tab of song setup in the editor and "N+1" must be added to the tags. For beatmaps containing multiple key modes, difficulties using this style must be denoted using N+1 notation (e.g. using 7K+1 instead of 8K on difficulty names).**
+- **Beatmaps can only use 4–10, 12, 14, 16, or 18 keys.**
+- **Each difficulty's name must denote its [playstyle](#common-terms), but only if its key count is above 10 or multiple playstyles are present in the beatmap.**
```
MChecaH, line 67: Few things. This should be reworded in a way that...
- Any Non-Standard playstyle
- Multiple key modes in a beatmapset
- Different playstyles of one same key mode

...need to add their key mode/playstyle in their difficulty name.

wiki/Ranking_criteria/osu!mania/en.md
```diff
@@ -58,13 +63,14 @@ Overall rules and guidelines apply to every kind of osu!mania difficulty. Rhythm
...
+- **Each non-[standard playstyle](#common-terms) present in a beatmap must be denoted in the beatmap's tags.**
+- **Difficulties using an [N+1 playstyle](#common-terms) must enable the `Use special style (N+1 style) for mania` option under the `Advanced` tab of the `Song Setup` window. The beatmap must also include the "N+1" tag.** For the purposes of this rule, only playstyles with key counts below 10 can be considered "N+1".[^n-plus-1]
```
MChecaH, line 69: `Difficulties` should probably be changed to `Beatmaps`. It's making reference to individual simfiles, but it's somewhat more correct to refer to them by Beatmaps rather than Difficulties to not mix the meaning of Difficulty

```suggestion
- **Beatmaps using an [N+1 playstyle](#common-terms) must enable the `Use special style (N+1 style) for mania` option under the `Advanced` tab of the `Song Setup` window. The beatmap must also include the "N+1" tag.** For the purposes of this rule, only playstyles with key counts below 10 can be considered "N+1".[^n-plus-1]
```

MChecaH (commented):
Few comments and annotations for the next rewrite.

wiki/Ranking_criteria/osu!mania/en.md
aryoadhi: - Your first line to state "any non-standard playstyle" requiring playstyle identifier on difficulty names directly conflicts with past RC, where "every single keycount/playstyle mapset" must not include playstyle identifier on difficulty names.
- Multiple key modes already covered on the new clause, on the condition "multiple playstyles are present in the beatmap."
- Different playstyles in one mode also covered with this PR. This to avoid circumvention of ranking 8K maps with mixed 8K and 7K+1 spread.

As for why it have extra requirement for 12K+ maps, it's because there are many playstyles can be derived, but only the mapper knows how to actually play these maps. It will be difficult to impossible to play if this intended playstyle isn't followed.

MChecaH: - The first point is something that was internally discussed. While the previous RC did not account for it, it is *now* been taking into consideration. A user is not going to know whether a beatmap with only one keymode being used in a non-standard playstyle unless they play the map. This is not intuitive, and adding the playstyle in such a case would be helpful.
- No. With the way playstyles are being treated differently from key counts, that is not the case. A multi-key mode set is not a multi playstyle set following the proposals' own definitions. Either this clause should disambiguate the terms, or they should be better clarified in the glossary. Or both, as proposed in the Review.
- Yes the current one does it. But that is the only thing it does.

I am not getting into 12+ playstyles. Simply commenting that the current wording is not clear for two of the three bulletpoints mentioned.

aryoadhi: - Not sure for the first point, but that does make sense. Does that mean for the future we should explicitly mention the playstyle?

I think you're right. We might need to put non-standards (mostly N+1s) descriptors.

MChecaH: For now, we have to follow what we have. We can't enforce rules that haven't been officially merged in the RC afterall. Once that changes then we can start applying it moving on.

wiki/Ranking_criteria/osu!mania/en.md
Walavouchey: the setting mentioned in the rule is per individual difficulty

MChecaH: It should probably be used for Beatmaps as a whole, specially now with the wording changes that would consider a "N+1" variation of a key mode its own playstyle. You don't really apply this arbitrarily on a per-difficulty basis. Maybe I'm looking too into it though.

aryoadhi: You can have both 7K+1 and 8K in one mapset, but you only apply special style tick to 7K+1 beatmaps. Applying this to the *whole* mapset defeats the purpose of special style because some mapsets that include N+1 may have a variant that is on different keycount, e.g 4K and 5K+1 mapset.

MChecaH: Hmm always had it wrong then, thinking that a Beatmap != Beatmapset when both meant the same thing. Then this is correct yep.

wiki/Ranking_criteria/osu!mania/en.md
cl8n: > A multi-key mode set is not a multi playstyle set following the proposals' own definitions.

I don't really see how you came to this conclusion. there's no such thing as a playstyle that can span multiple key modes, based on the definition in this PR

> This should be reworded in a way that [...] Any Non-Standard playstyle [...] need to add their key mode/playstyle in their difficulty name.

this was intentionally avoided so that no rules change for key modes below 12 -- if it should change, probably do it in separate pr

wiki/Ranking_criteria/osu!mania/en.md
cl8n: they're different things, how I intended it anyway. "key mode" is just the number of columns. "playstyle" is the way that the mapper probably intends you to play them

so the 8K keymode for example may be played with "8K" playstyle (standard), "7K1S" playstyle, etc... that's how I intended to write it anyway.

MChecaH: Yep I'm aware of the change. I'm moreso concerned at the way Key mode is perceived by the community. If a player was to see a beatmap listed as 7K, 4K, 5K, etc... they'd assume it is going to be played in the "default way"; in the way a Standard Playstyle is defined.

Would probably make more sense to specify that in Key mode outright.

wiki/Ranking_criteria/osu!mania/en.md
MChecaH: It would need to be discussed yes, but I personally don't see a reason why a playstyle lower than 10k should not be mentioned in the difficulty name just like 10k+ modes do.

While this is necessary for those modes, it still is something that lower keymodes can benefit from specifying as, again, players won't be scavenging through the tags to decypher what playstyle they're supposed to adapt to. Having it in the difficulty name is a surefire way for it to be understood at a first glance both from the website and client alike.

aryoadhi: It's almost always assumed that 4-10K uses standard playstyle. Even if the players played the wrong way, it's not as severe as 12K+ maps. For visibility of playstyles, I believe it will be seen on the map's tags (and maybe additionally description). For tags it will become more accessible on lazer.

It's true that we should discuss this in separate PR, because this PR's scope is on officially acknowledging 12K+ maps as rankable keycount/playstyles, not whether 4-10K requires identifier or not.

cl8n: > It would need to be discussed yes, but I personally don't see a reason why a playstyle lower than 10k should not be mentioned in the difficulty name just like 10k+ modes do.
> 
> While this is necessary for those modes, it still is something that lower keymodes can benefit from specifying as, again, players won't be scavenging through the tags to decypher what playstyle they're supposed to adapt to. Having it in the difficulty name is a surefire way for it to be understood at a first glance both from the website and client alike.

I agree with all of this, this PR is just not the place for it

as for the rest of your initial comment, those cases are already covered.

wiki/Ranking_criteria/osu!mania/en.md
cl8n: I think the usage of these terms should be more clear now. "standard playstyle" is only referred to once, specifically to distinguish from all playstyles, and now "key mode" and "playstyle" often appear alongside each other to help with readability for people who arent familiar (and don't *need* to be familiar) with any of this playstyle stuff.

wiki/Ranking_criteria/osu!mania/en.md
cl8n: marking as resolved but someone should mention this in the rc rewrite server thing.

maxuskazuma (commented):
Discuss everything with clay for the remaining changes and ultimately we come up with the agreement on this.

cl8n (approved):
depends on #10410 now for the links, otherwise lgtm

Walavouchey (approved):

PR #10143, opened 2023-10-07
Add wafer to `Centurions`

chromb:
berdlycore lover #1

## Self-check

- [N/A] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)


cl8n (changes requested):
es and fr dates aren't in the right format
also it feels like they missed an update somewhere because they are less specific with the years stat at the end...

chromb:
> feels like they missed an update somewhere because they are less specific with the years stat at the end...

Dunno exactly why the translators decided to put it in words instead of just like 3,5 but i've left it in the same precision they did.

cl8n (approved):

PR #10152, opened 2023-10-08
[TR] Add `Ranking`

SStoney6378:
## Self-check


ubinici:
Please refrain from translating stub articles in the future. There are many proper ones that are either outdated or unavailable. It would be a better practice to handle the outdated ones before delving into the other ones.

I'll drop a review at a later time, please do not merge/close the PR by then.

TicClick:
> Please refrain from translating stub articles in the future

while it's generally better to translate completed articles first, we have A LOT of stubs. realistically speaking, it's going to be a long time until all of them are fleshed out, so maybe some of stub translations are worth working on (I would advise against translating anything with `needs_cleanup: true` though)

wiki/Ranking/tr.md
```diff
@@ -0,0 +1,44 @@
+---
+stub: true
+etiketler:
```
WindowsMeosu, line 3:
```suggestion
tags:
```

WindowsMeosu (changes requested):

TicClick (approved):

PR #10160, opened 2023-10-09
CI: Update `node_modules` cache on pushes to `master`

FreezyLemon:
This PR adds a CI job that checks whether a `node_modules` cache entry for the current `package-lock.json` exists on the `master` branch.

PRs can use cache entries from the `master` branch, so adding this will speed up the CI for new PRs.

Examples from my fork:
- [Initial cache population](https://github.com/FreezyLemon/osu-wiki/actions/runs/6462542016/job/17544328515) (ca. 60s)
- [Push to master without cache update](https://github.com/FreezyLemon/osu-wiki/actions/runs/6462552336/job/17544355075) (ca. 15s)
- [PR benefitting from the cache](https://github.com/FreezyLemon/osu-wiki/actions/runs/6462585170/job/17544437386) (1s cache restoration vs 40s for `npm ci`)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -55,7 +55,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: node_modules
-          key: node-modules-${{ hashFiles('package.json', 'package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
 
       - name: install remark
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -57,7 +57,7 @@ jobs:
           path: node_modules
           key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
 
-      - name: install remark
+      - name: 'if cache outdated: install node.js dependencies'
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -57,7 +57,7 @@ jobs:
           path: node_modules
           key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
 
-      - name: 'if cache outdated: install node.js dependencies'
+      - name: 'if cache is outdated: install node.js dependencies'
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
 
--- a/.github/workflows/refresh-cache.yml
+++ b/.github/workflows/refresh-cache.yml
@@ -1,0 +1,41 @@
+name: refresh node_modules cache
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  cache-node-modules:
+    name: update cache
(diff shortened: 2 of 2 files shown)


FreezyLemon:
Actually, seeing as changes to the package lock file are really rare, I'll change this to run less frequently tomorrow. Please don't merge yet

FreezyLemon:
Updated to run roughly every two weeks (00:20 on day 1 and day 15 of each month).

Should be good to go now.

.github/workflows/continuous-integration.yml
```diff
@@ -55,9 +55,9 @@ jobs:
         uses: actions/cache@v3
         with:
           path: node_modules
-          key: node-modules-${{ hashFiles('package.json', 'package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
 
-      - name: install remark
+      - name: 'if cache outdated: install node.js dependencies'
```
TicClick, line 60:
```suggestion
      - name: 'if cache is outdated: install node.js dependencies'
```

.github/workflows/refresh-cache.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          path: node_modules
+          lookup-only: true
+          key: node-modules-${{ hashFiles('package-lock.json') }}-${{ steps.setup-node.outputs.node-version }}
+
+      - name: 'if cache outdated: install node_modules'
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
+        run: npm ci
+
+      - name: 'if cache outdated: save cache'
```
TicClick, line 36:
```suggestion
      - name: 'if cache is outdated: install node_modules'
        if: steps.cache-node-modules.outputs.cache-hit != 'true'
        run: npm ci

      - name: 'if cache is outdated: save cache'
```

TicClick (dismissed):
only non-substantial changes from me, looks good otherwise
for context, the message which prompted the PR: https://discord.com/channels/188630481301012481/218677502141399041/1161041955325747210

TicClick (approved):

cl8n:
I don't really see the benefit to having this on an arbitrary schedule rather than just every `master` push

FreezyLemon:
Running the CI job on every push to master feels like overkill..

The cache only needs to be "refreshed" when either
- package-lock.json is changed (in the last 12 months, this happened three times) or
- the last cache entry gets evicted (7 days of no new or changed PRs)

Pushes to master happen multiple times per day, right? Even if the job only runs for <30 seconds, it really is a bit unnecessary...

peppy:
Having to remember to manually purge caches each time `package-lock.json` changes is quite a high burden. I don't understand why it isn't done using standard practices rather than some weird schedule.

FreezyLemon:
There is no need for any manual action. As soon as the `package-lock.json` file gets changed, the cache entry is invalidated and not used for commits after that.

The only thing being moved backwards is the repopulation of the cache, which means that in a worst-case scenario, the cache entry could be missing for up to 2 weeks after a `package-lock.json` update.

Note that updates to `package-lock.json` are relatively rare. Seeing as this cache entry currently does not exist at all, I thought that it would not be worth running the CI job multiple times per day.

If you disagree, it's simple to change it back to `on every push to master`.

cl8n:
I don't think it's unnecessary to run on push to master, because that is exactly when the relevant cache key could change, even if rare. this workflow does much less work and runs much less often than our other CI, so I don't really see the issue from that angle

FreezyLemon:
Alright, reverted the change to a schedule. The job runs on every push to `master` now.

peppy (approved):

PR #10169, opened 2023-10-12
[KO] Translate country names in `OWC/2023`

Chronicler01:
## Self-check


Chronicler01:
Used criteria is ISO-3166-1

Korean : https://ko.wikipedia.org/wiki/ISO_3166-1
English : https://en.wikipedia.org/wiki/ISO_3166-1

Disclamier : According to ISO, Taiwan wrote "TAIWAN, PROVINCE OF CHINA", but in Korea, usually call it Taiwan. So i wrote as Taiwan

PR #10175, opened 2023-10-14
Update `Ranking criteria` with FA tag usage and AI content crediting

MChecaH:
This Pull request aims to clarify two previously discussed RC tweaks:
- **`Featured Artist` tag exclusivity:** The Featured Artist clarification has been [discussed by the community in a forum post](https://osu.ppy.sh/community/forums/topics/1801856?n=1) and internally. Both groups agreed on making the tag exclusive to Featured tracks only moving forwards.
- **AI Generated Background crediting:** Based on community discussion, we've decided internally to add a clarification on how AI Generated backgrounds should be credited. This doesn't aim to change the game's stance on AI generated content, but rather clarify how generated Backgrounds should be credited by users moving on.

## Self-check
- [x] The wording has been checked, ensuring it remains unambiguous.

### DIFF
--- a/wiki/Ranking_criteria/en.md
+++ b/wiki/Ranking_criteria/en.md
@@ -93,7 +93,7 @@ Important to understand before using:
   - `osu!` itself should never be used in the source field except in cases where the track was specifically created for osu! (also known as an [osu! original](/wiki/osu!_originals)), such as commissioned tournament tiebreaker tracks.
 - **Metadata that exceeds the field's limits (81 characters) must be shortened.** Start by dropping additional markers and if this still is not sufficient, indicate that the title has been shortened using `...` in a sensible place.
 - **If the artist or title fields were shortened to fit in field limits, the omitted information must be added to the tags.**
-- **The `featured artist` tag must only be added to tracks listed in the [Featured Artist catalogue](https://osu.ppy.sh/beatmaps/artists).** The tag is used to describe whether a track is licensed by osu!.
+- **The `featured artist` tag must be used for tracks listed in the [Featured Artist catalogue](https://osu.ppy.sh/beatmaps/artists) and for such tracks only.** The tag is used to describe whether a track is licensed by osu!.
 - **Tags must be related to the beatmap and not misleading for search results.** Tags describing the beatmap's style, song, storyboard, video, or background content are considered to be related to the beatmap.
 - **Song [genre and language](/wiki/Beatmap/Genre_and_language) must be added to the tags of a beatmap.** This is to enable users to search using these terms in-game like they do on the website. For instrumental tracks, "instrumental" is considered the language tag. Exceptions would be when language and/or genre are not clear, or multiple apply. In case of the latter, one fitting tag for each may be applied.
 
--- a/wiki/Ranking_criteria/en.md
+++ b/wiki/Ranking_criteria/en.md
@@ -42,7 +42,7 @@ Important to understand before using:
 ### Guidelines
 
 - **Creators of the media used in a beatmap should be credited in the beatmap's description.** Per the [Content Usage Guidelines](/wiki/Rules/Content_usage_guidelines), media used in beatmaps should be permitted for use in osu!, and their licensing will likely detail how to credit their creators. Common reasons to omit credits from the description include exceeding difficulty in sourcing, creators' wishes to remain uncredited, and presence of credits in the artist, title, or source fields.
-  - Beatmaps using AI-generated backgrounds should instead credit them as such in the description.
+  - AI-generated media in beatmaps should explicitly be attributed as being AI-generated, not illustrated or otherwise. This is to prevent misrepresentation of the creation process involved.
 - **Directly re-using your own Ranked difficulties in other Ranked beatmaps is discouraged.** This is to avoid unnecessary bloating of Ranked content.
 - **[Slider tick rate](/wiki/Beatmapping/Slider_tick_rate) should not be modified through the `.osu` file.** Most custom values result in unsnapped slider ticks, however tick rates 0.5, 1.333, and 1.5 have practical use and can be applied only when they do not cause slider ticks to be unsnapped.
 - **[Kiai](/wiki/Gameplay/Kiai_time) should start on a sound in the music.** Doing so otherwise causes the kiai flash to feel unrelated to the song.
--- a/wiki/Ranking_criteria/en.md
+++ b/wiki/Ranking_criteria/en.md
@@ -42,7 +42,7 @@ Important to understand before using:
 ### Guidelines
 
 - **Creators of the media used in a beatmap should be credited in the beatmap's description.** Per the [Content Usage Guidelines](/wiki/Rules/Content_usage_guidelines), media used in beatmaps should be permitted for use in osu!, and their licensing will likely detail how to credit their creators. Common reasons to omit credits from the description include exceeding difficulty in sourcing, creators' wishes to remain uncredited, and presence of credits in the artist, title, or source fields.
-  - Beatmaps using AI generated backgrounds should instead credit it as such in the description.
+  - Beatmaps using AI-generated backgrounds should instead credit them as such in the description.
 - **Directly re-using your own Ranked difficulties in other Ranked beatmaps is discouraged.** This is to avoid unnecessary bloating of Ranked content.
 - **[Slider tick rate](/wiki/Beatmapping/Slider_tick_rate) should not be modified through the `.osu` file.** Most custom values result in unsnapped slider ticks, however tick rates 0.5, 1.333, and 1.5 have practical use and can be applied only when they do not cause slider ticks to be unsnapped.
 - **[Kiai](/wiki/Gameplay/Kiai_time) should start on a sound in the music.** Doing so otherwise causes the kiai flash to feel unrelated to the song.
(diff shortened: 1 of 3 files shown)


wiki/Ranking_criteria/en.md
```diff
@@ -42,6 +42,7 @@ Important to understand before using:
 ### Guidelines
 
 - **Creators of the media used in a beatmap should be credited in the beatmap's description.** Per the [Content Usage Guidelines](/wiki/Rules/Content_usage_guidelines), media used in beatmaps should be permitted for use in osu!, and their licensing will likely detail how to credit their creators. Common reasons to omit credits from the description include exceeding difficulty in sourcing, creators' wishes to remain uncredited, and presence of credits in the artist, title, or source fields.
+  - Beatmaps using AI generated backgrounds should instead credit it as such in the description.
```
Walavouchey, line 45:
```suggestion
  - Beatmaps using AI-generated backgrounds should instead credit them as such in the description.
```

Walavouchey (changes requested):

wiki/Ranking_criteria/en.md
Walavouchey, line 45: would it make sense to put "media" here instead of "backgrounds"?

MChecaH: Back in the meeting we wanted to narrow the scope to backgrounds for now. Using "media" would potentially include music as well, and we discussed that songs/maps/etc... weren't mature enough to consider as of now.

MChecaH: Updated.

wiki/Ranking_criteria/en.md
```diff
@@ -92,7 +93,7 @@ Important to understand before using:
   - `osu!` itself should never be used in the source field except in cases where the track was specifically created for osu! (also known as an [osu! original](/wiki/osu!_originals)), such as commissioned tournament tiebreaker tracks.
 - **Metadata that exceeds the field's limits (81 characters) must be shortened.** Start by dropping additional markers and if this still is not sufficient, indicate that the title has been shortened using `...` in a sensible place.
 - **If the artist or title fields were shortened to fit in field limits, the omitted information must be added to the tags.**
-- **If the song mapped is a licensed one in the [Featured Artist](/wiki/People/Featured_Artists) library, `featured artist` must be added to tags.**
+- **The `featured artist` tag must only be added to tracks listed in the [Featured Artist catalogue](https://osu.ppy.sh/beatmaps/artists).** The tag is used to describe whether a track is licensed by osu!.
```
cl8n, line 96: the wording here no longer requires adding the featured artist tag. it needs to say more to the effect of "the tag is required if [...] and allowed only if [...]" (but theres probably a better way to word that)

MChecaH: If that is an issue, it could be worded similar to what was originally proposed like:
```
The `featured artist` tag must be added only if the tracks used are listed in the [Featured Artist catalogue](https://osu.ppy.sh/beatmaps/artists).** The tag is used to describe whether a track is licensed by osu!.
```

Moving the `only` ahead will make it clear that adding the tag is still a must, but only if that condition is met.

Will need to consult this with Wala however as they were the one to propose such wording.

wiki/Ranking_criteria/en.md
peppy: I don't get the differentiation. Please use "media" as it should t matter what the content is.

wiki/Ranking_criteria/en.md
```diff
@@ -42,6 +42,7 @@ Important to understand before using:
 ### Guidelines
 
 - **Creators of the media used in a beatmap should be credited in the beatmap's description.** Per the [Content Usage Guidelines](/wiki/Rules/Content_usage_guidelines), media used in beatmaps should be permitted for use in osu!, and their licensing will likely detail how to credit their creators. Common reasons to omit credits from the description include exceeding difficulty in sourcing, creators' wishes to remain uncredited, and presence of credits in the artist, title, or source fields.
+  - Beatmaps using AI-generated backgrounds should instead credit them as such in the description.
```
peppy, line 45: This is super unclear to the point of WTF.

The credit should go in the description instead of somewhere else? Bad sentence IMO.

But ignoring that, it is ambiguous in its aim. Need more context to even make a call on whether I even agree with this being added.

peppy (changes requested):
Blocking

wiki/Ranking_criteria/en.md
MChecaH: This was discussed in the meeting held on the 7th of October. It was catalyzed by a [recent discussion regarding how AI backgrounds should be credited](https://osu.ppy.sh/beatmapsets/2066427/discussion/-/generalAll#/3966141).

Currently, the aim of this change is to explicitly mention how AI Backgrounds should be credited moving on, and the importance of the term "generated" being used when doing so.

Other things such as website tags were discussed during the meeting, but extra efforts would require the game to take a stance for AI content, so for now we're only trying to change the way these are credited as otherwise there is no clause that makes mentions of this.

Please let me know if you need to look at the meeting notes where this was discussed and ping you accordingly!

wiki/Ranking_criteria/en.md
MChecaH: It can be done, but this may cause some confusion that was previously discussed. Do AI vocals on, for example, AI covers count? In practice, we would only be making sure the backgrounds and/or videos which is what we currently can work with.

Changing it to "media" would need further discussion, as this was not what was originally proposed. I'll gather some more opinions on it.

wiki/Ranking_criteria/en.md
Walavouchey: woops, didn't see this

```suggestion
- **The `featured artist` tag must be used for tracks listed in the [Featured Artist catalogue](https://osu.ppy.sh/beatmaps/artists) and for such tracks only.** The tag is used to describe whether a track is licensed by osu!.
```

MChecaH: Reads better on my end. Will wait for Clayton's OK before committing.

wiki/Ranking_criteria/en.md
peppy, line 45: If we're looking to make this as clear as day, I'd suggest something like this:

```suggestion
  - Beatmaps with AI-generated backgrounds should explicitly mention this in the description. This is to avoid any controversy from creators incorrectly claiming creative input (ie. claiming to have "illustrated" the background themselves).
```

peppy (changes requested):

wiki/Ranking_criteria/en.md
MChecaH: I'm keen on making the intent behind the clarification clearer.

Two things I'd change would be `controversy` to something less informal or "direct" as the purpose of crediting these works is also to be informative. Second thing would be changing `ie.` for `e.g.` which seems to be used in the RC when providing examples.

```suggestion
  - Beatmaps with AI-generated backgrounds should explicitly mention this in the description. This is to avoid creators from incorrectly claiming creative input (e.g. claiming to have "illustrated" the background themselves).
```

peppy: Fair points. I also thing it should be generalised to all imagery, not just backgrounds. How does this phrasing feel?

```suggestion
  - Beatmaps with AI-generated imagery should explicitly attribute it as being AI-generated, not "illustrated" or otherwise. This helps avoid any ambiguity over the creative input / process.
```

MChecaH: When it comes to phrasing I'd probably not be the best person to ask haha.

Using "imagery" may blur the line a little, as we aren't referring to the files themselves but the contents of backgrounds/videos. Say for example, a background with AI assets or tweaks would potentially fall there, even if it wasn't fully generated by a model. Is this something we want to account for? 

Otherwise, we could use "media" or a similar variation of that as you pointed out before to refer to files in their entirety.

Will need more eyes on `This helps avoid any ambiguity over the creative input / process.` as I cannot judge how well it fits wording wise. But the idea feels good.

peppy: Imagery was intended to include storyboard etc.

MChecaH: Hmm alright yeah fair. Not opposed to it if this is meant to encompass those kinds of assets.

Walavouchey: the wording doesn't need to be opinionated or embellished beyond the main point of marking ai content as ai content. i don't get why a point about "creative input" needs to be made in the same breath

```suggestion
  - AI-generated media in beatmaps should explicitly be attributed as being AI-generated.
```

both the original beatmap thread and internal nat thread also seem to partly revolve around not wanting people to say ai images are "illustrated", although if they're credited as "ai-generated" anyway then is such a remark really necessary?

peppy: It provides reasoning. Something which is sorely missing from the guidelines at the moment.

Walavouchey: well maybe something like this then

```suggestion
  - AI-generated media in beatmaps should explicitly be attributed as being AI-generated, not illustrated or otherwise. This is to prevent misrepresentation of the creation process involved.
```

basically the last suggestion here but more direct

peppy: Sure, I can agree on that one.

MChecaH: Looks good!

wiki/Ranking_criteria/en.md
MChecaH: Resolved with the most recent update.

wiki/Ranking_criteria/en.md
MChecaH: Resolved with the most recent update.

wiki/Ranking_criteria/en.md
cl8n: yeah wala's suggestion is good

cl8n (approved):

peppy (approved):

PR #10187, opened 2023-10-15
Replace "x" with "×" in resolutions in `Client/Playfield`

agatemosu:
## Self-check


Walavouchey (approved):
@TicClick thoughts? i remember you suggesting this for formulas

this could be applied elsewhere too, like the ranking criteria if it's not already the case

agatemosu:
I think that in normal text `×` should be used, and in code "x"

TicClick:
> @TicClick thoughts? i remember you suggesting this for formulas
> 
> this could be applied elsewhere too, like the ranking criteria if it's not already the case

yes to the PR, no to the ASC proposal

let me explain: while I surely like proper typography ([Ilya Birman](https://ilyabirman.net/typography-layout/)'s layout helps immensely), I recognize it will likely perform poorly when made a stylistic requirement, adding more hassle to writers' interactions with the wiki and more time to search where to copy the darn symbol from (readers will probably enjoy it — at everyone else's expense)

so, maybe make it optional, if-you-occasionally-remember-the-cross-product-sign-exists effort?

TicClick:
actually, I misread your comment, thinking you mentioned the article styling criteria, not the RC. yeah, sure, let's make it a batch edit, but the rest of my response stays

agatemosu:
look if it's good https://github.com/ppy/osu-wiki/compare/master...agatemosu:osu-wiki:xx

peppy:
I've never seen resolutions written like this

TicClick:
> I've never seen resolutions written like this. Please follow what operating systems actively use rather than using new symbols.

https://en.wikipedia.org/wiki/Image_resolution#Resolution_in_various_media and other articles on that topic, IMO it's similar to `—` vs `--` — these who can/want use the former, others resort to the latter

PR #3710, opened 2020-06-16
Update NPC2 for Quarterfinals

Mashyuf:

PR #10189, opened 2023-10-15
Update .markdownlint.json

agatemosu:
## Self-check

I don't know how I should update these:

- MD011: Was disabled in #3710 and it is not in `remarkrc`
- MD036: It does not detect hatnotes and notices, so isn't necessary

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,7 +1,7 @@
 {
     "MD003": { "style": "atx" },
     "MD004": { "style": "dash" },
-    "MD011": false,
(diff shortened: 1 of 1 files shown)


peppy (approved):
Seems ✅ based on tests still passing

peppy:
What was the force push for?

peppy (approved):

cl8n:
I think I disabled 11 to circumvent some bug so if that's working now then it's fine. rest looks good

PR #10197, opened 2023-10-16
Fix list indentation

agatemosu:
## Self-check

SKIP_OUTDATED_CHECK


TicClick:
I'm confused, what does this fix/change when done at such scale?

cl8n:
I think the technically-wrong list formatting may have been intentional to support the website's special image container for images alone in a paragraph. will check this one soon

cl8n:
the table thing seems pretty inconsequential to me but i guess asc does say this

> Empty cells must use a vertical bar (`|`) followed by two spaces and another vertical bar (`|`).

so it probably came up before...

remark-lint has nothing to complain about this and remark-stringify is just wrong so that's fun.

cl8n (changes requested):
pls split each kind of change into different PRs or at the very least different commits, I can't sanely review this when there are actual substantial fixes buried in 1000 lines of automated ones

also for the table spacing thing, it's probably better to leave them be and add relevant remark config instead. we should be able to use remark-stringify on the whole repo one day rather than addressing every minor formatting nitpick individually like this.

PR #10204, opened 2023-10-17
[DE] Remove `bestätigt` as a translation for `approved` from the ASC

FreezyLemon:
This translation isn't used anywhere in the wiki. `Approved` should just be used as-is.

## Self-check


Walavouchey:
@The-Last-Cookie

Walavouchey:
even if it's not used, are the other translations (ranked, qualified, etc.) used? why is "approved" an edge case?

FreezyLemon:
> even if it's not used, are the other translations (ranked, qualified, etc.) used? why is "approved" an edge case?

`ranked` is mostly used as-is, `gerankt` is an anglicism, not a translation.
An argument could be made for `qualified` and `loved` to also never be translated, but these translations make sense and are understandable. `bestätigt` is technically one correct translation, but I doubt users would understand what a `bestätigte Beatmap` is.

The-Last-Cookie:
@Walavouchey It's currently in debate in the German osu! discord and I will take a closer look at it as soon as I have time.

The-Last-Cookie:
In fact, I do agree that `bestätigt` is not a good translation.

While reading through the discussion, I prepared a list of words that I could *personally* see being used going forward. It was compiled with the suggestions in said discussion as well as my own considerations.

| English | German |
| :-: | :-- |
| graveyard | Friedhof |
| qualified | qualified, qualifiziert |
| pending | ausstehend |
| WIP | WIP, in Arbeit |
| loved | loved, (in der Community) beliebt, (von der Community) gewürdigt/geliebt |
| approved | approved, angenommen, zugelassen |
| ranked | ranked, gerankt, freigegeben |
| unranked | nicht freigegeben, nicht gerankt |

After taking a step back from this, I realise most players are preferring to keep the original words as proper nouns which I am fine with (for the most part at least). I can understand why one would rather use no translation for some of these terms, as there is no real fitting word for these in German that carry over the same meaning and expression (or are just pretty common by now like WIP).

In the end, we need to reach a consensus and if a major part of the community decides to not use what I suggested above, then so be it.

FreezyLemon:
This is way out of scope. Can we keep it to my teeny tiny change for now?

We can change the other translations later, thanks

The-Last-Cookie (approved):
I didn't have the intention to change this PR any further. My intention was to publish my opinion about the topic at hand.

So, my post in the discord still stands for discussion.

FreezyLemon:
Alright, let's move further discussion to Discord.

cl8n (approved):

PR #10211, opened 2023-10-19
[ZH-TW] Add `meta/group-info`

5idereal:
## Self-check

### DIFF
--- a/meta/group-info/zh-tw.yaml
+++ b/meta/group-info/zh-tw.yaml
@@ -1,0 +1,32 @@
+separator: 、
+alumni:
+  roles:
+    academy: osu!academy 創辦人
+    admin: 管理員
+    chat: 聊天室管理員
+    dev: osu! 開發者
+    media: osu!media 建立者
+    support: 支援團隊
+    tournaments: 錦標賽管理員
+gmt:
+  all_mods: 全域管理員
+  areas:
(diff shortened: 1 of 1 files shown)


YumeMuzi (approved):

meta/group-info/zh-tw.yaml
```diff
@@ -0,0 +1,32 @@
...
+    mapping: 製圖/Mod 圖社群管理員
+    skinning: Skin 社群管理員
+    leadership: 團隊領袖
+nat:
+  subgroups:
+    structural: 組織
+    evaluation: 評估
+    leadership: 領導
+languages:
+  partial: 部分 <language>
```
cl8n, line 32: is this space intentional? just making sure because ti seems like you don't use spaces in this language except around latin characters.

e.g. should it produce `部分 德語` or `部分德語`?

cl8n (changes requested):

meta/group-info/zh-tw.yaml
YumeMuzi: The space here should be removed, e.g. `部分德語`

cl8n (approved):

meta/group-info/zh-tw.yaml
5idereal: Thanks for the review!

PR #10221, opened 2023-10-19
CI: Remove unnecessary `git status` check

FreezyLemon:
There should not be any uncommited changes, so `git status` can be ignored here.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -21,10 +21,8 @@ jobs:
 
       - name: detect changed files
         run: |
-          INTERESTING_FILES=$(
-            # Changes committed to this PR so far
-            git diff --no-renames --name-only --diff-filter=d ${{ github.sha }}^
-          )
+          # Changes committed to this PR so far
+          INTERESTING_FILES=$( git diff --no-renames --name-only --diff-filter=d ${{ github.sha }}^ )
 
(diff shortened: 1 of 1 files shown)


Walavouchey (dismissed):
haven't tested

.github/workflows/continuous-integration.yml
```diff
@@ -22,12 +22,8 @@ jobs:
...
           INTERESTING_FILES=$(
-            sort -u < <(
-              # Changes that are not committed (staged + unstaged + untracked), but without deleted files
-              git status --short -v -v --no-renames --porcelain | awk '$1 != "D" { print $2 }'
-              # Changes committed so far (may overlap with the above)
-              git diff --no-renames --name-only --diff-filter=d ${{ github.sha }}^
-            )
+            # Changes committed to this PR so far
+            git diff --no-renames --name-only --diff-filter=d ${{ github.sha }}^
           )
```
TicClick, line 27: no need to have new lines at all:

```sh
# Changes committed to this PR so far
INTERESTING_FILES=$( git diff --no-renames --name-only --diff-filter=d ${{ github.sha }}^ )
```

(can't post a direct suggestion since it covers both original and changed lines)

TicClick (changes requested):

TicClick (approved):

PR #10246, opened 2023-10-21
[DE] Add "Difficulty spread"

FreezyLemon:
## Self-check

### DIFF
--- a/wiki/Beatmap/Difficulty_spread/de.md
+++ b/wiki/Beatmap/Difficulty_spread/de.md
@@ -1,0 +1,20 @@
+---
+tags:
+  - difficulty set
+  - progression
+  - difficulty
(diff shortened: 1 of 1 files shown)


wiki/Beatmap/Difficulty_spread/de.md
```diff
@@ -0,0 +1,20 @@
+---
+tags:
+  - difficulty set
+  - progression
```
The-Last-Cookie, line 4: Note to myself: The tag `progression` is not translated here because the literal translation `Verlauf` implies a time or spatial component, which both does not fit quite well.

The-Last-Cookie (approved):

cl8n (approved):

PR #10252, opened 2023-10-22
Reword comparisons of times and rhythms in RC

cl8n:
I noticed that they are pretty inconsistent at the moment. it seems in the majority of places it's preferred to use words like slow/fast/short/long, rather than low/high, when comparing times or rhythms. this also helps clarify some clashes in wording, like how in the phrases "1/1 or higher interval" and "1/6 or higher snap", "higher" has a completely opposite meaning.


wiki/Ranking_criteria/osu!catch/en.md
```diff
@@ -65,11 +65,9 @@ Difficulty-specific rules and guidelines do only apply to the difficulty level t
 
 | Difficulty | Basic-snapped dash | Higher-snapped dash | Basic-snapped hyperdash | Higher-snapped hyperdash |
 | :-- | :-- | :-- | :-- | :-- |
-| **Cup** | - | - | - | - |
-| **Salad** | 250 ms or higher | 125-249 ms | - | - |
-| **Platter** | 125 ms or higher | 62-124 ms | 250 ms or higher | 125-249 ms |
-| **Rain** | 125 ms or higher | 62-124 ms | 125 ms or higher | 62-124 ms |
-| **Overdose** | - | - | - | - |
+| **Salad** | 250 ms or slower | 125–249 ms |  |  |
```
aryoadhi, line 68: This shouldn't be changed, since we're talking about time, not speed. Keep the osu!catch terminology that uses absolute time (ms) as-is.

Walavouchey (approved):
for some values with clear units i don't think this matters too much, e.g. "below 3 Hz" or "above 200 BPM", but i don't disagree with the change

wiki/Ranking_criteria/en.md
```diff
@@ -32,7 +32,7 @@ Important to understand before using:
...
 - **There must not be any obscene imagery in the [background](/wiki/Beatmap/Background)/storyboard/video content.** This includes nudity, near-nudity, sexual references, violence, drug abuse, etc. See [Visual Content Considerations](/wiki/Rules/Visual_content_considerations) for more detailed rules.
-- **Difficulties that contain repetitive strobes, pulsing images, or rapid changes in contrast, brightness or colour in the storyboard or video must use an epilepsy warning.** If the warning interferes with gameplay, audio lead-in must be made longer. Strobing effects at 3 Hz (i.e. 3 flashes per second) and below are unlikely to cause concern. When in doubt, add the warning and confirm its necessity during the modding process.
+- **Difficulties that contain repetitive strobes, pulsing images, or rapid changes in contrast, brightness or colour in the storyboard or video must use an epilepsy warning.** If the warning interferes with gameplay, audio lead-in must be made longer. Strobing effects at 3 Hz (i.e. 3 flashes per second) or slower are unlikely to cause concern. When in doubt, add the warning and confirm its necessity during the modding process.
```
peppy, line 35: old sounds more correct

wiki/Ranking_criteria/en.md
```diff
@@ -212,7 +212,7 @@ This category contains explicit allowance statements of concepts and rules that
...
-- **The audio file of a song should not be artificially extended in order to meet a time limitation in the beatmap section of this criteria.** This can include (but is not limited to) looping sections of the audio file, lowering the [BPM](/wiki/Music_theory/Tempo) of the song or section of the song, or adding small amounts of music to the song without incorporating it throughout the entire song. This does not apply to [song compilations](/wiki/Beatmap/Song_compilation) or audio files less than the minimum rankable beatmap length.
+- **The audio file of a song should not be artificially extended in order to meet a time limitation in the beatmap section of this criteria.** This can include (but is not limited to) looping sections of the audio file, slowing the [tempo](/wiki/Music_theory/Tempo) of the song or section of the song, or adding small amounts of music to the song without incorporating it throughout the entire song. This does not apply to [song compilations](/wiki/Beatmap/Song_compilation) or audio files less than the minimum rankable beatmap length.
```
peppy, line 215: you don't "slow" tempo. you reduce or decrease it at best, but it still sounds wrong from a musical background.

the old seems fine, again.

wiki/Ranking_criteria/osu!catch/en.md
```diff
@@ -183,10 +181,10 @@ Difficulty-specific rules and guidelines do only apply to the difficulty level t
...
-- **1/8 and higher [hyperdashes](/wiki/Gameplay/Hyperdash) should not be used between consecutive object pairs.**
+- **1/8 and faster [hyperdashes](/wiki/Gameplay/Hyperdash) should not be used between consecutive object pairs.**
 - **[Edge dashes](/wiki/Gameplay/Edge_dash) may be used with caution for a maximum of three consecutive objects, and should not be used after [hyperdashes](/wiki/Gameplay/Hyperdash).**
 - **[Combos](/wiki/Beatmapping/Combo) should not exceed 16 objects including slider tails and repeats.** [Spinners](/wiki/Gameplay/Hit_object/Spinner) are an exception.
-- **Note density should follow a mostly 1/2 + 1/4 and/or 1/3 + 1/6 pattern.** 1/8 patterns and higher should be used sparingly.
+- **Note density should follow a mostly 1/2 + 1/4 and/or 1/3 + 1/6 pattern.** 1/8 and faster patterns should be used sparingly.
```
peppy, line 187: also think this is worse. it's not necessarily faster if there's caps, for instance. revert.

peppy (changes requested):
I'm not going to go through this whole thing because it feels like the same thing repeated, but disagree with a lot of these terminology changes.

wiki/Ranking_criteria/osu!catch/en.md
cl8n: this is essentially what prompted the change because currently this language is really inconsistent across RC -- a lot of places do use the wording I suggested already. it also bothered me that "higher" can mean two opposite things in incredibly similar contexts, when talking about beat snaps / rhythm density / etc. (see OP)

but more importantly, I am realising now that I interpreted this in a different way than you did. I thought that by "1/8 patterns or higher", it was talking about patterns that have consecutive notes with 1/8 gaps or smaller. but you seem to have read it as talking about any notes snapped to 1/8 or more precise snaps, which indeed may not support my suggested wording. maybe the wording of this whole line should be questioned... I'll see if I can talk to someone who would know for sure what the intention was

cl8n:
will probably apply peppy's first two suggestions, but I'm gonna wait for more reviews because I don't want to spend my time changing all these occurrences just to maybe have to do it again and again...

some NAT said they wanted to check this.

wiki/Ranking_criteria/osu!catch/en.md
peppy: Could be "1/8 and beyond" i guess?

Walavouchey: "denser/sparser rhythms" would also work

cl8n:
there's a larger rc rewrite over some of these sections happening so I'll just raise my points with them instead.

PR #10253, opened 2023-10-22
Clean up difficulty naming articles

cl8n:
the main point is that all of the previously black icon diff names that correspond to nothing in RC are removed from "Difficulty levels". they were only really there to show off the last color icon, but difficulties don't have specific bracketed icons anymore. only one had a proper name to begin with ("Hell Oni"), while other modes just got a `+` added to the second-top difficulty, which confused at least a few people into thinking that it was supposed to represent an actual difficulty name that mappers would use

coincidentally, this resolves https://osu.ppy.sh/community/forums/topics/1747458, but not in the way suggested

"Difficulty levels" is now also consistent with "Difficulty naming". I moved and cleaned up the "widely accepted alternatives" section of difficulty naming so that it's more prominently shown alongside the main 5 diff names for each mode

there are a lot of translation changes so you may have to view a diff only showing en.md files to review this one. also `SKIP_OUTDATED_CHECK` because some RC translations didn't need to be updated or outdated


Walavouchey (approved):
looks alright given the reasoning

PR #9331, opened 2023-05-08
Change osu!catch spread requirements

Darius-Wattimena:
Spread requirements change for osu!catch, based on changes decided in https://osu.ppy.sh/community/forums/topics/1747456

Moving all spread rules to their respective game mode RC might be better. Although I wasn't sure, so for now I added an extra item to the general RC referring to the catch RC like is done for mania.

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #10267, opened 2023-10-23
Remove defunct clause from osu! and osu!taiko RC spread rules

cl8n:
https://osu.ppy.sh/community/forums/posts/9309627

missed in #9331

SKIP_OUTDATED_CHECK https://github.com/ppy/osu-wiki/pull/10267#discussion_r1369701143


Hiviexd (dismissed):

wiki/Ranking_criteria/osu!taiko/zh.md
```diff
@@ -1,3 +1,8 @@
+---
```
YumeMuzi, line 1: `of each included game mode` is not translated into osu!taiko/zh.md (it doesn't make sense), so it may be better not to mark it as outdated?

Walavouchey (approved):

PR #10274, opened 2023-10-23
Update activity expectations for returning Beatmap Nominators in `Rules for Beatmap Nominators`

MChecaH:
On the 13th of October 2023, we [discussed internally](https://discord.com/channels/90072389919997952/1162458064100806857/1162458075903570021) how to handle inactivity of Beatmap Nominators rejoining in Good terms. The current ruling made it possible for them to be inactive for up to 90 days before action could be taken.

The extra clause expects returning BNs from meeting their activity quota on the first 30 days since they rejoin. It will be up to the NAT to manually judge each case of inactivity, and what reasonable action to be taken.

## Self-check

- [x] The wording used has been checked and discussed by the NAT


wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -26,6 +26,7 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
 - Beatmap Nominators are expected to stay below a maximum number of nominations over any 90-day period, with each mode having its own nomination cap. Exceeding this amount will lead to a warning, followed by a removal if they fail to manage their activity accordingly.
   - osu!, osu!catch, osu!mania: 75 unique nominations.
   - osu!taiko: 100 unique nominations.
+- Beatmap Nominators returning in good terms are expected to reach their mode's minimum activity requirements in the first 30 days since their rejoin. Staying inactive during this period of time may result in a warning or a removal, which will be evaluated on a case by case basis.
```
Walavouchey, line 29:
```suggestion
- Beatmap Nominators returning on good terms are expected to reach their mode's minimum activity requirements in the first 30 days since their rejoin. Staying inactive during this period of time may result in a warning or a removal, which will be evaluated on a case-by-case basis.
```

Walavouchey (approved):

Hiviexd (approved):

PR #10291, opened 2023-10-26
Fix table printing in remark-stringify

cl8n:
remark-stringify part of https://github.com/ppy/osu-wiki/pull/10197#issuecomment-1780553751 

- `npx remark <file with a table.md> -o` to test
- second and third commits are just spacing changes and you can ignore them for sake of reviewing


.github/workflows/refresh-cache.yml
```diff
@@ -12,7 +12,8 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          sparse-checkout: '.'
```
TicClick, line 15: does this still check out the root directory with`package.json` required for the Node.js deps caching action to work?

TicClick (approved):

.github/workflows/refresh-cache.yml
cl8n: it uses "cone mode" for sparse checkout which includes all of the files directly under each directory along the way (`.`, `meta`, `meta/remark`)

https://github.com/cl8n/osu-wiki/actions/runs/6679162109/job/18151011485

PR #10303, opened 2023-10-28
Add Usage permissions wiki article

Ephemeralis:
The list, any verbiage, policies or recommendations in this article are **heavily** WIP and should not be considered final.

Likely needs a lot more thought in terms of formatting this information and avoiding duplication (lots of it on the FA list column at the moment).

TODO:

- [x] entry paragraph/explanation
- [x] finalize usage list block
- [x] fix FA "bubble" not appearing in the list


wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
+---
+tags:
+  usage
+  permissions
```
cl8n, line 4: tags don't really do anything if they appear as complete words in the title already

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+---
+
+# Usage permissions
+
+## Opt-in
+
+If you are an artist and wish to have your stance regarding whether osu! users may make use of your work known publicly or privately, please feel free to reach out to either [the osu! support team via email](mailto:support@ppy.sh) or any member of the [osu!team](/wiki/People/osu!team) with as much information as possible regarding what work you may want to allow or disallow.
```
cl8n, line 11: the first clause before comma here is kind of hard to understand with how much it's trying to say... and also due to having the word "known" so far apart from the subject it's acting on ("your stance") i think. the "publicly or privately" part should be explained separately in a new sentence or something, I had to re-read this a few times to understand what was meant by that

so in general just try to simplify the sentence structure I guess. I think it's especially important for an article like this to be a clear and easy read.

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+
+## Audio
+
+All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are approved for use in the creation of osu! beatmaps. Tracks from a Featured Artist that do not appear in their listing are *not* licensed, and additional permissions may be required to use them.
```
cl8n, line 15: this feels more like a *second* sentence or paragraph of this section rather than the first -- something needs to first explain what this section is generally about (the table), and then this note about FA can come in to clarify how things work in that case.

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+
+| Artist                                   | Status                               | Exceptions                                                  | Notes                                                                                                               |
+| ---------------------------------------- | ------------------------------------ | ----------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------- |
+| -45                              ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/297)  |
```
cl8n, line 21: `![][FA]`

cl8n, line 21: if explained well enough above this table, i dont think all these duplicate notes about FA songs are necessary. it makes the mobile layout of this page terrible on top of just being kinda redundant

since that's currently where you link the fa pages, I think you could do something like this instead in the first cell
```markdown
[-45 ![][FA]](https://osu.ppy.sh/beatmaps/artists/297)
```

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+| 40mP                                     | 🔴 **Disallowed**                   |                                                             |                                                                                                                     |
+| 69 de 74                         ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/268)  |
+| 7_7                              ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/356)  |
+| a_hisa                           ![](FA) | 🟠 **Requires explicit permission** | Before uploading beatmaps, obtain the artist's permission.  | Artist can be reached via [email](mailto:hisaweb_info@yahoo.co.jp) or via [Bandcamp](https://a-hisa.bandcamp.com/). |
```
cl8n, line 29: shouldnt have fa badge i think?

cl8n, line 29: if you take my earlier suggestion about removing all the FA notes, then the "Exceptions" and "Notes" columns can probably just be combined to compact rows like this one.

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+
+Visual works refer to elements like beatmap backgrounds and are commonly retrieved from sites like Pixiv.
+
+In many cases, a visual artist will make it immediately apparent if they permit the use of their work elsewhere.  All visual works should be properly attributed and provide an appropriate link-back to their source. If a direct link cannot be used for whatever reason, another link that points to the visual artist's account on their most active platform should be used instead.
+
+If it is unclear whether a visual artist allows their work to be used on other platforms, we recommend not making use of any until appropriate usage conditions can be confirmed.
+
+## Notes
+
+[^monstercat-gold]: This artist is licensed in cooperation with [Monstercat](https://osu.ppy.sh/beatmaps/artists/255), and users may be able to obtain external media usage rights via a [Monstercat Gold](https://www.monstercat.com/gold) subscription. Please consult their site for specific details on this license.
```
cl8n, line 424: is this supposed to be a limitation or just a note for your convenience on where to get "external media usage rights"? it's under the "Exceptions" column of the table but it doesn't really seem like one.

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+| yuki.                            ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/4)    |
+| yukitani                         ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/167)  |
+| Yunosuke                         ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/244)  |
+| Yuyoyuppe / DJ'TEKINA//SOMETHING ![](FA) | 🟠 Allowed, with exceptions         | Tracks from or themed around Touhou should not be uploaded. | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/132)  |
```
cl8n, line 407: given the purpose of the page, I think the exception should be written in absolute terms rather than "should". something like "obtain the artist's permission for tracks themed around touhou" if that's what they were going for

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+| High Tea Music                   ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/26)   |
+| Hinkik                           ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/315)  |
+| Hino Isuka                       ![](FA) | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/272)  |
+| HoneyComeBear                    ![](FA) | 🟠 Allowed, with exceptions         | Do not use or upload non-FA tracks.                         | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/319)  |
```
cl8n, line 165: I would bold all of the "Status" column messages that are not simply "Allowed"

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+
+## Audio
+
+All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are approved for use in the creation of osu! beatmaps. Tracks from a Featured Artist that do not appear in their listing are *not* licensed, and additional permissions may be required to use them.
+
+**Important note**: external media (such as gameplay videos submitted to sites like YouTube, livestreams) is subject to different licensing requirements that must be negotiated in most circumstances with the artist themselves. Some artists have broad permissions or allowances for external media and we have made best efforts to include this information where appropriate, but please ensure you double-check using primary sources like the artist's own sites/platforms.
+
+| Artist                                   | Status                               | Exceptions                                                  | Notes                                                                                                               |
```
cl8n, line 19: maybe easier to find what you're looking for if this table were split into two, one for all the "Allowed" artists and one for everybody else

cl8n (changes requested):

wiki/Usage_permissions/en.md
cl8n, line 19: also tables need to be written in compact format i.e. don't try to align all of columns, just make each cell as small as possible, with 1 space of padding on each side.

Hiviexd (commented):
instead of using emojis, i'd suggest using the actual wiki icons alongside footnotes when appropriate, like this example below:



[article github link](https://github.com/ppy/osu-wiki/blob/master/wiki/Help_centre/Upgrading_to_lazer/en.md)  for reference

wiki/Usage_permissions/en.md
aryoadhi: Or better split into three, one for FA, two for non-FA (allowed and disallowed). A citation of proofs (such as chat/e-mail screenshot) would be great too.

peppy: I don't understand this. Alphabetical is best.

Case: If a user is unable to type a search term, at least then they can search a single list alphabetically, rather than three.

wiki/Usage_permissions/en.md
peppy: If the artist wanted users to obtain permission, then I think that would be written. This sounds like a hard ruling (@Ephemeralis will have to confirm).

Also "should" is quite an absolute term:

CleanShot 2023-10-30 at 03 01 22

wiki/Usage_permissions/en.md
peppy: The column should probably be renamed to "Notes".

wiki/Usage_permissions/en.md
peppy: Agree.

wiki/Usage_permissions/en.md
peppy: "any member of the osu! team" – no
"support@ppy.sh" – no

please change this to require explicitly contacting `music@ppy.sh`, no other avenues will be accepted.

```suggestion
If you are an artist and wish to add or update your entry on this listing, please feel free to reach out to our [music licencing team](mailto:music@ppy.sh).
```

wiki/Usage_permissions/en.md
aryoadhi: It depends on the artist themselves, do they hard rule disallows non-FA (must) or asked first before uploading (should)? Either case is fine depending on the degree. We should confirm tk each artists whether explicit requests needed or disallowed at all.

wiki/Usage_permissions/en.md
aryoadhi: Personally I would like to search non-FA artists randomly and listen to the song myself, since U like discovering songs in surprise. But you're right about this. If only there is a dynamic table to apply the filters in live...

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+
+## Opt-in
+
+If you are an artist and wish to have your stance regarding whether osu! users may make use of your work known publicly or privately, please feel free to reach out to either [the osu! support team via email](mailto:support@ppy.sh) or any member of the [osu!team](/wiki/People/osu!team) with as much information as possible regarding what work you may want to allow or disallow.
+
+## Audio
+
+All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are approved for use in the creation of osu! beatmaps. Tracks from a Featured Artist that do not appear in their listing are *not* licensed, and additional permissions may be required to use them.
+
+**Important note**: external media (such as gameplay videos submitted to sites like YouTube, livestreams) is subject to different licensing requirements that must be negotiated in most circumstances with the artist themselves. Some artists have broad permissions or allowances for external media and we have made best efforts to include this information where appropriate, but please ensure you double-check using primary sources like the artist's own sites/platforms.
```
peppy, line 17:
```suggestion
**Important note**: Posting or streaming gameplay on other sites may potentially be subject to different licensing requirements that need to be negotiated or understood. Please do your due diligence when broadcasting gameplay to an audience.
```

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+Visual works refer to elements like beatmap backgrounds and are commonly retrieved from sites like Pixiv.
+
+In many cases, a visual artist will make it immediately apparent if they permit the use of their work elsewhere.  All visual works should be properly attributed and provide an appropriate link-back to their source. If a direct link cannot be used for whatever reason, another link that points to the visual artist's account on their most active platform should be used instead.
+
+If it is unclear whether a visual artist allows their work to be used on other platforms, we recommend not making use of any until appropriate usage conditions can be confirmed.
+
+## Notes
+
+[^monstercat-gold]: This artist is licensed in cooperation with [Monstercat](https://osu.ppy.sh/beatmaps/artists/255), and users may be able to obtain external media usage rights via a [Monstercat Gold](https://www.monstercat.com/gold) subscription. Please consult their site for specific details on this license.
+[^official-promo-exclude]: Official osu! promotional media is excluded from this limitation.
```
peppy, line 425: this is never used.

wiki/Usage_permissions/en.md
agatemosu: Maybe one of [these notices](https://osu.ppy.sh/wiki/en/Article_styling_criteria/Formatting#notice) that appear in the ASC should be used instead

wiki/Usage_permissions/en.md
agatemosu: I would directly not mention featured artists that have no exception, and better put a paragraph saying that featured artists that are not on this list are always allowed.

wiki/Usage_permissions/en.md
peppy: That's not how this page works. It is opt in and already lists artists that want to be asked explicitly.

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+| I love you Orchestra             ![][FA] | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/117)  |
+| IAHN                             ![][FA] | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/3)    |
+| iFeature                         ![][FA] | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/210)  |
+| Igorr                                    | 🔴 **Disallowed**                   |                                                             |                                                                                                                     |
```
Nivalyx, line 172: small fix

```suggestion
| Igorrr                                    | 🔴 **Disallowed**                   |                                                             |                                                                                                                     |
```

Nivalyx:
Just a thought, would it be appropriate to have a specific note added next to Hatsuki Yura and Igorrr's listing respectively?

As much as the two of them don't allow for their songs to be used in osu!, they do have songs featuring them licensed in other artists' catalogue ([here](http://osu.ppy.sh/beatmaps/artists/tracks?is_default_sort=true&sort=update_desc&query=hatsuki) and [here](http://osu.ppy.sh/beatmaps/artists/tracks?is_default_sort=true&query=igorrr&sort=relevance_desc) respectively)

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,427 @@
...
+| The Gentle Men                   ![][FA] | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/142)  |
+| tieff                            ![][FA] | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/34)   |
+| Tiny Waves                       ![][FA] | 🟢 Allowed                          |                                                             | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/131)  |
+| tokiwa                           ![][FA] | 🟢 Allowed                          | [^monstercat-gold]                                          | Officially licensed songs can be found at their [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists/202)  |
```
Nivalyx, line 374: tokiwa is not monstercat (Tokyo Machine however is)

peppy:
@Ephemeralis reviews require attention.

Ephemeralis:
Tables have way less duplicated info now after factoring in formatting suggestions from reviewers. Might drop the notes column on disallowed entirely.

Walavouchey:
i've pushed some changes to clean up the tables and the article in general

[before](https://osu3.roanh.dev/wiki/en/Usage_permissions) | [after](https://osu1.roanh.dev/wiki/en/Usage_permissions)

wiki/Usage_permissions/en.md
```diff
@@ -0,0 +1,444 @@
...
+
+This section details which artists are willing to allow use of their audio works in osu!, and under what conditions (where appropriate).
+
+All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are approved for use in the creation of osu! beatmaps. Tracks from a Featured Artist that do not appear in their listing are *not* licensed, and additional permissions may be required to use them.
+
+### Allowed
+
+| Artist |  | Status |
+| :-- | :-- | :-- |
+| [![][FA]](https://osu.ppy.sh/beatmaps/artists/18) | [\*namirin](https://osu.ppy.sh/beatmaps/artists/18) | ![][true] |
```
Hiviexd, line 21: would make more sense to have the pill link to the [FA wiki article](https://osu.ppy.sh/wiki/en/People/Featured_Artists) instead of linking to the same artist page twice in a row.

Walavouchey: there's less utility in almost 400 links to the fa article than links to each individual artist. between having or not having a clickable icon, this is the best option imo

Hiviexd:
this looks to be in a more ready state now, probably fine to start moving forward with it?

peppy:
Let's get this out and see how it looks, then iterate from there.

peppy (approved):

Issue #8109, opened 2022-10-03
Figure out a replacement for "#1" links in tables


Walavouchey:
<img src="https://user-images.githubusercontent.com/36758269/193532946-5cf684f1-51e3-48c4-a3b2-dc5accb8c918.png" width=300 />

i think this convention started in tournament articles as a way to prevent spamming "link" or some other wording in tables, for situations where any other name for the link isn't applicable since it's covered by the table header. 

i can think of two solutions:

- an image
- a website-implemented widget

i'm imagining some kind of button- or card-looking thing, but if you have any good ideas they would be appreciated

TicClick:
I read this as a non-issue, simply because there are no alternative solutions: if it's not going to be `#1` all the time (is there even `#2`?), it will be something else just as repetitive. an image (I assume you are looking for a pretty `([ Match link ])` button or something) would require one to copy and paste the reference fron one article to another just the same way we did it for flags.

a website widget, however it may look, is a bit more useful, but I don't see much benefit in making it solely for the wiki purposes (not sure how often people share multiplayer links on forums -- supporting that widens the scope of work for whoever's in charge).

I guess this is a direct consequence of the layout chosen for the tournament results. since real link title (`Match link`) is moved to the column title row, the links themselves have no real title, while it should have been at least the name of the match:

```md
- [Germany 5 : 2 Australia](https://osu.ppy.sh/community/matches/100000001)
- [Malaysia 2 : 5 France](https://osu.ppy.sh/community/matches/100000002)
- [Austria 2 : 5 Norway](https://osu.ppy.sh/community/matches/100000003)
```

Walavouchey:
> is there even `#2`?

71 out of 17876 in tournament articles. even if there are two of them... we could just have two widgets/images

> it will be something else just as repetitive. an image (I assume you are looking for a pretty `([ Match link ])` button or something) would require one to copy and paste the reference fron one article to another just the same way we did it for flags

yes, it would still be repetitive since it's a match link, but it would require orders of magnitude less work than hundreds of flag references. the point of this is to have something more pleasing/sensible than a "`#1`".

> a website widget, however it may look, is a bit more useful, but I don't see much benefit in making it solely for the wiki purposes (not sure how often people share multiplayer links on forums -- supporting that widens the scope of work for whoever's in charge).

it would just be for the wiki, yes. as you said, other places than the wiki would not have any immediate need/use for this.

[...]

PR #10312, opened 2023-10-29
Improve format of tournament match history notes

cl8n:
applied to owc 2023 as an example, but I think this is a better format than the numbered links thing for any future tournament. https://discord.com/channels/188630481301012481/218677502141399041/1167996377515708436

### DIFF
--- a/wiki/Tournaments/OWC/2023/en.md
+++ b/wiki/Tournaments/OWC/2023/en.md
@@ -196,13 +196,13 @@ The complete sign-up list can be found [here](https://gist.github.com/LeoFLT/82e
 
 Saturday, 28 October 2023:
 
-| Team A |  |  | Team B | Match link | VOD link |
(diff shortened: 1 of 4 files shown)


Walavouchey (changes requested):
disagree. i don't think this is a proper solution, see https://github.com/ppy/osu-wiki/issues/8109

also we don't really use title texts on links anymore

cl8n:
i agree with pretty much everything you said in that issue -- why do u disagree with this change then? im not really following

Walavouchey:
i'd only agree if it's applied everywhere, but at that point i'd rather pick up the pace on resolving the linked issue (using icons instead)

cl8n:
closing so the bot stops commenting on everything, but I still feel like at the very least this is better than `#1` etc... the number means nothing right now (with only a few exceptions)

#10354 also does a very similar thing, for reference

PR #10315, opened 2023-10-29
Fix `epic man 2`'s flag in `osu!mania Modding & Mapping Hub` article

Hiviexd:


TicClick (approved):
oh I love that mistake (also, CA vs CN)

PR #10317, opened 2023-10-29
Improve explanation of the weekly hype limit

FreezyLemon:
The wording tries to make it clear that each hype will recharge one week after being used instead of the user recharging one hype per week.

## Self-check

### DIFF
--- a/wiki/Beatmap/Hype/en.md
+++ b/wiki/Beatmap/Hype/en.md
@@ -14,6 +14,6 @@ Beatmaps need at least 5 hypes (one complete bar of the hype train) to be eligib
 
 ## Limit
 
-Users start with 10 hypes. Each hype independently takes a week to recharge after being used on a beatmap, meaning all hypes are available again after one week.
+Users start with 10 hypes, and they are consumed when used on beatmaps. Each hype takes a week after use to become available again.
 
 <!-- TODO: images of the beatmap page and the modding page -->
--- a/wiki/Beatmap/Hype/en.md
+++ b/wiki/Beatmap/Hype/en.md
@@ -10,7 +10,7 @@ Every [pending](/wiki/Beatmap/Category#work-in-progress-and-pending) [beatmap](/
 
 ## Ranking requirement
 
-Beatmaps need at least 5 hype (one complete bar of the hype train) to be eligible for [nomination](/wiki/Beatmap_ranking_procedure#nominations), a step towards getting the map ranked.
+Beatmaps need at least 5 hypes (one complete bar of the hype train) to be eligible for [nomination](/wiki/Beatmap_ranking_procedure#nominations), a step towards getting the map ranked.
 
 ## Limit
 
--- a/wiki/Beatmap/Hype/en.md
+++ b/wiki/Beatmap/Hype/en.md
@@ -14,6 +14,6 @@ Beatmaps need at least 5 hype (one complete bar of the hype train) to be eligibl
 
 ## Limit
 
-Users start with 10 hypes, and they are consumed when used on beatmaps. Each hype takes a week to become available again.
+Users start with 10 hypes. Each hype independently takes a week to recharge after being used on a beatmap, meaning all hypes are available again after one week.
 
 <!-- TODO: images of the beatmap page and the modding page -->
--- a/wiki/Beatmap/Hype/fr.md
+++ b/wiki/Beatmap/Hype/fr.md
@@ -1,4 +1,6 @@
 ---
+outdated_translation: true
+outdated_since: ffc2b62f3415a9a54a838e55c845a7a0bcd8d681
 stub: true
 tags:
   - train
--- a/wiki/Beatmap/Hype/id.md
+++ b/wiki/Beatmap/Hype/id.md
@@ -1,4 +1,6 @@
 ---
+outdated_translation: true
+outdated_since: ffc2b62f3415a9a54a838e55c845a7a0bcd8d681
 stub: true
 tags:
   - train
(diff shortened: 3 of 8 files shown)


Walavouchey (dismissed):

minusQuantumNeko (approved):

agatemosu:
see the line 13 with `5 hype`

FreezyLemon:
> see the line 13 with `5 hype`

I'm not sure what you mean, could you explain?

agatemosu:
```Beatmaps need at least 5 hype``` to ```Beatmaps need at least 5 hypes```

**Comment by FreezyLemon** [comment]
I see, and agree. Using the countable noun there is better. Fixed

**Review by minusQuantumNeko — APPROVED** [review]

**Comment by cl8n** [comment]
> instead of the user recharging one hype per week.

the current wording doesn't leave any room to suggest this though. it seems more like a misreading than the wiki's fault

and then the new wording is basically same thing but more verbose to me; "independently" doesn't change or clarify anything, "recharge" is an odd word choice for something to become available

File: wiki/Beatmap/Hype/en.md [hunk 1]
```diff
@@ -10,10 +10,10 @@ Every [pending](/wiki/Beatmap/Category#work-in-progress-and-pending) [beatmap](/
 
 ## Ranking requirement
 
-Beatmaps need at least 5 hype (one complete bar of the hype train) to be eligible for [nomination](/wiki/Beatmap_ranking_procedure#nominations), a step towards getting the map ranked.
+Beatmaps need at least 5 hypes (one complete bar of the hype train) to be eligible for [nomination](/wiki/Beatmap_ranking_procedure#nominations), a step towards getting the map ranked.
 
 ## Limit
 
-Users start with 10 hypes, and they are consumed when used on beatmaps. Each hype takes a week to become available again.
+Users start with 10 hypes. Each hype independently takes a week to recharge after being used on a beatmap, meaning all hypes are available again after one week.
```
peppy, line 17:
```suggestion
Users start with 10 hypes, and they are consumed when used on beatmaps. Each hype takes a week after use to become available again.
```

wiki/Beatmap/Hype/en.md
```diff
@@ -10,10 +10,10 @@ Every [pending](/wiki/Beatmap/Category#work-in-progress-and-pending) [beatmap](/
 
 ## Ranking requirement
 
-Beatmaps need at least 5 hype (one complete bar of the hype train) to be eligible for [nomination](/wiki/Beatmap_ranking_procedure#nominations), a step towards getting the map ranked.
+Beatmaps need at least 5 hypes (one complete bar of the hype train) to be eligible for [nomination](/wiki/Beatmap_ranking_procedure#nominations), a step towards getting the map ranked.
```
peppy, line 13:
```suggestion
Beatmaps need at least 5 hype (one complete bar of the hype train) to be eligible for [nomination](/wiki/Beatmap_ranking_procedure#nominations), a step towards getting the map ranked.
```

Walavouchey: the previous was incorrect

Walavouchey:
> it seems more like a misreading than the wiki's fault

it apparently does cause misunderstandings (https://discord.com/channels/188630481301012481/218677502141399041/1168165581875335168). the part clarifying that it takes 1 week for all hypes to become available and not 10 should be kept if anything

wiki/Beatmap/Hype/en.md
peppy: I dunno, calling them "hypes" feels really corny.

When you build hype you don't call it hypes.

🤷

FreezyLemon: Hm.. I think the article should stick to one or the other.. using `5 hype` here and then `10 hypes` in the following paragraph is a bit weird, I think.

FreezyLemon:
> > instead of the user recharging one hype per week.
> 
> the current wording doesn't leave any room to suggest this though. it seems more like a misreading than the wiki's fault

I guess there was some context missing as to why I opened this (see Wavalouchey's comment).

> and then the new wording is basically same thing but more verbose to me; "independently" doesn't change or clarify anything,

Fair point. I was originally planning on using an even longer explanation, but I didn't want to blow up the text over what seemed like a small misunderstanding. I see now that the new explanation is no real improvement over the original one..

>  "recharge" is an odd word choice for something to become available

It seemed intuitive enough to use instead of using `become available` twice in that sentence. I'll change it

FreezyLemon:
Not sure how to proceed here. I didn't expect any pushback on these changes, but the fact that people are unhappy with the relatively small changes here makes me think that maybe it should just be left as-is.

wiki/Beatmap/Hype/en.md
```diff
@@ -10,10 +10,10 @@ Every [pending](/wiki/Beatmap/Category#work-in-progress-and-pending) [beatmap](/
...
 
 ## Limit
 
-Users start with 10 hypes, and they are consumed when used on beatmaps. Each hype takes a week to become available again.
+Users start with 10 hypes, and they are consumed when used on beatmaps. Each hype takes a week after use to become available again.
```
Walavouchey, line 17: as explained earlier, the premise is that this point wasn't obvious

```suggestion
Users start with 10 hypes, and they are consumed when used on beatmaps. Each hype takes a week after use to become available, meaning all 10 are available again after one week.
```

wiki/Beatmap/Hype/en.md
Walavouchey: "hype" is treated as a countable noun here, i.e. instances of hype posts in a beatmap discussion. if you don't want to do that and insist that "hype" can't be countable, you'd write "5 hype posts" or something similar instead

FreezyLemon:
My original changes weren't great, and I can't seem to come up with wording that I'd personally be happy with. I'll close; someone else should take a shot if they're interested enough.

PR #10325, opened 2023-10-29
Use direct links in `Beatmap Spotlights`

agatemosu:
## Self-check


cl8n (approved):
sure but i dont think this generally matters either...

agatemosu:
> sure but i dont think this generally matters either...

It was to remove "dev" and at the same time remove the redirection

agatemosu:
> sure but i dont think this generally matters either...

It was to remove "dev" and at the same time remove the redirection

PR #10352, opened 2023-11-03
Update Node.js used in CI workflow to version 20

FreezyLemon:
Node.js 16.X is [EOL](https://github.com/nodejs/release). 20.X is the current LTS release and will be supported until mid-2026.


sr229 (approved):

cl8n (approved):

cl8n:
im merging this because there's some features in node20 that remark will be making use of at some point -- on its own, someone dropping future support for a very stable piece of software that we weren't updating anyway is a non-issue, I think.

for any lurkers here the minimum required version is still 16.x

PR #10397, opened 2023-11-06
Merge `Usage permissions` and `Content usage guidelines` and move to `Rules/`

Walavouchey:


Walavouchey:
another option is to merge this article with https://osu.ppy.sh/wiki/en/Rules/Content_usage_guidelines as they cover related ground

wiki/Help_centre/Usage_permissions/Usage_permissions/en.md
```diff
@@ -6,6 +6,8 @@ osu! is centred around being able to freely make and share creations with one an
 
 For artists wishing to add or update their entry on this listing, please feel free to reach out to the osu! music licensing team at [music@ppy.sh](mailto:music@ppy.sh).
 
+Alternatively, artists are encouraged to contact the team about proper licensing of their work. See the [osu! music licensing](/legal/Music_licensing) article for more information.
```
peppy, line 9:
```suggestion
Please also consider contacting the team about negotiating licensing. See the [osu! music licensing](/legal/Music_licensing) page for more information.
```

Walavouchey:
(somehow managed to move the article incorrectly)

peppy:
Wonder if we should rename the article while here to either "Content usage permissions" or "Music usage permissions" or maybe even something else? Right now it doesn't really convey what it's containing too well.

Walavouchey:
content would work but not music since the article mentions visual works as well

peppy:
> content would work but not music since the article mentions visual works as well

Given we already have another wiki page for this, maybe it can be removed? Or just mentioned as something like "This page only covers music usage. If you are looking for guidelines on other content such as images / videos, please see [[Content Usage Guidelines]]]"

Sitoria:
little suggestion about visual part:

> Visual works refer to elements like beatmap backgrounds and are commonly retrieved from sites like Pixiv.

--> Visual works refer to elements like beatmap backgrounds, *storyboard* and are commonly retrieved from sites like Pixiv.

Ephemeralis:
This doesn't really feel like a `Help centre` article. If anything, it should be put somewhere close to where users might find information about beatmapping.

Walavouchey:
putting this into `Beatmapping` would be a stretch

the plan is to merge `Rules/Content usage guidelines` into this article as it covers the same ground. if not `Help centre`, then `Rules` might be a better home for this

peppy (approved):

PR #10419, opened 2023-11-09
[DE] Add `Primary metadata source`

FreezyLemon:
I did not translate the `Artist` etc. labels because the Song Setup window in stable is always English, even if other languages are selected in-client.

## Self-check


wiki/Beatmap/Primary_metadata_source/de.md
```diff
@@ -0,0 +1,29 @@
...
+  - Primäre Metadaten-Quellen
+---
+
+# Primäre Metadatenquelle
+
+Eine **primäre Metadatenquelle** ist eine Quelle für die Felder `Artist`, `Title` und `Source` in den [Metadaten](/wiki/Client/Beatmap_editor/Song_setup) einer [Beatmap](/wiki/Beatmap), die vom Künstler oder dem Musiklabel offiziell beworben werden.
+
+Die Webseite eines Künstlers oder seines Labels, CD-Scans und offizielle Inhalte auf Webseiten wie Bandcamp, YouTube und SoundCloud werden als primäre Quellen für Beatmap-Metadaten akzeptiert. Drittanbieterwebseiten wie Wikis, Datenbanken (vndb, vgmdb usw.) und Musikservices (Spotify, iTunes, Amazon usw.) geben die Wünsche des Künstlers nicht immer korrekt wieder und sollten nicht als Primärquelle verwendet werden.
+
+Beatmaps, die auf den [Ranked](/wiki/Beatmap/Category#ranked)-Status abzielen, müssen ihre Metadaten aus einer Primärquelle beziehen. In den [Ranking-Kriterien](/wiki/Ranking_criteria) werden weitere Regeln und Richtlinien zu Beatmap-Metadaten festgehalten.
```
The-Last-Cookie, line 29:
```suggestion
Beatmaps, die auf den [Ranked](/wiki/Beatmap/Category#ranked)-Status abzielen, müssen ihre Metadaten aus einer Primärquelle beziehen. In den [Ranking-Kriterien](/wiki/Ranking_criteria#metadata) werden weitere Regeln und Richtlinien zu Beatmap-Metadaten festgehalten.
```
Now this is a bit problematic. When the ranking criteria are translated at some point, the fragment at the end of the link is not added (because it will obviously be forgotten). That in turn has the consequence that the metadata section is only linked in the English wiki.

There are these options now:

1. Ommit the fragment and let the user search for the section themselves (not a fan of this)
2. Add the link fragment of the English wiki. This will cause the website to jump to the section in the English article automatically. If the article is translated at some point, the wiki link checker will add a notice that this link needs to be changed because there is no section called `Metadata` anymore. Then, we could change the link to German.
3. Add the link fragment of the supposedly German title. This will cause the English article to stay at the top and the user would need to search for the section themselves. However, the section will be accessed automatically as soon as the article is translated.

As you can see in the suggestion, I prefer the second option.

The-Last-Cookie (changes requested):
> I did not translate the `Artist` etc. labels

Understandable, I would have done the same.

wiki/Beatmap/Primary_metadata_source/de.md
```diff
@@ -0,0 +1,29 @@
...
+  - Offizielle Quellen
+  - Primäre Metadatenquellen
+  - Primäre Metadaten-Quellen
+---
+
+# Primäre Metadatenquelle
+
+Eine **primäre Metadatenquelle** ist eine Quelle für die Felder `Artist`, `Title` und `Source` in den [Metadaten](/wiki/Client/Beatmap_editor/Song_setup) einer [Beatmap](/wiki/Beatmap), die vom Künstler oder dem Musiklabel offiziell beworben werden.
+
+Die Webseite eines Künstlers oder seines Labels, CD-Scans und offizielle Inhalte auf Webseiten wie Bandcamp, YouTube und SoundCloud werden als primäre Quellen für Beatmap-Metadaten akzeptiert. Drittanbieterwebseiten wie Wikis, Datenbanken (vndb, vgmdb usw.) und Musikservices (Spotify, iTunes, Amazon usw.) geben die Wünsche des Künstlers nicht immer korrekt wieder und sollten nicht als Primärquelle verwendet werden.
```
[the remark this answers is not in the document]
FreezyLemon: `CD scans` are actual scans (as in, the digital files resulting from putting something into a scanner device) of the booklet/cover/whatever that is included with the physical CD, so it is not synonymous to edition or version in this context.

I'm aware of the difference between `Intention` and `Wunsch`, but I thought `Wunsch` sounded better in context, and the difference really isn't that important here. I don't mind changing it though

wiki/Beatmap/Primary_metadata_source/de.md
FreezyLemon: I dislike all options. Option 1 at least doesn't add a fragment that is stale or will become stale without a later change.

The (dis)advantages of 2 and 3 are evident. Adding an English fragment in the German translation (option 2) feels wrong. And guessing whatever the fragment might be (option 3) also feels wrong. So my instinct was to not translate these at all, but I don't care very much.

Walavouchey: option 2 is best for the moment, option 3 would be tracked as an error in tooling although it'd feel a bit weird because of that, and option 1 will leave a missing fragment that will be forgotten about

The-Last-Cookie (approved):

Hiviexd (approved):

PR #10420, opened 2023-11-09
[DE] Add `TAG`

FreezyLemon:
## Self-check


The-Last-Cookie:
- [ ] Requesting self-review. I'm not sure exactly when I get to do a review on this one.

FreezyLemon:
We could introduce a new term for `Comboset` that is only used in wiki articles when being specific or trying to disambiguate. not sure it's needed for this article though

The-Last-Cookie:
Yeah, I agree it's sufficient what we have now. The context should be clear in this case.

TicClick (approved):

PR #10426, opened 2023-11-10
Add custom IDs to beatmap categories

Swekka:
Required for https://github.com/ppy/osu-web/pull/10722

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

SKIP_REMARK SKIP_WIKILINK_CHECK SKIP_OUTDATED_CHECK

PR #10427, opened 2023-11-10
Add custom ids to beatmap categories

Swekka:
Required for ppy/osu-web/pull/10722

## Self-check


agatemosu:
if this applies only to English, the pr of osu-web remains the same 🤷‍♀️

Swekka:
That's true but I still think I probably shouldn't be editing articles of languages I don't speak. Let's wait for more opinions on this.

TicClick:
your edits don't change the translations -- I see no issue with your first PR, https://github.com/ppy/osu-wiki/pull/10426
(if anything I'm surprised by how few links were broken even though I expected a lot of articles to lead to `Beatmap/Category`)

Walavouchey:
> if anything I'm surprised by how few links were broken even though I expected a lot of articles to lead to `Beatmap/Category`

@TicClick well ci doesn't check unedited files

PR #10455, opened 2023-11-13
Update `osu! wiki maintainers`

iSlodinxOsu:
## Self-check

- Changed Inactive to Active since clayton is active right now.
👀


Walavouchey:
squashed commits

cl8n:
i just have a bad habit of checking more than i originally was involved in lol
i wont look active when those things are wrapped up (soon)

PR #10466, opened 2023-11-14
Replace trailing space line breaks with backslash line breaks and remove all other trailing whitespace

Swekka:
## Self-check


peppy:
Please provide the commands used to perform this operation.

peppy (changes requested):
Need automation

PR #10484, opened 2023-11-18
[DE] Add `Music theory/Layer`

The-Last-Cookie:
## Notes

I watched [What is Texture? - Music Theory](https://www.youtube.com/watch?v=rTKTpY6PJL8), but decided against using `Textur`[^typo] in the article at all in the end. It might be the correct word in German for `layer`, though I'm not entirely sure.

Both [range](https://en.wikipedia.org/wiki/Range_(music)) and [voice type](https://en.wikipedia.org/wiki/Voice_type) is `Stimmlage` in German. Thus I needed to adjust the corresponding sentence a bit.

## Self-check

[^typo]: For non-Germans: `Textur` is not a typo. It is written without an e at the end.


TicClick (approved):

PR #10495, opened 2023-11-19
Update `Global Moderation Team`

iSlodinxOsu:
## Self-check


aceticke:
A lot of this seems unconfirmed and inaccurate past basic team updates. For example, mangomizer doesn’t do any chat moderation, nor TKS nor 0x84f I believe? Adding languages should be done with confirmation first too, as it can be troublesome for the GMT themselves when dealing with something as serious as moderation.

iSlodinxOsu:
> A lot of this seems unconfirmed and inaccurate past basic team updates. For example, mangomizer doesn’t do any chat moderation, nor TKS nor 0x84f I believe? Adding languages should be done with confirmation first too, as it can be troublesome for the GMT themselves when dealing with something as serious as moderation.

yeah i apologize about it, perhaps it's the best to mention who shouldn't be on the list rn, i would like to fix it.

iSlodinxOsu:
i'll close the PR actually, sorry for the confusion!

PR #10513, opened 2023-11-21
Add "SKIP_REMARK" PR description tag

cl8n:
occasionally useful. feels like a weird way to go about skipping checks but this is the same system that we have in place to skip some others already so...


.github/workflows/continuous-integration.yml
```diff
@@ -56,6 +56,7 @@ jobs:
         run: npm ci
 
       - name: run remark on changed files
+        if: ${{ !contains(github.event.pull_request.body, 'SKIP_REMARK') }}
```
Walavouchey, line 59: mind either using `env` like the other skippable checks or make them also use this `if` thing

Walavouchey (approved):

PR #10521, opened 2023-11-22
Use stricter yamllint config

cl8n:
seems more in line with what we end up editing so might as well

- [x] osu-wiki-tools update https://github.com/ppy/osu-wiki/pull/10523

should be easy to review one commit at a time. the largest changes are simple find+replace

SKIP_REMARK SKIP_OUTDATED_CHECK SKIP_WIKILINK_CHECK


TicClick (dismissed):
lgtm for everything except `news/`

wiki/News_styling_criteria/en.md
```diff
@@ -69,7 +69,7 @@ Content
 ```
 
 - `layout` must always be set to `post`.
-- `title` must represent the article's full title. Markdown formatting is disallowed in this string. News post titles, unlike wiki article titles and other headings, should be in title case.
+- `title` must represent the article's full title. Markdown formatting is disallowed in this string. News post titles, unlike wiki article titles and other headings, should be in title case. Note that surrounding quotes may be necessary if the title contains colons (`:`) or number signs (`#`).
```
Walavouchey, line 72:
```suggestion
- `title` must represent the article's full title. Markdown formatting is disallowed in this string. News post titles, unlike wiki article titles and other headings, should be in title case. Note that surrounding quotes (`"`) may be necessary if the title contains colons (`:`) or number signs (`#`).
```

run-checks.sh
```diff
@@ -106,8 +106,7 @@ function main() {
   _test_wrapper "file size" "_docker bash scripts/ci/inspect_file_sizes.sh --target" "${interesting_files}"
   _test_wrapper "article style" "_docker bash scripts/ci/run_remark.sh --target" "${interesting_articles}"
 
-  yamllint_target_files=$( echo "${interesting_files}" | grep -e .yaml$ -e .yml$ -e .md$ )
-  _test_wrapper "YAML style" "_docker osu-wiki-tools check-yaml --config .yamllint.yaml --target" "${yamllint_target_files}"
+  _test_wrapper "YAML style" "_docker osu-wiki-tools check-yaml --target" .
```
Walavouchey, line 109:
```suggestion
  _test_wrapper "YAML style" "_docker osu-wiki-tools check-yaml"
```

Walavouchey (commented):
news look ok

run-checks.sh
cl8n: _test_wrapper wants a third "files" argument so it's written like this, I'll rewrite it better in a later pr.

.github/workflows/manual-sync.yml
```diff
@@ -1,6 +1,6 @@
 name: Sync with ppy:master
 
-on:
+"on":
```
peppy, line 3: I've never seen this style used in a yaml file for github workflows or otherwise. What's the deal.

cl8n: it's a weird gotcha that yaml keys can be parsed as other types and so this can be the boolean "true", which is why yamllint recommends quoting them like this. but checking again, I misunderstood which yaml parsers are used in our projects and it looks like this doesn't cause any problems so I'll change it back

Walavouchey (approved):

peppy:
@Walavouchey in the future could you please not approve/merge when I've left a review on a PR? It's generally bad manners and you should wait for reapproval from all involved parties before going ahead with a merge.

Walavouchey:
while i'll take that into account, we usually make this clear by self-requesting review or indicating in some way. i can't read your mind

peppy:
I think you might be misunderstanding. I already left a review.

CleanShot 2023-11-24 at 04 12 54


You can't "re-request review" when you're asleep, and someone dismisses your review and approves/merges without consultation.

PR #10526, opened 2023-11-23
Fix remaining link errors and enable link checking in all non-outdated articles

Walavouchey:
- [x] depends on #10525
- [x] another osu-wiki-tools bump to fix bug https://github.com/ppy/osu-wiki/pull/10565
- [x] another bug https://github.com/ppy/osu-wiki/pull/10569

now that link errors are so few we can enable more immediate and comprehensive feedback when someone renames headings or moves articles

SKIP_REMARK
SKIP_OUTDATED_CHECK


.github/workflows/continuous-integration.yml
```diff
@@ -80,9 +80,7 @@ jobs:
       - name: find broken wikilinks
         if: ${{ !contains(github.event.pull_request.body, 'SKIP_WIKILINK_CHECK') }}
         shell: bash
```
cl8n, line 82:
```suggestion
```

cl8n (changes requested):

run-checks.sh
```diff
@@ -108,7 +108,8 @@ function main() {
 
   _test_wrapper "YAML style" "_docker osu-wiki-tools check-yaml --target" .
 
-  _test_wrapper "link" "_docker osu-wiki-tools check-links --target" "${interesting_articles}"
+  # TODO: _test_wrapper expects a third argument but luckily --all overrides --target
+  _test_wrapper "link" "_docker osu-wiki-tools check-links --all --target" "${interesting_articles}"
```
TicClick, line 112: if the intention is to rework `_test_wrapper` so that it supports two arguments instead of three, or either two or three, it's better to say that explicitly rather than state a fact

TicClick (changes requested):
all good except for a single thing

run-checks.sh
Walavouchey:
```suggestion
  # TODO: _test_wrapper expects a third argument but luckily --all overrides --target. please rewrite to make this prettier
  _test_wrapper "link" "_docker osu-wiki-tools check-links --all --target" "${interesting_articles}"
```

Walavouchey:
indeed and the fix is awaiting osu-wiki-tools deployment (been slightly busy this weekend)

cl8n (approved):
changes look good, imo just skip the check and merge this one but I'll let you decide.

peppy:
> changes look good, imo just skip the check and merge this one but I'll let you decide.

won't it continue to trigger false positives if we do that?

cl8n:
yeah but it's already broken. I guess the scope of the check is expanded here though so it would be more annoying than before. anyway i will just be patient for walas update

cl8n (dismissed):

cl8n (approved):

PR #10546, opened 2023-11-26
Fix incorrect image file extensions (.png/.jpg)

FreezyLemon:
I noticed that some images are using the wrong file extensions (PNGs named .jpg, JPEGs named .png). This isn't that bad, all modern image viewers will detect the format based on the content, but it should still be fixed for obvious reasons.

I threw together a little script to find them (needs bash and [mediainfo](https://mediaarea.net/en/MediaInfo)):
```bash
#!/bin/bash
set -eu

shopt -s globstar

old_ext="$1"
expected_type="$2"

for f in **/*$old_ext; do
  t=$(mediainfo --Output='Image;%Format%' "$f")
  if [ -z "$t" ]; then
    # Some PNGs are considered a video, for some reason
    t=$(mediainfo --Output='Video;%Format%' "$f")
  fi

  if [ -z "$t" ]; then
    echo "Error: '$f' does not have a recognisable file type"
  elif [ ! "$t" = "$expected_type" ]; then
    echo "Mismatch: '$f' has type $t, not $expected_type"
  fi
done
```

It was quick enough to rename the files manually, so the script only prints the warnings.

SKIP_WIKILINK_CHECK (avoiding a current bug in osu-wiki-tools, I verified with older working versions that everything's good here -clayton)
SKIP_OUTDATED_CHECK


FreezyLemon:
Just got reminded that I still need to edit links to the images, I'll do that and then convert back.

peppy:
Any reason this is draft?

FreezyLemon:
There are still some CI failures which are caused by a bug in the CI tooling. The PR itself is done, but instead of skipping the CI checks, I'd rather wait until the tooling is fixed and then go for merging

cl8n (approved):

cl8n:
the bug in question strictly adds false errors, and doesn't silence any real ones, so you can ignore them. seems you have already fixed all of the links here.

peppy (approved):

PR #10560, opened 2023-11-29
Fix kamome sano and kanemiko's entries in `Content usage permissions`

pishifat:
i made a very bad typo and it made it here

context:
image2

and confirmation for the replacement
image (1)


Sitoria:
little bit offtopic: how about new FAs not on list?  insert it all at once on someday? or waiting for artist's response?

cl8n (approved):

peppy (approved):

PR #10561, opened 2023-11-29
Rewrite local test runner

cl8n:
- [ ] depends on #10557
- [ ] merge #10526 first for easier conflict resolve

I need to test this more but I'm just going to wait until osu-wiki-tools is fixed.

- Fix docker permission issues with files created in the container
- Run the docker container just once, instead of for each test
- More clear log and usage output (imo)
- Remove `bash` dependency from host
- Correct some of the recently-kind-of-broken test commands that I left for this PR
- Split run_checks into a few different script files that can be used on their own
- Generally more hardened script for weird edge cases


meta/check-all.sh
```diff
@@ -0,0 +1,61 @@
...
+  } | sort -u
+)"
+
+printf '%s\n' "$changed_files" | run_test 'file size' xargs -r meta/check-file-sizes.sh
+printf '%s\n' "$changed_files" | grep '\.md$' | run_test Remark xargs -r meta/remark.sh
+run_test 'YAML style' osu-wiki-tools check-yaml
+run_test link osu-wiki-tools check-links --all
+
+if test -n "$(git log master..)"; then
+  run_test 'outdated article' osu-wiki-tools check-outdated-articles
```
cl8n, line 60: note for myself this should have `--workflow` on it, but i cant add for same reason as in #10570

PR #10567, opened 2023-11-30
Uncomment `no_native_review_since` tags and explain in ASC

cl8n:
it will cause some extra info to be included on https://osu.wiki soon -- for now it is functionally the same as `no_native_review: true`

### DIFF
--- a/wiki/Article_styling_criteria/Formatting/en.md
+++ b/wiki/Article_styling_criteria/Formatting/en.md
@@ -106,12 +106,18 @@ tags:
 
 ### Translations without reviews
 
-*Note: Wiki maintainers will determine and apply this mark prior to merging.*
+*Note: Wiki maintainers will determine and apply this tag prior to merging.*
 
-Sometimes, translations are added to the wiki without review from other native speakers of the language. In this case, the `no_native_review` mark is added to let future translators know that it may need to be checked again. This tag must be written as shown below:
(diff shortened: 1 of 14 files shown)


wiki/Article_styling_criteria/Formatting/en.md
```diff
@@ -106,14 +106,20 @@ tags:
...
+Sometimes, translations are added to the wiki without review from other native speakers of the language. In this case, the `no_native_review` tag is added to let future translators know that it may need to be checked again:
 
 ```yaml
 no_native_review: true
 ```
 
+Alternatively, the `no_native_review_since` tag can be used to indicate the commit where the first unreviewed edit was introduced, as well as who edited the translation since then:
+
+```yaml
+no_native_review_since: 5539d9e8c943605a7be186dc3f5ab10569275b05 Doryan
```
Walavouchey, line 120: is there a functional use for the name? can't this just be checked from the commit history (with the commit to look at already being provided right here)?

cl8n: i was thinking that the credits may not necessarily align with the author or committer, like if applied as a suggestion on someone else's PR or something. and it seems like it could be helpful context to show upfront but I can't really say how translators will or won't refer to this. it's not really any more effort to write this down with the commit hash so I think it's okay to keep, but i also dont really care that much it can be removed if you think it's unnecessary

Walavouchey (approved):

PR #10577, opened 2023-12-01
Fix base commit of outdated articles check

cl8n:
`HEAD` is the right base commit here because it's a merge with the first parent at `master` and second parent at the tip of PR, and then `--outdated-since` was already correct, but I have to write it in explicitly now since it differs from base commit.

will see if i can make this less confusing to think about later...


Walavouchey (approved):
`--base-commit` coming after `--outdated-since` looks very backwards but if it works it works i guess

PR #10668, opened 2023-12-18
Update BN and NAT member listings

Skubers:
SKIP_OUTDATED_CHECK


cl8n:
where did you get the languages for the new bns? they don't match up very well with what's listed on mappersguild website

Skubers:
I didnt notice that there are languages on MG website, I just went with the countries states on their profiles

cl8n (changes requested):
aceticke will review but i cant assign cuz github dumb

aceticke (approved):
Looks fine but for a sanity check I'll run my script to pass over the URLs to make sure they are accurate. Also not sure about the reordering of languages in the NAT page, from how I've done it, it's been main language followed by subsequent languages, or sorted by fluency. @cl8n what do you think?

cl8n:
well I sorted them now based on what mappersguild shows but idk if it's "better"
I'm just gonna go with this because I feel like these two sources for bn languages should be in sync... and makes updating easier

cl8n (approved):

PR #10670, opened 2023-12-18
Add missing symbol in `osu!(lazer) Updates & Preparing For Ranked Play` news post

iSlodinxOsu:
## Self-check


Walavouchey:
it's already correct

PR #10685, opened 2023-12-23
Fix table formatting remark warnings

agatemosu:
## Self-check

- `SKIP_OUTDATED_CHECK` because some articles doesn't have that extra space


Walavouchey:
please post what script you've used for these changes

cl8n:
the list changes aren't quite right as explained in that long discord msg I mentioned you in earlier. the table is part of the list item and so should be indented. you can choose whether the list is spread (empty lines between each list item + block child of list item) or not (no empty lines) but probably keep spread since that's what they were already.

as for the other table formatting changes, they are not wrong and I'll merge them but I feel like I've already told you that this type of PR that can be automated by remark already is not very helpful...

agatemosu:
> please post what script you've used for these changes

I used the VS Code search and replaced manually

agatemosu:
> the list changes aren't quite right as explained in that long discord msg I mentioned you in earlier. the table is part of the list item and so should be indented. you can choose whether the list is spread (empty lines between each list item + block child of list item) or not (no empty lines) but probably keep spread since that's what they were already.

reverted

agatemosu:
I force-pushed because I wanted to see the diff of github desktop to make it faster

cl8n (approved):

Walavouchey (approved):

PR #10705, opened 2023-12-25
Improve readability of osu!taiko hit window formulae in `Beatmap/Overall difficulty`

yaggod:
For some reason osu!taiko OD formulas were weird formatted, especially the "great" one which was splitted in three formulas for different OD values despite being literally the same formulas. I haven't changed formulas for languages other than russian and english since i am not sure about its grammary


TicClick (dismissed):
seems to check out -- @Walavouchey is there a reason you wrote them like this?

Walavouchey:
when i added this in https://github.com/ppy/osu-wiki/pull/7290 i took it from lazer without simplifying after spaceman confirmed it was the same as stable
then, when i updated the judgement articles https://github.com/ppy/osu-wiki/pull/8245 i checked all formulas with stable and simplified them, but i didn't touch this article.

TicClick (approved):

PR #6058, opened 2021-09-04
[BG] Add `Main_Page`

YDu8VchU:
Bulgarian translation for wiki main page (bg.md)

PR #10754, opened 2024-01-01
[BG] Update `Main_page`

YDu8VchU:
## Self-check


TicClick (approved):
readded the `no_native_review` tag since the original article is also by you https://github.com/ppy/osu-wiki/pull/6058

PR #10760, opened 2024-01-05
Update yearly count in `Featured Artists`

pishifat:
i added a note for stub listings which didnt receive a proper announcement (like https://osu.ppy.sh/beatmaps/artists/393), but i'm not sure if that's a good way to handle it. input appreciated

### DIFF
--- a/wiki/People/Featured_Artists/en.md
+++ b/wiki/People/Featured_Artists/en.md
@@ -105,4 +105,4 @@ The [Featured Artist group page](https://osu.ppy.sh/groups/35) lists all such ac
 
 ## Notes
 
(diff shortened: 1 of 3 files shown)


cl8n:
are the 5 extra listings included in that 64 total? answer this question in the note as well :^)

otherwise seems good

Walavouchey:
not every new artist is tied to their own announcement (e.g. with one announcement of osu! originals from several artists) and not every announcement is tied to a specific artist (e.g. corsace or other labels)

this is more of just a case of the former and tbh i'm not too fond of tracking exceptions like this. i'd rather count them normally

cl8n (approved):

Walavouchey (approved):

PR #10799, opened 2024-01-16
Add "BREACH PROTOCOL" to the list of `osu! originals`

Nivalyx:
From the latest "[talking about osu!'s world cup grand finals tiebreakers](http://www.youtube.com/watch?v=0B6TyWML0zg)" video.

image

## Self-check


wiki/osu!_originals/en.md
```diff
@@ -990,3 +996,4 @@ These songs were later [announced](https://osu.ppy.sh/home/news/2023-02-01-twin-
 [^fa]: This song is made by a Featured Artist, but is not part of the Featured Artist catalogue.
 [^fa-feature]: This song features a Featured Artist, but is not part of the Featured Artist catalogue.
 [^fa-catalogue]: This song is part of the Featured Artist catalogue.
+[^breach-protocol-note]: Unlike other World Cup originals, this song is commissioned by a third-party for a purpose unrelated to osu!. However, it ended up debuting in the osu!mania 4K World Cup 2019 due to circumstances at the time.
```
Walavouchey, line 999: what is this supposed to mean? i'd rather confirm with the artist
other than waltertoro, there's no source for this

Nivalyx: ah yeah it is what it is as WalterToro said — the song ended up premiering first in osu! due to circumstances at the time

i feel like it's worth adding to the page tho, since by definition an "osu! original" is ["songs that we've commissioned or had made for us or that have otherwise appeared first on osu!"](https://discord.com/channels/188630481301012481/218677502141399041/1023076686851485767)

Nivalyx: this is a comment from [-mint-](http://osu.ppy.sh/users/8976576) (mwc 4k 2023's head of mappool selector) from the video btw, just in case it's worth as a consideration material

image

Walavouchey: if we list it i don't think the footnote should be necessary, and i wouldn't want to keep adding random footnotes for minor details

Nivalyx: ah, okay, that's fair enough i suppose

removed the footnote

TicClick (approved):

PR #10800, opened 2024-01-16
Update `Countries that participated in osu! tournaments`

iSlodinxOsu:
## Self-check

Feel free to comment something if i mess something up.


Walavouchey (commented):
you should probably update the comment on line 10

iSlodinxOsu:
Will update the translations once i have time.

cl8n (approved):

cl8n:
well I will be honest I didn't check all of the data because this seems like a nightmare. I did add a column for MWC 2022 4K, which was missing. this is probably a just-automate-it type of page if we want to keep supporting it

as for the translations, I did just automate them (`osu-wiki translate-tournament-countries`), it should hold up for at least the most basic edits to this page like adding new tournaments

PR #10815, opened 2024-01-19
Add `Simplified ranking criteria`

pishifat:
the ranking criteria is too long and dense for anyone to reasonably read (which makes sense, since it's supposed to cover all potential issues and exceptions to those issues in an airtight way)

this article is a readable version of the ranking criteria that only includes
- concise rc issues that aren't too niche
- subjective stuff to think about when making a rankable map

it's not intended to cover everything on the ranking criteria, and it's not intended to have loophole-free wording. it's for common sense readers to use sorta like a rankability checklist

still missing:

- [x] taiko criteria
- [x] catch criteria
- [x] mania criteria

PR #10825, opened 2024-01-20
Misc. Project Loved updates

cl8n:
cleaning up a few random thigns i noticed and modernizing some of the requirements


wiki/Community/Project_Loved/en.md
```diff
@@ -23,32 +23,25 @@ Creators of said beatmaps are also able to request them to be removed from the L
...
-- [Some points from the Metadata section of the ranking criteria](/wiki/Ranking_criteria#metadata):
-  - Song title, artist, and source should be correct in all cases.
-  - Beatmap tags should properly list all contributors (guest difficulty creators, storyboarders, etc.).
-- [Song Content Rules](/wiki/Rules/Song_content_rules)
-- [Visual Content Considerations](/wiki/Rules/Visual_content_considerations)
-
-Additionally, in order to keep the audio quality of the song at a reasonable level, beatmap audio files should have an average bitrate of at least 128 kbps.
-
-Any problems with these rules are usually sorted out while the beatmap is up for voting.
+Contributors to Project Loved often suggest technical changes to maps before they are Loved, such as improvements to audio quality, [timing](/wiki/Beatmapping/Timing) accuracy, or [metadata standardisation as suggested by the ranking criteria](/wiki/Ranking_criteria#metadata). The mapper is strongly encouraged to apply these changes, but they are not required to.
```
aceticke, line 36: should add back in the song content rules and visual content considerations, these are not part of the metadata check but instead the moderation check and are required changes before the map can be loved, unless the osu! team decides otherwise.

wiki/Community/Project_Loved/en.md
```diff
@@ -23,32 +23,25 @@ Creators of said beatmaps are also able to request them to be removed from the L
...
 
 ### osu!-specific criteria
 
-Starting from February 2023, beatmaps of the osu! game mode considered for Loved will follow these new categories as well as the rules and guidelines below, which are similar to the ranking criteria, but adapted for the more lenient process Project Loved is known for.
+Beatmaps of the osu! game mode considered for Loved follow these categories as well as the rules and guidelines below, which are similar to the [ranking criteria](/wiki/Ranking_criteria), but adapted for the more lenient process Project Loved is known for.
```
aceticke, line 40:
```suggestion
Beatmaps of the osu! game mode considered for Loved follow the following categories as well as the rules and guidelines below, which are similar to the [ranking criteria](/wiki/Ranking_criteria), but adapted for the more lenient process Project Loved is known for.
```
could be wrong, but "these" can mean that the categories can be before or after the statement, i think "the following" is better for clarity?

wiki/Community/Project_Loved/en.md
```diff
@@ -64,20 +57,14 @@ Every month, beatmaps are chosen according to 9 categories, with the goal of ens
...
+For any questions regarding what goes in each category, contact the [osu! Loved captains](https://osu.ppy.sh/groups/31?mode=osu).
 
 #### Rules
 
 **All rules are exactly that: rules.** They are not guidelines and may not be broken under any circumstance.
 
-- **The beatmap must abide by the [beatmap submission rules](/wiki/Rules#beatmap-submission-rules).**
-- **The beatmap must not be currently in the Qualified, Ranked or Approved categories.** Ideally, a map should not be going for Ranked and Loved at the same time, as they are mutually exclusive.
 - **The beatmap must be submitted at least 1 year before Loved nomination, except for beatmaps that break ranking criteria due to their concept.** Submitting a beatmap solely for Loved is discouraged when Ranked might be a better option. Exceptions include sliderator usage, Aspire maps, storyboard gameplay gimmicks, etc.
 - **The beatmap must have at least 30 favourites.** This is a minimum threshold to indicate a positive and existing community reaction to the beatmap. Brigading for favourites is discouraged, although captains cannot verify whether that happens.
```
aceticke, line 67:
```suggestion
- **The beatmap must have at least 30 favourites.** This is a minimum threshold to indicate a positive and existing community reaction to the beatmap. Brigading for favourites is discouraged, although captains cannot verify whether that happens in most cases.
```
sometimes its extremely obvious or the effort to do so has been made public

wiki/Community/Project_Loved/en.md
```diff
@@ -64,20 +57,14 @@ Every month, beatmaps are chosen according to 9 categories, with the goal of ens
...
-- **The beatmap host cannot be restricted.** Guest difficulties from restricted users will be judged on a case-by-case basis once the map is considered for nomination by a Loved captain.
-- **At least one difficulty of the beatmap relevant to the Loved submission process must not auto-fail.** Leaderboards are pointless for beatmaps with broken score submission. For the same reason, this also forbids beatmaps that are relevant solely because of an unpassable difficulty, even if they have other difficulties. If the beatmap has a difficulty that auto-fails, a new difficulty with the problematic hit objects removed needs to be also included, preferably removing the original.
-- **Every difficulty of a beatmap must have a minimum drain time of 20 seconds.** This is lower than the 30-second minimum of the ranking criteria for the sake of being less limiting.
 - **A beatmap must have at least some form of mapping contribution from the person submitting it.** This specifically avoids "beatmap compilations" which, even with the original mappers' permissions, do not represent the kind of original work showcased in Loved.
```
aceticke, line 68: clarify? this would only apply to pending, wip or graveyarded beatmaps in the compilations, since ranked, loved or approved maps are disallowed by principle of the rules already mentioned above.

aceticke (changes requested):

wiki/Community/Project_Loved/en.md
cl8n: those are part of submission rules

wiki/Community/Project_Loved/en.md
cl8n: resolved ondiscord

cl8n:
both of those are in the osu section tho

wiki/Community/Project_Loved/en.md
cl8n: Nvm changed this part to clarify

huuishuu:
> both of those are in the osu section tho

Are the rules & guidelines not meant to be universal for nominations? Because that implies that the rules do not apply to any other mode aside from osu!

cl8n:
yeah the "Rules" and "Guidelines" lists are both under osu section, no other mode wrote that kind of thing. I moved the ones we do actually apply to all modes up to the "Eligibility" section

huuishuu:
> yeah the "Rules" and "Guidelines" lists are both under osu section, no other mode wrote that kind of thing. I moved the ones we do actually apply to all modes up to the "Eligibility" section

I don't like how these points then really imply that they don't apply to other modes, since they're only listed under the osu! section, despite the fact that these do get picked up on other modes as well.

- Beatmap metadata should be as accurate as possible. Beatmaps that are nominated for Loved voting go through a final metadata check by specialised members of the community. Correct metadata makes searching beatmaps easier when looking for a song. When in doubt, contact a Loved captain.

- All difficulties in the beatmap should be finished. This means reaching the start and end point of the song, excluding unmappable sections, intros, or outros. For example, in a 5-minute song with multiple difficulties, if the Extra and Hard difficulties are 4:30 in length, the Insane cannot be 1 minute long. If all difficulties have large amounts of the song left unmapped, the audio file should be cut accordingly.

Perhaps this should be talked over with the other mode captain teams as well. Metadata checks are done for *every* nominated map and unfinished difficulties do get excluded from voting in other modes as well.

cl8n:
in general the osu rules feel verbose to me, and yes other gamemodes may be lacking stuff written here (I'm not sure), all of that is for some other time though. not trying to cover the entire article with this pr

MChecaH:
[Gave my two cents back when it was proposed.](https://discord.com/channels/188630481301012481/384640068847009792/1055465687532978246)

In general, I don't really like the approach that was taken to have a *"Ranking Criteria v2"*, mainly because it falls on the same pitfalls as the RC does by not being all that approachable.

The RC is actively being simplified in two different ways, and if we want to keep these rules we may want to approach it the same was as [the current attempt at simplifying the RC](https://github.com/ppy/osu-wiki/pull/10815) is doing.

Maybe having a separate article that goes over the eligibility for all modes could work? Regardless, I think these should need to be simplified in some way, shape, or form.

wiki/Community/Project_Loved/en.md
```diff
@@ -23,32 +23,29 @@ Creators of said beatmaps are also able to request them to be removed from the L
...
+- It is not in the [Ranked](/wiki/Beatmap/Category#ranked), [Approved](/wiki/Beatmap/Category#approved), or [Qualified](/wiki/Beatmap/Category#qualified) categories
 - Its creator does not disapprove of it being Loved
-- Its creator is not restricted
+  - If any guest contributors disapprove of their part being Loved, their part will be excluded from Loved by appropriate means. In some cases, such as [collabs](/wiki/Beatmap/Beatmap_collaborations), this may make the entire beatmap ineligible for Loved.
+- Its creator is not [restricted](/wiki/Help_centre/Account_restrictions)
 - At least one difficulty can be passed
+- Every nominated difficulty has a [drain time](/wiki/Beatmap/Drain_time) of at least 20 seconds
 
-Although many Pending and Work-in-progress beatmaps meet these criteria, few are voted into the Loved category due to the more selective nomination and voting requirements.
+Although many [Work in Progress and Pending](/wiki/Beatmap/Category#wip-and-pending) beatmaps meet these criteria, few are voted into the Loved category due to the selective nature of the captains' picks, and the high community approval requirement by voting.
```
aceticke, line 34: hmm maybe this could be simplified to newer maps, feel like people make minor changes to maps all the time to keep them in pending for attention

aceticke (changes requested):
good apart from one minor thing

wiki/Community/Project_Loved/en.md
cl8n: i dont think it was meant to imply anything about the age of the map, it's just missing "graveyard" for some reason, i will fix that instead

cl8n:
@MChecaH i think cutting it down is the way to go, but as I said to huu as well let's do that later -- just want to get some small changes in for this PR

aceticke (approved):

TicClick (approved):

PR #10840, opened 2024-01-25
Fix small mistake in `History of osu! 2007`

Tanza3D:
For the first change: Judging by the fact that the song select has the "Play Level", it's easy to make the assumption that this is in fact the level select itself, and not just the edit level select. This is also proven by [this version on Osekai Snapshots](https://osekai.net/snapshots/?version=81)

For the second change: For the version shown in the image just above, the same selection carousel was used inside the edit menu, once again shown in [this version](https://osekai.net/snapshots/?version=185), and also judging by the fact that this UI was never used solely as the edit level select. This orange version was however used for a few days as the primary song select, still, so unsure about entirely removing it.

First actual PR on osu-wiki so hope I've done things right here :)

## Self-check


wiki/History_of_osu!/2007/es.md
agatemosu, line 1: Selección de canciones

wiki/History_of_osu!/2007/fr.md
agatemosu, line 1: Sélection de musiques

agatemosu (commented):
spanish and french can be changed to that

cl8n:
the orange one should probably not be deleted here, just moved if necessary and given a more accurate description

wiki/History_of_osu!/2007/id.md
JuanM4T, line 1: Pemilihan Lagu

wiki/History_of_osu!/2007/ru.md
JuanM4T, line 1: Выбор песни

wiki/History_of_osu!/2007/pt-br.md
JuanM4T, line 1: Song Selection

wiki/History_of_osu!/2007/nl.md
JuanM4T, line 1: Muziekselectie

JuanM4T (changes requested):
the other languages can also be swapped with those.

idk why Portuguese keeps the English but I've kept it in English for coherence's sake with the original.

TicClick (approved):

PR #10842, opened 2024-01-26
Fix incorrect format for UK flags in `o!RT` tournament page

NiceAesth:
## Self-check

Sorry for this one, I had only just noticed it.


TicClick (approved):
and this is why I should really start running [my own damn flag checker](https://github.com/TicClick/scissors) on pull requests I review

Issue #2589, opened 2019-09-20
Reformatting skin element list pages


dorothy3242:
![image](https://user-images.githubusercontent.com/33286824/65292950-6bc83080-db27-11e9-9c3c-d000e96825ef.png)

Current organization of skin elements is very tedious to parse, probably need to aim for more tables organized similar to old wiki.

cl8n:
did u want to do this urself ?

sr229:
@dorothy3242 I suggest when opening issues like this, you have to implement fixes yourself since some of us are busy in other areas of the wiki.

peppy:
That's not how github issues work.

RockRoller01:
So, I have given this a try and run into some formatting problems. You can not put multiple lines into a single field, making it so that you can not put the notes into the table at all. This is fine for things like e.g. the default numbers since they all share the same notes
![image](https://user-images.githubusercontent.com/35633093/108159357-22e32580-70e7-11eb-9a97-9bd6bd3751c2.png)
But for things like the hitcircle elements its a major problem, leaving you with an essay long list
![image](https://user-images.githubusercontent.com/35633093/108159338-18c12700-70e7-11eb-9ef2-d3a00fd0d4a3.png)

If anyone got an idea on how to deal with this please let me know, otherwise I will drop this, since this would just transform the article from one mess into another

PR #10863, opened 2024-01-30
Clean up format of `Skinning/Sounds` and add Nightcore sample info

CloneWith:
This might be part of the solution of #2589 

## Self-check


Walavouchey (commented):
- why are only some of these in tables? the ones which have additional information could have a "Notes" column for instance
- expand contractions ("aren't" --> "are not", etc.)
- "osu! client", "osu! cookie" instead of "osu!client" and "osu!cookie"
- the article should probably be renamed to something more grammatically correct than "Sounds skinning"

will check in more detail later

CloneWith:
> why are only some of these in tables?
Some of sounds are similar in filename and usage, then I tried to merge them into one line, for example the hitsound sets and the `key-press-*` set.

> "osu! client", "osu! cookie" instead of "osu!client" and "osu!cookie"
> the article should probably be renamed to something more grammatically correct than "Sounds skinning"
Just kept them without changes when rewriting, will change later.

cl8n:
I have very far down a todo list that I wanted to write about these files:
> nightcore-kick.wav
> nightcore-hat.wav
> nightcore-clap.wav
> nightcore-finish.wav

idk if I'll ever get to it but maybe it's relevant to do in this PR.

CloneWith:
Have read osu! lazer source code and tried on stable by myself. It's really odd that I didn't find `nightcore-hat.wav` being used in stable (maybe?) but did in lazer, and the [corresponding wiki page](https://osu.ppy.sh/wiki/Nightcore)) seems unclear about this, not even mentioning it.

Don't know

cl8n:
in stable
- the files are sourced from your skin, but not beatmap skin
- nightcore-hat plays only when slidertickrate is a multiple of 2
- nightcore-finish plays only when not omitting the first barline
- non-4/4 works differently, to be hoenst it's so confusing and these skin elements are so niche that I wouldn't even bother documenting non-4/4

otherwise it works same as lazer

cl8n (changes requested):
rest of the changes seem fine, there's still a lot to improve about this article but the new layout here is certainly better

just fix NC and that's good for this PR

cl8n (approved):

PR #10867, opened 2024-01-30
[PT-BR] Add upppy

flavin27:
## Self-check


TicClick:
hi, I have cleared the second checkbox since I don't see a review from a Portuguese-speaking person. whether you are looking for one or not, let me know -- I'll either merge it in as-is with `no_native_review: true`, or wait until a native speaker approves your work

TicClick (approved):

PR #10870, opened 2024-01-31
Fix dead image link in `osu!monthly, November 2022` news post

0x84f:
monthly 1: privated video
monthly 5: dead banner embed
monthly 8: dead VOD link

speaking to wala about whether the HTML tags need to be reverted to symbols or not (looks like they work as tags on osu-web)
Example:
image
https://osu1.roanh.dev/home/news/2022-12-01-osumonthly-5


news/2022/2022-12-01-osumonthly-5.md
```diff
@@ -107,8 +107,6 @@ Word on the street is that there will be more Hush-Hush medals coming soon, so i
 
 ### SillyFanGirl's BR eSports nomination
 
-![https://twitter.com/premioesportsbr/status/1593590309631295488?s=20&t=1thfKY6-BpTCFDsbtltDdw](https://pbs.twimg.com/profile_banners/983759809156669440/1663599648/1500x500)
```
Walavouchey, line 110: this is why you don't hotlink images aaaaaaaaa

archive here https://web.archive.org/web/20230108000953im_/https://i.ppy.sh/2e2651e1d1b90751eb22f84336e2eb077016019d/68747470733a2f2f7062732e7477696d672e636f6d2f70726f66696c655f62616e6e6572732f3938333735393830393135363636393434302f313636333539393634382f3135303078353030

news/2022/2022-08-01-osumonthly-1.md
```diff
@@ -62,10 +62,6 @@ Not content to allow osu! to vacuum up all the glory, the members of the osu!tai
 
 Showcasing some of the best drum bashers the community has to offer, EGTS has a storied history of being one of the most competitively and visually impressive contests around, sporting enough dons and kats to even make the Taikonator blush.
 
-The GTS team's insanely talented designer [OsuMe65](https://osu.ppy.sh/users/852867), has also made a trailer for the tournament. Check it out and leave your thoughts down below!
```
Walavouchey, line 65: especially disagree with deleting text like this

Walavouchey (changes requested):
disagree with the html. should be the same as in the wiki (where html doesn't work)

i don't think dead links matter too much for news posts (?) and i'd argue to keep them. news articles are timed publications

news/2022/2022-08-01-osumonthly-1.md
0x84f: Do you think it should be left as is then?

news/2022/2022-12-01-osumonthly-5.md
0x84f: Yea dunno how this got through i was thinking of removing it entirelt but luckily you found an archive. I will add this tomorrow

0x84f:
> disagree with the html. should be the same as in the wiki (where html doesn't work)
> 
> i don't think dead links matter too much for news posts (?) and i'd argue to keep them. news articles are timed publications

Based on your response do you think we should just fix that one banner and call it a day then?

As for the html tags I will change this on browser tomorrow, was just curious

Walavouchey:
> Based on your response do you think we should just fix that one banner and call it a day then?

i think so

Walavouchey (approved):

PR #10872, opened 2024-02-01
Update `Ranking Criteria` to allow red lines to be used for out-of-range slider speeds

quilaosu:
Original forum thread: https://osu.ppy.sh/community/forums/topics/1682839

Initially, this led to plans to change osu!'s code to reduce the minimum base slider velocity from 0.1x to 0.01x. However, that change has been on hold for over a year due to a backwards compatibility issue. (relevant github threads: https://github.com/ppy/osu-stable-issues/issues/1072 -> https://github.com/ppy/osu/pull/22152). Because of this, we'd like to proceed with the RC change for now.

There's currently [a map](https://osu.ppy.sh/beatmapsets/2121566#osu/4458007) waiting to be ranked which uses a red line to make its last slider slow enough.

If peppy doesn't agree with this RC change, the mapper requests a one-time exception to the rule be granted, because the mapper and one of the nominators may not be around when the code changes are implemented.


cl8n:
blocked on needing more activity on the thread. aceticke said they asked nat to check it. probably also needs developer input?

TicClick:
cross-posting [status update from aceticke](https://discord.com/channels/188630481301012481/218677502141399041/1209747195910426675):

> [7:23 AM] aceticke: hello update on https://github.com/ppy/osu-wiki/pull/10872
> [7:23 AM] aceticke: response from the nat is positive leaning, but still overall mixed
> [7:24 AM] aceticke: perferably, most want dev implementation to allow a larger range of inherited sv values (green lines) possibly in lazer 
> [7:25 AM] aceticke: though i think we'd need to see if thats feasible and seek dev opinion on the proposal in general, so ill probably open a discussion on the ppy/osu side
> [7:25 AM] aceticke: safe to say it should stay blocked for now

cl8n:
I'm going to close this because it's been blocked for so long and not active, but feel free to re-open or recreate later if necessary

PR #10878, opened 2024-02-02
Update `Skinning/Interface`

Redo7:
`songselect-top.png` has incomplete information about it's behavior, which is a bit hard to obtain with A/B testing and guesswork.
The missing information isn't very relevant, and would only provide exact values for something which may never end up being used for any creative purpose. The asset can easily be used without it.

## Self-check


TicClick (approved):

TicClick:
just noticed it's waiting for clayton's review

Walavouchey (approved):

PR #10839, opened 2024-01-25
Add `osu!catch Convert Cup Open 2023` wiki article

Nivalyx:
Wiki article for the recently concluded `osu!catch Convert Cup Open 2023` tournament, on the behalf of [Urabe Mikoto](http://osu.ppy.sh/users/1272422).

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #10881, opened 2024-02-03
Add `Beatmap/Converts`

Nivalyx:
noticed that the wiki doesn't have a dedicated page detailing converted beatmaps yet while working on [#10839](http://github.com/ppy/osu-wiki/pull/10839) earlier


wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+tags:
+  - convert beatmaps
+  - converted beatmaps
+---
+
+# Converts
+
+![The \"Show converted\" toggle in osu!(lazer)](img/show-converted-button.png "The \"Show converted\" toggle in osu!(lazer)")
+
+In osu! gameplay, **converts** (or *converted beatmaps*) are [beatmaps](/wiki/Beatmap) of a specific [game mode](/wiki/Game_mode) that have been "converted" by the osu! client in-game in order to be playable in another game mode they were not originally created for. All converts are automatically generated based on certain parameters present in the base beatmap.
```
TicClick, line 11: several things to unpack here:
- "gameplay" has nothing to do with it, and come to think of it, "In osu!" does too
- since only osu! game mode beatmaps can be converts, why not say that instead of "a specific game mode"?
- while the first sentence is factually correct, it gives an impression that these difficulties are permanent (as in, they coexist with regular ones), which isn't the case -- at least to me, they're automatically generated _versions_ of the original difficulties. how about you put focus on the second sentence instead?

wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+
+As it stands, osu! supports conversions of beatmaps from the [osu! game mode](/wiki/Game_mode/osu!) into [osu!taiko](/wiki/Game_mode/osu!taiko), [osu!catch](/wiki/Game_mode/osu!catch), and [osu!mania](/wiki/Game_mode/osu!mania).
+
+## Conversions
+
+### osu! to osu!taiko conversion
+
+*See also: [osu!taiko mapping/osu! conversion notes](/wiki/Game_mode/osu!taiko#osu!-conversion-notes)*
+
+Converts between osu! and osu!taiko are generated according to the rhythm, hitsound, and slider velocity changes in the beatmap. Depending on how the [hit objects](/wiki/Gameplay/Hit_object) are being hitsounded, the conversion between the two game modes generally goes as follows:
```
TicClick, line 21: "converts between X and Y" gives an impression that it could be both way -- why not write "from X to Y" instead?

TicClick: also I would say hitsounds play the most important role here, so one sentence instead of two:
```suggestion
osu!taiko converts are generated according to the placement of hitsounds, rhythm, and slider velocity changes in a beatmap:
```

wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Circles, slider heads, slider tails, and slider repeats that bear no additive hitsounds (e.g. no whistle, clap, or finish) are converted into *red notes* (Don).
+- Circles, slider heads, slider tails, and slider repeats that are hitsounded with a whistle, a clap, or both are converted as *blue notes* (Katu).
+- Circles, slider heads, slider tails, and slider repeats that are hitsounded with a finish are converted into *large red notes*.
+- Circles, slider heads, slider tails, and slider repeats that are hitsounded with a finish AND a whistle, a clap, or both are converted into *large blue notes*.
```
TicClick, line 26: this could be a table, reads better, less repetitive

wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+- Sliders with significant lengths are converted into *drumrolls*.
+- Spinners are converted into *osu!taiko spinners* (dendens/shakers).
```
TicClick, line 28: while we don't have an article on shakers, it's the preferred term (and "spinners are converted into osu!taiko spinners" isn't very well-written)
```suggestion
- Spinners are converted into shakers (dendens).
```

wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+
+### osu! to osu!catch conversion
+
+*See also: [osu!catch mapping/osu! conversion notes](/wiki/Game_mode/osu!catch#osu!-conversion-notes)*
+
+Converts between osu! and osu!catch are generated according to the rhythm and the *horizontal placement* (e.g. the x-axis coordinate) of the [hit objects](/wiki/Gameplay/Hit_object) in the beatmap. The conversion between the two game modes goes as follows:
```
TicClick, line 34:
```suggestion
osu!catch converts are generated according to the rhythm and *horizontal placement* (e.g. the x-axis coordinate) of the [hit objects](/wiki/Gameplay/Hit_object) in a beatmap:
```

wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Circles, slider heads, slider tails, and slider repeats are converted into *fruits*.
+- Slider ticks are converted into *juice drops*.
+- Slider paths are converted into *droplets*.
+- Spinners are converted into *bananas*.
```
TicClick, line 39: this could benefit from links to corresponding articles

wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+
+### osu! to osu!mania conversion
+
+Converts between osu! and osu!mania are generated mostly according to the rhythm of the beatmap, and are supposedly much more random than the other converts in its algorithm. While there are no consensus as to how osu! to osu!mania conversion works precisely, it generally goes as follows:
```
TicClick, line 43:
```suggestion
osu!mania converts are generated mostly according to the rhythm of the beatmap, and are supposedly much more random than converts in other game modes. While there is no consensus as to how osu! to osu!mania conversion works precisely, it generally goes as follows:
```

wiki/Beatmap/Converts/en.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Circles and shorter sliders (e.g. [1/4 sliders](/wiki/Client/Beatmap_editor/Beat_snap_divisor)) are converted into *regular notes*.
+- Spinners and longer sliders are converted into *long notes*.
+
+In addition to the above, osu!mania converts are unique in a sense that the resulting convert can either be [4K, 5K, or 7K](/wiki/Gameplay/Game_modifier/xK) depending on the difficulty of the base beatmap.
```
TicClick, line 48:
```suggestion
In addition to the above, osu!mania converts are unique in a sense that the resulting convert can either be [4K, 5K, or 7K](/wiki/Gameplay/Game_modifier/xK) depending on the difficulty of the original beatmap.
```

TicClick (changes requested):

TicClick:
overall, thanks for writing this all down -- one of the articles we're lacking

wiki/Beatmap/Converts/en.md
Nivalyx: changed the first sentence but i'll be keeping the "generally goes as follows" part i think, as (to my knowledge) there are some edge cases where the conversion is happening differently

wiki/Beatmap/Converts/en.md
Nivalyx: i tried to make it a table but i don't feel like it works, there are a bit too many variables to be arranged in a neat table 😦

wiki/Beatmap/Converts/en.md
Nivalyx: deleted the "in osu! gameplay" part

i wrote this section as such because i vaguely remember peppy saying that osu!(lazer) is planning to support conversion between modes e.g. taiko --> mania going forward

will try to ask peppy whether this is true or is this is just my memory hazing up tho...

Nivalyx:
@TicClick i think that's everything addressed, can you give this one a check again please?

TicClick (approved):

PR #10902, opened 2024-02-07
Update `Official tournament support` with new badge guidelines

ChillierPear:
- [x] approval from another tc member

### DIFF
--- a/wiki/Tournaments/Official_support/en.md
+++ b/wiki/Tournaments/Official_support/en.md
@@ -168,8 +168,8 @@ Profile badge images must adhere to the following standards:
 - It must be cleanly designed and of reasonable visual quality.
 - It must not include any sort of sponsorship or promote anything besides the tournament.
 - It must not include any characters or logos which were not specifically made for the tournament and adheres to the [Content usage guidelines](/wiki/Rules/Content_usage_permissions).
-    - This does not include assets which are open-source or public-use (i.e. fonts, icons, etc).
-    - If assets that do not fall into the above category are used in a design, explicit approval from the original artist must be provided before approval.
+  - This does not include assets which are open-source or public-use (i.e. fonts, icons, etc).
+  - If assets that do not fall into the above category are used in a design, explicit approval from the original artist must be provided before approval.
 
 Please consult the following examples of approved badge designs:
 
--- a/wiki/Tournaments/Official_support/en.md
+++ b/wiki/Tournaments/Official_support/en.md
@@ -167,13 +167,13 @@ Profile badge images must adhere to the following standards:
 - It must clearly display the logo, motif, or name of your tournament.
 - It must be cleanly designed and of reasonable visual quality.
 - It must not include any sort of sponsorship or promote anything besides the tournament.
-- It must not include any assets which were not specifically made for the tournament and adheres to the [Content usage guidelines](/wiki/Rules/Content_usage_permissions).
+- Any assets must be specifically made for the tournament and adhere to the [content usage guidelines](/wiki/Rules/Content_usage_permissions).
   - This does not include assets that are open-source or public-use (i.e. fonts, icons, etc).
   - If assets that do not fall into the above category are used in a design, explicit approval from the original artist must be provided before approval.
 
 Please consult the following examples of approved badge designs:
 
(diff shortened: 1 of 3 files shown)


wiki/Tournaments/Official_support/en.md
```diff
@@ -167,6 +167,9 @@ Profile badge images must adhere to the following standards:
 - It must clearly display the logo, motif, or name of your tournament.
 - It must be cleanly designed and of reasonable visual quality.
 - It must not include any sort of sponsorship or promote anything besides the tournament.
+- It must not include any characters or logos which were not specifically made for the tournament and adheres to the [Content usage guidelines](/wiki/Rules/Content_usage_permissions).
```
aceticke, line 170:
```suggestion
- It must not include any media which was not specifically made for the tournament and must adhere to the [Content usage guidelines](/wiki/Rules/Content_usage_permissions).
```
thoughts on this wording

Walavouchey: "media" isn't really a correct substitution. if you want one word it'd be something like "artworks" or "designs"

(also note the plural)

aceticke: perhaps assets then, since id imagine things like “hype videos” as you call it would be covered too

ChillierPear: Please keep in mind that these changes are being specifically made under the "Badges" header so we're not trying to cover every design possible here.

aceticke: Ah my mistake, then assets alone seems like good wording I think to cover many bases

wiki/Tournaments/Official_support/en.md
```diff
@@ -167,6 +167,9 @@ Profile badge images must adhere to the following standards:
 - It must clearly display the logo, motif, or name of your tournament.
 - It must be cleanly designed and of reasonable visual quality.
 - It must not include any sort of sponsorship or promote anything besides the tournament.
+- It must not include any assets which were not specifically made for the tournament and adheres to the [Content usage guidelines](/wiki/Rules/Content_usage_permissions).
```
Walavouchey, line 170: "It must not include any assets which [...] adheres to the content usage guidelines"

```suggestion
- Any assets must be specifically made for the tournament and adhere to the [content usage guidelines](/wiki/Rules/Content_usage_permissions).
```

wiki/Tournaments/Official_support/en.md
```diff
@@ -167,6 +167,9 @@ Profile badge images must adhere to the following standards:
...
+  - This does not include assets that are open-source or public-use (i.e. fonts, icons, etc).
```
Walavouchey, line 171:
```suggestion
  - This does not include assets that are open source or public-use (i.e. fonts, icons, etc).
```

Walavouchey (changes requested):

wiki/Tournaments/Official_support/en.md
ChillierPear: Indifferent to actual wording, original was chosen as it is more consistent with the rest of the article points formatting-wise.

cl8n: avoiding the double-negative is def more important in this case, it can be confusing because "and adheres [...]" could be read as coming *after* the entire "not [...]" clause rather than inside of it

wiki/Tournaments/Official_support/en.md
cl8n, line 171: i dont think this should be written as a sub-bullet because it isn't really related. you could also have assets that aren't meant for public use but were made for the tournament for example. also, fonts and icons don't generally fall into this category so idk why they are used to explain this

looking at the next sub-bullet as well, I feel like all of this is kind of unnecessary to write when you've already linked to content usage guidelines, which gets the same points across

ChillierPear: content usage guidelines do not cover paid font usage or licensing which these are meant to cover it is very barebones.

Walavouchey (approved):

Albionthegreat:
I confirm that these changes were discussed and approved by the tournament committee.

PR #10924, opened 2024-02-09
Add an action to automatically add `area:` labels

agatemosu:
https://github.com/actions/labeler

Addresses https://discord.com/channels/188630481301012481/218677502141399041/1170639204393422878

## Self-check


.github/labeler.yml
```diff
@@ -0,0 +1,45 @@
...
+
+"area:legal":
+  - changed-files:
+      - any-glob-to-any-file:
+          - wiki/Legal/**
+
+"area:meta":
+  - changed-files:
+      - any-glob-to-any-file:
+          - "*"
```
TicClick, line 14: this will mark everything as `area:meta`, no?

TicClick (changes requested):

TicClick:
good job regardless (action: https://github.com/actions/labeler)

.github/labeler.yml
agatemosu: only files in root

agatemosu: only files in root

TicClick (approved):

TicClick:
think it's good to go, but I'd like a second pair of eyes in case I missed something (the docs are convoluted)

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,12 @@
+name: Label pull request
+on:
+  - pull_request_target
```
cl8n, line 3:
```suggestion
on:
  pull_request_target:
    branches:
      - master
    types:
      - opened
```
dont need other branches and also i think i would find this annoying if it added back labels that I intentionally removed, so just `opened`

.github/labeler.yml
```diff
@@ -0,0 +1,46 @@
+"area:help centre":
+  - changed-files:
+      - any-glob-to-any-file:
+          - wiki/Help_centre/**
```
cl8n, line 4: doesnt really matter but i think this is supposed to indent more like
```suggestion
"area:help centre":
  - changed-files:
    - any-glob-to-any-file:
      - wiki/Help_centre/**
```
(and all the others)

.github/labeler.yml
```diff
@@ -0,0 +1,46 @@
...
+          - news/.markdownlint.json
+          - news/.remarkrc.js
+          - wiki/Tournaments/README.md
+          - wiki/Tournaments/TEMPLATE.md
+          - wiki/Tournaments/TEMPLATE-SERIES.md
+
+"area:news":
+  - changed-files:
+      - any-glob-to-any-file:
+          - news/**/*.md
```
cl8n, line 26: why does this one get `.md` when others dont?

also probably include the news images directory

cl8n (changes requested):

.github/labeler.yml
agatemosu: Because the folder has config files of formatting, although it can be deleted if not necessary

.github/labeler.yml
agatemosu: My formatters (Prettier and Red Hat) say that it is formatted like this, I searched and [this page (of prettier)](https://prettier.io/docs/en/configuration.html#configuration-overrides) is an example of this

cl8n (approved):

PR #11013, opened 2024-02-23
Add `Tournament Staff` to user groups

ChillierPear:


TheRealStevie:
Is there a group listing for this? If not, it should mention above that it's not public alongside chat bots and announce.

wiki/People/User_group/en.md
```diff
@@ -5,7 +5,7 @@ tags:
 
 # User group
 
-*Note: The Chat Bots and Announce user group listings are not public.*
+*Note: The Chat Bots, Tournament Staff, and Announce user group listings are not public.*
```
Hiviexd, line 8: a footnote that's referenced in the relevant table rows would be cooler imo

Hiviexd (approved):

PR #11049, opened 2024-02-29
[ID] Update and retranslate `Beatmap category`

Nivalyx:
(the current `id.md` is actually missing some information and a couple of links from the original `en.md` page but for some reason it's not marked as outdated)


TicClick (approved):

PR #11093, opened 2024-03-10
[ES] Add `osu! tournament client/osu!tourney/Multiplayer usage`

UnknownOsu:
## Self-check


agatemosu (changes requested):
image not necessary, the image shows the order, not the text

TicClick (approved):
@agatemosu @UnknownOsu as long as ^ was the only comment, you can merge this one in freely

UnknownOsu:
I still think the image was fine.

In the end, the wiki is being translated into Spanish, I think it wouldn't hurt if the images had as much Spanish as possible. But well, as you see best.

I checked it again to see if there is any error, but I didn't see anything serious that requires a change. I don't know if agatem is going to review it to see if there is anything that seems wrong in the article, so... I guess someone can merge it when they can

TicClick (approved):

PR #11098, opened 2024-03-11
Fix incorrect date in `Tournaments`

The-Last-Cookie:
Dates were taken from the wiki article: [Tournaments/UCUP/2022](https://osu.ppy.sh/wiki/Tournaments/UCUP/2022)

## Self-check


Walavouchey (changes requested):
we count tournament duration from qualifiers (first match) until grand finals (final match)

Walavouchey (approved):

PR #11105, opened 2024-03-14
Fix wrong hit window in `Overall difficulty`

minisbett:
Maybe I'm wrong but from what I can see, the 50 hitwindow formula for osu!mania has been wrong all along. `188 - 3 * od` is the formula for the miss hitwindow (which prevents spamming keys), not the 50 hitwindow. The current formula is `151 - 3 * od`.

Reference source:
https://github.com/ppy/osu/blob/master/osu.Game/Rulesets/Scoring/HitWindows.cs#L24
Note: These are default hitwindow values for all rulesets, but only mania actually refers to them. Taiko and Standard have their own hitwindows specified and catch does not have any.

## Self-check


Walavouchey (approved):
correct (and it's correct in https://osu.ppy.sh/wiki/en/Gameplay/Judgement/osu%21mania, but that hasn't been ported over into this article)

peppy (approved):

PR #11112, opened 2024-03-15
Update spelling for "No Mod" and "Free Mod" in `Article styling criteria/Writing` and `Game modifier`

Walavouchey:
people keep pointing these out but they've been standardised to use spaces at least in world cup articles


cl8n (approved):

cl8n:
I still feel like the right solution here is to not make them strict, they're not consistent in-game or across any community things

but will merge cuz this is Not Worse:tm:

PR #11122, opened 2024-03-17
Update extended edit rule to account for recreating official versions in RC

Okorin:
- adds conclusion from https://osu.ppy.sh/community/forums/topics/1891796?n=1
- same clause is already in place for cut versions that recreate official versions


cl8n (dismissed):
maybe mirroring the wording of the cut exception would clarify the intent, but maybe that reads too clunky. good to me either way

cl8n (approved):

Okorin:
@peppy if i recall, you wanted to have at least a look into RC changes before they go live. This one adds a clause to allow something similar that is already the case for cut versions

let me know if this can be merged

peppy:
I don't get it.

If it's not the artist's version, then it should be distinguished against.

What is this rule even achieving.

peppy (commented):
block

Okorin:
@peppy 

thanks for reviewing. The change was suggested because we have the same clause for applying `(Cut Ver.)` (see italic sentence)

- **Unofficial cut versions of songs must add a `(Cut Ver.)` marker at the end of the current title.** If a length marker is already in the title of the track, `(Cut Ver.)` would replace it. This is to distinguish unofficial cuts of a song from full length versions. _Songs that are shortened in ways that nearly match their official versions, and songs that are a full loop of a looping track will not be considered cut._

Since the rule about unofficial extended edits does not include the clause, doing the same thing to recreate an official version would require an `(Extended Edit)` marker as per 
- **Unofficial extended versions of songs must add an `(Extended Edit)` marker at the end of the current title.** If a length marker is already in the title of the track, (Extended Edit) would replace it. This is to distinguish unofficial extended versions of a song from official extended versions.

this is, as far as we can tell, unintended and this PR is a small fix to address that. Can you elaborate instead of outright blocking this?

Okorin:
the re-creation part is only really done if 
1. the only available version of the song is in a shit quality and the tools to recreate a better audio file existed
2. the "long version" never released, but was announced
  - basically song is lost media

Adding that marker for artifical, user-created extensions when that wasn't the point of the edit was what people agreed on being a bit silly in the forum thread, hence we are here now

peppy:
Any user edits should be *explicitly mentioned* regardless of "how close" they match.

So if anything, the other weird clause should be removed?

cl8n:
i think removing the other one is okay too (basically, none or both), but i don't know the history behind the other rule so that might be trampling over some earlier consensus

peppy:
sounds really weird to me and opposite of my expectations. i guess i won't block the change if everyone disagrees though 🤷

Walavouchey:
i've said the same thing as above in the thread, but here's another point

the main use case provided in the thread is this

> -> LILIUM's full version has never been released, but it has played in the show
> => current logic would make the fan-recreation of the fullversion an extended edit instead of just "LILIUM"
> => I don't think this is accurate to do in this case as the extension wasn't made to make a longer song, but was made to just have the same song that was used in the show
> ==> cases like this should be considered the original they're trying to recreate as that is the goal here

if the version was never released, how can one know that a fan recreation is the same as the original?

why should a map on osu! pretend it contains someone's "unreleased" track?

people have [pointed this out in the thread](https://osu.ppy.sh/community/forums/topics/1891796?n=12) but those concerns seem to have been ignored??

Screenshot_20240325-165421

Okorin:
> if the version was never released, how can one know that a fan recreation is the same as the original?
> 
> why should a map on osu! pretend it contains someone's "unreleased" track?
> 
> people have [pointed this out in the thread](https://osu.ppy.sh/community/forums/topics/1891796?n=12) but those concerns seem to have been ignored??

this particular example was addressed by App in [this post, which the asker agreed to](https://osu.ppy.sh/community/forums/posts/9461111). 

The `(TV Size)` of `LILIUM` would be forced to use the appropriate marker, though you're correct that the title just being `LILIUM` as an assumption as all we could dig up was an announcement about a version called `LILIUM` existing. We could find no official release of a song version matching what the long version played in the show is like. If it did release it is lost media now (2004 show), but there's no confusion about which version the map in question actually is

Anyway, maybe the example could have been a bit better, though the idea was still agreeable to people i think. 

For completeness, we created a thread to ask if we should remove the way this is handled for cut versions instead to gauge what people think about it:
https://osu.ppy.sh/community/forums/topics/1901188?n=1

Will keep that open for a few days to get people's opinions

peppy:
Can I just make a call of common sense here?

Any song, edited in ANY WAY from the original content, NEEDS A MARKER of some kind in the title saying it has been edited.

End of story. Adjust rules as required for this to happen.

Okorin:
No? The mapping community doesn't want this.

this behaviour was established more than 4 years ago, refers to 3 proposals that had extensive discussion (https://github.com/ppy/osu-wiki/pull/4181) and I even asked the current community if this is worth keeping

So far the consensus seems to be it's worth keeping.

This was supposed to be a 2 minute change. Why is this happening?

peppy:
i've shared my opinion and given what i believe to be valid reasonning.

do with this what you want; i'll mute the thread and look away

ShiiTsuin:
> Any user edits should be _explicitly mentioned_ regardless of "how close" they match.
> 
> So if anything, the other weird clause should be removed?

I'd like to invite you to look through (recently) ranked TV Size and Full Versions of songs and see if you can spot which ones are actually edits and which ones are actually official versions. I would bet money you couldn't correctly identify even half of edits.

Lets suppose a hypothetical situation. I want to map some anime opening. I have a full version, and a poor quality rip from the anime. Editing the full version to make the tv size would force me to use a marker that does not clearly tell players it's a tv size (or imitation of one), whereas using the original tv size may break audio ranking critiera for distortion or bitrate. What's preferred in this situation?

And suppose I do not disclose that I have edited a full ver - and nobody else knows I have - who's to say I didn't just find a high quality tv size source in the first place? This seems like a likely scenario in the event that you try and distinguish outright the difference between an official version and its imitation. If you're expecting BNs to check in these cases then you're putting a metric ton of audio checking work onto them whcih seems unreasonable.

For what its worth, I do agree that we should indicate the presence of any edits (where feasible). I just think that in cases that the RC edits proposed by oko (and the existing rule you oppose) address, it'd be more appropriate to indicate the edit elsewhere than the song title. 



It's a massive shame that you're bowing out despite clearly being passionate about this though, think there's plenty of room for discourse to be had. Also means I wasted my time writing this, but still posting it just cuz I had different things to say than I did on the forum.

Walavouchey:
the whole premise seems backwards, that not only do we insist on ranking songs for which there are no official releases, but that the ranking criteria should encourage making edits in order to imitate such songs.

> I'd like to invite you to look through (recently) ranked TV Size and Full Versions of songs and see if you can spot which ones are actually edits and which ones are actually official versions. I would bet money you couldn't correctly identify even half of edits.
> [...]
> And suppose I do not disclose that I have edited a full ver - and nobody else knows I have - who's to say I didn't just find a high quality tv size source in the first place?

and what's to say someone didn't make the edit in such a way that there's some random noticeable difference (that was missed before ranking), either deliberately or by accident?

what happens if an official version is released or found later?

do people not specify the source of where they obtained the song?

the can of worms seems to have already been opened. it's scuffed as hell. that's not even mentioning the question of licensing.

> For what its worth, I do agree that we should indicate the presence of any edits (where feasible). I just think that in cases that the RC edits proposed by oko (and the existing rule you oppose) address, it'd be more appropriate to indicate the edit elsewhere than the song title.

hard agree, in the map description at the very least. i highly encourage a proposal about this

but oh well, this change is not blocked per what peppy said

Okorin:
Okay, after more discussion, we changed the `Note:` on cut version and added a similar one for the `Extended Edit` to clarify that this is only applicable when the version you can create is either indistinguishable or nearly indistinguishable from the official thing

wiki/Ranking_criteria/en.md
```diff
@@ -109,8 +109,9 @@ Important to understand before using:
 - **If a song is used in a television program, web series, or direct-to-video series, such as an opening/ending/insert song, use a `(TV Size)` marker at the end of the current title.** If there is an existing TV size marker in the title, replace it with `(TV Size)`.
 - **If a song title contains short or game version markers, the markers must be standardised to `(Short Ver.)` and `(Game Ver.)` respectively.**
 - **Unofficial cut versions of songs must add a `(Cut Ver.)` marker at the end of the current title.** If a length marker is already in the title of the track, `(Cut Ver.)` would replace it. This is to distinguish unofficial cuts of a song from full length versions. Songs that are shortened in ways that nearly match their official versions, and songs that are a full loop of a looping track will not be considered cut.
```
cl8n, line 111:
```suggestion
- **Unofficial cut versions of songs must add a `(Cut Ver.)` marker at the end of the current title.** If a length marker is already in the title of the track, `(Cut Ver.)` must replace it. This is to distinguish unofficial cuts from full length versions.
```
(see next suggestion)

wiki/Ranking_criteria/en.md
```diff
@@ -109,8 +109,9 @@ Important to understand before using:
...
-  - Note: If an unofficial cut contains matching sections in the same order and is roughly the same length as an official TV size, short version, or game version, it will instead count as an official cut and use the appropriate marker. Covers and remixes do not count.
-- **Unofficial extended versions of songs must add an `(Extended Edit)` marker at the end of the current title.** If a length marker is already in the title of the track, `(Extended Edit)` would replace it. This is to distinguish unofficial extended versions of a song from official extended versions.
+  - Note: This only applies to songs that can be cut to match official versions in ways that are nearly identical. If the version of the song is noticeably different, the `(Cut Ver.)` marker must be used instead. Covers and remixes do not count.
```
cl8n, line 112:
```suggestion
  - **Songs that are intentionally cut to recreate an official version are not considered cut versions.** This only applies if the audio is nearly indistinguishable from the official version.
  - **Songs that are a full loop of a looping track are not considered cut versions.**
```

cl8n (commented):
continuing from dm, lmk what you think about this. something similar would apply to extended version rules as well. I think this reads a bit easier and makes the intent more clear

cl8n (approved):

cl8n:
@Walavouchey this part of RC is concerning how to title these edits when they do come up, and not questioning the practice itself, which I think is what a lot of your comment is about

> and what's to say someone didn't make the edit in such a way that there's some random noticeable difference (that was missed before ranking), either deliberately or by accident?

if nobody notices an edit such that the wrong metadata gets ranked, that is definitely an error, but not one that RC itself can really fix. the burden on the modders to identify unofficial edits comes from the wanting metadata to differentiate official&unofficial versions *at all*, not these few extra exceptions that are the topic of this PR

the main idea ur bringing up can be a separate thread, but at least personally I am doubtful that anything but the status quo will be popular on this issue

PR #11176, opened 2024-03-26
Update `Monthly Beatmapping Contest`

fowwo:
## Changes
- ~~**Replaced the contest listing links with the news post links.** The contest listings for all previous contests no longer exist.~~
- **Added the results of the latest contests.**

As far as translation goes, ~~the only change was "Contest Listing" → "Contest" ("Liste des concours" → "Concours"), and~~ the only additions were the latest months ("Noviembre" and "Diciembre").

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check).


Walavouchey:
err.. the contests disappearing probably isn't intentional

@peppy (see op and diff)

peppy:
That doesn't sound right, will investigate.

fowwo:
The contest listings are back, so I've reverted the links back to the contest listings.

pishifat (approved):

PR #11233, opened 2024-04-11
Clean up and extend `Reporting bad behaviour/Abuse`

Userlog2:
I have had the "no response" scenario happen to myself.

## Self-check


wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -60,14 +62,20 @@ If the tweet is from an account which is linked to an osu! player's profile via
...
-## Someone I know is being abused by a member of the volunteer teams (NAT, GMT) but is too afraid to speak up!
+## Someone I know is being abused by a member of the volunteer teams (NAT, GMT), but is too afraid to speak up!
 
 Send an email to [abuse@ppy.sh](mailto:abuse@ppy.sh) containing everything you know with as much evidence as you can get.
 
 If you don't have any evidence but only stories or rumours that you believe are substantial enough to act on, please send us an email anyway. We take complaints like this very seriously and will look deeply into every report.
+
+## My email has been ignored!
+
+If you haven't received a response to your email after a couple of weeks, please send a follow-up email.
```
iSlodinxOsu, line 79: perhaps add pe@ppy.sh?

iSlodinxOsu (changes requested):

wiki/Reporting_bad_behaviour/Abuse/en.md
Blooshing: It can just be stated to send up a follow email, since peppy and ephemeral are the only ones that see abuse@ppy. No need to double email them. Keeping it as such is fine.

iSlodinxOsu: Right, that's what i thought the same, i did reported something to abuse@ppy.sh, but after like more than a month (maybe?) peppy messaged me back with pe@ppy.sh, and i believe my report was lost somewhere and he found it out later.

Blooshing: Its not unknown that people can email peppy about osu! things since he has publicly announced his email, but when it comes to matters like these its best to continue through the right channels so procedures can flow smoothly.

Userlog2: Agree with Blooshing, but I think it would be simplest to just ask @peppy what his opinion is on this matter.

peppy: abuse forwards directly to me. sending an email to pe@ppy.sh will not change anything.

Userlog2: I think this answers the question - if using abuse@ppy.sh gets a faster response and is otherwise the same, it should be used.

Blooshing (approved):
lgtm

Userlog2:
(misclicked, if it's fine leave it, if it isn't revert it)

TicClick (approved):
it's alright as long as you don't click it regularly, which creates extra empty commits
(not that we curate the history log excessively, but still)

PR #11263, opened 2024-04-17
Split artist permissions table based on status in `Content usage permissions`

Hiviexd:
current artist list is a pain to navigate through and it's really hard to see at a glance which artists are disallowed/allowed with exceptions, so i propose to group artists by status while still maintaining alphabetical order when relevant, that way it's much more readable.

[wiki preview](https://osu3.roanh.dev/wiki/en/Rules/Content_usage_permissions)

### DIFF
--- a/wiki/Rules/Content_usage_permissions/de.md
+++ b/wiki/Rules/Content_usage_permissions/de.md
@@ -34,388 +34,395 @@ Dieser Abschnitt beschreibt, welche Künstler die Nutzung ihrer Songs in osu! er
 
 Alle Songs, die auf der jeweiligen [Featured-Artist](https://osu.ppy.sh/beatmaps/artists)-Seite stehen, sind für die Verwendung in osu!-Beatmaps freigegeben. Songs von einem Featured Artist, die nicht auf seiner Seite erscheinen, sind *nicht* lizenziert und erfordern für die Nutzung möglicherweise zusätzliche Genehmigungen.
 
+#### Erlaubt
+
+| Künstler |  | Status |
+| :-- | :-- | :-- |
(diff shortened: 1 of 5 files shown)


peppy:
I dunno, personally I prefer alphabetical. Interested in more opinions on the matter.

Userlog2:
Agree with this PR as it would make the wiki page a lot more effective at serving its main purpose: making clear what songs may or may not be used. 
It still looks very clean, as within the categories, all the artists are alphabetized anyway, and there are only 3 categories.

Walavouchey:
very much agree with this change and i'd additionally prefer 3 tables instead of one (under separate, linkable headings)

Albionthegreat:
I would also prefer having 3 tables instead of one, under separate, linkable headings. All 3 tables being sorted alphabetically.

There are 2 ways that need to be catered for. The first, if you know the artist, and want to find out if they are in the listing. And the second, if you have a list of songs from various artist, in tournaments for instance, and you want to check if any of the artists are not permitted.

For the first, you can easily do ctrl+f, or go to the alphabetical location in the list(s). For the second, you'd need to manually go through the entire list. This has the added risk of accidentally skipping over an artist, to the detriment of the user or tournament. Having all artists that disallow their songs grouped together makes this easier in my opinion. Though it would need to be made abundantly clear that there are 3 tables.

peppy:
The preview seems to have stopped working.

Let's try with three tables then, if that will satisfy everyone.

Hiviexd:
i've switched to 3 tables and fixed the [web preview](https://osu3.roanh.dev/wiki/en/Rules/Content_usage_permissions).

peppy:
It's a bit of a pity that the tables' columns no longer line up.

peppy (approved):

PR #11265, opened 2024-04-17
Update number of provided top replays per beatmap difficulty

Joehuu:
Example (out of top 1000 a bit (1004) but probably it's pending deletion): https://osu.ppy.sh/scores/1057219585
Another close one (825) if the above is deleted after: https://osu.ppy.sh/scores/1506502404

Unsure how the 500 was found initially in https://github.com/ppy/osu-wiki/commit/a1c56faa4ae6d2ba10579e0e2c9971b41fe4fa33#diff-d64250a82e24059efd19be5a973cf10f2be8a90456649f7dedf6d7012c29d963R96 (2021, but [people noticed it was top 1000 on reddit](https://www.reddit.com/r/osugame/comments/anucjn/til_you_can_click_on_the_3_dots_next_to_top/) before that)

## Self-check


TicClick (approved):
> Unsure how the 500 was found initially

my fuzzy and likely outdated memory. right now, on the spot, I can find mentions of the 1000 as early as [2017](https://osu.ppy.sh/community/forums/posts/6045125) -- with further luck and amount of scavenging you can probably find the release note (or _could_ if the old development log was still accessible) both for that value and for 400/500 that existed before

PR #11275, opened 2024-04-18
Switch badge images to assets.ppy.sh links in `Tournaments` for World Cups

Albionthegreat:
## Self-check

Badges are done inconsistently on the wiki, with some posts referring to assets.ppy.sh, while others have the badge in an image folder. In an effort to make sure only applied badges end up on the wiki and to increase consistency, this PR removes the badge images as standalone files and instead refers to the assets.ppy.sh link. I'm intending to take a look at community tournaments at a later time, so this PR only includes world cup badges.


cl8n:
I'm not sure how I feel about purposely introducing dependencies to the wiki when there aren't any problems with hosting it here. not that there are big problems with this change either, just feels kind of... unnecessary

Walavouchey:
there are some advantages

- if a badge somehow disappears or gets redesigned, it'll get reflected on the wiki automatically. tooling could notify of this down the line (if query strings need to be added for caching or if the image link needs to be removed)
- there's no way to link a badge image if it's not applied yet. i don't typically check for this, but it can sometimes lead to denied badges being forgotten about and being left over in the article

hosting on the wiki is only a secondary source of truth, and it's not like assets.ppy.sh is third-party, so i only see positives in this

cl8n:
the wiki *should* require someone intentionally editing the wiki to change anything imo. these badges are hosted for a different purpose and may be edited/moved/deleted without regard for how the wiki uses or refers to them. for the latter point, I don't think it's a good thing to prevent content from existing, it just needs appropriate context like everything else e.g. I might describe a badge reward as "tentative" in the article if it has no guarantee at the time of writing

TicClick:
↑ I agree with the arguments -- while hotlinking creates uniformity, IMO it's worth preserving historical value and do the reverse instead (keep the badges local to the respective articles)

Walavouchey:
closing for the above reasons

PR #11308, opened 2024-04-21
Fix accuracy requirement table column order in `Gameplay_differences_in_osu!(lazer)`

gi-b716:
## Self-check

harmonize the order of columns


Walavouchey (approved):
makes sense

gi-b716:
thx

PR #11329, opened 2024-04-24
[DE] Update `ASC/Writing` and `Community/How_you_can_help`

The-Last-Cookie:
## Self-check


wiki/Article_styling_criteria/Writing/de.md
```diff
@@ -219,8 +217,8 @@ Manche Wörter haben Varianten. Ihre bevorzugte Schreibweise ist wie folgt:
 
 Diese Worte müssen wie folgt verwendet werden (der Abstand muss übereinstimmen):
 
-- `NoMods`, `NoMod` oder `NM`
-- `FreeMods`, `FreeMod` oder `FM`
+- `No Mod` oder `NM`
+- `Free Mod` oder `FM`
 - `Tiebreaker`
```
The-Last-Cookie, line 222: Not sure how we want to handle this word in German 🤔 

For reference:

> (not a proper noun, i.e. not capitalised within sentences)

cl8n (approved):

wiki/Article_styling_criteria/Writing/de.md
cl8n: well if it is a proper noun then you don't need to include that note

PR #11354, opened 2024-05-04
Correct Poland pick in 5WC Conclusion

hburn7:
## Self-check

Poland didn't pick DT3 (DT3 was never picked), they picked NM3. Double checked this against the ref sheet.


hburn7:
Threw in another fix (Beyond reality was marked as grand finals TB in one place)

Walavouchey (approved):

PR #11430, opened 2024-05-21
[ID] Standardise writing of index pages

Nivalyx:
rn the translations of the index pages (e.g. `http://osu.ppy.sh/legal/id` and `http://osu.ppy.sh/wiki/id/Client`) are done quite differently from one another, so i thought i'd try to introduce some consistency here since in the original `en.md` pages the variations are rather minimal

## Self-check


wiki/Beatmap/Pattern/osu!/id.md
```diff
@@ -0,0 +1,8 @@
+---
+stub: true
+---
+
```
cl8n, line 4: missing title

cl8n (changes requested):

wiki/Beatmap/Pattern/osu!/id.md
Nivalyx: whoops

cl8n (approved):

PR #11435, opened 2024-05-22
Add `osu! keycaps` news post

pishifat:
[preview](https://osu3.roanh.dev/home/news/2024-05-22-osu-keycaps)

- [x] kapco website fixes
- [x] finalize giveaway details
- [x] store listing
- [x] tweet link
- [x] redate


wiki/shared/news/2024-05-22-osu-keycaps/header.jpg
Walavouchey, line 1: this would be the `@2x`

peppy:
I think the main banner needs fixing since it looks like the caps are red still? (the issue i pointed out yesterday).

news/2024/2024-05-22-osu-keycaps.md
```diff
@@ -0,0 +1,80 @@
...
+## Giveaway
+
+*but wait, there's more!*
+
+**The KapCo** was generous enough to give us NUMBER sets of keys for free, so we'll be doing a joint giveaway!
+
+There are two ways to enter:
+
+1. Retweet [our tweet](LINK)
+2. Follow [@thekapco](https://www.instagram.com/thekapco/) on Instagram, and like/comment on [this post](LINK)
```
peppy, line 70: how does this work? are they requiring we do this? how do we choose the winners if it's on their account?

also i dunno how easy it is to get a list of people that retweet these days because x is a pile of dicks.

pishifat: not tied into anything yet. kapco said their goal with the giveaway was to get more instagram follows, but after thinking about it more, i don't think we need to directly provide it, so ill revise this soon

peppy: hmm, as long as they are okay with this. i'm fine with having the giveaway only be on twitter or instagram (even theirs) but we should choose one location for it to be at, rather than multiple.

wiki/shared/news/2024-05-22-osu-keycaps/banner@2x.jpg
Walavouchey, line 1: the banners could really use some compression

i'm not entirely a fan of this layout with the keyboard in between the logos, but if anything the logos could be a tad bigger

image

pishifat: tanza is cooking, will update when it's ready (hopefully before kapco fixes things, since that's the current bottleneck)

Walavouchey (approved):

PR #11491, opened 2024-05-29
Document exact scale of osu!mania score bar

mcendu:
This pull request documents the exact scale of the score bar in osu!mania under stable, as discovered by skinners of the skinship Discord. This can be useful for intergrating the scorebar into an osu!mania stage design.

## Self-check


wiki/Skinning/Interface/en.md
```diff
@@ -1118,7 +1118,8 @@ Notes:
 Notes:
 
 - This element has no size restrictions.
-- When used in [osu!mania](/wiki/Game_mode/osu!mania), this element is rotated 90 degrees anti-clockwise and is placed at the bottom right of stage.
+- When used in [osu!mania](/wiki/Game_mode/osu!mania), this element is rotated 90 degrees anti-clockwise,
```
TicClick, line 1121: we don't add line breaks where unnecessary even if they are not rendered -- that is my only complaint

Pacuwu: Try editing it

Walavouchey:
can you link to where this was discovered? there's nothing in the stable code that says the sprite should be scaled in any particular way, so i'm wondering if you're comparing to a particular coordinate system

mcendu:
> can you link to where this was discovered? there's nothing in the stable code that says the sprite should be scaled in any particular way, so i'm wondering if you're comparing to a particular coordinate system

[A comparison using screenshots](https://discord.com/channels/591469672747106304/591696564578156575/1239612122388889630) is made to get the result of around 0.7 compared to std and other modes. Skinship's discord invite link is <https://discord.skinship.xyz/>.

wiki/Skinning/Interface/en.md
mcendu: Done. I'm too used to code and doc conventions enforcing line wraps.

cl8n:
0.7 is correct

Walavouchey (approved):
[looks like i didn't look hard enough](https://github.com/peppy/osu-stable-reference/blob/0e425c0d525ef21353c8293c235cc0621d28338b/osu!/GameModes/Play/Components/HpBarMania.cs#L24)

PR #11526, opened 2024-06-03
Update `Completionists` list

minusQuantumNeko:
[Source of data](https://score.kirino.sh/completionists)

## Self-check


0x84f:
the wiki page says that the list presented is verified by the osu team, and the data source you are linking to is a 3rd party. have you spoken to anyone about updating this?

cl8n:
if it's only counting people given the badge then yea leave as-is. if it's counting everyone that people would typically consider to have "completed all maps" then this is obviously missing a lot of people prior to 2023 (anecdotally I remember when I started playing ctb that a lot of experienced players had played all ranked maps, my guess is that this was true in every gamemode when they had less than a few thousand maps or so, and also when ranked score was the main ranking metric)

cl8n:
^additionally this PR isn't changing the part that says the list is for the "verified by osu! team" completionists, so yeah this won't work as-is. it'll either need to change what it's listing or have osu! team update their own list

PR #11563, opened 2024-06-12
Add new Metadata article to RC

Okorin:
# Summary
- While a relatively small sample size, i'll try going with separate articles for drafts where appropriate (we also talked about this in the latest NAT meeting). Forum conversation and poll in : https://osu.ppy.sh/community/forums/topics/1931114?n=1
  - Had to rearrange the main RC article, took some creative liberty:
    - The info included is still the same, only mentions to other general rules as being part of the beatmap modding process (like VCC or content usage permissions) were added 
    - metadata is added as its own article with a reference in the table of contents
 since the changes aren't intended to change ANY behaviour and are instead about formatting, I skipped making a separate discussion thread for moving this around
- Adds draft from https://osu.ppy.sh/community/forums/topics/1894663?n=1 as its own article

The [comment about `(Slowed Down Ver.)`](https://osu.ppy.sh/community/forums/posts/9563682) is not included in this draft for now for the following reason: These songs aren't common enough to warrant their own mention and they'd be covered by "make a sensible marker" via the `#### Ver.` guideline introduced.

Some american english may still be left and some phrasing may still include a bit simpler language. I personally think the simpler language is fine


wiki/Ranking_criteria/Metadata/en.md
```diff
@@ -0,0 +1,266 @@
...
+- **Combined Markers**
+  - If the edit is sped up *and* cut, use `(Sped Up & Cut Ver.)` or `(Nightcore & Cut Ver.)`.
+  - Other combined markers can be used after a discussion is held to discuss their necessity.
+
+##### Markers that you must standardise, but not always add
+
+- **If there are similar markers in a song's title, replace them with the relevant marker from the list below.** 
+- **If there are no markers, you can only add them when there are multiple versions of a song to indicate the correct version.**
+- `(Short Ver.)`
+  - Usually used to signal that a longer version actually exists. Use this when there is an existing length marker such as `-Short Ver-`, `Short Ver.`, `~Short Version~` and `Movie` or `Game Ver.` do not apply.
```
NoffyNoffeh, line 144:
```suggestion
  - Usually used to signal that a longer version actually exists. Use this when there is an existing length marker such as `-Short Ver-`, `Short Ver.`, `~Short Version~`. This does not apply to `Movie` or `Game Ver.` markers.
```

wiki/Ranking_criteria/Metadata/en.md
```diff
@@ -0,0 +1,266 @@
...
+### Rules
+
+- **Tags must be related to the beatmap**, such as describing the style, song, storyboard, video, or background content. Misleading tags must be avoided.
+- **Names with spaces between single characters like `-[M o c h a]-` need to be tagged as `-[M_o_c_h_a]-`**
+- **Tags must include the following items when applicable:**
+  - **[Guest difficulty](/wiki/Beatmap/Guest_difficulty) creators, storyboarders, skinners and hitsounders.**
+  - **`Featured Artist`, if the track is listed in the [Featured Artist catalogue](https://osu.ppy.sh/beatmaps/artists).** You must not use the tag if the song is not in the featured artist catalogue.
+  - **At least one song genre and one language tag.** 
+    - For instrumental tracks, `instrumental` is the language tag.
+    - For tracks in artificially created languages add `conlang` to tags and use its name as the language tag.
```
NoffyNoffeh, line 196:
```suggestion
    - For tracks in artificially created languages add `conlang` to tags as the language tag. If the conlang has its own name, add that to tags as well.
```

what "its name" refers to was unclear and clunky. how about this?

Okorin: i have never heard of an unnamed conlang, cuz the nature of conlangs is to make up names for stuff 

that said i think that may be better?

Okorin: how about just `and use the conlang's name as the language tag.` instead

wiki/Ranking_criteria/Metadata/en.md
```diff
@@ -0,0 +1,266 @@
...
+### Guidelines
+
+- **Logos should not be used as reference for capitalisation.** As logos are often stylised, apply regular title case[^title-case] unless other textual metadata supports it.
+- **When a remix or cover appears to have a typo in the song name compared to the original track, use the original spelling instead, unless the typo is an intentional stylisation.** If it is unclear, a [discussion](/wiki/Beatmap_discussion#discussions) should be held to come to a consensus on which to use. Below are examples with a remix of a song titled `triangles`:
+  - `triangls` should be `triangles`
+  - `triANGELS (angelic remix)` should be left as-is
+  - `3angle5` should be left as-is
+  - `Triangles` can be either `Triangles` or `triangles`
+- **When multiple metadata options are available:**
+  - Aim to match Ranked or Loved beatmaps. Follow what is most recent and common, then verify that metadata is correct and fix as needed.
```
NoffyNoffeh, line 27:
```suggestion
  - Aim to match Ranked beatmaps. Follow what is most recent and common, then verify that metadata is correct and fix as needed.
```

Loved isn't as able to enforce every requested change as it used to be, so may be best to change this accordingly.

Okorin:
@peppy You asked me to ping you on the conversation to check this out

peppy (approved):
Looks logical enough to me. Note that I have not done a pass with a fine comb; this is just a review showing my approval to go-ahead-with-this-direction.

cl8n (approved):
i never got around to finishing the detailed review i was promising here but dont want to hold this up any longer. ive read through all of it though and i dont remember having many suggestions that meaningfully affect how the rules are read, it was mostly cleanup stuff and wiki-isms. i may still get back to that, just not worth blocking over

for reference this has also been reviewed in some form by other people before it was even PRd

peppy:
Let's get this in as-is and work from there. Do we need to worry about the missing outdated markers?

cl8n (approved):

PR #11570, opened 2024-06-13
Use `master` instead of HEAD for `--base-commit` in outdated article check

Walavouchey:
fixes issue reported on discord https://discord.com/channels/188630481301012481/218677502141399041/1250894073951289446

turns out i forgot what clayton said in https://github.com/Walavouchey/osu-wiki-tools/pull/22#issuecomment-1872171172 when updating in https://github.com/ppy/osu-wiki/pull/11438


cl8n:
^just that should be okay, it's the same as the defaults now.

cl8n (approved):

TicClick (approved):

PR #11574, opened 2024-06-13
Community Mentorship Program 2024 Summer newspost

Okorin:
adds newspost for the summer cycle 2024, currently scheduled for saturday 2024-06-15 

changing this schedule is fine, but we'd prefer to not go that much later because as-is we likely have to skip a season (in the naming) next cycle around already


Walavouchey (approved):
all good

PR #11587, opened 2024-06-16
[ZH] Add Bilibili streaming in `Livestreaming osu!`

CloneWith:
## Self-check

This part of content is Chinese-specific, independent from the original article and other translations.


cl8n:
leave a comment (`<!-- text here -->`) explaining the intended differences and additions compared to EN, so that future translators have an easier time coordinating changes when EN updates

proxyerium (approved):

cl8n (approved):

PR #11632, opened 2024-06-21
[ZH] Update `Skinning/Interface`

CloneWith:
## Self-check


TicClick (approved):
seems to be too small to require a native review

PR #11651, opened 2024-06-24
[DE] Update `meta/group-info`, `Twitter` and `osu! tournament client/Setup`

The-Last-Cookie:
## Self-check


cl8n:
branch name isn't important but I feel as if when you resort to just naming it "something", you may want to think if there was a good reason to put a bunch of unrelated things in one branch in the first place...

cl8n (approved):

PR #11662, opened 2024-06-27
Update group member listings

Laurakko:
## Self-check


Walavouchey:
force-pushed to revert an unrelated file on request

peppy:
Is this done using a script?

cl8n (approved):

cl8n:
> Is this done using a script?

one to check that EN lists are accurate (just the members, not necessarily the additional info) and another to apply EN changes to all of the translations, yes

PR #11707, opened 2024-07-08
[FR] Update `group-info`

DoryanR:
I'm back !

## Self-check


cl8n:
for `no_native_review_since`, if the person who did the last review is yourself, it should stay as-is

cl8n (approved):

PR #11719, opened 2024-07-12
Update lists of GMT and alumni

CivilWiki:
I only edited en.md :pray:

## Self-check


cl8n (approved):

cl8n:
I removed all of the outdating commits bc these tables don't follow the same rules. read the notes in en.md above the tables

PR #11727, opened 2024-07-12
Update en.md

Orangemoody:
its true (i think, might have the wrong date)

## Self-check


cl8n:
will do this in #11805 instead

PR #11736, opened 2024-07-14
Add Round of 8 support for specific types of draft tournaments

hburn7:
## Self-check

- [x] Tournament Committee review

The Tournament Committee has voted on this rule change. Draft tournaments are eligible for Round of 8 support so long as they are open rank and feature a minimum team size of 8. The team agrees that this support should be limited to one open-rank division should the tournament feature multiple divisions. This affects both international and region-restricted draft tournaments.

Personal note: I probably didn't codify the division/region-restricted portion that well. To avoid confusion, I feel it should be abundantly clear that open rank **does not** imply the tournament needs to be international.


wiki/Tournaments/Official_support/en.md
```diff
@@ -29,6 +29,10 @@ Community-run tournaments which abide by the following hard criteria are eligibl
 - The tournament series must run no more than two times per year.
 - The tournament must operate on a Round of 16 double-elimination format at a minimum.
   - Larger formats, such as group stages and Swiss brackets, may also be used so long as they start with at least 16 teams and, if applicable and if starting with only 16 teams, break to a double-elimination bracket afterwards.
+  - Tournaments that feature a draft format may operate on a Round of 8 double-elimination bracket provided the tournament meets the following additional criteria:
+    - The tournament is open rank.
+    - The team size is 8 or more.
+    - Note: In tournaments with multiple divisions, this rule applies to only one open-rank division. Open rank does not imply international.
```
cl8n, line 35:
```suggestion
    - Note: In tournaments with multiple divisions, this rule applies to only one open-rank division.
```
this last sentence confused me because it doesn't seem like it would be an assumption in the first place. rank restrictions and region restrictions are unrelated (?)

cl8n (approved):

Albionthegreat (approved):
Changes have been discussed and approved by the tournament committee.

PR #11798, opened 2024-07-20
Remove mostly incorrect info in `COE 2024` article

Walavouchey:


peppy:
@Walavouchey any more context on this? was it copied from 2023 without thought? i think at least some of the staff list is still correct, so i guess that will come back at some point?

Walavouchey:
@peppy indeed it will be back at the end, but i was told by meyer to remove the entire thing as it was apparently 50% incorrect. ended up keeping the bare minimum

peppy (approved):

PR #11845, opened 2024-07-25
Update recommended user for forum post reports in `Reporting abuse`

Walavouchey:
this swaps ephemeral to venix (gmt leader). if there are other suggestions do tell


0x84f:
i assume that venix is aware of this and accepted to do this?

TicClick:
> i assume that venix is aware of this and accepted to do this?

yes

TicClick (approved):

PR #11868, opened 2024-07-30
Add VSCode line ending setting for github.dev

### LINKED ISSUE #11861: Configure github.dev environment to use LF for new files
idk if that's even possible but something to look into. feels really bad for this to be causing issues because github.dev is probably a common choice among people who don't know enough about git to solve them

cl8n:
the web version of vscode doesn't take hints for this setting from gitattributes or editorconfig, so it has to be set explicitly like this (or at least I didn't find a better way). this has the unfortunate side effect of tracking the workspace settings file, so any further workspace-level changes any vscode user makes will have to be "assume unchanged" in the git index -- for some reason there is only one possible workspace settings file name (https://github.com/microsoft/vscode/issues/40233), so I found no better way to avoid this either.

resolves #11861


peppy:
> this has the unfortunate side effect of tracking the workspace settings file

Probably better than not having it at all.

peppy (approved):

PR #11877, opened 2024-07-31
Update developers listing

cl8n:
some more that the recent update didn't catch and also update lists from github

### DIFF
--- a/wiki/People/Developers/en.md
+++ b/wiki/People/Developers/en.md
@@ -21,7 +21,6 @@ The osu! development effort is lead by ::{ flag=AU }:: **[peppy](https://osu.ppy
 
 ### Game client developers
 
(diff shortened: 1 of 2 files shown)


Hiviexd (approved):

peppy:
Please stop this stuff.

I have no idea what's even going on with this listing anymore.

Why are pelpe being removed like spaceman? Why are people being added? where from?

cl8n:
<https://osu.ppy.sh/wiki/en/People/Developers#team-members> is where the devs in the usergroup are, including spaceman since the last update -- <https://osu.ppy.sh/wiki/en/People/Developers#contributors> lists additional contributors from the 3 main repos that aren't listed as official dev

I haven't changed anything about how this page works in this PR, just updating it

PR #11881, opened 2024-08-01
[Proposal] New osu! macro rules

minisbett:
I'm looking for feedback & suggestion on how to improve the overall readability and understandability without sacrificing much details. If you have anything to bring up, feel free to do so here on GitHub or in my Discord DMS. (`minisbett`)

---

### This is my proposal for a rework & specification of the macro rules for osu!.

Over a month ago I've reached out to peppy regarding concerns about the current way the rules regarding macroing & co. are 
outlaid, and that after almost 7 years a rework of them, especially regarding recent events, many people I've talked to think it is more than overdue to give them a rework. He was open to a proposal, and after abit over a month of brainstorming, figuring out wording and finding the balance between coverage, technical details and accessibility for the mainstream I've come up with this.

Peppy told me that last week there has been changes made to the rules already, but despite them seeming to try to cover what is important I'm concerned about how vague they are, as well as being hard to understand even for technical people like me. Therefore I decided to come up with this proposal anyways.

### My main goals with this are the following:
1. A very general but accurate definition of a macro for the purpose of the rules.
2. A small glossary for readers to have an easy time understanding terms and prevent misinterpretation.
3. Clearly structured rules, worded and reasoned, making them easy to understand, accessible and allowing for them to be extended in the future if necessary.
4. The rules are decorated with examples of allowed and disallowed things (while trying to be as easy to understand as possible)
5. Proividing a preamble and closing words mentioning what I think is important to mention (eg. non-keyboard input devices are to be used at your own risk, as they are hardly directly affected by current & proposed rules)

After careful consideration I decided to specifically reference to some features from keyboard brands in the footnotes (Wooting, Razer, SteelSeries) as I believe many people are aware of their functionality and it further contributes to build an understanding of the rules.

I put a lot of care into the rules being hard to misinterpret, as I've seen people before being confused about rules , putting out wrong assumptions or just arguing against them to justify things they are doing.

---

I'm open for discussion for the wording and details of this proposal, but I personally am pretty happy with the current state, as I believe they provide everything we need to proceed with this topic. One thing I further thought about was a visualization of certain scenarios through gifs, similar to how you know it from instruction manuals etc. on what to do and what not to do, as a visual workup of these rules might be beneficial for further understanding or accessibility. For now, this will be marked as a draft until I've received more feedback here directly, as I'd like to see that first to be more confident about this.

What I'd also be open to is adding a TL;DR for the rules that is much more vague, but probably enough for the average player.

I have already been reaching out to the osu! community in various places and receiving positive feedback from everyone (even including a Wooting employee), it seems to be good to understand even for less technical players and is seen as a desired addition.

—mini

## Self-check


peppy:
I think what we had was already enough. If you have issues with the phrasing, can you start by proposing a fix? I find the proposal here far more complicated and hard to follow than what we have, which is KEY DOWN = key down, KEY UP = key up.

wiki/Rules/Macro_rules/en.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Glossary
+
+**Key**: A "key" refers to a physical key on your keyboard that can be moved upwards or downwards.
+
+**Key state**: A key state is the current state of a key as the osu! client perceives it. A key can be either be considered "pressed" or "released", regardless of the physical situation of the key.
+
+**Movement**: A movement is the continuous physical movement of a key in **one direction** (upwards or downwards). A direction change ends the movement and starts the next one. Whenever a key is "moved" in the rules, it refers to this definition.
+
+**Action**: An action is the act of the state of a key changing, regardless of what causes it.
```
peppy, line 24: means nothing to me.

make it "game key" or "game action" like the original text.

minisbett:
I've done some changes after receiving more feedback from various people.

A `key state` is now defined as what the osu! client sees. This means that one physical movement = 1 action on that same = 

> I think what we had was already enough. If you have issues with the phrasing, can you start by proposing a fix? I find the proposal here far more complicated and hard to follow than what we have, which is KEY DOWN = key down, KEY UP = key up.

It is certainly more complicated, but I feel like it really is necessary. I've already talked with people about the new changes, and honestly people don't understand it and/or are not happy with it. It's very broad, and even I not having fully understood it believe there's far too many loop-holes or oversights there.

I didn't really try to modify the new changes as adding what I think is necessary would bloat that part of the rules a lot, and I still struggle to understand everything even now. I don't think the current live implementation is the way to go. I'm open to dumb down a lot of what I did here, if we can figure out a way to maintain most of the quality.

By no means is the average user expected to read and understand all of this. As seen in my last commit, generally no one needs to worry about these details if they're using a normal keyboard in a normal way. But based on feedback from the community, there is the need for details, so that there's a deterministic answer to most things and not a "oh u should figure out if it's allowed or not by yourself", because for some people that's too vague.

peppy:
I'm not going to review this yet. There's enough wrong/unclear with this that I'd want others to put in the effort (or maybe just re-read it a few days later with fresh eyes yourself and spot the issues).

minisbett:
> I'm not going to review this yet. There's enough wrong/unclear with this that I'd want others to put in the effort (or maybe just re-read it a few days later with fresh eyes yourself and spot the issues).

Agree. By no means to I think this is perfect, I just think it requires a lot more thoughts. I think I will be trying to dumb things more down to make it easier to understand

wiki/Rules/Macro_rules/en.md
minisbett: I've changed this and got completely rid of the "key state" thing, and I think this significantly improved the overall simplicity already

cl8n:
what cases do you think this is making more clear that were "vague" in the recent clarification to macro rules? I can't see how the 2 rules in this new article are any different than the summarized version already present

calling out proprietary terms is good for convenience (answers what I imagine to be the most common questions here, e.g. "is DKS allowed?") but I'm thinking it's probably not outweighing the fact that the reader now has a whole article to sift through instead of just a few sentences

minisbett:
> what cases do you think this is making more clear that were "vague" in the recent clarification to macro rules? I can't see how the 2 rules in this new article are any different than the summarized version already present
> 
> calling out proprietary terms is good for convenience (answers what I imagine to be the most common questions here, e.g. "is DKS allowed?") but I'm thinking it's probably not outweighing the fact that the reader now has a whole article to sift through instead of just a few sentences

As I mentioned these more detailed rules are most likely uninteresting to the average player, but I've been experiencing demand in a detailed and more covering version of the rules that stops any misinterpretation or loop holes as good as possible. The main complaints I've heard about the rules have not really been tackled by the recent rewording.

About the "whole article", peppy expressed to me before that he seems to not be rejective about something in this sort of format, he said "it shouldn't take more than a couple of paragraphs".

I wish for there to be a straight forward rule saying pretty much what is currently said, but also provide a more guiding version that satisfies an apparently significant part of the community. There's new technology and weird shenanigans people come up with all the time and people disliked that pretty much anything was a fight of opinions in the community whether something is allowed or disallowed or whatnot until people start messaging peppy & co about it.


I do acknowledge the feedback that I've already received via other means of communication and will come back to improve this, or maybe find a middle way. (eg. keeping the original, but still referring to this more detailed article so that a clear official attitude towards things around input exists and is easily visible)

cl8n:
> [...] more covering version of the rules that stops any misinterpretation or loop holes as good as possible.

this is the part I'm skeptical of -- what can be misinterpreted or have holes poked in it in master, that is fixed in this PR? the points are demonstrated in examples sure, but they don't change in meaning, nor are they further clarifying anything you couldn't have understood unambiguously from what's in master right now.

fwiw I'm not saying this article is useless; presenting examples and relating the rules to real-world features/products is nice for convenience and helping give insight as to why the rules have been written this way. I'm just not convinced that the current rules are flawed in some way that this is supposed to fix, and so I don't think the current rules should be edited or removed. if it were up to me I guess I'd have the new article linked in a way that makes it sound like a supplemental explainer, rather than required reading to understand the rules

minisbett:
>  I guess I'd have the new article linked in a way that makes it sound like a supplemental explainer, rather than required reading to understand the rules

Yeah, this this as I mentioned before where I realistically see something like this, as all of this is too much for the average player that is not affected or has the risk of being affected in any way.

Walavouchey:
gave this a read and i don't see anything new here that the original wording already doesn't encompass. overspecifying rules with specifics and examples has its demerits
- there's overhead in keeping it up-to-date and people will continue to ask for clarifications whenever new corner cases appear that don't line up with the examples
- more words = more people don't read them and ask for tl;drs instead
- it obfuscates the actual intent behind the rule (fair play), which is imo already explained well enough ("If a program is doing something to help you play the game that you should be doing yourself, it isn't okay!")

PR #11886, opened 2024-08-03
Fix incorrect slash character usage for local file paths in `osu! tournament client`

Nivalyx:
noticed this while working on [#11782](http://github.com/ppy/osu-wiki/pull/11782)

basically as it stands the page incorrectly uses forward slashes (`/`) to refer to local files instead of backslashes (`\`)

- [x] ~~*(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)~~ (*shouldn't be required*)


agatemosu:
It should be `/`.

Sincerely, a Windows user.

cl8n:
i was going to say, even on windows i dont think it's a problem to use `/` for the purpose of navigating to folders or starting programs (but i also don't have a way to test rn so idk for sure)

Nivalyx:
hmm okok

PR #11907, opened 2024-08-05
Fix rework link on `The Followpoint: Flaro, the Queen of Speed`

felipemarins:
The hyperlink to Xexxar's rework was wrongly linking to the bonus pp rework when it is supposed to link to the rework that should fix speed pp.

## Self-check


peppy:
Not linking a random external site sorry.

felipemarins:
Sorry. As far as I know, there isn't a github pr/discussion on upstream osu repositories tracking the current progress of this rework. I could link the rework branch on xexxar's fork, but it wouldn't work as an explanation for players wanting to know about it.

PR #11908, opened 2024-08-05
Rename tournament bans to "indefinite" in `Official tournament support`

Userlog2:
Permanent tournament bans were revised to make them "indefinite" somewhat recently.

I believe it's common sense that badges and official banners have to follow the same rules as any other image on osu!.

## Self-check

- [X] check from tournament committee

### DIFF
--- a/wiki/Tournaments/Official_support/en.md
+++ b/wiki/Tournaments/Official_support/en.md
@@ -148,7 +148,7 @@ Participants in officially supported tournaments are expected to adhere to the [
 
 Additionally, players are expected to have *reasonably participated* in a given tournament in order to qualify for any prizes it may award — simply qualifying as a member of a winning team is not enough.
 
-At the account support team's discretion, offending players may be issued timed or indefinite tournament bans depending on the severity of their behaviour. Repeat offenders may risk incurring an indefinite tournament ban or a restriction.
+At the account support team's discretion, offending players may be issued timed or indefinite [tournament bans](/wiki/Help_centre/Tournament_bans) depending on the severity of their behaviour. Repeat offenders may risk incurring an indefinite tournament ban or a restriction.
 
 ## Other procedures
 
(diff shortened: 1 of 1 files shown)


peppy:
Not sure where this change is coming from so going to need context.

Walavouchey:
for the tournament ban wording, context is https://github.com/ppy/osu-wiki/pull/9552 and https://github.com/ppy/osu-wiki/pull/9770. seems like updating here was missed

vcc applies everywhere anyway (as mentioned in the first line of [the article](https://osu.ppy.sh/wiki/en/Rules/Visual_content_considerations)), it just hasn't been relevant in areas like news and probably tournament badges/menu banners

cc @tournament-committee

Userlog2:
I was primarily thinking of [this](https://osu.ppy.sh/community/forums/topics/1798871) when updating the wording to "indefinite".

As for the VCC, while it feels a bit obvious, it's also stated in, for example, the RC and fanart contest news posts. I don't think it hurts to mention VCC for tournament banners/badges as well.

MChecaH:
Brought it up to the other Tournament Committee members. Though this is just my opinion.

In short, the inclusion of the VCC is kind of redundant as *all* visual content in the platform is subject to it. The wording change from permanent to indefinite seems alright. Semantics, but I see how people could be confused otherwise.

MChecaH (approved):
Changes look good. TComm seems OK with them likewise.

wiki/Tournaments/Official_support/en.md
```diff
@@ -148,7 +148,7 @@ Participants in officially supported tournaments are expected to adhere to the [
 
 Additionally, players are expected to have *reasonably participated* in a given tournament in order to qualify for any prizes it may award — simply qualifying as a member of a winning team is not enough.
 
-At the account support team's discretion, offending players may be issued timed or permanent tournament bans depending on the severity of their behaviour. Repeat offenders may risk incurring a permanent tournament ban or a restriction.
+At the account support team's discretion, offending players may be issued timed or indefinite tournament bans depending on the severity of their behaviour. Repeat offenders may risk incurring an indefinite tournament ban or a restriction.
```
Albionthegreat, line 151:
```suggestion
At the account support team's discretion, offending players may be issued timed or indefinite [tournament bans](/wiki/Help_centre/Tournament_bans) depending on the severity of their behaviour. Repeat offenders may risk incurring an indefinite tournament ban or a restriction.
```
Linking to the tournament bans article in help centre would be a worthwhile addition.

Albionthegreat (changes requested):

PR #11933, opened 2024-08-11
Add `History of osu!/Online rankings/osu!catch`

minusQuantumNeko:
## About sources
I mostly relied on Web Archive and Reddit posts when collecting this data, and in some cases also linked posts from number 1's on Twitter. On Web Archive I mostly checked [rankings page](https://osu.ppy.sh/rankings/fruits/performance) and [this wiki page](https://osu.ppy.sh/wiki/en/History_of_osu!/Online_rankings), but also players' profiles to check their rankings graph and username at the time.

## Self-check


wiki/History_of_osu!/Online_rankings/osu!catch/en.md
```diff
@@ -20,10 +20,18 @@ Over the years, there were three different systems used for ranking players:
 
 Because of this, this list only includes players who have reached rank #1 on the main leaderboard that was used at the time. *Italicised* dates and numbers indicate that the corresponding player is continuing their reign as the osu!'s rank #1 holder either from the previous year or into the following year.
 
+## 2018
+
+| Player | From | To | Sources |
+| --: | :-- | :-- | :-- |
+| ::{ flag=RU }:: [skvix](https://osu.ppy.sh/users/4447639) | *2018-04-14* | *2019-04-06* | [\[1\]](https://web.archive.org/web/20180414205044/https://osu.ppy.sh/rankings/fruits/performance) |
```
cl8n, line 27: while the rankings page is a fine source for showing that they were rank 1 at the time of the snapshot, it doesn't show that the "From" and "To" dates are accurate

cl8n, line 27: the asterisks on the dates are meant to show that the date continues into the previous or next year, so on this line, "2018-04-14" should not be italicised

the last entries of 2019, 2020, 2021 have same issue as well

wiki/History_of_osu!/Online_rankings/osu!catch/en.md
```diff
@@ -0,0 +1,100 @@
...
+| ::{ flag=KR }:: [YesMyDarknesss](https://osu.ppy.sh/users/4158549) | 2021-04-14 | 2021-04-22 | [\[1\]](https://web.archive.org/web/20210414210651/https://osu.ppy.sh/rankings/fruits/performance) |
+| ::{ flag=HK }:: [AutoLs](https://osu.ppy.sh/users/9155988) | 2021-04-22 | 2021-04-27 | [\[1\]](https://web.archive.org/web/20210422102910/https://osu.ppy.sh/rankings/fruits/performance) |
+| ::{ flag=KR }:: [YesMyDarknesss](https://osu.ppy.sh/users/4158549) | 2021-04-27 | 2021-04-29 | [\[1\]](https://web.archive.org/web/20210427172037/https://osu.ppy.sh/rankings/fruits/performance) |
+| ::{ flag=HK }:: [AutoLs](https://osu.ppy.sh/users/9155988) | 2021-04-29 | 2021-05-05 | [\[1\]](https://web.archive.org/web/20210429090335/https://osu.ppy.sh/rankings/fruits/performance) |
+| ::{ flag=KR }:: [YesMyDarknesss](https://osu.ppy.sh/users/4158549) | *2021-05-05* | *2024-07-21* | [\[1\]](https://web.archive.org/web/20230308031543/https://osu.ppy.sh/wiki/en/History_of_osu!/Online_rankings) [\[2\]](https://web.archive.org/web/20210608065616/https://osu.ppy.sh/users/4158549) [\[3\]](https://web.archive.org/web/20210505073934/https://osu.ppy.sh/rankings/fruits/performance) |
+
+## 2024
```
cl8n, line 58: if just copying the osu page then there should be sections for 2022 and 2023 which contain a continued listing for yesmydarkness (e.g. https://osu.ppy.sh/wiki/en/History_of_osu%21/Online_rankings/osu%21#2017)

wiki/History_of_osu!/Online_rankings/osu!catch/en.md
```diff
@@ -0,0 +1,100 @@
...
+| --: | :-- | :-- | :-- |
+| ::{ flag=RU }:: [skvix](https://osu.ppy.sh/users/4447639) | *2018-04-14* | *2019-04-06* | [\[1\]](https://web.archive.org/web/20180414205044/https://osu.ppy.sh/rankings/fruits/performance) |
+
+## 2019
+
+| Player | From | To | Sources |
+| --: | :-- | :-- | :-- |
+| ::{ flag=RU }:: [skvix](https://osu.ppy.sh/users/4447639) | *2018-04-14* | 2019-04-06 | [\[1\]](https://web.archive.org/web/20180414205044/https://osu.ppy.sh/rankings/fruits/performance) |
+| ::{ flag=KR }:: [Motion](https://osu.ppy.sh/users/3885626) | 2019-04-06 | 2019-08-09 | [\[1\]](https://web.archive.org/web/20190406161641/https://osu.ppy.sh/rankings/fruits/performance) |
+| ::{ flag=KR }:: [YesMyDarknesss](https://osu.ppy.sh/users/4158549) | 2019-08-09 | 2019-08-18 | [\[1\]](https://www.reddit.com/r/osugame/comments/co0ruo/after_2_years_224hyperw7_has_reclaimed_rank_1_on/) [\[2\]](https://web.archive.org/web/20190810115254/https://osu.ppy.sh/rankings/fruits/performance) |
```
cl8n, line 35: this should either use their name at the time (`[224]Hyperw7`) or a combination of current and former name (`YesMyDarknesss (*[224]Hyperw7*)`). the latter is what the osu page does so probs do that

also i did not check all of the users here for this so there may be more.

cl8n (changes requested):

wiki/History_of_osu!/Online_rankings/osu!catch/en.md
minusQuantumNeko: Not sure if there's any way I could fix this

TicClick:
@Bazzadwarf [wanted to suggest a few extra links with data](https://discord.com/channels/188630481301012481/218677502141399041/1352062394901205175), but otherwise I've no idea where to begin looking outside of the IA

Bazzadwarf (commented):
Posting some thoughts here at the request of @TicClick.

I had originally conducted some [research](https://docs.google.com/spreadsheets/d/1EkK3t_I6Mf7ox4iSk6_tjYyr_1zha5zCNCdE17oYlOs/edit?usp=sharing) for the other games modes after completing the page for osu!. The sticking points that I came across mostly came down to a lack of usable data for the Ranked Score and ppv1. 

While a timeline for ranked score could *potentially* be possible using the same method as was used for osu!, the number of users that would have to be scraped is fairly large and of an unknown quantity, it's doable but it would take some time to sort out. (I may take a look at it again over the summer but that is no guarantee).

On the other hand, with ppv1 I have no idea where to even start. The data is basically none existent and I am not in the know enough with catch to know which users I should start digging into on the web. Due to the large gaps in data it is hard to even do a dirty estimation. 

With the small research that we did, I know with some did hold catch rank 1 at some point in previous systems, use as you see fit:

## Ranked Score
- L-Lawliet (2009)
- mikhe (2010)
- kenji (2011)
- Dark_Diego (2012)

## ppv1
- kenji (2012)
- Dusk (2013)

My two cents for the direction to take the direction of these pages in the short term would be to get pages up that only cover ppv2, the information for these

PR #11939, opened 2024-08-12
[ES] Update `Tournaments/Countries that participated in osu! tournaments`

UnknownOsu:
## Self-check


cl8n:
also I used a program to translate the country names when I was updating this, you might want to review those to see if they match your expectations.

it used whatever Node.js 18's `Intl.DisplayNames` implementation is using and then these exceptions
https://github.com/cl8n/osu-wiki-bin/blob/e3e494854f50b8706c5bdb89d4fe2b4622e77a9c/src/commands/translate-tournament-countries.ts#L45-L49

UnknownOsu:
Yeah, I checked all the names, there is no problem with them

Unnombredeusuarioextremadamentelargo (approved):

TicClick (approved):

PR #11940, opened 2024-08-13
Update `People/Tournament_Committee`

MChecaH:
This update is meant to mirror some information present in [internal documentation](https://pif2.ephemeral.ink/tc-documentation), but hasn't been exposed to the public. These are important things that users should know of, and their enforcement has not been possible as the information was not freely available.

Additionally, to hold the Committee accountable and to inform users, an estimated deadline on when can a user receive a response to their application has been added.

Both values are currently being internally discussed, alongside other information that may need to be mirrored from internal docs to the wiki.

## Self-check

- [x] The Tournament Committee has reviewed the changes. [A vote to decide the exact numbers is currently on going](https://discordapp.com/channels/90072389919997952/625402846489346058/1272723657428435020).


wiki/People/Tournament_Committee/en.md
```diff
@@ -97,6 +97,8 @@ New members to the Tournament Committee must meet the following basic criteria b
 
 Whether a prospective member applies via email or is brought forward by an existing member, the process is the same. The Tournament Committee holds an in-depth discussion on the candidate. Special attention is paid towards their activity in the tournament scene, stature, community contributions, and how they would benefit the committee. Following said discussion, an anonymous vote is held to determine whether to add the candidate to the committee.
 
+Users that fail to gather enough positive votes will be denied, and won't be able to apply for a period of 6 months from the moment they receive a response to their application which will be delivered no later than 2 weeks from the day they apply.
```
cl8n, line 100: I'm assuming this part only goes for people who actually *applied*, right? the previous paragraph makes it sound like this whole section is about both applicants and ppl who were just suggested by existing staff

wiki/People/Tournament_Committee/de.md
```diff
@@ -1,7 +1,7 @@
 ---
 tags:
   - TC
-outdated_since: 9661014ab900317b13040f5f3a2d36ce6b154b07
+outdated_since: 79edb3a3685e1e9f35c34c9733b2cfbff5eb2752
```
cl8n, line 4: don't change outdated_since if it's already there
you can think of these as saying the article is "outdated since the changes to EN at \<time\>", new changes to EN don't make the article any less outdated.

cl8n (changes requested):

wiki/People/Tournament_Committee/en.md
MChecaH: In both cases, both would have the same cooldowns plus app response times. The only thing that would potentially change is the way the additions are communicated. Otherwise, the internal processes are all the same.

wiki/People/Tournament_Committee/de.md
MChecaH: That makes a lot of sense that 3:48 AM me would probably not have properly understood.

Ran the CI script and it threw the outdated error. Will roll back.

wiki/People/Tournament_Committee/en.md
cl8n: so it's possible that I could apply for the first time and be told "sorry, still have 3 months of your cooldown left from the last time we brought you up", even though that wasn't my decision? that doesn't sound right to me but if it's actually how it is then okay.

MChecaH: Will have to ask TComm members about the nuances of it. By the internal docs, I do not see a concrete interpretation on whether the voting is with the user's previous consent, if it's held and then the user is asked to be added afterwards, or what. Forwarding.

hburn7: For now, let's change this to:

> Applicants that fail to gather enough positive votes will be denied and will not be able to reapply for a period of 6 months. This cooldown begins the moment an applicant receives a response to their application. This notification will be delivered no later than 1 week from the date of their application.

hburn7:
imo, we need to change existing wording such that the "users who are brought up internally" part isn't there anymore. Every prospect should apply via email. No user would go on cooldown without a formal application via email.

wiki/People/Tournament_Committee/en.md
MChecaH: Applicants can now **only** apply via email. This has been made clear. If any member in the team wants to bring up a candidate, they should still apply via email. No exceptions. Under this premise, the cooldown on failed applications will always apply.

An [internal vote](https://pif2.ephemeral.ink/discussions/214) was also held. The 6 month cooldown will stay as-is.

wiki/People/Tournament_Committee/en.md
```diff
@@ -95,7 +95,9 @@ New members to the Tournament Committee must meet the following basic criteria b
...
 - Pass a check for recent account infringements
 
-Whether a prospective member applies via email or is brought forward by an existing member, the process is the same. The Tournament Committee holds an in-depth discussion on the candidate. Special attention is paid towards their activity in the tournament scene, stature, community contributions, and how they would benefit the committee. Following said discussion, an anonymous vote is held to determine whether to add the candidate to the committee.
+Once a prospective member applies by sending an email to [tournaments@ppy.sh](mailto:tournaments@ppy.sh), the Tournament Committee holds an in-depth discussion on the candidate. Special attention is paid towards their activity in the tournament scene, stature, community contributions, and how they would benefit the committee. Following said discussion, an anonymous vote is held to determine whether to add the candidate to the committee.
+
+Users that fail to gather enough positive votes will be denied, and won't be able to apply for a period of 6 months from the moment they receive a response to their application which will be delivered no later than 1 week from the day they apply.
```
cl8n, line 100:
```suggestion
Applicants that fail to gather enough positive votes will be denied, and won't be able to apply again for 6 months from the time of response to their application. The response will be sent no later than 1 week from the day the application was submitted.
```
mostly random wording suggestion and then split the last clause into a new sentence cuz it's getting kind of rambly and changing the subject

cl8n (changes requested):
also im assuming it's intended that this doesn't actually say how many "votes" are needed...?

hburn7 (approved):
Changes have been reviewed and approved by TC.

TicClick (approved):

cl8n (approved):

PR #11963, opened 2024-08-14
[FR] Update `Help centre/Upgrading to lazer`

DoryanR:
## Self-check


TicClick (approved):
tiny enough to warrant a pass without a native review

PR #11972, opened 2024-08-15
Fix misleading deprecation notice in skin.ini version 2.6

### LINKED ISSUE #3176: Skin.ini Changelog
Info regarding 2.6 and 2.7 got added recently #2876 
It says that play-warningarrow got removed, however, thats not true. It still works like it used to. If one of the other arrow types is in the folder, it will use that one instead. So if you have arrow-warning it wil use that one for warnings, but play-warningarrow for all the other warnings. 
Also, arrow-generic is used for coop multiplayer scoreboards, I however was not to verify this: https://github.com/ppy/osu-stable-issues/issues/351 

It also should be noted that 2.6 and 2.7 dont have any actual function. They were implemented across all versions, so there is no actual change between 2.6/7 and 2.5

cl8n:
resolves #3176 (min effort)

the whole skin ini changelog needs to be rewritten to properly explain what behavior each version causes, rather than what skin features were introduced at the same time as each version (which are often backward and forward compatible). I have some other skin.ini updates to get back to at some point and I'll probably revisit this at that time

### DIFF
--- a/wiki/Skinning/skin.ini/en.md
+++ b/wiki/Skinning/skin.ini/en.md
@@ -89,7 +89,7 @@ If your `skin.ini` does not specify a `Version`, it will default to this version
 **Allow per-type skinning of arrows**
 
 - Adds `arrow-generic.png`, `arrow-warning.png` and `arrow-pause.png`.
-- Removes `play-warningarrow.png`.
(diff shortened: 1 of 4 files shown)


peppy (approved):
🤷

PR #11985, opened 2024-08-17
Add ``osu! Philippines Nationals 2024 - osu!mania 4K/7K`` tournament article

SurfChu85:
## Self-check


wiki/Tournaments/oPHN/2024_mania/en.md
```diff
@@ -0,0 +1,643 @@
...
+## Prizes
+
+The winners of each division will receive a profile badge.
+
+![](img/phnm-badge_4k.png "PHNM 4K winner badge")
+![](img/phnm-badge_7k.png "PHNM 7K winner badge")
+
+## Organization
+
+The osu! Philippines Nationals 2024 - osu!mania 4K/7K tournament is ran by various community members.
```
iSlodinxOsu, line 35:
```suggestion
The osu! Philippines Nationals 2024 - osu!mania 4K/7K tournament was run by various community members.
```

iSlodinxOsu (changes requested):

TicClick (approved):
surprisingly for me, "Philippine ..." also seems to be a valid phrase

epic-githubber-2-em2:
Heyo, this is "epic man 2" -- sorry if this seems abrupt but I think there is a small misconception in the wiki page here.

image

On here (This is the "Qualifiers" for the 4K side of this tournament) it says that my map (The Contortionist - Reimagined) was used on 1.05x rate, although it was only the base rate that was used and the ID that is used on the map link would originally redirect to the base rate diff, although because I added rates a week after the qualifiers were done the ID swapped to being on the 1.05x rate specifically instead of the original difficulty that was used. Would it be possible to make a small correction here so the information is a bit more accurate and to avoid confusion?

PR #12021, opened 2024-08-23
Mention that mod ID 512 replaces unused Taiko mod 

Oreeeee:
In some 2008 builds of osu!, Taiko instead of being a "game mode" like it is now, used to be a mod, and used the ID 512. After Taiko became a game mode, mod ID 512 went unused until around 2011(?) when the mod ID was reused for Nightcore. I feel like this should be noted for historical reasons.

## Self-check


cl8n (approved):

cl8n:
from the perspective of the average reader that doesn't already know about this I feel like it just begs for more details or explanation. but since this is copying the NoVideo comment and the whole article is kind of low quality anyway... I guess it's fine on its own for now

PR #12023, opened 2024-08-23
[FR] Update `Tournaments/Official support`

DoryanR:
## Self-check


TicClick (dismissed):
(already marked as lacking native review)

PR #12034, opened 2024-08-25
Fixed SV value calculation for missing timing point en.md

darshsingh116:
Updated Misinformation in osu wiki on the time it takes to complete the slider calculation


cl8n (changes requested):
see discord, I don't understand what this is trying to fix

PR #12053, opened 2024-08-27
[FR] Update `Reporting bad behaviour/Handling foul play`

DoryanR:
## Self-check


TicClick (approved):
looks small enough to warrant skipping the native review stage

PR #12057, opened 2024-08-27
Fix redirect of `Community/osu!_Discord_server` across the wiki

DoryanR:
I noticed this on #12054 due to CI errors.

Capture d'écran 2024-08-27 193012

Changes made with Visual Studio code, using the search tool :

Capture d'écran 2024-08-27 192549

SKIP_OUTDATED_CHECK

## Self-check


TicClick (approved):

pishifat (approved):

Walavouchey:
maybe we shouldn't be using redirects to sections if they're always going to break in translations.. ?

making all identifiers english is another way to solve it

DoryanR:
> maybe we shouldn't be using redirects to sections if they're always going to break in translations.. ?
> 
> making all identifiers english is another way to solve it

I've put the identifiers in English for this problem, as I'm familiar with it.

PR #12067, opened 2024-08-29
Update `played` filter in `Beatmap search`

agatemosu:
- [x] change is deployed (in the next lazer update)


agatemosu:
https://github.com/ppy/osu/commit/fde790c014179ab88a381918dc3b8e5354f8173d

peppy (approved):

cl8n (approved):

agatemosu:
i think that we should move the bool explanation of lazer (from `played`) and stable (from `unplayed`) to another place

cl8n:
> i think that we should move the bool explanation of lazer (from `played`) and stable (from `unplayed`) to another place

maybe if they are used in more than one place later

cl8n (approved):

PR #12142, opened 2024-09-15
Remove `Guides/Videos from YouTube` article

### LINKED ISSUE #12133: Change YouTube downloader tools to working ones
ytmp3.cc doesn't work (when I click it, it hangs).
https://youtube-dl.org doesn't work either - it was taken down for legal reasons (something about copyright, based on the [ruling](https://openjur.de/u/2466945.html))
EDIT: While https://ytdl-org.github.io/youtube-dl/ still works, none of the executable downloads work.

New tools need to be added to replace ytmp3.cc. Make sure these are safe to use (virus-free).

Hiviexd:
should close #12133

people seem to agree this isn't an appropriate article to have on the wiki. (see issue for context)


Joehuu (approved):
Can agree. This is more of a general Google type of thing.

cl8n:
I know it's been done before (although I can't remember when most recently), but it feels like deleting a longstanding article with no replacement should leave behind something that's not just a 404.

PR #12166, opened 2024-09-17
[ID] Cleanup/Retranslate `Chat console`

Nivalyx:
this is the only ID translation flagged as `needs_cleanup` across the entire wiki, and after actually cleaning it up i... can understandably see why (as it is apparently a relic from the mediawiki days with all the funky formattings and such)


TicClick (approved):

PR #12216, opened 2024-09-25
Fixed Gameplay listing

Y2nn1c:
Added missing link to Play Styles which was missing in en/fr/es Gameplay wiki.
Furthermore put the German entry into alphabetical order while also correcting the spelling on some to share the same title as the linked Wiki entry.
I do not need a language check since I just copied the headers from the linked articles.

## Self-check


wiki/Gameplay/en.md
```diff
@@ -20,6 +20,7 @@
 - [Mind block](/wiki/Gameplay/Mind_block)
 - [Notelock](/wiki/Gameplay/Judgement/Notelock)
 - [Pass](/wiki/Gameplay/Pass)
+- [Play Style](/wiki/Gameplay/Play_style)
```
Walavouchey, line 23: the title of the article in question isn't in title case

```suggestion
- [Play style](/wiki/Gameplay/Play_style)    
```

Y2nn1c: Will change, should I also change Play Time to Play time then since I did it because it was capitalized there as well and it seems to be the only instance of capitalization

PR #12237, opened 2024-09-28
Update `MWC 4K 2024` with partial results

LeoFLT:
Also updates the schedule

### DIFF
--- a/wiki/Tournaments/MWC/2024_4K/en.md
+++ b/wiki/Tournaments/MWC/2024_4K/en.md
@@ -115,17 +115,10 @@ The complete sign-up list can be found [here](https://gist.github.com/LeoFLT/f3e
 
 ## Match schedule: Grand Finals
 
-### Saturday, 28 September 2024
-
-| Team A | Team B | Match time | Twitch stream |  |
-| --: | :-- | :-- | :-: | :-: |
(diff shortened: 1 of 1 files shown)


wiki/Tournaments/MWC/2024_4K/en.md
```diff
@@ -443,7 +444,7 @@ Sunday, 25 August 2024:
 
 ### Qualifiers
 
-The final standings for the Qualifier stage can be found in the following [spreadsheet](https://docs.google.com/spreadsheets/d/1Al5x3RJQfKD9FiEV9lJAmX0S_ajMdndogNM7_quP3PM/edit?gid=754304763#gid=754304763).
+The final standings for the Qualifier stage can be found in the following [spreadsheet](https://docs.google.com/spreadsheets/d/1Al5x3RJQfKD9FiEV9lJAmX0S_ajMdndogNM7_quP3PM/edit?gid=754304763#gid=754304763).\
```
Walavouchey, line 447: we only stack hatnotes and notices without the paragraph gap.. ?

LeoFLT: That would be inconsistent with the VOD annotations for other stages (as originally suggested by you btw)

Walavouchey: the spreadsheet lines aren't worded like the kinds of notices we stack, they don't seem to be standardised, and they don't have the whole sentence linked like the other vod and mappack notices. similar to the [sign-up list mention](https://osu.ppy.sh/wiki/en/Tournaments/MWC/2024_4K#participants), these look like regular paragraphs

also notices are supposed to stick below the heading, above any regular paragraphs

Walavouchey (approved):

PR #11673, opened 2024-06-30
Update `CWC 2024` with tournament results

LeoFLT:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] ~~*(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)~~ (translations utilize country names as listed on each respective locale)

PR #12240, opened 2024-09-29
Update `MWC 4K 2024` with tournament results

LeoFLT:
Adds tournament results and assets (badge/podium images). Also updates the `Tournaments` index page for up-to-date translations.


Walavouchey (approved):

wiki/Tournaments/MWC/2024_4K/img/mwc4k2024-podium.jpg
Joehuu, line 1: I was hoping that the tooling used to generate these images was updated to use twemoji flags when I pointed it out last time in https://github.com/ppy/osu-wiki/pull/11673#discussion_r1660313837...

LeoFLT: Seems like an oversight on our part, I'll fix it in a bit. Our tooling for this is basically Figma, but the aspect ratio for the flags as suggested in #11673 didn't really line up with the ones we use currently so wasn't trivial to swap to the new ones.

PR #12258, opened 2024-10-02
Trash

Y2nn1c:
Trash


wiki/Gameplay/Input_device/Mouse/img/mouse_drift_animated.mp4
Joehuu, line 1: I don't think mp4 is supported in wiki (there no mp4 files in this repo), and I believe gif works better here so it repeats.

Nivalyx: image

Nivalyx: (no idea if this works on articles but news posts have frequently use youtube embeds)

Hiviexd: this should ideally be a gif, so we can safely store it in the wiki without worrying about 3rd party sources potentially breaking in the future (i.e. video gets deleted/taken down for whatever reason).

cl8n: > (no idea if this works on articles but news posts have frequently used youtube embeds)

html is disabled in wiki articles

peppy:
Neither of these diagrams make any sense to my brain. I checked for a second opinion and they also couldn't follow.

cl8n:
I agree the static image is hard to understand, I would scrap that. the animated one gets the idea across to me, but it's a bit abstract, and also the aesthetics clash hard with the wiki. I kinda want to try making an animation that shows this same concept but literally using osu to explain it

PR #12264, opened 2024-10-03
Document custom game modes from osu!(lazer) in `Game mode`

Tenexxt:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #12272, opened 2024-10-05
Document custom game modes from osu!(lazer) in `Game mode`

Tenexxt:
## Self-check


Tenexxt:
How did i just realise right after commiting that i forgot to add the lazer interface png back -_-
Ill just add it back if there will be anything else to correct, worst thing possible is that ill merge 2 commits instead of 1 if everything (exept that) is fine.

Tenexxt:
Didn't expect for me to _somehow_ do something wrong with tagging the translations as outdated... _again_ ... _this is the 3rd time i did this wrong, wow..._

I'll fix it later with other changes to avoid spamming with commits tho

cl8n:
https://github.com/ppy/osu-wiki/pull/12264#issuecomment-2394945215

PR #12274, opened 2024-10-05
Fix wording of minimum luminosity guideline in `osu! ranking criteria`

cyndifusic:
this is a one-word change - where it says "Dark colours like these impact readability of approach circles with low background dim", the author clearly meant _high_ background dim. low background dim would make it easier to read dark approach circles

## Self-check


cl8n (approved):

cl8n:
thanks for the correction. okoratu also checked this

PR #12313, opened 2024-10-15
Fix Netherlands team listing in OWC 3

osulaport:
Added 2 missing players from Netherlands team, they were participants in group stage and round of 16.


osulaport:
You can ask me directly for verification within osu messaging -laport

cl8n (approved):

PR #12326, opened 2024-10-17
Update `Global Moderation Team` member roles

0x84f:
corrected roles within the team and replaced outdated #portuguese moderators

the entire "group by languages moderated" table needs to be updated but its ambiguous with all the various roles whether someone should be listed there or not so i'll just leave that for another time


cl8n:
you dont need to translate these lists manually. someone else missed that recently too so i wonder if the notes arent visible enough or something (currently above the tables in EN)

> the entire "group by languages moderated" table needs to be updated but its ambiguous with all the various roles whether someone should be listed there or not so i'll just leave that for another time

not sure how more recent editors do it, but in the past I just asked every new gmt (something like "do you actively moderate any specific channels?")

cl8n (approved):

PR #12347, opened 2024-10-22
Update `osu!supporter` user profile images

Hiviexd:
screenshots now match current web (use swipe/onion skin if you're reviewing)


Joehuu (commented):
Should update the language-specific images (i.e. usercard), or argue that they shouldn't be a thing.

IMO, I'm leaning towards slightly against because any slight visual change would need x6 image updates.

cl8n (approved):

PR #12363, opened 2024-10-29
Update drain time thresholds for spread rules in `osu!mania ranking criteria`

MChecaH:
Discussed in [the following forum thread](https://osu.ppy.sh/community/forums/topics/1993866?n=1).

The community would like to try a small reduction on the current thresholds given the positive outcome the last reduction had. This comes after [collecting some data on the number of difficulties made for each of the spread thresholds](https://osu.ppy.sh/community/forums/topics/1993866?n=43).

The same query should be made in 6-12 months after this change is merged to determine whether the outcome of it has been positive. If the number of spreads (particularly those favouring lower difficulties) is reduced in a substantial manner, a revert may be discussed.

## Self-check

- [x] The changes have been properly discussed, and the [proposing changes](https://osu.ppy.sh/wiki/en/Ranking_criteria#proposing-changes) protocol was followed.
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review) Only already up to date translations were modified. No review is required as numbers alone are being changed.


cl8n:
> The same query should be made in 6-12 months after this change is merged to determine whether the outcome of it has been positive. If the number of spreads (particularly those favouring lower difficulties) is reduced in a substantial manner, a revert may be discussed.

since a few people wanted this to be a precondition of merging at all, I think you should decide on a date of review and mention that in the thread. I have a feeling nobody would actually remember in a year from now otherwise

cl8n (approved):

MChecaH:
> I think you should decide on a date of review

Already mentioned by two NATs, including myself. We will review this either in 6 months, 1 year, or both to contrast how things are going. Wisest choice is to do it on both, and rewind if the 6 month back shows less than desirable results beyond reasonable doubt. We'll keep track of this ourselves.

PR #12377, opened 2024-10-31
Remove comma in `osu! World Cup 2024: Qualifiers Recap` news post

iSlodinxOsu:


Walavouchey:
oxford comma usage isn't something we keep consistent in news post. with and without both work

Hiviexd:
oxford comma good

PR #12379, opened 2024-10-31
[ID] Cleanup `Help centre`

Nivalyx:


TicClick (approved):

TicClick:
I have spot-checked the blame, and this looks like an improvement over the original PR #8496, so I decided to skip the native review requirement

PR #12381, opened 2024-11-01
Update activity requirements in Beatmap Nominator rules

Hiviexd:
the following changes were a result from multiple NAT discussions and an internal VC meeting.

- minimum activity has been made consistent across all modes, which allowed for clearer wording
- removed maximum activity

we've already made the required bnsite changes and have started enforcing this a short while ago

main BN page translations technically still use the correct link from their respective rules outdated translations so SKIP_OUTDATED_CHECK


wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -18,27 +18,20 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
...
-| Hybrid | 3 nominations, with at least 1 per proficient game mode |
+Beatmap Nominator activity is measured via nominations[^multi-key-mode], and is evaluated during regular BN evaluations, which happen around every 90 days.
+
+### Minimum activity requirements
+
+- Beatmap Nominators must nominate at least **6** beatmaps over a 90-day period.
+  - Falling below **4** nominations in a 90-day period will result in a removal.
+  - Falling below the **6** nomination requirement with **4** or more nominations will result in an activity warning.
+- Hybrid Beatmap Nominators must nominate at least **3** beatmaps per proficient game mode over a 90-day period.
+  - Falling below **2** nominations per proficient game mode in a 90-day period will result in a removal.
```
aceticke, line 29:
```suggestion
  - Falling below **2** nominations in any given game mode in a 90-day period will result in a removal from that game mode.
```
smthn like this might be better

Hiviexd: applying to the point under it too

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -18,27 +18,20 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
...
-Each mode has its own required minimum for monthly activity.
-
-| Game mode | Requirement |
-| :-- | :-- |
-| osu!, osu!taiko | 3 nominations |
-| osu!catch, osu!mania[^multi-key-mode] | 2 nominations |
-| Hybrid | 3 nominations, with at least 1 per proficient game mode |
+Beatmap Nominator activity is measured via nominations[^multi-key-mode], and is evaluated during regular BN evaluations, which happen around every 90 days.
+
+### Minimum activity requirements
```
cl8n, line 23: this heading feels kind fo useless

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -18,27 +18,20 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
...
+
+- Beatmap Nominators must nominate at least **6** beatmaps over a 90-day period.
+  - Falling below **4** nominations in a 90-day period will result in a removal.
+  - Falling below the **6** nomination requirement with **4** or more nominations will result in an activity warning.
```
cl8n, line 27:
```suggestion
  - An activity warning will be issued when failing to meet this rule.
  - Falling below **4** nominations will always result in removal from the Beatmap Nominators.
```
maybe? it felt weird for me for thsi rule to be formatted like: here's a rule, actually here's a second rule, here's what happens when you break the first one

something similar would apply to the next lines about hybrid

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -18,27 +18,20 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
...
+- Hybrid Beatmap Nominators must nominate at least **3** beatmaps per proficient game mode over a 90-day period.
+  - Falling below **2** nominations in a given game mode in a 90-day period will result in a removal from that game mode.
+  - Falling below the **3** nomination requirement in a given game mode with **2** or more nominations will result in an activity warning in that game mode.
+- When warned for activity, minimum activity requirements must be met over the course of one month.
```
cl8n, line 31: I don't think you mean minimum activity for the entire month because it's impossible to instantly start meeting the minimum once you've been warned

is it supposed to be, meeting the activity minimum *after* a month?

(also small thing but 30 days is probably easier to understand here than 1 month)

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -18,27 +18,20 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
...
+- Failing to meet the required minimum when being warned for it within a period of 1 year will result in a removal from the Beatmap Nominators.
+
+When a Beatmap Nominator's evaluation is created, and they still need to meet their minimum activity requirements, they will receive a reminder via the [NAT bot](https://osu.ppy.sh/users/6616586) and will be given **7** days to meet their quota.
```
cl8n, line 34: not enough details to understand how this part affects all of the warning rules above, or at least it's really confusing

cl8n (changes requested):

wiki/People/Beatmap_Nominators/Rules/en.md
Hiviexd: basically, if you get warned for activity (let's say you have 4 nominations), your activity needs to be at least 6 nominations in the past 90 days on the day your next eval is created (which will happen 30 days after the warning)

wiki/People/Beatmap_Nominators/Rules/en.md
Hiviexd: it's worded this way to make the threshold very clear/explicit

wiki/People/Beatmap_Nominators/Rules/en.md
Hiviexd: i don't really understand how this is confusing. when an eval is created and the BN hasn't met their activity, they're given a 7-day leeway to reach their activity minimum before we finalize a decision, and are reminded via the nat bot in the form of a DM.

wiki/People/Beatmap_Nominators/Rules/en.md
cl8n: > on the day your next eval is created (which will happen 30 days after the warning)

that's not what "over the course of 30 days" means. should be "after 30 days" (or other wording to the same effect)

Hiviexd: will do "after 30 days"

wiki/People/Beatmap_Nominators/Rules/en.md
cl8n: I can't tell if it means you have 7 days to produce a 90-day period where you meet the quota, or if you have 7 days to make up the difference between the quota and whatever you nominated in the 90 days leading up to the eval. and if you ultimately fail the check after 7 days, does your "activity warning" start at the time of the eval or 7 days past that?

your reply now makes it sound like this is all a roundabout way to write "97-day period" :thinking:

Hiviexd: > or if you have 7 days to make up the difference between the quota and whatever you nominated in the 90 days leading up to the eval

this

> your reply now makes it sound like this is all a roundabout way to write "97-day period"

ehhh, kinda? intention was 90 days with a 7-day grace period to catch up, given evals tend to last 14 days.

wiki/People/Beatmap_Nominators/Rules/en.md
Walavouchey: i agree with clayton here

wiki/People/Beatmap_Nominators/Rules/en.md
```diff
@@ -18,27 +18,18 @@ Sustained behaviour in conflict with these expectations is grounds for dismissal
...
-| Game mode | Requirement |
-| :-- | :-- |
-| osu!, osu!taiko | 3 nominations |
-| osu!catch, osu!mania[^multi-key-mode] | 2 nominations |
-| Hybrid | 3 nominations, with at least 1 per proficient game mode |
+Beatmap Nominator activity is measured via nominations,[^multi-key-mode] and is evaluated during regular BN evaluations, which happen around every 90 days.
+
+- Beatmap Nominators must nominate at least **6** beatmaps over a 90-day period.
+  - An activity warning will be issued when failing to meet this rule.  
+  - Falling below **4** nominations will always result in removal from the Beatmap Nominators instead.
```
Hiviexd, line 25:
```suggestion
  - Falling below **4** nominations will always result in removal from the Beatmap Nominators.
```

cl8n (approved):
i guess i wont hold this up on my comments since it's better than before and these rules gotta be written somewhere

I still think a lot of the edge cases aren't sound though, and are silently assuming that the reader knows more about how evals work. which might be the case for BNs but not other readers (or me)

PR #12382, opened 2024-11-01
Update `TMWC 2024` article with results

Hiviexd:
rewards are confirmed now


wiki/Contests/TMWC/2024/en.md
```diff
@@ -35,48 +31,58 @@ Registrations will be open from May 17 until June 8, 2024.
 | Event | Timestamp (UTC) |
 | --: | :-- |
 | Round 4 mapping phase | 2024-07-27 until 2024-08-10 |
```
cl8n, line 33: "Round 4 and 5 mapping phase" if they're just the same

same for rest of table and 6+7

cl8n (changes requested):
also it seems like the podium image may compress better as PNG. cant try it myself without the original though

cl8n (approved):

PR #12388, opened 2024-11-03
Add link to "Past osu!store items" on `Store archive`

Nivalyx:
(so that the article don't end up being orphaned, e.g. it has some other article linking to them)


The-Last-Cookie:
I think [History of osu! § Other records](https://osu.ppy.sh/wiki/en/History_of_osu%21#other-records) is also a fitting place.

Nivalyx:
> I think [History of osu! § Other records](https://osu.ppy.sh/wiki/en/History_of_osu%21#other-records) is also a fitting place.

Fair enough, added 👍

peppy (approved):

PR #12393, opened 2024-11-03
Remove detail about maps moving out of Loved in `Beatmap category`

iSlodinxOsu:


cl8n:
in this case I think the part in parenthesis is just unnecessary

cl8n (approved):

cl8n:
also dont skip ci checks if you have no special reason to :thinking:

PR #12394, opened 2024-11-03
Update `OWC 2024` for the Round of 16 stage

LeoFLT:
Adds RO32 results/RO16 schedules and mappool. Updates the staff listing with new mappers.


wiki/Tournaments/OWC/2024/en.md
```diff
@@ -171,7 +212,7 @@ The complete sign-up list can be found [here](https://gist.github.com/LeoFLT/838
   2. [senya - Sakuretsu Irony (Satellite) \[Satellite\]](https://osu.ppy.sh/beatmapsets/2271602#osu/4838687)
   3. [Two Door Cinema Club - Cigarettes In The Theatre (Lesjuh) \[Insane\]](https://osu.ppy.sh/beatmapsets/37461#osu/120461)
 - Free Mod
-  1. [kano - Houkago Stride (Clementine) \[RLC\]](https://osu.ppy.sh/beatmapsets/79878#osu/228679)
+  1. [kano - Houkago Stride (RLC) \[RLC\]](https://osu.ppy.sh/beatmapsets/79878#osu/228679)
```
cl8n, line 215: ok i guess but then why is the map below this listed as sahuang if it's actually Regou's map
is there a consistent standard for this?

cl8n (changes requested):

wiki/Tournaments/OWC/2024/en.md
LeoFLT: the pattern we've been following is to credit the actual mapper vs the set host, which should be standard nowadays for obvious reasons (IMO at least, but seeing as osu-web is crediting the guest difficulty owner instead of set owner, shouldn't be even a topic of discussion). Fixed this inconsistency where applicable in a72715531e625c7fa68d951e8f549185cbcf0bf4

cl8n: i wouldnt say it's "obvious", for me at least it's way easier to ID sets when it's the host's name instead (and this line is a great example of that). so to me there's a tradeoff between identification and crediting

cl8n (approved):

PR #12413, opened 2024-11-09
Update Centurion listing

momoyosu:
## Self-check


Walavouchey (approved):
note that we don't typically keep updating usernames on the wiki, but since this one matches the [pack listing](https://osu.ppy.sh/beatmaps/packs/TM22) i suppose it makes sense

momoyosu:
got it

PR #12478, opened 2024-11-24
Cleanup `Beatmap packs`

Nivalyx:
Wanted to fix a single typo initially, but upon further reading I noticed that certain sections in this article could be written more properly so I went to clean that up as well

⛷️ 


Nivalyx:
applied all changes, should be ready for merging now @Hiviexd

Hiviexd (approved):

PR #12486, opened 2024-11-26
Update osu!taiko pp committee member list

Lawtrohux:
Natelytle and Horiiizon have departed the osu!taiko pp committee, this moves them into the former section.


cl8n (approved):

cl8n:
i trust this is correct but the team on github doesn't seem to have been updated accordingly

PR #12492, opened 2024-11-27
Define public dataset expiry window in `Tournaments/Official_Support`

hburn7:
Specify that public datasets for automated tools should not be more than 30 days old.

- [x] TC approval
- [x] Outdate translations
- [x] Update last-updated date.


wiki/Tournaments/Official_support/en.md
```diff
@@ -124,7 +124,7 @@ Registrant filtering, not to be confused with [screening](#tournament-screening)
 
 In addition to the expectations for programs above, tools or systems used to filter registrants or seed players must also abide by the following practices:
 
-- If a dataset is used, it must be up to date, publicly available and downloadable. Documentation must be provided explaining how to recreate the dataset and reproduce the results using it — including code where applicable.
+- If a dataset is used, it must be publicly available, downloadable, and no more than 30 days out of date. Documentation must be provided explaining how to recreate the dataset and reproduce the results using it — including code where applicable.
```
Walavouchey, line 127: can you give an example of what "dataset" refers to here? data on what? it doesn't mean anything to an outsider

hburn7: Dataset essentially means a database of osu! match data which these programs reference in their algorithms. So the context here is "If the program you want to use has a database it builds its performance network from, it must be public".

I think dataset is used twice here to mean two different things. The first use of dataset is talking about what I write above. The second use is, I think, talking about the actual performance network.

There are community tools out there / in development which convert a large amount of osu! match data into some meaningful measure of performance on a per-player basis. i.e. "Player X has 500 rating and Player Z has 200, therefore Player X should win much more often." This rule is in place to allow third parties to validate that the database of osu! data is valid and that the "rating ladder", if you will, isn't being tampered with. Without this rule, hypothetically, the maintainers of such programs could be artificially manipulating the osu! match data fed into the system. They could, in theory, also be artificially boosting or weakening certain players to manipulate who is eligible for certain tournaments.

hburn7:
Added a footnote for dataset, discussed with @ThePooN on data archival period and disclosure of the seed / filter dates. Will look to get TC signoff over the next days.

wiki/Tournaments/Official_support/en.md
```diff
@@ -124,7 +124,8 @@ Registrant filtering, not to be confused with [screening](#tournament-screening)
 
 In addition to the expectations for programs above, tools or systems used to filter registrants or seed players must also abide by the following practices:
 
-- If a dataset is used, it must be up to date, publicly available and downloadable. Documentation must be provided explaining how to recreate the dataset and reproduce the results using it — including code where applicable.
+- If a [^dataset] is used, archives must be publicly available, downloadable, timestamped, and have the most recent archive be no more than 30 days out of date. Archives must remain available for at least 6 months. Documentation must be provided explaining how to reproduce the results using it — including code where applicable.
```
Walavouchey, line 127: i don't think we need a footnote with a dictionary definition, but just a way to understand what is being said

```suggestion
- If a dataset of osu! multiplayer matches (or otherwise) is used, archives must be publicly available, downloadable, timestamped, and have the most recent archive be no more than 30 days out of date. Archives must remain available for at least 6 months. Documentation must be provided explaining how to reproduce the results using it — including code where applicable.
```

wiki/Tournaments/Official_support/en.md
```diff
@@ -299,5 +300,6 @@ Yes! See [Official beatmapping contest support](/wiki/Contests/Official_support)
 
 ## Notes
 
+[^dataset]: A collection of data, including databases, typically comprised of osu! match data.
```
Walavouchey, line 303:
```suggestion
```

wiki/Tournaments/Official_support/en.md
hburn7: Agree with this

wiki/Tournaments/Official_support/en.md
```diff
@@ -124,7 +124,8 @@ Registrant filtering, not to be confused with [screening](#tournament-screening)
 
 In addition to the expectations for programs above, tools or systems used to filter registrants or seed players must also abide by the following practices:
 
-- If a dataset is used, it must be up to date, publicly available and downloadable. Documentation must be provided explaining how to recreate the dataset and reproduce the results using it — including code where applicable.
+- If a dataset of osu! multiplayer matches (or otherwise) is used, archives must be publicly available, downloadable, timestamped, and have the most recent archive be no more than 30 days out of date. Archives must remain available for at least 6 months. Documentation must be provided explaining how to reproduce the results using it — including code where applicable.
```
Albionthegreat, line 127:
```suggestion
- If a dataset of osu! multiplayer matches, player scores or otherwise is used, archives must be publicly available, downloadable, timestamped, and have the most recent archive be no more than 30 days out of date. Archives must remain available for at least 6 months. Documentation must be provided explaining how to reproduce the results using it — including code where applicable.
```
Though I recognise "or otherwise" is present here as a catch all, I would also like to see player scores mentioend, since  multiplayer matches and user scores combined cover 98% of cases.

hburn7: Would a `,` be needed before `or`?

hburn7: Applied

Albionthegreat (approved):
The changes were discussed and approved by the tournament committee.

Walavouchey (approved):

hburn7:
Albion said KR translation needs to be outdated still, I won't get to that for a few hours. If auto merge sends it then I'll handle it in another PR.

Walavouchey (approved):

PR #12509, opened 2024-12-01
Update grading system heading in `Gameplay differences in osu!(lazer)`

Hiviexd:
opted for consistent wording with https://github.com/ppy/osu-wiki/blob/a1d6efbb2f2ed80a1f6da7fb7018a5a2d1f2db45/wiki/Client/Release_stream/Lazer/Gameplay_differences_in_osu!(lazer)/en.md?plain=1#L66

Couldn't really figure something that encapsulates both the acc threshold changes and the removal of judgments while not being too wordy. Open for suggestions.

### DIFF
--- a/wiki/Client/Release_stream/Lazer/Gameplay_differences_in_osu!(lazer)/en.md
+++ b/wiki/Client/Release_stream/Lazer/Gameplay_differences_in_osu!(lazer)/en.md
@@ -17,7 +17,7 @@ Instead of pausing gameplay while filling up the health bar, health is restored 
 | Intentionally changed | ![No][false] |
 | Needs further consideration | ![Yes][true] |
 
-### All game modes except osu!catch use the same grading system
+### Differences in grading system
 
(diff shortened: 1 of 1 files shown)


wiki/Client/Release_stream/Lazer/Gameplay_differences_in_osu!(lazer)/en.md
```diff
@@ -17,7 +17,7 @@ Instead of pausing gameplay while filling up the health bar, health is restored
 | Intentionally changed | ![No][false] |
 | Needs further consideration | ![Yes][true] |
 
-### All game modes except osu!catch use the same grading system
+### Differences in grading system
```
Walavouchey, line 20:
```suggestion
### Differences in grading systems
```

Walavouchey (approved):
> Couldn't really figure something that encapsulates both the acc threshold changes and the removal of judgments while not being too wordy. Open for suggestions.

the section explains itself so it's ok i think

peppy (approved):

PR #12513, opened 2024-12-01
Update `OWC 2024` mappool links

LeoFLT:
Swaps remaining temporary links for their permanent versions. Updates the staff listing again to account for one missing mapper


wiki/Tournaments/OWC/2024/en.md
```diff
@@ -153,12 +153,12 @@ The complete sign-up list can be found [here](https://gist.github.com/LeoFLT/838
 
 ### Grand Finals
 
-**[Download the mappack here (270 MB, *temporary link*)](https://drive.google.com/file/d/1mIk-iRaqd_p2lH1FUoWSTC79K9QDpUC5/view?usp=sharing)**\
+**[Download the mappack here (268 MB)](https://drive.usercontent.google.com/download?id=1mIk-iRaqd_p2lH1FUoWSTC79K9QDpUC5&export=download)**\
```
Walavouchey, line 156: this is still temporary if it's a drive link (we use beatmap packs nowadays)

LeoFLT: Technically a complete pack, just not uploaded on assets since the tiebreaker has a larger than acceptable storyboard (which we still wanted to ship while storyboarders reduce its size) but here it is bcf62ad0fe4072e4c43b67c67c91562e9213690e

Walavouchey (approved):

PR #12551, opened 2024-12-13
include lazer in grade page

sparkhere-sys:
the grading in osu! Lazer in std is very similar to osu!mania, iirc the exact same

### DIFF
--- a/wiki/Gameplay/Grade/en.md
+++ b/wiki/Gameplay/Grade/en.md
@@ -59,3 +59,5 @@ From lowest to highest, the possible grades are D, C, B, A, S, and SS. Silver S 
 | B | Over 80% accuracy |
 | C | Over 70% accuracy |
 | D | Anything else |
+
+In osu! Lazer, the criteria in osu! is the same as in osu!mania.


wiki/Gameplay/Grade/en.md
```diff
@@ -59,3 +59,5 @@ From lowest to highest, the possible grades are D, C, B, A, S, and SS. Silver S
 | B | Over 80% accuracy |
 | C | Over 70% accuracy |
 | D | Anything else |
+
+In osu! Lazer, the criteria in osu! is the same as in osu!mania.
```
iSlodinxOsu, line 63:
```suggestion
In osu!(lazer), the criteria in osu! is the same as in osu!mania.
```

iSlodinxOsu (changes requested):

TicClick:
two things to mention here:
- please don't use `SKIP_OUTDATED_CHECK` when changes need to be propagated to translations
- your change reads out of place -- my interpretation is that it has to do with the osu! game mode, not osu!mania, unless I misunderstand it

peppy:
I don't think this works too well, needs more thought.

PR #12572, opened 2024-12-17
Clarify that songs featuring disallowed artists that are licensed as part of other artist's FA listing are fine to use in `Content usage permissions`

Nivalyx:
Discussed in #nat just now.

Recently there has been some confusion regarding whether the song `Kiss As A Vanity feat. Hatsuki Yura`, which is currently licensed as part of [Tokyo.MeltiMelt's FA listing](http://osu.ppy.sh/beatmaps/artists/308), is allowed to be Ranked given that Hatsuki Yura herself is in the list of disallowed artists.

This change adds a clarification to the `Content usage permissions` wiki page that on such cases, the FA listing precedes over the restriction as it has always been.


peppy:
Will leave to others to decide whether it is or not. To me it feels 100% redundant/obvious 🤷 .

peppy (commented):
indifferent

MChecaH:
Probably biased as I'm used to the current wording, but the extra sentence just makes it more confusing to read IMO.

It already states *all* tracks in the listing are good to use. Licenses are on a per song basis, not whether a disallowed artist is featured in them or not. Feels redundant and prone to users asking what the extra words are making reference to. If anything, a better change to emphasize this could be to bold the entirety of `**All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are approved for use in the creation of osu! beatmaps.**`

wiki/Rules/Content_usage_permissions/en.md
```diff
@@ -32,7 +32,7 @@ While music is frequently copyright-restricted, sound effects used for custom hi
 
 This section details which artists are willing to allow use of their audio works in osu!, and under what conditions (where appropriate).
 
-All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are approved for use in the creation of osu! beatmaps. Tracks from a Featured Artist that do not appear in their listing are *not* licensed, and additional permissions may be required to use them.
+All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are approved for use in the creation of osu! beatmaps regardless of the standing of their additional performers (if any). Tracks from a Featured Artist that do not appear in their listing are *not* licensed, and additional permissions may be required to use them.
```
Walavouchey, line 35: mentioning the case of "additional performers" seems oddly specific and probably doesn't cover everything that it's supposed to clarify

for the record, i do agree with peppy. how can a track be licensed (in the fa listing), but somehow be disallowed from ranking? this is already clarified in the first sentence here

```suggestion
All tracks listed on any [Featured Artist listing](https://osu.ppy.sh/beatmaps/artists) are licensed and approved for use in the creation of osu! beatmaps, regardless of any blanket disapprovals in these lists. Tracks from a Featured Artist that do not appear in their listing are *not* licensed, and additional permissions may be required to use them.
```

Nivalyx:
Yeah, I agree that this might create just a bit of redundancy — but people have been legitimately confused by this more than once (I remember back then [Altai was also confused as to what he can and can't do with this](http://osu.ppy.sh/s/2085134) as Igorrr was also disallowed)...

...so I suppose this clarification would be nice to have nevertheless 🐇 

Either way @Walavouchey's wording sounds good for what the PR is trying to achieve so I'm going with that I guess

Walavouchey (approved):

peppy (approved):

PR #12586, opened 2024-12-22
Adjust Z-sum/Z-score definition on `osu! Indonesia Tournament #14 (2024)`

Nivalyx:
(the tournament management wanted the formula to be explained this way before handing out badges apparently)

image

## Self-check


cl8n (approved):
this seems kind of implied to me but okay (and idk if I trust my high-school-level understanding of stats)

PR #12599, opened 2024-12-26
Add Daily Challenge medals to `Dedication` category in unlock requirements

minusQuantumNeko:
## Self-check


minusQuantumNeko:
Not sure if `All-Mode` is the term you want to use, maybe could just change to "Daily Challenge" cuz that's literally it atm, up to you really

cl8n (approved):

cl8n:
> Not sure if `All-Mode` is the term you want to use, maybe could just change to "Daily Challenge" cuz that's literally it atm, up to you really

yeah. 
also idk why "Skill & Dedication" got split, now it looks weird that it isn't all one type of medal in here :thinking:

PR #12643, opened 2025-01-06
Move and rename osu! originals article

Walavouchey:
[web preview](https://osu1.roanh.dev/wiki/en/Community/Bespoke_music)

a discussion and poll was conducted among the team recently about whether "osu! original" should encompass only officially licensed music (fa listing) or anything made within osu! (current definition), and most were of the former opinion

whether to keep remixes/bootlegs was also in the discussion but the outcome was tied (it'd also incur the question of what is considered a remix, and it'd likely imply that many licensed remixes wouldn't be labelled as original), so that is not taken into consideration for now

this doesn't update the listing with new tracks, and the text on the graph will be updated in conjunction with the next update

the signifier for originals in the listing may need some design input if the current solution doesn't stick. it can't be too wide like the pink pill on the fa listing, and separating osu! originals into new tables doesn't make sense


peppy (approved):

cl8n:
"bespoke" is definitely not the word I'd use to describe these then if the goal is to make it sound less affiliated with osu. most of the time I see that word now it refers to something that was commissioned

peppy:
The goal is to not use the "osu! original" term. The tracks generally *are* commissioned so I think it makes sense?

PR #12661, opened 2025-01-12
Update `Ranking Criteria` with new difficulty name rules.

MChecaH:
As discussed, the majority of users seemed to dislike the rule. As such, it will be removed. Many preferred its removal rather than settling with a guideline.

## Self-check

- [x] The changes have been [discussed and agreed upon by the community](https://osu.ppy.sh/community/forums/topics/2019747?n=1).
- [x] The changes have been [reviewed and agreed internally](https://discord.com/channels/90072389919997952/1327548349682942022/1327758835531645008).


Hiviexd (approved):

peppy:
I dunno. I'm not approving this because I don't agree with it. But I also won't block it because who am I to judge I guess.

Espionage724:
The idea of an Expert map named "Easy" because the mapper's name is "EasyMapperXx192" is silly. I feel people only want this for noritety vs any direct in-game benefit, and feel it'll eventually be abused with people with sly nicknames.

NoffyNoffeh:
> The idea of an Expert map named "Easy" because the mapper's name is "EasyMapperXx192" is silly. I feel people only want this for noritety vs any direct in-game benefit, and feel it'll eventually be abused with people with sly nicknames.

Just because usernames can be referenced as the difficulty name in this PR does not mean other difficulty name rules are gone. Difficulty names are still not allowed to be misrepresentative, so the case you're describing still isn't allowed.

PR #12663, opened 2025-01-12
Fix broken links on `History of osu!` pages

tsunyoku:
## Self-check


cl8n (approved):
looks like a script gone wrong

PR #12668, opened 2025-01-14
Use proper form URL for Tournament Commitee applications news post

venix12:
pretty sure it's supposed to be that, as of form auth workflow :thinking:

Current URL points to the same form without OAauth step, asking to access it with ``form-auth`` link.


Walavouchey (approved):

hburn7:
Tested, flow is much better this way - thank you.

PR #12676, opened 2025-01-17
Added `ladder` to the table of contents of videos for the tournament client

zAngarn:
This one is not mentioned anywhere, but it is still used by the game (as seen [here](https://github.com/ppy/osu/blob/master/osu.Game.Tournament/Screens/Ladder/LadderScreen.cs))

Now, for the english translation I used `Bracket` instead of `ladder`, because the earlier is the one that is used on the button label in the tournament client, while the later is only used within the codebase. Maybe the name of the file should be changed internally to `bracket` to make it more intuitive instead of ladder, but that is up for discussion for the EN file.

The spanish one should be right (I'm a native spaniard), so i'll leave it up to other translators to add the missing line in the table

## Self-check


zAngarn:
Added the outdated label to other languages.

zAngarn:
Will mark the last task as done when someone confirms whether or not it should be ladder or bracket.

TicClick (approved):
approval from LeoFLT: https://discord.com/channels/188630481301012481/218677502141399041/1333461422306492416

PR #12714, opened 2025-01-28
[VI] Add `Help centre/Silences`

kussu0998:
## 

### DIFF
--- a/wiki/Help_centre/Silences/vi.md
+++ b/wiki/Help_centre/Silences/vi.md
@@ -1,3 +1,7 @@
+---
+no_native_review: true
+---
+
 # Sự im lặng
 
 **Im lặng** là một thuật ngữ được sử dụng để mô tả hình phạt giới hạn phương tiện giao tiếp của người dùng trong cộng đồng osu!. Các thành viên của [nhóm điều hành toàn Cầu](/wiki/People/Global_Moderation_Team) và [nhóm đánh giá đề cử](/wiki/People/Nomination_Assessment_Team) thường đưa ra hình phạt im lặng này để duy trì một môi trường sạch sẽ, cả trên trang web lẫn trong game.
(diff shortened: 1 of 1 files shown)


kussu0998:
I don't know how to fix the error...

kussu0998:
hi @TicClick , may I ask how to fix this link error?

TicClick:
> hi @TicClick , may I ask how to fix this link error?

my bad, didn't see the notification about your question. it's easy — you missed the leading slash (`/wiki/`, not `wiki/`), and as soon as you add it, the checks should pass. let me know if you need further help, especially if it was the only thing that made you delete the changes (they should be equally easy to restore)

TicClick:
lemme help really quick

TicClick (approved):
the issue was that the Silence article had been moved to a completely different location, along with its images. since you created the branch way before that, in _your_ copy of the wiki it was still at its old location:

```
wiki/
|- ...
|- Silence/
   |- img/
   |- en.md
   |- vi.md
   |- ...
|- ...
```

what would happen during the merge is that it would have recreated the old location with your translation being the only file there:

```
wiki/
|- Help_centre/
   |- Silences/
      |- img/
      |- en.md
      |- ...
|- ...
|- Silence/
   |- vi.md
|- ...
```

so, the images in your translation would be broken, and _that_ was the reason for the error -- the system checks in which state the wiki would be after the merge.

anyway, I fixed this by updating the branch and moving the article to the new location

(and we should merge this in anyway -- doesn't seem like your PR is getting reviewed by other Vietnamese folks)

TicClick:
to clarify, it's not your fault -- the case was fairly complex, because it both included an uncommon situation (whole article moved, a file added to the old location) and exposed technical details that aren't very straightforward and irrelevant to the wiki

PR #12747, opened 2025-02-02
Add `poolingcore` to `Community/Discord_servers`

hburn7:
## Self-check


peppy:
@Walavouchey have you done a quick check on the general moderation / culture on the server? All looks good?

Walavouchey:
did earlier and it looks good, is run by prominent tournament people

PR #12802, opened 2025-02-11
Adjust vote count in `Skin of the Year 2024: Voting Open` news post

Walavouchey:
so i missed the memo specifying 12 votes instead of 15

- [ ] if adjusting the contest page is not feasible (due to people already voting) we can just roll with it


Hiviexd (approved):

0x84f:
AFAIK its been determined that the vote count cant change on the voting page now so this can be merged

peppy:
has it? i did change it already 🤔

PR #12823, opened 2025-02-15
[ID] Update `Community`

Nivalyx:


TicClick (approved):
probably too small to require a native review

PR #12834, opened 2025-02-17
Add menu tip about 'laser' writing

The-Last-Cookie:
[This commit](https://github.com/ppy/osu/commit/1174f46656510e7524af30d5218fd48b7d99b0d9) added a new menu tip detailing the difference between "lazer" and "laser".

I believe that this should be added to `Menu tips`.

I added the translations from Crowdin where available.

## Self-check


Walavouchey (commented):
hmm the source link in the comment above the list could be updated as well

`https://github.com/ppy/osu/blob/master/osu.Game/Localisation/MenuTipStrings.cs`

peppy:
I really have doubt about the value this page holds :shrug

Walavouchey (approved):

Walavouchey:
@peppy maybe you wouldn't be as fond of those program/game manuals that describe everything about the interface. we have that for all the stable screens, which can sometimes be useful for new people who don't know all the intricacies yet. many features are "hidden" or sometimes hard to notice, [here's an example](https://clips.twitch.tv/SillyAbstemiousWebCoolStoryBob-cCk5owD5aQ4BoQ9T)

this particular menu tip is just trivia, but that's cool too

PR #12853, opened 2025-02-22
Note that some parts of timing points in .osu files are optional

staticssleever668:
## Self-check


TicClick:
> In fact, lazer checks most of them as optional. [ppy/osu@`89e3c55`/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs#L489-L517](https://github.com/ppy/osu/blob/89e3c551ffe1a0513fb0e08c3a3d75bc07083cc2/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs#L489-L517) I have only marked last two as optional as that is what I've witnessed personally in older beatmaps.

can't currently test this, but what happens if you go by the parser's code and remove these attributes from random beatmaps' timing points? does this result in default values from the code or not?

if not, I'll merge the change as-is

peppy:
If the client isn't doing this, then it doesn't need to be documented IMO.

PR #12894, opened 2025-03-02
Update official tournament support article's last updated date

hburn7:
Sets the last updated date of the TC official support wiki page to March 02, 2025. We did technically change a procedure with #12892 so would probably be good to have this date reflect that.


hburn7:
Unsure if I need to set this for the translations too.

PR #12913, opened 2025-03-05
Update `osu! ranking criteria` with revised slider rule on Expert difficulties

NoffyNoffeh:
This was approved and even allowed in a ranked map following the proposal, but fell through for making it into the actual wiki. This PR simply fixes that to provide clarity for future maps.

Relevant RC thread: https://osu.ppy.sh/community/forums/topics/1979244?n=1

## Self-check


Walavouchey (commented):
seemingly almost the same thing. i'd have to see examples of what new sliders this allows to understand what changed

edit: ah yaspo pointed out a good example in the forum thread https://yaspo.s-ul.eu/lKfHuwlR. the main difference is that the cursor doesn't have to strictly be within the slider follow circle

Walavouchey (approved):

peppy (approved):

PR #12918, opened 2025-03-06
Update `osu!supporter` with new features

MChecaH:
[Preview](https://osu3.roanh.dev/wiki/en/osu!supporter).

Updates the supporter wiki article with the new osu!supporter exclusive perks for the recently added Teams feature. Discussed [in the `#osu-wiki` channel](https://discord.com/channels/188630481301012481/218677502141399041/1345433434222231665).

## Self-check

- [ ] Outdate translations.

### DIFF
--- a/wiki/osu!supporter/en.md
+++ b/wiki/osu!supporter/en.md
@@ -95,6 +95,7 @@ osu! offers more relaxed limits on various online features to supporters:
 | Value | Regular limit | osu!supporter limit |
 | :-- | :-: | :-: |
 | [Pending beatmaps](/wiki/Beatmap/Category#wip-and-pending) | `4 + min(ranked beatmaps, 4)`, up to **8**[^pending-beatmaps-ref] | `8 + min(ranked beatmaps, 12)`, up to **20**[^pending-beatmaps-ref] |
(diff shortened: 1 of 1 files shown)


wiki/osu!supporter/en.md
```diff
@@ -164,6 +165,7 @@ The giver can also hide this event from their own profile by ticking `Hide all o
 
 Additionally, both the giver and the recipient will receive email notifications mentioning the purchase.
 
-## References
+## Notes
 
 [^pending-beatmaps-ref]: [Forum post by peppy (2021-09-14) in "Increase the number of pending beatmap slots"](https://osu.ppy.sh/community/forums/posts/8294132)
+[^team-size-note]: Each supporter will add 4 slots to their team's base size of 8, up to a cap of 256.
```
Walavouchey, line 171: why is this note needed 🤔
isn't it just restating the formula? the other table rows don't have that

MChecaH: The difference is that the other limits aren't really "Dynamic". The only similar one being Pending beatmaps. However this is not a limit individual to the user, so I just wanted to make it clear that the limit depends on more supporters, other than the user.

PR #12954, opened 2025-03-13
Add asset visualisation tool to `Official support` and `News styling criteria`

Hiviexd:
- [ ] TC eyes

### DIFF
--- a/wiki/Tournaments/Official_support/en.md
+++ b/wiki/Tournaments/Official_support/en.md
@@ -218,6 +218,8 @@ Please consult the following examples of approved badge designs:
 
 ![](https://assets.ppy.sh/profile-badges/ysc4-2021@2x.png) ![](https://assets.ppy.sh/profile-badges/perennial-2021@2x.png) ![](https://assets.ppy.sh/profile-badges/TATHC2023@2x.png)
 
+Use [this badge visualisation tool](https://tcomm.hivie.tn/assets-previewer?tab=badges) to help you verify that your badge complies with the above standards.
+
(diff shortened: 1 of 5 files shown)


hburn7 (approved):
tc eyes

Albionthegreat (approved):

wiki/Tournaments/Official_support/en.md
```diff
@@ -271,6 +273,8 @@ Please consult the following examples of acceptable banner designs:
 
 ![](https://assets.ppy.sh/main-menu/egts-2021/gf.png)
 
+Use [this banner visualisation tool](https://tcomm.hivie.tn/assets-previewer?tab=in-game-banners) to check how the banner would look in-game.
```
Albionthegreat, line 276:
```suggestion
Use [this banner visualisation tool](https://tcomm.hivie.tn/assets-previewer?tab=in-game-banners) to check how the banner would look in game.
```
This is not a compound adjective, so it doesn't need the hyphen.

Albionthegreat (changes requested):
Should probably also change the "last updated" date, since the page has seen modifications.

wiki/Tournaments/Official_support/en.md
Walavouchey: it is something that has always been spelt with a hyphen on the wiki regardless 🤔
https://osu.ppy.sh/wiki/en/Article_styling_criteria/Writing#terminology

Walavouchey (approved):

wiki/Tournaments/Official_support/en.md
Hiviexd: would prefer to stick to in-game unless we change our terminology for this

PR #12961, opened 2025-03-14
`Ranking criteria` allow mapset host to indicate possession in a difficulty name

Okorin:
Forum thread with conclusions: 
https://osu.ppy.sh/community/forums/topics/2052052

Since the difficulty name rules were a bit spread out we are attempting to reword the following into a **Guideline**:

DELETED:
> **A [beatmap host](/wiki/Beatmap/Beatmap_host) cannot indicate possession in a difficulty's name.** (e.g. Beatmap Host's Insane). Conflicts caused by beatmapping multiple songs with the same metadata and [collaborative difficulties](/wiki/Beatmap/Beatmap_collaborations) are the only exceptions. [Guest difficulties](/wiki/Beatmap/Guest_difficulty), however, may indicate possession with its creators' username or nickname.

MERGED INTO THE GUIDELINE:
> **Avoid incomprehensible username combinations to indicate possession of a [collaborative difficulty](/wiki/Beatmap/Beatmap_collaborations).** If it's unclear whose usernames are combined, simplification is recommended.

> **Avoid difficulty names with descriptive elements not clearly related to a [guest difficulty](/wiki/Beatmap/Guest_difficulty) creator or a level of difficulty.** (e.g. Beatmap Creator's Tragic Love Extra)

The current wording takes the feedback given in https://osu.ppy.sh/community/forums/posts/9841035 into account


Okorin:
BTW how is it not pointing out a conflict on `en.md`

Okorin:
ah, got it, it had no overlapping wording anywhere so it is just conflict-less

wiki/Ranking_criteria/en.md
```diff
@@ -92,16 +92,19 @@ Important to understand before using:
...
+- **Difficulty names should not be misleading.**
+  - **Descriptive difficulty names should have a relation to the song.** When adding other terms such as `Tragic Love` to a difficulty name, it should relate to the song.
+  - **When combining multiple usernames into a single collaborator name for a [guest difficulty](/wiki/Beatmap/Guest_difficulty), make sure the name remains comprehensible.** For example `Mamayucha` being a [collaboration difficulty](/wiki/Beatmap/Beatmap_collaborations) between `Mani`, `Mari`, `Yuzu` and `Mocha` loses many of the collaborators' features and listing their full usernames, replacing the collaborative username with `Collab` or a descriptor like `osu! Mascots' Expert` may be clearer.
+  - **The host should not include difficulties such as `Host's Insane`, unless this is done in order to rank the same metadata twice or strongly encouraged by a naming scheme related to the song.**
+  
+  If it is unclear, a discussion should be held to come to a consensus on whether the difficulty name can be used or should be changed.
```
Walavouchey, line 107:
```suggestion
  - **The host should not include difficulties such as `Host's Insane`, unless this is done in order to rank the same metadata twice or strongly encouraged by a naming scheme related to the song.** If it is unclear, a discussion should be held to come to a consensus on whether the difficulty name can be used or should be changed.
```

that or

`- **Difficulty names should not be misleading.** If it is unclear, a discussion should be held to come to a consensus on whether the difficulty name can be used or should be changed.`

not sure which one was meant here

Walavouchey (changes requested):

wiki/Ranking_criteria/en.md
Okorin: both suggestions are wrong, they are supposed to visually be on the same level as `Difficulty names should not be misleading` as you've guessed

Okorin: to specify: the sentence was at the end and under the list because if none of the bulletpoints apply to your usecase and you're not sure, have a discussion before ranking the map was the idea, moving it up kinda makes it read worse to me

peppy: what does "rank the same metadata twice" mean? it's very weird terminology.

Okorin: @peppy https://github.com/ppy/osu-stable-issues/issues/714 basically this usecase is when people are generally fine with including themselves as a difficulty owner in the format `Host's <Difficulty Level>` 

i.e. 
- https://osu.ppy.sh/beatmapsets/474731#osu/1014240 Ranked in July 2016
- https://osu.ppy.sh/beatmapsets/488619#osu/1045500 Ranked in September 2016 had to use `Doormat's Easy` to avoid osu! thinking the difficulty is from the ranked set 

I didn't know a better way to word this while keeping it brief. Maybe I should elaborate in a footnote?

peppy:
Generally against this change because it makes the guidelines more complicated than before. I think we should be simplifying if anything.

wiki/Ranking_criteria/en.md
Walavouchey: the point is that it messes up list spacing ([image](https://github.com/user-attachments/assets/e2e450e3-9fe5-449e-84d5-6ed9851d3e11)), which is probably why lists aren't formatted this way elsewhere on the wiki. if it's to be on the same level then it should be just another bullet.

```suggestion
  - **The host should not include difficulties such as `Host's Insane`, unless this is done in order to rank the same metadata twice or strongly encouraged by a naming scheme related to the song.**
  - If it is unclear, a discussion should be held to come to a consensus on whether the difficulty name can be used or should be changed.
```

Okorin:
Okay, applied the changes from reviews i think.

> Generally against this change because it makes the guidelines more complicated than before. I think we should be simplifying if anything.

I think it doesn't make it more complicated, if anything it merges 3 separate rules you had to keep track of across the text into smaller to read chunks and groups them together. As far as I can see the existing rules merged into it are simplified in wording without changing in meaning

MChecaH:
Posting my thoughts as requested in [Internal](https://discord.com/channels/90072389919997952/1094408191665770597/1377000555511156807).

If I had to nitpick, the last bulletpoint is somewhat redundant. The point is that guidelines can be discussed as needed if there are any discrepancies. I do believe that grouping everything in its own block with examples potentially makes it clearer though. Plus that way all rules/guidelines pertaining to difficulty names are neatly bundled together.

Okorin:
OK. I've taken out the strong discouragement against using `Host's Insane`. The change of letting the host indicate possession makes little sense if we just move it to a guideline because that makes the reasons when it's allowed to break this very unclear. Any arguments we came up internally kinda went nowhere because right now this flat out isn't allowed and just allowing it would be fine

wiki/Ranking_criteria/en.md
```diff
@@ -92,16 +92,16 @@ Important to understand before using:
...
-- **Avoid incomprehensible username combinations to indicate possession of a [collaborative difficulty](/wiki/Beatmap/Beatmap_collaborations).** If it's unclear whose usernames are combined, simplification is recommended.
-- **Avoid difficulty names with descriptive elements not clearly related to a [guest difficulty](/wiki/Beatmap/Guest_difficulty) creator or a level of difficulty.** (e.g. Beatmap Creator's Tragic Love Extra)
+- **Difficulty names should not be misleading.**  
+  - **Descriptive difficulty names should have a relation to the song.** When adding other terms such as `Tragic Love` to a difficulty name, it should relate to the song.
+  - **When combining multiple usernames into a single collaborator name for a [guest difficulty](/wiki/Beatmap/Guest_difficulty), make sure the name remains comprehensible.** For example `Mamayucha` being a [collaboration difficulty](/wiki/Beatmap/Beatmap_collaborations) between `Mani`, `Mari`, `Yuzu` and `Mocha` loses many of the collaborators' features and listing their full usernames. Replacing the collaborative username with `Collab` or a descriptor like `osu! Mascots' Expert` may be clearer.
```
Walavouchey, line 104: wrong conjunction

```suggestion
  - **When combining multiple usernames into a single collaborator name for a [guest difficulty](/wiki/Beatmap/Guest_difficulty), make sure the name remains comprehensible.** For example `Mamayucha` being a [collaboration difficulty](/wiki/Beatmap/Beatmap_collaborations) between `Mani`, `Mari`, `Yuzu` and `Mocha` loses many of the collaborators' features compared to listing their full usernames. Replacing the collaborative username with `Collab` or a descriptor like `osu! Mascots' Expert` may be clearer.
```

Walavouchey (changes requested):

Walavouchey:
@peppy current state of this pr simplifies things more clearly, please take another look

peppy (approved):

Walavouchey (approved):

PR #13040, opened 2025-03-26
Update `Community Contributors`

iSlodinxOsu:
## Self-check


Hiviexd (approved):

wiki/People/Community_Contributors/pl.md
```diff
@@ -1,5 +1,6 @@
 ---
 outdated_translation: true
+outdated_since: 776e337b5ed79c751285d76678634b6f0291ce1d
```
TicClick, line 3: this is an incorrect change -- if a translation is already outdated, but there's no `outdated_since`, don't add it, because otherwise you change it from "it's outdated for we don't know how long" to an arbitrary point in time. translators looking up the diff between this and the head revision will have false expectations about what they need to update the translation with, while in reality it needs a side-by-side comparison

PR #13085, opened 2025-04-04
[ZH] Add `History of osu!/2010`

jimmy-sketch:
## Self-check

@osu-translate-zh


TicClick (approved):

PR #13111, opened 2025-04-09
Merge `oBWC` into `o!bwc`

Walavouchey:
this way round because the provided news post abbreviated the contest as "o!bwc" so i'm assuming that'll be used going forward


wiki/redirect.yaml
```diff
@@ -2157,3 +2157,10 @@ osu!_originals: Community/Bespoke_music
 
 # Renamed in #12800
 people/user_group: People/User_groups
+
+# Moved in #13111
+contests/obwc: Contests/o!bwc
+contests/obwc/1: Contests/o!bwc/1
+contests/obwc/2: Contests/o!bwc/2
+contests/obwc/3: Contests/o!bwc/3
+contests/o!bwc/2025: Contests/o!bwc/4
```
TicClick, line 2166: why, wouldn't it be better to simply rename the article instead?

TicClick (approved):

wiki/redirect.yaml
Walavouchey: that's what i did?
renamed 2025 -> 4 for consistency, then added redirect

peppy (approved):

PR #13126, opened 2025-04-12
Update nomination rules for hybrid mapsets

### LINKED ISSUE #12514: `Beatmap_ranking_procedure` needs updated to reflect new(ish) nomination requirements for multiple game modes
Page can be found [here](https://osu.ppy.sh/wiki/en/Beatmap_ranking_procedure). Title is self-explanatory. New requirements can be found [here](https://github.com/ppy/osu-web/pull/10984).

The-Last-Cookie:
From [osu-web#10984](https://github.com/ppy/osu-web/pull/10984):

>The new rules for nomination on hybrid sets are:
>
>2 nominations required for the main ruleset and 1 on every other ruleset.
>There must be one full nominator on every ruleset.
>
>The main ruleset is the one with the most difficulties
>If the number of difficulties are the same, then the one with the most diffs by the host is the main ruleset.
>If 1 and 2 do not result in a clear main ruleset, then any ruleset that satisfies the previous criteria is used.

Closes #12514 

## Self-check

### DIFF
--- a/wiki/Beatmap_ranking_procedure/de.md
+++ b/wiki/Beatmap_ranking_procedure/de.md
@@ -9,6 +9,8 @@ tags:
   - Qualifizierung
   - Nominierung
   - gerankt
+outdated_translation: true
+outdated_since: dcd2a38fb52e9427609446990b0219aa37128dc7
 ---
 
 # Ranking-Verfahren für Beatmaps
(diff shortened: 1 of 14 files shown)


The-Last-Cookie:
Superseeded by #13211

PR #13142, opened 2025-04-14
Simplify approach rate formulas.md

Esnos33:
Formulas in approach rate can be simplified by using equivalent transformations to much easier to understand form.
## Self-check

### DIFF
--- a/wiki/Beatmap/Approach_rate/en.md
+++ b/wiki/Beatmap/Approach_rate/en.md
@@ -22,23 +22,19 @@ The duration of a hit object that stays visible on the screen (without mods) ran
 See the infographic and formulas below for an overview:
 
 ```
-                                       X = hit/collect
-             p r e e m p t             ↓
-├───────────────────────┬──────────────┤
-0%       fade_in          100% opacity
+               p r e e m p t             X = hit/collect
+0%                        66%            ↓
+├──────────────────────────┬─────────────┤
+         fade_in            full opacity
 ```
 
 The hit object starts fading in at `X - preempt` with:
 
-- AR < 5: `preempt = 1200ms + 600ms * (5 - AR) / 5`
+- AR < 5: `preempt = 120ms * (15 - AR)`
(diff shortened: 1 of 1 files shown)


peppy:
I dunno, this doesn't do much for me.

Esnos33:
Instead of two tables there is only one table, formulas are easier to calculate in mind and it is clear when hitcircles fade in to 100%, I thought this is better.

TicClick:
I agree with the fade-out edit, but would like to keep initial formulas as well -- while becoming easier to calculate, they partially lose the expressiveness

Esnos33:
If you like formulas, then maybe I can simplify formulas like this?
```
1200ms + 600ms * (5 - AR) / 5 -> 1200ms + 120ms * (5 - AR)
1200ms - 750ms * (AR - 5) / 5 -> 1200ms - 150ms * (AR - 5)
```
I really don't know why you need `/ 5`

TicClick:
@peppy would that be fine by you?

wiki/Beatmap/Approach_rate/en.md
```diff
@@ -22,23 +22,19 @@ The duration of a hit object that stays visible on the screen (without mods) ran
...
 - AR = 5: `preempt = 1200ms`
-- AR > 5: `preempt = 1200ms - 750ms * (AR - 5) / 5`
+- AR > 5: `preempt = 1200ms - 150ms * (AR - 5)`
 
-The amount of time it takes for the hit object to completely fade in is also reliant on the approach rate:
-
-- AR < 5: `fade_in = 800ms + 400ms * (5 - AR) / 5`
-- AR = 5: `fade_in = 800ms`
-- AR > 5: `fade_in = 800ms - 500ms * (AR - 5) / 5`
+Hit object competely fade in at 2/3 of preempt time.
```
peppy:
```suggestion
Hit object is at final (full) opacity at 2/3 of preempt time. 
```

peppy (approved):

PR #13193, opened 2025-04-27
Add historical information about the .osr file format

Oreeeee:
## Self-check

Introduced some historical information about the .osr file format:
- Dates
- Backwards compatibility
- Online ID changes
- More information about re-used mod IDs

I'm not sure how to point to specific headers in articles (I want the History of osu! article links to point to specific months for example), so that's to be fixed.

### DIFF
--- a/wiki/Client/File_formats/osr_(file_format)/en.md
+++ b/wiki/Client/File_formats/osr_(file_format)/en.md
@@ -15,7 +15,7 @@
 
 ## Format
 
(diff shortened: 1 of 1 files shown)


Walavouchey (changes requested):
include sources please

TicClick:
@Oreeeee when (if?) this will be worked on in fugure, please reopen the pr and address the above

PR #13234, opened 2025-05-08
Change the mention of "rhythm game" to "game" in `Difficulty naming schemes`

HannahBruch:
Touhou project is listed as one of the games in this section, but is not primarily a series of rhythm games. I changed the text "rhythm games" to just "games." The translations need to be reviewed.

## Self-check


Nivalyx:
this change is fair enough but instead of guessing how the change would translate to other languages usually we just outdate them by using the `outdated_translation` tags as follows :

image

(so that the respective translators who are actually fluent in their own language can adjust the translation by their own as they see fit)

Nivalyx:
@HannahBruch feel free to ping me once you've added the `outdated_translation` tags, i'll get this merged after that

Nivalyx (approved):

PR #13320, opened 2025-06-03
`Ranking criteria - osu!`: Move Bonus score requirement to diff-specific, exclude Expert

Okorin:
- conclusion thread: https://osu.ppy.sh/community/forums/posts/9381470
- went with Wala's wording (just remove the numbers), bonus score implies it's not just a score that gets no bonus

`Spinners must be long enough for Auto to achieve 1000 bonus score.` -> `Spinners must be long enough for Auto to achieve bonus score.`

since the rule needs to apply to all diffs besides experts it went from the general part of the osu!RC to the diffspecific parts of the levels it would apply to (Easy/Normal/Hard/Insane)

### DIFF
--- a/wiki/Ranking_criteria/osu!/en.md
+++ b/wiki/Ranking_criteria/osu!/en.md
@@ -13,7 +13,6 @@ Overall rules and guidelines apply to every kind of osu! difficulty. Rhythm-rela
 #### Rules
 
 - **Hit objects must never be off-screen in 4:3 aspect ratios.** Hit objects that are even partially off-screen can create reading difficulties. Test play your beatmap to confirm this.
-- **Spinners must be long enough for Auto to achieve 1000 bonus score.** Shorter spinners do not allow adequate spin time.
 - **Each beatmap must use at least two different custom combo colours unless the default skin is forced.** The combo colours must not blend with the beatmap's background/storyboard/video in any case. This is so hit objects are always visible to the player and custom skin's combo colours do not blend with the background accidentally.
 - **All actively clicked parts of objects must have at least one audible [hitsound](/wiki/Beatmapping/Hitsound) that does not blend with the song.** Players do not receive enough feedback otherwise. Using [keysounds](/wiki/Beatmapping/Hitsound#keysound) as hitnormals without other distinct additional hitsounds, for example, is not allowed.
 - **Difficulties must convert to other game modes without breaking star rating/performance points.** In certain cases, a `.osu` file can be formatted improperly, causing converted difficulties to incorrectly display star rating and reward inaccurate performance points.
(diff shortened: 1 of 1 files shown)


AJT127:
just so we are aware, this permits spinners that are essentially forced 100s for people without pretty good hardware

Okorin:
Easy - Insane: Auto must achieve bonus score (so 50 in lazer or 1000 in stable) 
Extra / Expert: kinda anything goes, we could add that auto must have a 300 judgement on the spinner to expert i guess?

AJT127:
> Is this implying that 500 bonus score is the new requirement on all difficulties? 

my testing found that auto gets 400 on some spinners (if you're talking about stable) - if the limit was 500 that would work better though imo because a spinner where auto gets 400 is like basically impossible for a human with bad framerate to get a 300 on. 

> And that expert can just have silly spinners that autocomplete?

I don't think the wording above implies this but actually i'm completely fine with autocomplete spinners too since it's a forced 300 and there are no playability issues, but i guess that might have to be discussed idk what others think

> Easy - Insane: Auto must achieve bonus score (so 50 in lazer or 1000 in stable)

isn't this overriden by the spinner length guidelines of easy/normal/hard - for insane this seems fine tho.

Okorin:
When writing this, the intention was for bonus score to mean the `50` and `1000` that you see on screen. If that wasn't clear (the old rule just had the number 1000), we may just have it easiest if we just add the numbers back

Nyanaro:
an idea I just had for this;

I believe the most elegant way to solve any kind of confusion for what is defined as bonus score would be to just have a footnote explaining that it refers to the 1000/50 popups the player receives while spinning.

cihe13375:
Sorry for bumping an old thread, but as I run into this problem when mapping myself (and I believe in any case this rule should be reworded) so I want to add something here.

The problem is that, the spinner length needed for "auto getting bonus score" is different between stable and lazer. Such examples are pretty easy to craft, and does not need 1ms-level accuracy; one example is 620~670ms at OD8.5. So the current wording is ill-defined.

If we don't "fix" lazer spinners to make them 1:1 with stable, I think the only viable option is to use raw length instead, e.g. spinner must be at least 600ms long (maybe a function of OD).

Personally I don't have a strong opinion on the rule change itself, but once I ran into this (mapset verifier says the same):
{1967CF9E-26E7-44A7-9569-A422F0487116}
So I have to export this map to stable to check if auto can get 1000 points, which I did, but it's not very convenient.

btw, as E/N/H diffs already have guidelines on the spinner length, I feel that adding another clause about the spinner length for them is a bit weird (but it's not a huge problem).

bloeiendelgtbbt:
bump

PR #13327, opened 2025-06-06
Group updates june 25

Laurakko:
## Self-check


TicClick (approved):

Laurakko:
some areas of focus for the gmt articles are not translated in even the old versions of the articles, such as "General moderation" "Team Leadership". not sure they were ever translated, and just let into the articles as is lol

Laurakko:
> > some areas of focus for the gmt articles are not translated in even the old versions of the articles, such as "General moderation" "Team Leadership". not sure they were ever translated, and just let into the articles as is lol
> 
> i believe team leadership was translated from back when venix was the only leader, if i recall right i just copy-pasted the existing text to kobold and my areas

might be just that some articles don't have it translated because it never was, in those languages? at least some have the last commit from clayton from last year as where it was updated to "team leadership" for you and kobold, but not all languages were translated

Laurakko:
i mean, the translate-groups command was ran, but it didn't translate all of them. i still don't understand how it works lol

TicClick (approved):

PR #13326, opened 2025-06-05
Update Lazer migration page to specify that certain medals are unobtainable on Stable

ziv-vy:
## Description

Certain medals are exclusive to Lazer. This clarifies it by replacing stable's medal score from true to partial.

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #13333, opened 2025-06-06
[ID] Update `Upgrading to lazer`

Nivalyx:
To go along with the recent PR ([#13326](http://github.com/ppy/osu-wiki/pull/13326)) that was pushed earlier this morning, should be safe to merge right away I guess since the changes are very minor

- [x] ~~*(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)~~ (*shouldn't be required*)


Joehuu (approved):
There's already a `no_native_review: true` so merging.

PR #13347, opened 2025-06-08
[IT] Update `Beatmap/Category`

hhacki:
## Self-check


TicClick (approved):
a change too small to add `to_native_review: true` for, so I skipped this part

PR #13354, opened 2025-06-09
Corrected text in "sliderendcircle" and "sliderendcircleoverlay" sections

BunnyTub:
- The sliderendcircle and sliderendcircleoverlay entries say that their images override the hit circle at the beginning of the slider.
- The wiki says that both will override the start of the slider, which is incorrect before this change.

### DIFF
--- a/wiki/Skinning/osu!/en.md
+++ b/wiki/Skinning/osu!/en.md
@@ -345,7 +345,7 @@ Notes:
 
 Notes:
 
-- Overrides `hitcircle.png` for the start of the slider, if skinned.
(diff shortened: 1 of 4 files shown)


peppy (dismissed):

BunnyTub (commented):
(comment was incorrect)

BunnyTub:
@0x84f Thank you for letting me know about the translation stuff, will remember if there is a next time I try and make a commit.

peppy:
This is a case where this change REALLY should be squashed, but the translation thing mens it can't be. I'll manually fix this time.

peppy (approved):

PR #13364, opened 2025-06-13
Update `Upgrading to lazer` feature comparison

Joehuu:
Checks off song select grouping modes and editor cross-compatibility (the whole `.osz`/`.olz` is done already) for lazer.

## Self-check


Joehuu:
Bumping as song select v2 is the default now.

peppy (approved):

PR #13365, opened 2025-06-13
[ZH] fix(/wiki/Help_centre/zh.md): nothing outdated

LaoshuBaby:
English version only change plural format for a English word, but Chinese don't use such way to express plural so nothing need change. So I remove outdated tag. 

https://github.com/ppy/osu-wiki/commit/b4207575b7ba29a8456d3bd762278fae68e0131f#diff-83ac327018a86174fde156b2791d6ef7c64a8f538ff96e802d907c606c17cc44L1

## Self-check

### DIFF
--- a/wiki/Help_centre/zh.md
+++ b/wiki/Help_centre/zh.md
@@ -7,8 +7,6 @@ tags:
   - missing
   - 帮助
   - 问题
-outdated_translation: true
-outdated_since: 712bbdeb6c5c3e1c40c7d6b44cf61df76a6ab8ff
 ---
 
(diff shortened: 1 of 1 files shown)


TicClick (changes requested):
you are looking at the wrong file -- there is now a new table row dedicated to silences: https://github.com/ppy/osu-wiki/commit/b4207575b7ba29a8456d3bd762278fae68e0131f#diff-576399d98712e10329c0a3d6c5a015ffb048b2dcabc2457d0ada1e96e8044b05

PR #13395, opened 2025-06-18
Change `Beatmap Nominators/Beatmap Nominator Test` to past tense

Userlog2:
This article has been outdated for over a year: 
image

(is it possible to link to one of those BN site announcement posts?)

The page should be reformatted in line with other historical articles, and links to the modern application process should be added. Suggestions for a new format of the article are needed.
 

## Self-check

- [x] The article has been put in line with other historical articles


Userlog2:
> > (is it possible to link to one of those BN site announcement posts?)
> 
> don't think so, but i think you can link this instead http://fxtwitter.com/osu_nat/status/1773573625045529068 @Userlog2

can't see the thread without an account, can you send the tweet specifically mentioning that BN tests are gone?

PR #13402, opened 2025-06-19
Add Newbie Mania Tournament's songs to the list and also update on sugosugiii's songs that is now part of the FA catalogue

Walker37712:


Walker37712 (commented):
add outdated_translation tags

Nivalyx:
Hiya, thank you very much for your interest in keeping the article up-to-date with new tracks.

However, due to the ever-expanding nature of it, this article is not meant to be updated manually by individual PRs — instead, everything here is linked to an external spreadsheet managed by the wiki maintainers, in which from there all the new tracks are added to the page from time to time in batches via an automated script.

For now though, I've added your songs to the spreadsheet, and they should appear on the article later along with the upcoming batch. Many thanks once again!

(cc : @Walavouchey)

PR #13409, opened 2025-06-20
Update BN, GMT, NAT, and osu!Alumni listings (2025-06-20)

Laurakko:
## Self-check


UnknownOsu:
The tasks and languages of the NAT members are all in English, I don't know what might have happened, but just letting you know.

It seems that something happened with your [previous pull request](https://github.com/ppy/osu-wiki/pull/13327/files#diff-0d952ebf22f487936e531c3f97eccc2c4759235f194dcc2aacfe261da4e9a1f3) and nobody noticed

TicClick (dismissed):

TicClick:
I took a closer look and I think it's because the duties are non-standardized now. let me reintroduce these.

TicClick (dismissed):

UnknownOsu:
Languages are still the same, in case you didn't notice

TicClick:
> Languages are still the same, in case you didn't notice

yup, I missed that. will work on it separately, because the gmt page lags behind for enough time already and I'd rather get this in sooner

TicClick (approved):

PR #13416, opened 2025-06-23
Add an action for outdating translations post-merge

TicClick:
## why?

this is both to support [squash commits](https://discord.com/channels/188630481301012481/218677502141399041/1377124578970566717) [^squash-commits] and free other editors from the need to outdate the translations by hand

## what?

when a PR is merged (or _someone_ commits directly to `master`), the action will:
- run on the HEAD commit of `master`
- outdate affected translations automatically, adding `outdated_since: <hash of a merge/squash merge/standalone commit>`
- commit and push this to `master` (no, this is not recursive[^not-recursive])

before merging a PR, a maintainer can decide which files should _not_ be outdated by adding one of the following lines to the PR's description:

```shell
# Individual articles.
DO_NOT_OUTDATE: wiki/Path/To/Article/en.md

# All files inside a directory, including ALL nested articles.
DO_NOT_OUTDATE: wiki/Article
DO_NOT_OUTDATE: wiki/

# Fnmatch masks.
DO_NOT_OUTDATE: wiki/*/es.md

# Paths with shell-style brace expansion.
DO_NOT_OUTDATE: wiki/{Article,Other_article}/es.md
DO_NOT_OUTDATE: wiki/{Article,Other_article}/{es,jp}.md

# Several rules on a line.
DO_NOT_OUTDATE: wiki/Article, wiki/Other_article/es.md, wiki/*/ru.md
```

[^squash-commits]: typically, hashes in `outdated_since: <hash>` point to the first commit of a feature branch. while regular merge commits preserve these, squash commits destroy them and invalidate commit references.
[^not-recursive]: from https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow: "When you use the repository's `GITHUB_TOKEN` to perform tasks, events triggered by the `GITHUB_TOKEN`, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run"

PR #13440, opened 2025-06-29
Retry outdating the translations up to 3 times to account for new upstream changes

TicClick:
another follow-up to #13416 that should make it retryable and a bit more resilient


TicClick:
if anything the app still struggles with protection bypass -- maybe it's not in the bypass list of a `master` protection rule, or needs a rule of its own? https://github.com/ppy/osu-wiki/actions/runs/15959543896/job/45010184303

I don't have visibility into the repository's settings, but can try and experiment with my fork first if you'd prefer it that way.

peppy:
Rather than adding a loop, I believe you can ensure two actions of the same type don't run in parallel (if that's your concern).

TicClick:
no, I thought of making it do the retries by itself if it clones the repository and ends up with a stale copy of `master` repeatedly, so that you don't have to click `retry` every time (I guesstimate 3 failures in a row as unlikely, hence 3 retries)

TicClick:
oh, and I also found out how to allow the app to push. for that, you need to go to https://github.com/ppy/osu-wiki/settings/rules and create a ruleset similar to the below (not sure what your exact settings are):

```json
{
  "id": 6428309,
  "name": "protect-master",
  "target": "branch",
  "source_type": "Repository",
  "source": "TicClick/osu-wiki",
  "enforcement": "active",
  "conditions": {
    "ref_name": {
      "exclude": [],
      "include": [
        "refs/heads/master"
      ]
    }
  },
  "rules": [
    {
      "type": "non_fast_forward"
    },
    {
      "type": "pull_request",
      "parameters": {
        "required_approving_review_count": 1,
        "dismiss_stale_reviews_on_push": true,
        "require_code_owner_review": true,
        "require_last_push_approval": false,
        "required_review_thread_resolution": true,
        "automatic_copilot_code_review_enabled": false,
        "allowed_merge_methods": [
          "merge",
          "squash"
        ]
      }
    }
  ],
  "bypass_actors": [
    {
      "actor_id": 5,
      "actor_type": "RepositoryRole",
      "bypass_mode": "always"
    },
    {
      "actor_id": 1493879,
      "actor_type": "Integration",
      "bypass_mode": "always"
    }
  ]
}
```

this way, it'll still display output of the server commit hook, but the push will go through: https://github.com/TicClick/osu-wiki/actions/runs/16008358100/job/45160496392

PR #13447, opened 2025-07-01
Fix incorrect infos in `Bespoke_music`

RushFTK:
- fix incorrect notes (GF - > F for 1st custom song)
- fix incorrect videos (but GF haven't dedicated VOD yet, idk using GF showcase VOD with timestamp worked.)

## Self-check


Walavouchey (approved):
feel free to send youtube links once the songs have been uploaded on their own (would prefer not having timestamped vods or similar)

PR #13287, opened 2025-05-25
[EN,RU] Add `Community/Moscow osu! Event` index page and first iteration article

Rainbowtaves:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

Any reviews to these articles will be strongly appreciated. Also I want to summon @TicClick to this PR for checking both language files

PR #13448, opened 2025-07-01
Fix broken formatting in `Community/Moscow osu! Event`

TicClick:
#13287 

## Self-check


wiki/Community/Moscow_osu!_Event/en.md
```diff
@@ -24,8 +24,8 @@
 
 Until 2023, large-scale osu!-related events were rarely held in Russia. Usually, these were small community meetups of 10–20 people, but many were interested in more significant events.
 
-Inspired by [cavoe's osu! event](/wiki/Community/cavoe's_osu!_event), the `
-*#russian osu cup tournament* teams and the [play-osu.ru](https://play-osu.ru) communities decided in 2023 to organise their own event in Moscow, primarily dedicated to the [osu! game mode](/wiki/Game_mode/osu!). In subsequent iterations, other modes were included as well.
+Inspired by [cavoe's osu! event](/wiki/Community/cavoe's_osu!_event), the 
+`#russian` osu cup tournament teams and the [play-osu.ru](https://play-osu.ru) communities decided in 2023 to organise their own event in Moscow, primarily dedicated to the [osu! game mode](/wiki/Game_mode/osu!). In subsequent iterations, other modes were included as well.
```
Walavouchey: does this actually refer to the irc channel? shouldn't it be "osu!" if it's not referring to a specific event name?

also the paragraph shouldn't span two lines

TicClick: looking at https://osu.ppy.sh/community/forums/topics/2050180?n=1, it's a stylised name of the Russian community rather than that of the channel (I don't think I've ever seen anyone from the staff chatting there), and "osu" is taken from the title of the tournament as-is

I wrapped it in a code block because I was unsure what to made of your suggestion https://github.com/ppy/osu-wiki/pull/13287#discussion_r2174004245 (the one that added the newline) -- italic doesn't even apply to event names, so I compromised

Walavouchey: either italics or quotes work to signify that this is supposed to be interpreted as a name (though linking the thing would work better)

the new line was a mistake

```suggestion
Inspired by [cavoe's osu! event](/wiki/Community/cavoe's_osu!_event), the "#russian osu! cup" tournament teams and the [play-osu.ru](https://play-osu.ru) communities decided in 2023 to organise their own event in Moscow, primarily dedicated to the [osu! game mode](/wiki/Game_mode/osu!). In subsequent iterations, other modes were included as well.
```

TicClick: I'm confused, I don't think I've seen such a use of italics

PR #13487, opened 2025-07-12
[RU] Remove outdated and/or poorly translated tournament articles

TicClick:
IMO these aren't worth keeping at all


Walavouchey (approved):
sounds good, trusting your judgement

PR #13602, opened 2025-08-13
Update `Storyboard/Scripting/Commands` with proper examples

KuroiseTNP:
Change the 2nd example using hexadecimal colors because putting those values in will result in an error. The only way to put hexadecimal colors in is to transform them to RGB (only accepted format).

## Self-check


Nivalyx (approved):
should be good to merge

PR #13628, opened 2025-08-21
Add note to `Locus 2025` regarding Free Your Heart's disqualification

Finadoggie:
This was mentioned during the livestream. However, I believe it should be mentioned in the wiki page as well for documentation's sake.

## Self-check


wiki/Contests/Locus/2025/en.md
```diff
@@ -459,3 +459,4 @@ The results were as follows:
 
 [^win-rate]: Win rate is the ratio of total wins over losses against other entries during the Elo calculation.
 [^vote-count]: Each voter is assigned 5 randomly selected entries to rank in each category. The vote count is the amount of voters assigned to a particular entry who also submitted their vote.
+[^free-your-heart-dq]: This map was disqualified from the graphics category due to the beatmap's background being plagiarized.
```
Walavouchey: should be more accurate, since it's not like it's literally copy-pasted

```suggestion
[^free-your-heart-dq]: This map was disqualified from the graphics category due to the beatmap's background being too heavily inspired by an existing artwork.
```

PR #13647, opened 2025-08-24
replace `daily-challenge-menu.png` due to explicit chat message

Tenexxt:
The reason i want to replace this screenshot is because I was notified that one of the chat messages isn't as "wiki friendly" as it should, and while this isn't a extremely explicit message *In My Opinion*, I also don't belive a message of that kind should be here considering the nature of the wiki.

since i don't want to get accidentally banned for saying explicit sentances on github (...again), i'll just say it was sent by bhnhjbv and its near the bottom of the chat.

also i noticed the leadderboard has a new design so thats an update on that side too.

I also updated the english and polish screenshot description 
(~~not touching other languages because i don't know them and don't want to mess it up~~ nevermind Hivie did that for me

(the polish translation is still outdated, but for now this is what i'll do with it)

### DIFF
--- a/news/2025/2025-08-24-new-featured-artist-gryscl.md
+++ b/news/2025/2025-08-24-new-featured-artist-gryscl.md
@@ -1,0 +1,49 @@
+---
+layout: post
+title: "New Featured Artist: GRYSCL"
+date: 2025-08-24 16:30:00 +0000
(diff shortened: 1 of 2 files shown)


peppy (approved):
🤷

PR #13651, opened 2025-08-25
[PL] Update `daily challenge`

Tenexxt:
A simple update really, not many things we're missing

- Stated that maps are no longer only from FA's but mostly are (like in the english ver.)
- Added new main contibutors of the project


TicClick (approved):
I don't think there will be any reviewers for Polish any time soon, so  let's just get this in

PR #13652, opened 2025-08-25
Minor edits to `Tournament bans`

Userlog2:
Comma: "Generally speaking, almost all players[...]" 
An infringement of the rules is the violation itself, not the penalty for said violation (as I see it); that would be a silence, restriction, tournament ban, etc.
If there is no list of things before "etc.", "such as" works better IMO 
"other means" -> "for other reasons" because - and this is subjective - users wouldn't see the means by which a tournament ban is applied, but they can infer the reason(s), therefore "reasons" is more relevant to the general user

### DIFF
--- a/wiki/Help_centre/Tournament_bans/en.md
+++ b/wiki/Help_centre/Tournament_bans/en.md
@@ -10,7 +10,7 @@ Due to the high-stakes nature of tournament play, tournament bans are treated ve
 
 ## What causes someone to receive a temporary tournament ban? {id=why-temporary}
 
-Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only been considered a minor infringement outside of the competitive play setting.
+Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only resulted in a less severe punishment outside of the competitive play setting.
 
 As a real-world example, such behaviour would include drawing or writing hate symbols/slurs (such as swastikas) using cursor smoke in streamed matches.
 
@@ -32,7 +32,7 @@ This includes situations like:
 
 ## Can I appeal a tournament ban? {id=appeal}
 
-Temporary tournament bans as applied via standard appeal terms or for other reasons cannot be appealed.
+Temporary tournament bans as applied via standard appeal terms or other means cannot be appealed.
 
 Indefinite tournament bans may be appealed after at least **two years** (24 months) have elapsed since they were first applied. Users contesting an indefinite tournament ban will be expected to have demonstrated positive involvement in the broader osu! community, in tournaments or otherwise, and have a spotless behavioural record during this timeframe. In the event of a failed appeal, a user must wait at least **one year** (12 months) from the date of their last appeal before appealing again.
 
--- a/wiki/Help_centre/Tournament_bans/en.md
+++ b/wiki/Help_centre/Tournament_bans/en.md
@@ -4,15 +4,15 @@
 
 A tournament ban prevents a player from playing in official or officially supported tournaments. It also limits to what extent a player can staff or otherwise offer assistance in these tournaments.
 
-Tournament bans are very rarely handed out on their own, but generally speaking almost all players who return under a standard appeal for an [account restriction](/wiki/Help_centre/Account_restrictions) are subject to one of at least a year in length.
+Tournament bans are very rarely handed out on their own, but generally speaking, almost all players who return under a standard appeal for an [account restriction](/wiki/Help_centre/Account_restrictions) are subject to one of at least a year in length.
 
 Due to the high-stakes nature of tournament play, tournament bans are treated very seriously and are not subject to the kind of leniency an account restriction might ordinarily receive.
 
 ## What causes someone to receive a temporary tournament ban? {id=why-temporary}
 
-Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only resulted in an infringement outside of the competitive play setting.
+Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only been considered a minor infringement outside of the competitive play setting.
 
-As a real-world example, such behaviour would include drawing or writing hate symbols/slurs (swastika, etc.) using cursor smoke in streamed matches.
+As a real-world example, such behaviour would include drawing or writing hate symbols/slurs (such as swastikas) using cursor smoke in streamed matches.
 
 All users returning to the game after a restriction under the standard appeal terms are also subject to a [minimum temporary tournament ban of 1 year](/wiki/Help_centre/Account_restrictions#reasons), which may be lengthened at the discretion of the [account support team](/wiki/People/Account_support_team).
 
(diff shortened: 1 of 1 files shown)


wiki/Help_centre/Tournament_bans/en.md
```diff
@@ -4,15 +4,15 @@
 
 A tournament ban prevents a player from playing in official or officially supported tournaments. It also limits to what extent a player can staff or otherwise offer assistance in these tournaments.
 
-Tournament bans are very rarely handed out on their own, but generally speaking almost all players who return under a standard appeal for an [account restriction](/wiki/Help_centre/Account_restrictions) are subject to one of at least a year in length.
+Tournament bans are very rarely handed out on their own, but generally speaking, almost all players who return under a standard appeal for an [account restriction](/wiki/Help_centre/Account_restrictions) are subject to one of at least a year in length.
```
MChecaH, line 7: ~~Incorrect usage of the new comma. The previous one was fine already joining two independent clauses together.~~

~~You would only use this comma sandwich to add a clause in between, sort of as you'd do with em dashes.~~

wiki/Help_centre/Tournament_bans/en.md
```diff
@@ -4,15 +4,15 @@
...
 
 Due to the high-stakes nature of tournament play, tournament bans are treated very seriously and are not subject to the kind of leniency an account restriction might ordinarily receive.
 
 ## What causes someone to receive a temporary tournament ban? {id=why-temporary}
 
-Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only resulted in an infringement outside of the competitive play setting.
+Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only been considered a minor infringement outside of the competitive play setting.
```
MChecaH: Sentence is different. The new wording makes it seem like the infringement would only have been "considered", whereas the original states that the rule would've been broken outside of the competitive setting. Minor change, but:

1- Original is fine as is.
2- Changes the sentence's meaning by implying the rule break would only have been considered rather than acted on (as it should, it's a *severe* violation of the rules, and thus would've been dealt with promptly)

wiki/Help_centre/Tournament_bans/en.md
```diff
@@ -24,15 +24,15 @@ This includes situations like:
 
 - As above, any use of third-party cheating tools or utilities to gain an advantage
 - Making use of multiple accounts to enter a tournament
-- Attempting to evade a silence or infringement by using another account
+- Attempting to evade a silence or other penalty by using another account
```
MChecaH, line 27: I don't get the change. Yes, even if semantically it is making reference to the violation, that is precisely the point. Evading the violation itself is an issue, just as much as evading a previous punishment is. Do not feel like it needs to be changed.

wiki/Help_centre/Tournament_bans/en.md
```diff
@@ -24,15 +24,15 @@ This includes situations like:
...
 - Account sharing within/during an officially supported tournament
 - Conspiring to obtain or otherwise using private information to gain or confer an advantage (such as accessing mappool information before other teams, etc.)
 - Abusing positions of trust to gain or confer an advantage (manipulating schedules, unfairly changing team compositions, or disqualifying participants without cause or reasoning)
 - Organising targeted abuse campaigns towards participants or organisers well outside the realm of criticism
 
 ## Can I appeal a tournament ban? {id=appeal}
 
-Temporary tournament bans as applied via standard appeal terms or other means cannot be appealed.
+Temporary tournament bans as applied via standard appeal terms or for other reasons cannot be appealed.
```
MChecaH: I do not get the change, even as explained in the PR's description. Original wording looks, reads, and sounds just fine — or even better. `...via standard terms or (via) other means...` the second `via` is implicit. Moreover, if a user is Tournament Banned, they do not need to know the means. Those are transparent to them. Only the punishment concerns them.

MChecaH (changes requested):
Review of the suggested changes, and whether the original meaning is being changed.

wiki/Help_centre/Tournament_bans/en.md
MChecaH: I have been told I am Wrong. Will eat my own words. Introductory adverbs (or sentences compounded by one) should have the extra comma. This is fine.

wiki/Help_centre/Tournament_bans/en.md
Walavouchey: op is right: infringement/violation is not the same penalty/consequence, and "evading violations" doesn't semantically make sense (unless you mean avoiding making violations, which of course wouldn't be punished)

wiki/Help_centre/Tournament_bans/en.md
Walavouchey: the original wording isn't that good since "rule violations result in infringements" doesn't make sense. the infringement would be the rule violation itself

if the point is that severe rule violations cause tournament bans even if they'd only result in silences (or otherwise less severe consequences) elsewhere, then the new version is still a little off since it's saying a "severe violation" is "considered minor" which is a contradiction, so you'd word it like this

```suggestion
Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only resulted in a less severe punishment outside of the competitive play setting.
```

ofc it's ambiguous whether that's actually what was intended - just inferring based on context

in any case the premise is that a rule *was* broken, since we're talking about rule *violations* in the first place. it can then be considered minor or severe, but that's not saying anything about whether it'd be acted upon

wiki/Help_centre/Tournament_bans/en.md
Walavouchey: i agree that there's nothing wrong with the original wording

Walavouchey (approved):

wiki/Help_centre/Tournament_bans/en.md
```diff
@@ -4,15 +4,15 @@
...
+Tournament bans are very rarely handed out on their own, but generally speaking, almost all players who return under a standard appeal for an [account restriction](/wiki/Help_centre/Account_restrictions) are subject to one of at least a year in length.
 
 Due to the high-stakes nature of tournament play, tournament bans are treated very seriously and are not subject to the kind of leniency an account restriction might ordinarily receive.
 
 ## What causes someone to receive a temporary tournament ban? {id=why-temporary}
 
-Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only resulted in an infringement outside of the competitive play setting.
+Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have typically only resulted in a less severe punishment outside of the competitive play setting.
```
hburn7:
```suggestion
Any severe violation of the [community rules](/wiki/Rules) that takes place during an [officially supported tournament](/wiki/Tournaments/Official_support) is likely to cause a tournament ban, even if it would have resulted in a less-severe punishment outside of the competitive play setting.
```

Walavouchey: ok but no hyphen

hburn7: I hate my language...

Userlog2: fixed

TicClick (approved):
@peppy your call

TicClick:
alright looks like this is getting discarded in favour of https://github.com/ppy/osu-wiki/pull/13719

Issue #13113, opened 2025-04-10
"Higher snaps"


iamnotcoding:
In the file _wiki/Beatmap/Pattern/osu!mania/Stream/en.md_
`Streams with notes appearing in intervals faster than one beat are typically referred to as bursts. They commonly use snaps higher than 1/4.`
Here **higher than 1/4** means more frequent intervals than 1/4 such as 1/8, 1/16...
`Jumpstreams are streams that involve jumps. Usually, the stream itself is on 1/4 snaps while the jumps land on higher snaps.`
However in this sentence **on higher snaps** means less frequent intervals such as 1/2 1/1...
Are those two sentences actually confusing or am I just bad at english?

peppy:
You're correct that "higher" is being used in two different ways here. Maybe changing it to use "faster" or "slower" is more appropriate?

PR #13656, opened 2025-08-27
Fix ambiguous phrasing of "higher snapping"

Userlog2:
Resolves issue #13113 - but I am not a mania player/mapper, so I'm not entirely sure.

## Self-check

- [x] The changes are checked by someone familiar with osu!mania


MChecaH (approved):
Changes are fine. They fix the current ambiguity.

peppy (approved):

PR #13693, opened 2025-09-06
[DE] Update `Rules/Content usage permissions`

The-Last-Cookie:
## Self-check


Userlog2:
The dates need an update - the last update was on September 1. See also #13672

The-Last-Cookie:
> The dates need an update - the last update was on September 1. See also #13672

Thanks 👍

PR #11130, opened 2024-03-19
Update `Completionists`

iSlodinxOsu:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [ ] Apply the badge for the completionist

PR #13695, opened 2025-09-06
Archive osu! Completionists and separate completionist badge from completionist accomplishment

ziv-vy:
The badge has not been awarded to any users since 2023 despite multiple users meeting the criteria. Not even users like osu420 for which malicious intentions were not suspected by anyone have failed to get the badge on osu!std. Archiving the badge is best to avoid scenarios where users complete a ruleset and are told at the very end that they don't deserve it.

## Self-check

### DIFF
--- a/wiki/People/Completionists/en.md
+++ b/wiki/People/Completionists/en.md
@@ -6,16 +6,25 @@ tags:
 
 # Completionists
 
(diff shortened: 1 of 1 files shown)


Tenexxt:
As said by Hivie in pull #11130 curently the list isn't updated because they want to automate the reward process instead of always pulling and entering data manualy.

I'm on mentioning that the list is incomplete and thats getting worked on, but saying that the badge is discontinued is a huge overstatement afaik.

-# Small disclaimer - I didn't look further for more messages with an anwser for the question of the badge (or even acknowledging completionists) being canceled as IMO this one is satisfactory.

PR #13709, opened 2025-09-08
Add Andrea re-addition date to `Staff log 2012`

pishifat:
previous log included Andrea's removal from BAT in 2012, plus another removal in 2013. this is closest source to log the re-addition in-between (ss shows red name, "you're back!", and date): image


Andrea1992wr:
Yeah I sent to pishi after verifying, I checked through my archived puush files and I have more from a similar date I could also send if needed.

PR #13717, opened 2025-09-15
Use precise formula in `Beatmap/Circle_size`

secretbnuuy:
`Beatmap/Circle_size` says circle radius equals `54.4 - 4.48 * CS` osu pixels. at CS=5, that would equal `32`, but the correct value (in both stable and lazer) is actually `32.01312`. both clients also display the (rounded) correct value when hovering over difficulty numbers.

the correct formula is: `(54.4 - 4.48 * CS) * 1.00041`. ([a comment in lazer's source code](https://github.com/ppy/osu/blob/master/osu.Game/Rulesets/Objects/Legacy/LegacyRulesetExtensions.cs#L48) explains why)

alternatively, `* 1.00041` could be said in a footnote.

rationale:
i don't think the wiki should say "Circle size is derived through the following formula", when the following formula is actually an approximation, because that's misleading.

## Self-check

### DIFF
--- a/wiki/Beatmap/Circle_size/en.md
+++ b/wiki/Beatmap/Circle_size/en.md
@@ -15,7 +15,7 @@ tags:
 
 In [osu!](/wiki/Game_mode/osu!), circle size changes the size of [hit circles](/wiki/Gameplay/Hit_object/Hit_circle) and [sliders](/wiki/Gameplay/Hit_object/Slider), with higher values creating smaller hit objects. [Spinners](/wiki/Gameplay/Hit_object/Spinner) are unaffected by circle size. Circle size is derived through the following formula:
 
-`r = 54.4 - 4.48 * CS`<!-- multiplied by 1.00041 in the end to account for some bug in old replays -->
+`r = (54.4 - 4.48 * CS) * 1.00041`<!-- multiplied by 1.00041 in the end to account for some bug in old replays -->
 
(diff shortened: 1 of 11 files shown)


wiki/Beatmap/Circle_size/en.md
```diff
@@ -15,7 +15,7 @@ tags:
 
 In [osu!](/wiki/Game_mode/osu!), circle size changes the size of [hit circles](/wiki/Gameplay/Hit_object/Hit_circle) and [sliders](/wiki/Gameplay/Hit_object/Slider), with higher values creating smaller hit objects. [Spinners](/wiki/Gameplay/Hit_object/Spinner) are unaffected by circle size. Circle size is derived through the following formula:
 
-`r = 54.4 - 4.48 * CS`<!-- multiplied by 1.00041 in the end to account for some bug in old replays -->
+`r = (54.4 - 4.48 * CS) * 1.00041`<!-- multiplied by 1.00041 in the end to account for some bug in old replays -->
```
MChecaH, line 18: Before merging though, I think it'd be worth explaining where this ratio comes from in plain text rather than a hidden comment. Believe the ratio should be explained as to not let users wondering "what is this for". Especially when we're already commenting all the other variables in the formula.

peppy:
I agree that this should not be a hidden comment. Please add the explanation to the actual wiki pages.

Hiviexd:
Reopened in #14084

PR #13732, opened 2025-09-18
Add Azu's Tourneys to Tournaments list

agatemosu:
## Self-check

- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review) (i don't know, it should be reviewed? or if i should only edit en)


TicClick:
> i don't know, it should be reviewed? or if i should only edit en

there's nothing language-specific save for different date formats, which appear to be correct

PR #13752, opened 2025-09-24
`Ranking criteria - Metadata` - Allow alternate casing for markers that go against Ranked Consistency 

Okorin:
Pushes conclusion from [this thread](https://osu.ppy.sh/community/forums/topics/2040439?n=1) ([conclusion post](https://osu.ppy.sh/community/forums/posts/10017972))

This was kinda implicitly fine right now but gray-area so it just clarifies what's happening i think


Okorin:
no, the purpose is to avoid misunderstanding that you're allowed to do this, because ranked consistency would imply if someone in the past used `ARTIST - TITLE (TV Size)`, you'd be beholden to the same and cannot do `ARTIST - TITLE (TV SIZE)`, that the other sections of this article allow people to do this is in different contexts so explicitly allowing it again was the point here, removing it from the other sections of this rule page makes no sense to me

MChecaH (approved):

peppy (approved):

Issue #13754, opened 2025-09-24
Articles for "Miscelanneous official projects"


Nivalyx:
Related to #13751 that I was working on earlier.

As of now, quite a bunch of miscellaneous official "side projects" are still yet to be documented properly in the wiki, namely:

- [x] upppy (#13751)
- [ ] rajio (i have this one ready and will pr tomorrow most likely)
- [ ] ptransl
- [ ] puush

(*there's probably more, but these four are the ones that I can think of*)

The idea is to ultimately have these projects documented and grouped in one single folder named "Miscelanneous official projects". As discussed earlier in #osu-wiki I do plan to have all of these covered soon, but the research for this takes time and so most likely the PRs will be coming one by one individually.

Nivalyx:
yeah now that i think of it something like just "miscellaneous projects" would probably work better ;~;

Nivalyx:
oh yeah i suppose the pages can go there, but that index page looks like it needs some cleaning up lol

i'll see what i can do tho, for now i'll just focus on writing the missing pages i guess (moving things around is the easy part anyways ;~;)

Nivalyx:
Closing this as everything's been merged 🫡

PR #13762, opened 2025-09-25
Move `upppy` to `Community/Projects`

Nivalyx:
As discussed here → http://github.com/ppy/osu-wiki/issues/13754#issuecomment-3332973965

Should be safe to be merged right away as it only involves moving files from one location to another.

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check


TicClick (changes requested):
you still need to add an entry to the redirect.yaml file

PR #13780, opened 2025-09-27
Add translation for gameplay diff & remove external editing note from lazer storage desc

und3rlife:
## Self-check


und3rlife (commented):
There is no reimport neccesary, you can just mount it externally.

PR #13781, opened 2025-09-27
Add lazer notelock & Ru translations on lazer gameplay diff

und3rlife:
## Self-check


und3rlife:
previous PR was closed for some reasons, there's new with suggestions included from previous PR

wiki/Gameplay/Judgement/Notelock/en.md
```diff
@@ -30,3 +30,11 @@ Since objects in regular osu! beatmaps are meant to be cleared in chronological
 ## Prevention
 
 From a mapper's perspective, notelock could be prevented by carefully choosing the OD value of a [difficulty](/wiki/Beatmap/Difficulty) according to its object density and BPM. On beatmaps with BPM of 200 and higher, OD 5 and higher is recommended. For a more detailed breakdown, refer to the "[Avoiding notelock at high BPM](https://osu.ppy.sh/community/forums/topics/334458)" guide.
+
+## Notelock in osu!(lazer)
+
+In comparison to osu!(stable), notelock was made more lenient in osu!(lazer) because the first hit object would stop locking the second hit object once the first hit object's time window reached an offset of 0 ms.
```
peppy: This doesn't read great in english. Needs a complete rewrite.

frag1le384: can you give any suggestions how can I make this more readable?

MChecaH: See if this sounds better. Please let me know if this is the actual behaviour of the new notelock, or if there still is part of the timing window right after the object is clicked late that is taken into account for it. Additionally, I'm not too sure about the `once the first hit object reaches __an offset of 0 ms__...` part. Could be reworded to something clearer.

```suggestion
When compared to osu!(stable), notelock was made more lenient in osu!(lazer). This was mitigated by making the timing window significantly more forgiving: once the first hit object reaches an offset of 0 ms, the next object is no longer locked and can be clicked immediately. In short, late hits from a previously missed object will no longer lock the note right afterwards.
```

Added a small summary at the end to keep it briefer. The .gif below should drive the point home.

und3rlife:
received your changes, I'll finish the Russian translation a bit later and we can merge it i guess

wiki/Gameplay/Judgement/Notelock/en.md
frag1le384: Screenshot_2025-09-29-13-44-24-42_40deb401b9ffe8e1df2f1cc5ba480b12
are you sure about it's "late", not "early"?

frag1le384: > I'm not too sure about the `once the first hit object reaches __an offset of 0 ms__...` part. Could be reworded to something clearer.

I tested this with a two-note map, ar 0, od 0, and halftime at 0.5x, so I was sure that after closing the first note AR, the second one was immediately allowed to hit them.

MChecaH: Yeah, "late" is correct here. It is making reference to late hits of the first missed note, and that those late hits no longer overlap with the timing window of the second note. Your experiment is not incorrect.

frag1le384: > Yeah, "late" is correct here. It is making reference to late hits of the first missed note, and that those late hits no longer overlap with the timing window of the second note. Your experiment is not incorrect.

didn't quite understood this, can you attach an example please?

MChecaH: It is not that complicated. [Here are two notes, one after the other](https://mch.s-ul.eu/ymDlCQmq).

Clicking the first object now would be considered *early*. Clicking the second object would also be considered *early*. Moreover, the note would be locked both in stable and lazer.

Then, [we click the object as it reaches it "clicked" state](https://mch.s-ul.eu/QuV5ObZn).

From thereon, any attempt to hit the first object would be considered *late*, while the second object is still considered *early*. This is where behaviour differs between clients. In Stable, you would notelock the second object for as long as the latter half timing window of the first note overlaps with the first half of the second object. That means the *late* part of the window of the first object overlaps with the *early* timing window of the second.

Asked peppy for a real quick review before. The wording should be fine. Hope this explains what that "late" is making reference to.

und3rlife:
closing this pr to reopen a new separate

PR #13804, opened 2025-10-07
Update "Gameplay differences in osu!(lazer)" article with recent changes

bdach:
Noticed that this article had some outdated or not-quite-correct claims in passing.

## Self-check


wiki/Client/Release_stream/Lazer/Gameplay_differences_in_osu!(lazer)/en.md
```diff
@@ -33,7 +33,7 @@ Meanwhile, osu!(lazer) has these accuracy cutoffs now:
 
 | Grade | osu! / osu!taiko | osu!catch | osu!mania |
 | :-: | :-- | :-- | :-- |
-| SS | 100% | 100% | 100% |
+| SS | 100% | 100% | all judgements GREAT or PERFECT |
```
Walavouchey: do PERFECT judgements contribute to accuracy anyway?

if not, "100% (GREAT or PERFECT)" would be nicer since the table lists accuracies otherwise

bdach: PERFECT contributes 305 to accuracy portion, GREAT contributes 300. So all GREATs would be like 98.36%. But you can't put 98.36% as the threshold, because you can get 98.36% by getting PERFECTs and some GOODs, which could also be 98.36% or above, but not give SS.

I consider what I wrote to be the most accurate description of how mania SS works in lazer right now.

MChecaH: As a mania player, this should be understandable enough. Would be unreliable to use accuracy, and I can't come up with any better way to describe it.

We could always link to the [osu!mania judgement system](https://osu.ppy.sh/wiki/en/Gameplay/Judgement/osu%21mania) article just in case a player doesn't understand what the word "judgement" even is.

```suggestion
| SS | 100% | 100% | all [judgements](/wiki/Gameplay/Judgement/osu!mania) GREAT or PERFECT |
```

bdach: Have applied the above

Walavouchey (approved):

PR #13836, opened 2025-10-16
Add Dropinx's title to `Users with unique titles`

minusQuantumNeko:
## Self-check


MChecaH (changes requested):
Translations' status must remain unchanged. Since the 6th of June, a [workflow was added by TicClick](https://github.com/ppy/osu-wiki/pull/13416) that would allow files to be automatically outdated once a Pull Request is merged.

`es.md` and `zh.md` need to be reverted so the Action can do its thing.

MChecaH (approved):

PR #13845, opened 2025-10-17
[DE] Remove redundant word on `Main page`

The-Last-Cookie:
Also fits what is shown as the article title

## Self-check


The-Last-Cookie:
Above mentioned change does not conflict with the change here.

The-Last-Cookie:
@MChecaH This can be merged.

PR #13863, opened 2025-10-22
[DE] Make some phrasing sound more natural

The-Last-Cookie:
You can use both `die 2006 veröffentlicht` and `die im Jahr 2006 veröffentlicht`, but not `die in 2006 veröffentlicht`. It sounds very weird. Probably a lazy translation resulting in this left-over.

## Self-check


MChecaH (approved):
Small enough not to need a second native review. Changes look sound.

PR #13876, opened 2025-10-25
[RU] Update `Ranking criteria` and add `Metadata`

NikoN1nja:
## Self-check
- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #13907, opened 2025-11-02
[RU] Cleanup `Metadata`

NikoN1nja:
This PR follows #13876 . While rewatching the article on the wiki itself it turned out I made a lot of mistakes during translation. Found 22 of them, but there's probably a bit more.

Edit: 79. I think that's gotta be all of them.

## Self-check


MChecaH (approved):
Approved, still no native review.

PR #13921, opened 2025-11-06
Update `How to resolve BSS issues`

Nivalyx:
found this article when i was working on [#13916](http://github.com/ppy/osu-wiki/pull/13916) earlier, thought i'll update this since it could use some updating 🕊️ 


wiki/Guides/BSS_issues/en.md
```diff
@@ -4,35 +4,39 @@ tags:
   - beatmap submission system issues
 ---
 
+::: Infobox
+![Screenshot of the warning message](img/errormessage.png "The warning message as shown in the editor")
```
MChecaH, line 10: The info box looks *really* wacky when being at the very top of the page. See the [preview](https://osu1.roanh.dev/wiki/en/Guides/BSS_issues). Would suggest having it right after `How to resolve BSS issues`, but not before.

wiki/Guides/BSS_issues/en.md
```diff
@@ -4,35 +4,39 @@ tags:
...
+:::
+
 # How to resolve BSS issues
 
-If the **[Beatmap Submission System](/wiki/Beatmapping/Beatmap_submission)** (***BSS***) won't let you submit your own beatmap because it says that you do not own the map then do the following:
+When submitting a beatmap, the **[Beatmap Submission System](/wiki/Beatmapping/Beatmap_submission)** (***BSS***) may occasionally prevent you from uploading your beatmap by warning that `This beatmap you're trying to submit isn't yours!`. This is  often caused by an incorrectly set parameter in the [.osu file](/wiki/Client/File_formats/osu_(file_format)) or a bugged submission cache, and can be resolved by following these steps:
```
MChecaH, line 13: Bit confused if this is an accurate statement given the name of the guide. There are multiple ways the upload can fail. The warning `This beatmap you're trying to submit isn't yours!` is but one of them. Therefore, having the guide be called `How to resolve BSS issues` in plural would then not be precise.

Considering this issue is quite common, we can have a section for it, which is what it is being done right below (although it's mixing different troubleshooting steps. May want to split those). Would just make it a little bit more clear, however. Having it in the intro would be misleading.

Or we can change the guide's title and focus on this one specific issue, though imo that won't make the guide scalable for other common issues to come.

wiki/Guides/BSS_issues/en.md
```diff
@@ -4,35 +4,39 @@ tags:
...
-6. Set the `BeatmapSetID` to `-1`.
-7. Make sure the beatmap folder name does not start with a string of numbers.
-   - E.g. `1000 - Song Name` should be `Song Name`.
-8. Delete the file(s) in the `SubmissionCache` folder. This is located in the osu! installation folder.
+4. Make sure that your username is already written correctly in the `Creator` field. If it isn't, update this field with your current username.
+5. Set the `BeatmapID` field to `0`.
+6. Set the `BeatmapSetID` field to `-1`.
+7. Make sure that the beatmap's folder name does not begin with a string of numbers. If it does, rename this folder accordingly.
+   - For example, `1000 - Song Name` should be renamed into e.g. `ABCDE - Song Name`.
+8. Delete all the files in the `SubmissionCache` folder. This folder can be found in your osu! installation directory.
```
MChecaH: We could make this more clear by adding the relative path from the osu! folder. Additionally, we should add a tip saying that the folder where `SubmissionCache` can be found is hidden by default. Users need to be told to show their hidden files if they want to see the `Data` directory inside the osu! folder.
```suggestion
8. Delete all the files in the `SubmissionCache` folder. This folder can be found in your osu! installation directory.
   - This folder will be hidden inside the `Data` directory. The following [article](https://support.microsoft.com/en-au/windows/file-explorer-in-windows-ef370130-1cca-9dc5-e0df-2f7416fe1cb1) explains how to open this file from the file explorer.
```

wiki/Guides/BSS_issues/en.md
```diff
@@ -4,35 +4,39 @@ tags:
...
 9. Open osu! again and try to upload your beatmap.
 
-If you are still having issues, follow these steps:
+You should be able to submit your beatmap by now. If you are still having issues by then, follow these steps:
```
MChecaH:
```suggestion
You should be able to submit your beatmap by now. However, if you are still having issues, try following these steps:
```

wiki/Guides/BSS_issues/en.md
```diff
@@ -4,35 +4,39 @@ tags:
...
 
-1. Export the beatmap using the beatmap editor.
-   - `File` then `Export Package`.
-2. Exit the beatmap editor and then delete the beatmap in osu!.
-3. Exit osu!
-4. Go to the exported beatmap and change the extension from `.osz` to `.zip` (you will need file extensions to be visible from the windows folder options to do this).
+1. Export the beatmap from inside the beatmap editor by going to `File` and then `Export Package`.
+2. Exit the beatmap editor and delete the beatmap in osu!.
+3. Exit osu!.
+4. Go to the exported beatmap and change its extension from `.osz` to `.zip`. You will need file extensions to be visible from the Windows folder options in order to do this.
```
MChecaH: *Technically* not needed. The file can directly be opened via a user's zip manager of choice (7Z, Winrar, etc...). Not sure if it's worth mentioning, but just as a heads-up that this may be a faster option.

MChecaH (commented):
All in all the guide is missing some stuff, plus it only goes over osu!(stable)'s BSS. Would be nice to eventually flesh this out, should the opportunity arise. For the time being though, the updates won't hurt.

MChecaH:
Those comments were more so food for thought, not blocking or scope creeping the current PR.

As to how I'd go around it, the path of least resistance would be changing the article to `How to resolve _common_ BSS issues` then list the most common issues (in this case the name *is* a very common issue, especially due to how name changes work) and an accompanying solution. That way if we want to add other cases we can add them later, plus it would make the article's guide a bit less misleading as it is not supposed to be the end-all be-all of BSS troubleshooting.

PR #13936, opened 2025-11-09
Update `OWC 2025` with Group stage results, RO16 schedules

LeoFLT:
Adds GS results, RO16 schedules and mappool


wiki/Tournaments/OWC/2025/en.md
```diff
@@ -210,7 +247,9 @@ The complete sign-up list can be found [here](https://gist.github.com/LeoFLT/a66
 
 ## Match results
 
-### Group stage - Week 2
+### Group stage
```
Walavouchey, line 250: table order is now scrambled in this combined section (read dates from top to bottom)

LeoFLT: nice catch, fixed in 7d9aa65c63db8d32fa31ac83efc0531849eac9f8

PR #13955, opened 2025-11-12
[EN/ES] Add new filter to `Beatmap search`

wetwayne:
## Self-check

Add the 'ln' and 'lns' filters present in osu! lazer and osu! web to the wiki.

### DIFF
--- a/wiki/Beatmap_search/es.md
+++ b/wiki/Beatmap_search/es.md
@@ -10,15 +10,13 @@ tags:
   - filtros de beatmaps
   - búsqueda del cliente
   - búsqueda del sitio web
-outdated_translation: true
-outdated_since: 9938f5a037e7e2f9fb9270624edc4ff648deb0d3
 ---
 
 # Búsqueda de beatmaps
 
 *Véase también: [Beatmap § Descargar beatmaps](/wiki/Beatmap#descargar-beatmaps)*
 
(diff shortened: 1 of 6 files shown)


wiki/Beatmap_search/en.md
```diff
@@ -71,6 +71,7 @@ Beatmap metadata fields can be compared against specific values for more granula
 | `circles` | Number of circles in the beatmap |
 | `sliders` | Number of sliders in the beatmap |
 | `key`, `keys` | Number of keys (osu!mania and converted beatmaps only) |
+| `ln`, `lns` | Percent of notes that are long notes (osu!mania and converted beatmaps only) |
```
MChecaH: What's the source for this? As far as I've been able to test between lazer and the web client (both in game and native web beatmap searches) this filter did not work. Only searches done in the actual lazer client (not even stable) could be filtered by their long note percentage.

MChecaH (commented):
Needs to be validated

wiki/Beatmap_search/en.md
wetwayne: This is a filter that was added to Lazer relatively recently and is not present in stable. I thought it was present in the website too but couldn't get it to work a moment ago, I can remove the sections mentioning that it's on the website.
Usage is like `ln>50` or `lns>50` for charts with greater than 50 percent long notes, also works with the other operators. I tested this in the search tool used for searching already downloaded beatmaps, not in the "osu! direct" section used for downloading beatmaps.
The code in lazer that handles this is here:
https://github.com/ppy/osu/blob/cb9d9734d692e40750c597bc5d939dabbb70a39a/osu.Game.Rulesets.Mania/ManiaFilterCriteria.cs#L93

MChecaH (approved):

PR #14066, opened 2025-12-06
[FR] Update `People/Beatmap Spotlight Curators`

pingusurmars:
Updated the list of players in the BSC
Just to be sure, I didn't remove the outdated_translation: true & outdated_since since I wasn't sure to do so or not.

## Self-check


TicClick:
>  Just to be sure, I didn't remove the outdated_translation: true & outdated_since since I wasn't sure to do so or not.

if you translate all changes that happened to the original article, definitely do this (I have already done that for you)

TicClick (approved):

PR #14135, opened 2026-01-01
Update `Reporting abuse` and explain how to admit own misconduct

Userlog2:
New section is dependent on how peppy feels about it. Otherwise, I fixed some grammar, added IDs, added a notice about osu!(lazer), and replaced the link to Venix with a link to the GMT group page.

- [x] Wait for #14117 to be merged (which will fix the CI failure)
- ~Explain what would be considered abuse, inappropriate advances, harassment, etc.~ peppy doesn't want that

## Self-check


msmcculley:
In agreeance with Randome here, linking to just one gmt can be troublesome since they arent expected to be online 24/7 and not even be regularly active enough to respond to something that can be critical for the user. Linking to the group page can at least tell the user which gmt are active / offline.

MChecaH:
Not much to add other than what has been said already. Linking to one user is just going to cause more problems than they're worth. Ideally, linking to the Groups page or People/GMT wiki article are far better options so users can browse all available GMTs.

Also, ever since [the incident](https://discord.com/channels/90072389919997952/1388494055364890664/1388494065607381024) we have no formal leadership. No need to maintain the "de-facto head of the GMT" anymore.

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -12,35 +12,37 @@ The abuse email address is only seen by [peppy](https://osu.ppy.sh/users/2) (the
...
 
-## Someone's made inappropriate sexual advances or remarks towards me!
+## Someone's made inappropriate sexual advances or remarks towards me! {id=sexual-advances}
 
 Scroll through the sub-headings below to find out what to do, depending on where this all happened.
 
-### In the in-game chat
+### In the in-game chat {id=chat}
+
+*Caution: The `!report` command **does not work** in osu!(lazer), and such reports will be visible to everyone! Use [the report menu](/wiki/Reporting_bad_behaviour#lazer-reports) instead.*
```
MChecaH: Big nitpick, but this message is inconsistent with what is being suggested [in the other PR](https://github.com/ppy/osu-wiki/pull/14117).

```suggestion
*Caution: Reporting a user via the `!report` command **does not work** in osu!(lazer), and such messages will be visible to everyone! Instead, [use the report menu](/wiki/Reporting_bad_behaviour#lazer-reports).*
```

Or the other way around. Both sound fine, even if I do prefer the phrasing already reworded by Wala.

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -60,19 +62,27 @@ Please be aware that the person will see you have blocked them if they check you
...
 
-## A member of one of the volunteer teams (NAT, GMT) is approaching/talking/acting towards me inappropriately!
+## A member of one of the volunteer teams (NAT, GMT) is approaching/talking/acting towards me inappropriately! {id=volunteer}
+
+Send an email to [abuse@ppy.sh](mailto:abuse@ppy.sh) containing as much information as possible **immediately**. This includes screenshots, chat logs, and basically anything else you can attach to an email.
 
-Send an email to [abuse@ppy.sh](mailto:abuse@ppy.sh) containing as much information as possible **immediately**. This includes screenshots, chat logs, basically anything you can attach to an email.
+If you can't find the messages for any reason, but remember at least roughly when they were sent and are sure they were sent in osu!'s chat, include this information instead — the abuse team has full access to public and private chat logs.
```
MChecaH: `...the abuse team has full access to public and __private__ chat logs`

Want to keep this open so @peppy gives it a look. Do we want/need this information to be public? Think this is a given by now, but it could be a cause for concern for the more ""privacy-savvy"" users. Would at the very least mention how this information is only accessible by very, very, very, very few individuals.

I'm probably being overly cautious, so this is your call.

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -60,19 +62,27 @@ Please be aware that the person will see you have blocked them if they check you
...
+## Someone I know is being abused by a member of the volunteer teams (NAT, GMT) but is too afraid to speak up! {id=friend}
 
 Send an email to [abuse@ppy.sh](mailto:abuse@ppy.sh) containing everything you know with as much evidence as you can get.
 
-If you don't have any evidence but only stories or rumours that you believe are substantial enough to act on, please send us an email anyway. We take complaints like this very seriously and will look deeply into every report.
+If you don't have any evidence, but only stories or rumours that you believe are substantial enough to act on, please send us an email anyway. We take complaints like this very seriously and will look deeply into every report.
+
+## I think I have acted inappropriately towards someone and want to admit it {id=own-advances}
+
+Send an email to [abuse@ppy.sh](mailto:abuse@ppy.sh) detailing any of your behaviour that you think may have been inappropriate. Include as many screenshots and chat logs as you can, and try to explain the context for why the behaviour was inappropriate if it is relevant.
```
MChecaH: From what I know, it's perhaps best these people are directed to `accounts@ppy.sh` as described in the [account support team wiki article](https://osu.ppy.sh/wiki/en/People/Account_support_team#accounts@ppy.sh). The abuse inbox should be used only for the most important reports, when users believe this is the **only** way to get the help they need. Not really as a self-report inbox.

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -60,19 +62,27 @@ Please be aware that the person will see you have blocked them if they check you
...
+
+Your honesty and help in solving such cases is heavily appreciated, but given the inherent severity of harassment and sexual misconduct, a reduced punishment cannot be guaranteed.
```
MChecaH: I'd, personally, remove this sentence. Not only is it assuming what the user may or may not have done, but it also feels somewhat discouraging(?) for those who would potentially be willing to admit their fault. Seems counterproductive.
```suggestion
```

MChecaH (commented):
Few details

wiki/Reporting_bad_behaviour/Abuse/en.md
Userlog2: FWIW, I know this because I requested a data export and got all my past messages, including DMs.

wiki/Reporting_bad_behaviour/Abuse/en.md
Userlog2: I'm not sure, since I think admitting to misbehaviour to `accounts@ppy.sh` is primarily meant for cheating and foul play, not abuse. Will ask @peppy for a 2nd opinion.

wiki/Reporting_bad_behaviour/Abuse/en.md
Walavouchey: i wouldn't quite get the concern. yeah osu! doesn't have end-to-end encryption for messages (i think), but i don't think it's mentioned anywhere that it would have. besides, it would be a mistake to assume that any chat service does end-to-end encryption anyway, as it's by no means common

also, "abuse team" just means peppy, no?

peppy:
Is this ready for review? One unchecked point in OP.

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -12,45 +12,55 @@ The abuse email address is only seen by [peppy](https://osu.ppy.sh/users/2) (the
...
+There is no "statute of limitations" for abuse on osu! — do not hesitate to report something because of how long ago it happened.
+
+## Someone's made inappropriate sexual advances or remarks towards me! {id=bad-remarks}
 
 Scroll through the sub-headings below to find out what to do, depending on where this all happened.
 
-### In the in-game chat
+### In the in-game chat {id=chat}
+
+*Caution: Reporting a user via the `!report` command in osu!(lazer) or via the website will cause the report to be **visible to everyone!** Instead, [use the report menu](#lazer-reports) or report them via [their profile page](#profile-page).*
```
Walavouchey: the section is in another page

```suggestion
*Caution: Reporting a user via the `!report` command in osu!(lazer) or via the website will cause the report to be **visible to everyone!** Instead, [use the report menu](/wiki/Reporting_bad_behaviour#lazer-reports) or report them via [their profile page](/wiki/Reporting_bad_behaviour#profile-page).*
```

Userlog2:
> Is this ready for review? One unchecked point in OP.

Not sure if you would want a section like "What counts as an inappropriate advance/sexual misconduct?", since the wiki page is not clear about it — I would need to write one based on the news post, if so.

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -12,45 +12,67 @@ The abuse email address is only seen by [peppy](https://osu.ppy.sh/users/2) (the
...
+
+## What is considered sexual misconduct? {id=what}
+
+Any behaviour that would likely violate another user's sexual integrity or make them feel harassed is considered unacceptable. Examples include, but are not limited to:
+
+- Requesting sexual photos or acts from another user
+- Sending another user sexual content without reasonable cause
+- Making another user feel pressured to enter or remain in a romantic relationship
+
+**Sexual or romantic conduct in private will usually not be considered problematic if both parties consent to it.** However, if a user is underage (i.e. less than 18 years old), sexual conduct with them may be considered problematic regardless of their consent, especially if the age difference between the users is more than a year or two. If you are not sure about someone's age, ask them, and if you are still not sure, refrain from sexual or romantic conduct towards them.
```
peppy: Where did all this directive come from? I'd have to spend quite a bit of brainpower deciding whether we want to list this here.

I'd *much* rather we are just very explicit that "if someone says something that makes you uncomfortable, report it" and leave the onus of decision making on the person replying to the emails.

wiki/Reporting_bad_behaviour/Abuse/en.md
peppy: It shouldn't be mentioned here, and access is limited (we don't keep them around forever and regularly change purge rules, plus they can be deleted on user request). And yes, the team is just me.

wiki/Reporting_bad_behaviour/Abuse/en.md
Userlog2: Derived some of it (like the "one-party-under-18" guideline) from the wiki; will remove for now. My intent was to prevent offences by laying down more clear rules for what is OK and what is not, but that might not make sense in a "How to report" article.

Should I keep the "Actively using osu! to look for partners is prohibited" thing?

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -12,45 +12,55 @@ The abuse email address is only seen by [peppy](https://osu.ppy.sh/users/2) (the
...
+There is no "statute of limitations" for abuse on osu! — do not hesitate to report something because of how long ago it happened.
+
+## Someone's made inappropriate sexual advances or remarks towards me! {id=bad-remarks}
 
 Scroll through the sub-headings below to find out what to do, depending on where this all happened.
 
-### In the in-game chat
+### In the in-game chat {id=chat}
+
+*Caution: Reporting a user via the `!report` command in osu!(lazer) does not work, and the report will be **visible to everyone!** Instead, [use the report menu](/wiki/Reporting_bad_behaviour#lazer-reports).*
```
peppy: This is no longer the case. `!` commands are [now private](https://github.com/ppy/osu-web/pull/12717). I'm not 100% sure it will work though, so please test it.

Userlog2: Fixed.

wiki/Reporting_bad_behaviour/Abuse/en.md
```diff
@@ -60,21 +70,25 @@ Please be aware that the person will see you have blocked them if they check you
...
 
-Send an email to [abuse@ppy.sh](mailto:abuse@ppy.sh) containing as much information as possible **immediately**. This includes screenshots, chat logs, basically anything you can attach to an email.
+Send an email to [abuse@ppy.sh](mailto:abuse@ppy.sh) containing as much information as possible **immediately**. This includes screenshots, chat logs, and basically anything else you can attach to an email.
 
 If anything you want to include makes you feel embarrassed, please blank it out. 
 
 We'll look into it as soon as we can.
 
-## Someone I know is being abused by a member of the volunteer teams (NAT, GMT) but is too afraid to speak up!
+## Someone I know is being abused by a member of the volunteer teams (NAT, GMT) but is too afraid to speak up! {id=friend}
```
TicClick: nit: `friend` is not the best option, as it implies abuse coming from a friend due to the way other anchors are chosen -- maybe `friend-abused` or `represent-a-friend`?

TicClick (commented):
as a moderator I'm okay with the changes proposed, but it's not my call to make really, so deferring

wiki/Reporting_bad_behaviour/Abuse/en.md
Userlog2: changed to `friend-cannot-report`

wiki/Reporting_bad_behaviour/Abuse/en.md
Userlog2: image

[he speaks his clear words](https://discord.com/channels/188630481301012481/218677502141399041/1508704246516945058)

peppy (approved):

PR #14352, opened 2026-02-11
Update `Beatmap ranking queue` with new Qualified time limit

MChecaH:
After some [internal discussion](https://discord.com/channels/90072389919997952/1451321209831555194), in an attempt to mitigate the impact from users arbitrarily stalling a map's ranking date, we've added a **hard limit** to how long maps may be stalled with open suggestion/problem posts.

This change...
- ...puts a hard cap to how long a discussion may be active for in a Qualified map before the NAT is forced to either disqualify the set so the issues can get some closure, or close the suggestions if they're non-blocking.
- ...allows some leeway for those users who really want to rank their sets on a predetermined date.

## Self-check

- [x] Reviewed by the NAT
- [x] Add entry to RC to link the article
- [x] Make a BN announcement to go along with the merging of this post (can be left last)
- [x] edit to unclog task-list-completed


TheMoMan (approved):

MChecaH:
> What happens to maps with vetoes or Content Reviews that block them? If they start late, it could force a DQ despite the earlier reform aimed at stopping DQs for ongoing Cars/vetoes.

This is already covered in [here](https://osu.ppy.sh/wiki/en/Rules/Content_voting_process). You are mixing a mapping-related reason to halt the map (this pertains to this PR) and a moderation-related reason to block the map (that's the link I sent above). Tldr is that those maps would stay Qualified.

For the purposes of this article I wouldn't start adding "if else if else if else" with all special cases. Moderation in itself *is* a special case and we cannot account for potential edge cases that may occur. So for what the average mapper and modder is concerned, those mentioned cases should not be a cause for concern.

Vetoes are receiving a rework soon, so their wiki article will be changed to better reflect that in due time.

Hiviexd (approved):

Walavouchey (approved):

PR #14476, opened 2026-03-06
Rules: Update Anime Streaming Service Example

NoffyNoffeh:
Because animelab has not existed for, like, 5 years.

No actual rules or wording is touched in this, so I hope directly changing the example this way is OK.

DO_NOT_OUTDATE: *


TicClick (dismissed):
(added `DO_NOT_OUTDATE: *` because this doesn't look like a change substantial enough to outdate translations)

Joehuu (approved):

MChecaH (approved):

peppy (approved):

PR #14540, opened 2026-03-24
Change triggers for labeler and welcome github actions

Walavouchey:
even though [both](https://github.com/garg3133/welcome-new-contributors) [readmes](https://github.com/actions/labeler?tab=readme-ov-file) include it in the examples, apparently `pull_request_target` might not be required? works when testing in https://github.com/Walavouchey/osu-wiki/pull/38, with my fork's `master` branch having this change applied. the action logs don't complain anything and it seems to apply labels correctly

of course, that's a pr on the same repo, but if that isn't a significant detail then it'd be pretty great since `pull_request` [isn't as privileged](https://github.com/actions/labeler?tab=readme-ov-file) (but note that neither of the workflows check out any files anyway). i dunno what the easiest way to test this would be other than trying it here

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,6 +1,6 @@
 name: Label pull request
 on:
-  pull_request_target:
+  pull_request:
     branches:
       - master
     types:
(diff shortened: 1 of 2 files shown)


TicClick (approved):
(discussed this 1:1 before the PR hence the approval)
let's get this in and roll back if something breaks (I personally don't consider these critically important actions)

PR #14536, opened 2026-03-23
[DE] Update `Reporting bad behaviour`

The-Last-Cookie:
- [x] Waiting for #14555 to be merged
- [x] Apply changes from #14555

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #14555, opened 2026-03-26
Revert osu!(lazer) and osu!web reports being described as unsafe in ``Reporting_bad_behaviour``

Userlog2:
Sorry for forgetting about this for so long @The-Last-Cookie re: #14536 

## Self-check


wiki/Reporting_bad_behaviour/en.md
```diff
@@ -22,11 +22,9 @@ For anything else, see the instructions below.
...
-
 ![](img/report-command.jpg "Example of using the !report command")
 
-Chat misbehaviour is the most common infringement. To report it, use [the `!report` command](https://osu.ppy.sh/community/forums/topics/34843), preferably in the channel where the infringement(s) took place. Don't feel stressed: if your message starts with `!report` and a space, and you sent it from within osu!(stable) or an IRC client, **it is only visible to moderators** and no one else in the chat. You can also use the report function via the osu!(stable) client's [in-game overlay](#stable-reports) as depicted below if you are still concerned.
+Chat misbehaviour is the most common infringement. To report it, use [the `!report` command](https://osu.ppy.sh/community/forums/topics/34843), preferably in the channel where the infringement(s) took place. Don't feel stressed: if your message starts with `!report` and a space, **it is only visible to moderators** and no one else in the chat. You can also use the report function via the osu!(stable) client's [in-game overlay](#stable-reports) as depicted below, or via [the offender's profile page](#profile-page), if you are still concerned.
```
msmcculley, line 27: the "it is only visible to moderators" reads like the moderators are able to see the command in chat rather than the report. As it stands right now any and all "!" commands are purged from lazer/web chat. At least that is how I read it, so idk if it would be better to just say that the command is not visible in chat.

Walavouchey: reports are still only visible to moderators, and the current wording sounds pretty clear to me. i don't think adding technical details of how chat commands work is necessary here, and idk how to convey what you're describing in a way that doesn't just make this part longer for little gain

Userlog2: To the average user, there is no practical difference, and I think the GMT would be looking at internal documentation rather than a public-facing wiki page if this becomes relevant.

MChecaH (approved):

Issue #3152, opened 2020-02-03
Add Article Stubs (Hit Objects)


Loctav:
See the main issue https://github.com/ppy/osu-wiki/issues/3143
See the project: https://github.com/ppy/osu-wiki/projects/3

Here you find all article stubs that need to be created. Some articles might already exist. Some articles can be deemed unnecessary as singular article and be summed up into a bigger article. 

Each article should only contain information of one topic. 

You can claim articles beforehand. Just comment here.

- [ ] **Hit Object**
- [ ] Slider
- [ ] Slider Tick Rate
- [ ] Slider Anchor
- [ ] Slider Velocity
- [ ] Spinner
- [ ] Bananas
- [ ] Hit Circle
- [ ] Approach Circle
- [ ] Hit Note
- [ ] Hitsound
- [ ] Keysound
- [ ] Hold
- [ ] Release
- [ ] Don and Kat
- [ ] Drumroll
- [ ] Finisher
- [ ] Fruit
- [ ] Drop
- [ ] Droplet
- [ ] Barline
- [ ] Rest Moment

Pennek:
I'll be refactoring the Ranking Criteria Glossary into stubs and crosslinking them over the next couple of days. Do we want to throw all of these articles into the Beatmapping directory for now?

Also @MegaApplePi do you mind if I create a stub for Sliders since I saw you working on it at #3161 ?

MegaApplePi:
@Pennek go ahead

Gamingzone123:
a bit of an update on which stubs still need to be written

approach circle has a stub https://osu.ppy.sh/wiki/en/Hit_object/Approach_circle

hitsound has a stub https://osu.ppy.sh/wiki/en/Beatmapping/Hitsound

edge dash has a stub https://osu.ppy.sh/wiki/en/Gameplay/Edge_dash

Issue #7139, opened 2022-04-05
Prepare the wiki for osu!(lazer)


TicClick:
# the issue

most of the articles listed on the wiki only describe how the current stable build functions. this goes for things such as gameplay mechanics or chat ([osu!supporter](https://osu.ppy.sh/wiki/en/osu%21supporter) colour for instance).

as lazer may (and will) differ significantly from stable, at least with respect to certain features, we need to figure out how to handle the wiki update in advance.

# related

- https://github.com/ppy/osu-wiki/issues/6590
- https://github.com/ppy/osu-wiki/issues/4686
- user count: https://status.ppy.sh/

TicClick:
a couple ideas from myself:

- a separate section for every such article. quick & cheap, but will definitely disrupt the flow of almost any article
- a separate namespace or a query string for all articles, such as `https://osu.ppy.sh/wiki/lazer/en/Main_Page` or `https://osu.ppy.sh/wiki/en/Main_Page?release=lazer`
- a separate **subarticle**, such as `https://osu.ppy.sh/wiki/en/Beatmap/Difficulty/Lazer` (imo the worst solution, because it will nearly double the amount of articles and likely make editing harder)

peppy:
I think it should be done in a way that we can switch the primary source of information when the majority of players are using lazer, with the old client articles being moved to `(Legacy)` or similar.

Walavouchey:
> a separate namespace or a query string for all articles, such as `https://osu.ppy.sh/wiki/lazer/en/Main_Page`

I think this could be a very elegant solution, but it doesn't necessarily need web support.

In practice, this would be a separate folder in the repo (`lazer` and then later `legacy`) with a mirrored file structure for the articles there.

- If the only difference is images, we can store them in the new folder (e.g. `/wiki/lazer/Client/Beatmap_editor/Beat_Snap_Divisor/img/*`) if we want to prepare them beforehand.
- If there are partial differences, make a stub explaining the differences with a hatnote pointing to it from the original article. After switching, the now "legacy" stub would explain differences in stable.
- If the whole article is stable-specific (e.g. interface articles), the lazer-specific version would just be a complete article.
- If something exists only in lazer, they just get their own article (game modifier articles for instance, with `Game_modifier` either having `(lazer only)` in the listing or having a new version of itself).

I don't think there's a need for a lazer-specific main page

peppy:
[...]

PR #14560, opened 2026-03-27
Add osu!taiko `drumrolls`

The-Last-Cookie:
Related: #3152 

- [x] TODO: Mention big drumrolls?

## Self-check


wiki/Gameplay/Hit_object/Drumroll/en.md
```diff
@@ -0,0 +1,20 @@
+---
+tags:
+  - hit objects
+---
+
+# Drumroll
+
+![](img/taiko_drumroll.png "Drumroll, with the judgement circle on the left side")
+
+**Drumrolls** are [hit objects](/wiki/Gameplay/Hit_object) in [osu!taiko](/wiki/Game_mode/osu!taiko) [beatmaps](/wiki/Beatmap). They appear as yellow horizontal bars and move from right to left on the conveyer belt. The user is required to hit either part of the drum for each white tick found inside the drumroll once they are above the white **judgement circle**.
```
Walavouchey: only the main words being defined should be bolded

```suggestion
**Drumrolls** are [hit objects](/wiki/Gameplay/Hit_object) in [osu!taiko](/wiki/Game_mode/osu!taiko) [beatmaps](/wiki/Beatmap). They appear as yellow horizontal bars and move from right to left on the conveyor belt. The player is required to hit either part of the drum for each white tick found inside the drumroll once they are on the white *judgement circle* to the right of the drum.
```

wiki/Gameplay/Hit_object/Drumroll/en.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Judgement
+
+Hitting drumrolls correctly will not refill the [health bar](/wiki/Gameplay/Health). When missed, drumrolls reward no score.
+
+Each successful hit on the white ticks gives a static 300 score and adds to the `DRUM TICK` judgement.
```
Walavouchey: the 300 is only relevant for stable (if it's correct, haven't checked). in lazer, they only give 10 points

```suggestion
Each successful hit on the white ticks gives a static 300 bonus score.
```

wiki/Gameplay/Hit_object/Drumroll/en.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Trivia
+
+- In [Taiko no Tatsujin](https://en.wikipedia.org/wiki/Taiko_no_Tatsujin), drumrolls require the player to hit the drum as fast as possible.
```
Walavouchey:
```suggestion
- In [Taiko no Tatsujin](https://en.wikipedia.org/wiki/Taiko_no_Tatsujin), drumrolls require the player to hit the drum as fast as possible instead of every [1/4 tick](/wiki/Client/Beatmap_editor/Beat_snap_divisor).
```

Walavouchey (changes requested):

wiki/Gameplay/Hit_object/Drumroll/en.md
```diff
@@ -0,0 +1,20 @@
+---
+tags:
+  - hit objects
+---
+
+# Drumroll
+
+![](img/taiko_drumroll.png "Drumroll, with the judgement circle on the left side")
```
The-Last-Cookie:
```suggestion
![Drumroll, with the judgement circle on the left side](img/taiko_drumroll.png "Drumroll, with the judgement circle on the left side")
```

Walavouchey:
are you going to mention the big drumrolls? iirc the only difference is that they give more score (both with and without kiai). i don't have stable handy to check but it's pretty simple to do

wiki/Gameplay/Hit_object/Drumroll/en.md
```diff
@@ -0,0 +1,20 @@
...
+
+![Drumroll, with the judgement circle on the left side](img/taiko_drumroll.png "Drumroll, with the judgement circle on the left side")
+
+**Drumrolls** are [hit objects](/wiki/Gameplay/Hit_object) in [osu!taiko](/wiki/Game_mode/osu!taiko) [beatmaps](/wiki/Beatmap). They appear as yellow horizontal bars and move from right to left on the conveyor belt. The player is required to hit either part of the drum for each white tick found inside the drumroll once they are on the white *judgement circle* to the right of the drum.
+
+## Judgement
+
+Hitting drumrolls correctly will not refill the [health bar](/wiki/Gameplay/Health). When missed, drumrolls reward no score.
+
+Each successful hit on the white ticks gives a static 300 bonus score.
```
The-Last-Cookie:
```suggestion
Each successful hit on the white ticks gives a static 300 bonus score. Big drumrolls work the same as the regular ones, but each tick awards a static 720 bonus score instead.
```

The-Last-Cookie:
In each osu!taiko hit object article, the `judgement` section only describes stable's behaviour for now. I am unsure of what structure you would like to see in terms of stable vs. lazer in hit object articles, and also regarding what information should be presented in these articles.

Personally, I think that a hit object article should lay out information on what the respective hit object is, how it behaves, and how the player interacts with it.

Nevertheless, I believe the three osu!taiko hit object articles are a good start for now.

@Walavouchey

Walavouchey:
the path forward for lazer articles is described in https://github.com/ppy/osu-wiki/issues/7139
there would be a separate version with "(lazer)" appended to the title, which would ofc include lazer screenshots and whatnot

Walavouchey (approved):

PR #14567, opened 2026-03-30
[PL] Update `meta/group-info`

kn00pers:
## Self-check


meta/group-info/pl.yaml
Walavouchey, line 1: the `outdated_since` and `outdated_translation` tags at the top should be removed since you've brought the file up to date

kn00pers (commented):
Updated

TicClick (approved):

PR #14595, opened 2026-04-06
[DE] Update `Beatmap/Circle size`

The-Last-Cookie:
## Self-check


TicClick:
(looks ok according to my limited proficiency)

TicClick (approved):

PR #14617, opened 2026-04-13
Fix commit hash in `Beatmap/Packs`

The-Last-Cookie:
Related to #14589

When merged, every commit is squashed into one, new commit. The `outdated_since` tag needs to reference this new commit hash.

This is why translation *must not* be outdated manually.

cc @Sitoria @Nivalyx 

## Self-check


Sitoria:
ok my bad again sorry for double pr
i'll never do that again

Walavouchey (approved):

The-Last-Cookie:
> ok my bad again sorry for double pr i'll never do that again

No worries. Commits being squashed is relatively new, so you were probably not aware of this. Now you know.

PR #14645, opened 2026-04-23
Add new alignment tags to `BBCode`

someonecantcode:
https://github.com/ppy/osu-web/pull/12896 
As a side note, the alignment tags are not only limited to text. They can align any BBCode element (images, videos, etc.).

## Self-check


wiki/BBCode/en.md
```diff
@@ -182,6 +182,14 @@ Within the osu! forums, code blocks are most often used to post source code for
 
 The `[centre]` tag is used to align text to the centre of a box. This is most often used for stylistic effect in titles, headers, or poems. If placed inside of or surrounding a `[quote]` tag, the text inside the quote block will be centred, but the stylistic lines and such will not.
 
+### Right
+
+```
+[right]text[/right]
+```
+
+The `[right]` tag is used to align text to the right of a box. This is most often used for right-to-left languages alongside stylistic effect in titles, headers, or poems. If placed inside of or surrounding a `[quote]` tag, the text inside the quote block will be right aligned, but the stylistic lines and such will not.
```
Walavouchey:
```suggestion
The `[right]` tag is used to align text to the right of a box. This is most often used for right-to-left languages alongside stylistic effect in titles, headers, or poems. If placed inside of or surrounding a `[quote]` tag, the text inside the quote block will be right-aligned, but the stylistic lines and such will not.
```

Walavouchey: what are "stylistic lines"

Walavouchey (commented):
what about `[left]`?

The-Last-Cookie (commented):
Personally, I would have suggested to move all 3 tags into one `alignment` section. Otherwise, the tag description would essentially be copied 2 times.

someonecantcode:
> what about `[left]`?

oops, I missed it. The only thing that doesn't make sense is that all elements are automatically left aligned in the first place.
When nesting alignment tags, the inner tag will overrides the outer tag, making a new line with the new alignment, which kind of serves no purpose as it is identical to just not nesting them.

```
[right][left]yes[/left]no[/right]

yes
[right]no[/right]
```

> Personally, I would have suggested to move all 3 tags into one `alignment` section. Otherwise, the tag description would essentially be copied 2 times.

Agreed. I will rewrite and add a new section to reduce redundancy.

wiki/BBCode/en.md
SisypheOvO: this means the vertical line at the left side of a quote block. This sentence is copied from the `[centre]`'s description right above.

SisypheOvO:
> "The `[right]` tag is used to align text to the right of a box."

This seems misleading because it obviously can be used directly without a `[box]` tag outside.

> "This is most often used for right-to-left languages alongside stylistic effect in titles, headers, or poems."

This is also not quite precise because 
1. firstly RTL languages may not be the main reason we are adding this tag for website bbcode render.
2. descriptions that works for the `[centre]` tag will not be fully appliable for `[right]` so you have to think twice before explaining.
3. we usually want to align anything you can imagine to the right of the profile, not just specific languages, titles and something like that. For example I've seen at least 2 beatmap descriptions trying to implement a similar style for the "credits" part or something else (before we added the `[right]` tag). Besides that, I did this in my profile, and both 2 lines do not belong to the things you've listed (which is mostly because this sentence is simply describing `[centre]` instead of `[right]`):

    image

someonecantcode:
> "descriptions that works for the [centre] tag will not be fully appliable for [right] so you have to think twice before explaining."

Hello, I tested the `[quote]` tag on all three alignment tags and the vertical line produced is consistently always left aligned so the description applies for all three. I tried to keep everything more accurate alongside their original state.

SisypheOvO:
> Hello, I tested the `[quote]` tag on all three alignment tags and the vertical line produced is consistently always left aligned so the description applies for all three. I tried to keep everything more accurate alongside their original state.

oh yeah i mean the "stylistic effect in titles, headers, or poems" part when I'm saying that. The `[quote]` vertical line surely won't change. Sorry for the misunderstanding.

wiki/BBCode/en.md
```diff
@@ -15,6 +15,34 @@ A set of correct and incorrect usages of nested tags is described below:
 - `[centre][b]text[/b][/centre]` is correct
 - `[b][centre]text[/b][/centre]` is incorrect
 
+## Alignment
+
+Alignment tags are used to control the horizontal positioning of text and BBCode elements. If placed inside of or surrounding a `[quote]` tag, the text inside the quote block will be aligned accordingly with the alignment tag, but the stylistic lines and such will not.
```
Walavouchey: probably should be described more clearly

```suggestion
Alignment tags are used to control the horizontal positioning of text and BBCode elements. If placed inside of or surrounding a `[quote]` tag, the text inside the quote block will be aligned accordingly with the alignment tag, but the vertical line of the quote block will remain on the left.
```

someonecantcode:
I did something wrong with git and so I'm just going to close it and make another branch. Apologies.

PR #13933, opened 2025-11-09
Rewrite Twitter page to X (branding update + full translations)

iSweat-exe:
### Summary

This pull request replaces the outdated **Twitter** page with a new one for **X**, reflecting the current branding and account links used by the osu! team and community.

### Changes

- Replaced all `twitter.com` links with `x.com`
- Renamed the page directory from `wiki/Twitter` → `wiki/X`
- Updated all language variants with new branding
- Added or updated translations for:
  - 🇬🇧 English
  - 🇫🇷 French
  - 🇩🇪 German
  - 🇪🇸 Spanish
  - 🇮🇹 Italian
  - 🇵🇹 Portuguese (pt-br)
  - 🇳🇱 Dutch
  - 🇷🇴 Romanian
  - 🇵🇱 Polish
  - 🇷🇺 Russian
  - 🇧🇬 Bulgarian
  - 🇹🇭 Thai
  - 🇹🇷 Turkish
  - 🇻🇳 Vietnamese
  - 🇯🇵 Japanese
  - 🇰🇷 Korean
  - 🇨🇳 Simplified Chinese
  - 🇹🇼 Traditional Chinese
  - 🇮🇩 Indonesian

### Motivation

Twitter was rebranded as **X** in 2023.  
The osu! wiki still contained outdated references to the old platform.  
This rewrite aligns the wiki with the current branding and ensures multilingual consistency.

### Self-check

- [x] Changes validated with `remark-lint` (`no issues found`)
- [x] Verified all images and references resolve correctly
- [x] The structure follows osu! wiki’s contribution standards
- [x] All translations reviewed and synchronized

---

**Author:** iSweat  
**Date:** 2025-11-09  
**Type:** Content / Localization / Branding

PR #14658, opened 2026-04-26
[EN] Update `Twitter`

mousesalt:
updating it so its looking fresh..

## Self-check


TicClick:
sorry, this does not look like a substantial update, and I don't think we're interested in renaming Twitter to X either as per https://github.com/ppy/osu-wiki/pull/13933#issuecomment-3507376020

PR #14679, opened 2026-05-02
Fixed 3 beatmap links in bespoke music page

kadoen:
Some wrong beatmap links
Apart from these there is one case where the original set was deleted, and ranked in a new set, but it probably makes sense to keep linking to the original set as the WC page does.


Walavouchey (approved):
i'll check that the spreadsheet is also updated later

PR #14685, opened 2026-05-04
Update `osu!mania Mapping World Cup 2026` wiki article

rmdhnilham99:
## Self-check


wiki/Contests/MMWC/2026/en.md
```diff
@@ -12,16 +12,16 @@ The **osu!mania Mapping World Cup 2026** (***MMWC 2026***) is a country-based, m
 
 ## Schedule
 
-| Event | Timestamp (UTC) |
+| Event | Timestamp (0 UTC) |
```
MChecaH:
```suggestion
| Event | Timestamp (00:00 UTC) |
```

wiki/Contests/MMWC/2026/en.md
```diff
@@ -40,9 +40,66 @@ The osu!mania Mapping World Cup 2026 is run by various community members.
...
 | Designer | ::{ flag=ID }:: [Syadow-](https://osu.ppy.sh/users/3484485) |
 | Statistician | ::{ flag=ID }:: [Reihynn](https://osu.ppy.sh/users/16630515) |
+| Video Editor | ::{ flag=ID }:: [Mirsaaa](https://osu.ppy.sh/users/15328093) |
 | Commentators | ::{ flag=SG }:: [Raveille](https://osu.ppy.sh/users/1388767), ::{ flag=PH }:: [SurfChu85](https://osu.ppy.sh/users/4469895) |
 | Judges | ::{ flag=VN }:: [\[TCD\] Dzar03](https://osu.ppy.sh/users/16712231), ::{ flag=DO }:: [Antalf](https://osu.ppy.sh/users/8793773), ::{ flag=PH }:: [chocomilku-](https://osu.ppy.sh/users/16687756), ::{ flag=AU }:: [fvrex](https://osu.ppy.sh/users/11863699), ::{ flag=CN }:: [gzdongsheng](https://osu.ppy.sh/users/8660315), ::{ flag=GB }:: [Hydria](https://osu.ppy.sh/users/808176), ::{ flag=RU }:: [Kurisu Makise](https://osu.ppy.sh/users/2790640), ::{ flag=CA }:: [Ryax](https://osu.ppy.sh/users/8949769), ::{ flag=KR }:: [secXcscX](https://osu.ppy.sh/users/13543418), ::{ flag=US }:: [Unpredictable](https://osu.ppy.sh/users/7560872) |
 
+## Participants
+
+| | Country/Continent | Members | Note |
```
MChecaH:
```suggestion
|  | Country/Continent | Members | Note |
```

MChecaH (changes requested):

wiki/Contests/MMWC/2026/en.md
Walavouchey: wuh? plain `UTC` is standard

MChecaH: Host wanted it to be specified, was confused by the entire day (23:59) rather than the beginning of which.

If that was the case maybe it'd been better to set the previous day as the last one, but it seemed to cause confusion. Check the thread for more info. I do not mind either way so long it is crystal clear.

rmdhnilham99: doesn't matter for me to use UTC as well, but question: if mappers need to submit before June 12th 00:00 UTC

should we write 2026-06-12 or 2026-06-11 ?

MChecaH: To me it would make more sense to use 2026-06-11. Most users would assume the bounds are included in its entirety unless specified otherwise.

Walavouchey: > Host wanted it to be specified, was confused by the entire day (23:59) rather than the beginning of which.

oh, i was confused because we normally only write the time zone there (UTC / UTC+0), which is what i thought you meant.

put the timestamp after the date in the table rows e.g. `2026-06-11 (23:59)` (and it should preferably be 23:59 in this case). alternatively, add a line to clarify the matter right after the table

MChecaH (approved):

PR #14689, opened 2026-05-04
Attempt to fix image bug in `Help_centre/Silences`

Userlog2:
Because otherwise [the image is stretched when you load it](https://discord.com/channels/188630481301012481/218677502141399041/1500893081711546579). It *might* be cache-related.

## Self-check

DO_NOT_OUTDATE


Walavouchey (dismissed):

Walavouchey (commented):
maybe do the same thing in the translations too?

Walavouchey (approved):

peppy (approved):

PR #14719, opened 2026-05-12
Add key layout info to `osu!mania skinning`

Greenest-Guy:
## Self-check


wiki/Skinning/osu!mania/en.md
```diff
@@ -431,6 +431,24 @@ Notes:
   - If the columns' widths differ: the smallest one is scaled correctly and the others are compressed to match its height.
 - Notes can be manually stretched or compressed via the `WidthForNoteHeightScale` command in the [skin.ini](/wiki/Skinning/skin.ini) file.
 
+---
+
+### Default Key Layout
```
TicClick: think it would be better to have it appear at the beginning of the article, no?

TicClick: also, the heading must be in sentence case, so `Default key layout`

Greenest-Guy: Fixed the heading. Also I thought it would make more sense to place the layouts after all the key types etc were defined.

TicClick: I don't have a strong preference here, so here goes

TicClick (approved):

PR #14721, opened 2026-05-13
Skip re-running passed CI checks when PR tags haven't changed

SisypheOvO:
Currently if it is just a normal title change / description change, the CI will be quickly skipped.

Result as you can see (a much faster 3s simple check):

image

The speed of normal CIs will also be massively inproved because checkout with fetch-depth 0 is no longer used.

---

This PR
- Adds a ~conditional and reusable~ ~workflow~ job to determine whether the main CI should run, optimizing CI usage by skipping unnecessary runs.
- Integrates it into the main CI pipeline 
- Removes redundant conditional checks from individual steps ~(we don't need `!cancelled()` specifically)~
- Use Rest API to get changed-files list, and restore them later using `jq`, replacing the `checkout with fetch-depth 0` - `git branch/fetch` - `git diff` workflow

---

The ~conditional workflow~ job
- maintains a flag
- checks the type of `github.event.pull_request.action`
- if the type is `edited`, checks if there are specific PR body flag changes (`SKIP_REMARK`, `SKIP_WIKILINK_CHECK`)


TicClick (changes requested):
I don't like the implementation for two reasons:
1) feels like such a check should belong to the main workflow
2) it clones the repo which adds overhead (~30-50s) essentially eliminating the benefits of the PR

that being said, it's still a vaild idea -- do you want to try and reimplement it at least without the clone using, for example, [actions/github-script](https://github.com/ppy/osu-wiki/blob/08cb204145832ef0ab7066bb2541a66543f4310e/.github/workflows/post-merge-outdate.yml#L33)?

SisypheOvO:
> 1. feels like such a check should belong to the main workflow

fixed

> 2. it clones the repo which adds overhead (~30-50s) essentially eliminating the benefits of the PR

fixed

> do you want to try and reimplement it at least without the clone using, for example, [actions/github-script](https://github.com/ppy/osu-wiki/blob/08cb204145832ef0ab7066bb2541a66543f4310e/.github/workflows/post-merge-outdate.yml#L33)?

yeah!

I also did some other changes towards the final idea, all in commits above.

and I think [actions/github-script](https://github.com/actions/github-script) used here and there can all be bumped to v7 to use node20 (or more, also bump `node-version: 20` in step `setup-node` in order to use node24). This can be done in another PR.

TicClick:
need to think about this a bit more, will post later

TicClick (changes requested):
https://github.com/SisypheOvO/osu-wiki/pull/3

TicClick (approved):

PR #14792, opened 2026-06-01
Delete custom fork syncing action and related instructions

TicClick:
(GitHub can do that just fine now)

## Self-check


Walavouchey (approved):
makes sense. almost forgot this existed which hopefully is indicative

PR #14794, opened 2026-06-02
Update and extend lazer mod articles

AndiTafel:
### This PR updates all the articles about lazer mods (`Game_modifer_(lazer)`).

First of all, I fixed these errors:

- `HD/FR`, `TC/DP`, `AP/TD`, `TP/DC` and `NR/IN` are incompatible.
- `AC` allows for minimum accuracy up to 99.9%.
- `CO`, `SW`, `AL` and `SG` are ranked.
- `CL` is available in osu!mania.
- `HD` and `FL` incompatibilities updated by mode.
- osu!mania key mods have a score multiplier of `0.90x`.
- `SY` is available in osu!catch.

Here I also updated the translations [ES, ZH], see https://github.com/ppy/osu-wiki/pull/14839.

---

Furthermore, I extended most articles:

- Added at least a brief description to every single mod, so now none of the articles are empty anymore.
- Added `Customisation` for all mods that were missing it (including screenshots).
- Rewrote and updated `Game_modifer_(lazer)` a bit.

On top of that I fixed some inconsistencies and typos and rewrote some things, but since there are 60 different mods in lazer, I'm sure I also made some mistakes myself. I appreciate anyone who takes the effort to review this.

These changes are not included in any translations (in particular, this PR doesn't touch the translation files at all).

---

Along with this, two other PRs implement some more changes:

- Update all mod icons to the current designs, see https://github.com/ppy/osu-wiki/pull/14837.
- Add `Simplified Rhythm`, `Moving Fast` and `Dual Stages` as these are still missing from the wiki entirely, see https://github.com/ppy/osu-wiki/pull/14836.

---

*Note:* I also wrote some explanations on how the new score multipliers work, but these are not included in any of these PRs. I will make a new PR to add those once the current changes have been reviewed and merged.

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #14798, opened 2026-06-02
Update mod icons

AndiTafel:
osu-web is using these icons, so I don't see why osu-wiki shouldn't.

I got them from [here](https://www.figma.com/design/gjt79VMELJMl0CdTOmF7ng/Master-Library?node-id=18783-2090&p=f&t=9iv2rB47uB1cO6AQ-0) but changed TC to be red and SV2 to be yellow as in the client.

I moved the previous icons to an ``old`` directory inside ``mods``, idk if that's necessary.

## Self-check


AndiTafel:
Included in https://github.com/ppy/osu-wiki/pull/14794

PR #14837, opened 2026-06-13
Update lazer mod icons

AndiTafel:
### Part of https://github.com/ppy/osu-wiki/pull/14794.

- I got them from [here](https://www.figma.com/design/gjt79VMELJMl0CdTOmF7ng/Master-Library?node-id=18783-2090) but changed `TC` to be red and `SV2` to be yellow as in the client.
  - Also updated the link in ``wiki/Gameplay/Game_modifier_(lazer)/img/mods/README.md``.
- I moved the previous icons to an `old` directory inside `mods`, idk if that's necessary.

(This is incorrectly labeled as `area:translations`.)

## Self-check


TicClick (approved):

Walavouchey:
there's no reason to keep the old icons, is there?

AndiTafel:
> there's no reason to keep the old icons, is there?

I guess not, though I can't remove the `old` folder now, since this is already merged. But go ahead and delete it if it's not needed.

TicClick:
it's a lot of icons and I think we may potentially need them (idk what for, though), hence the decision. if you have a strong case for the removal, go ahead

PR #12966, opened 2025-03-15
Updated "Cookie" article

mdlcgithub:
This PR modifies the article "Cookie", which is a sub-article of "Interface", which itself is also a sub-article of "Client".
- The notable change is the addition of a history for the osu! cookie.
This PR isn't checked by any reviewer and has only modified the English language variant, so BEWARE.

## Self-check

- [x ] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [ ] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #14850, opened 2026-06-15
Complete `Client/Interface/Cookie` history with the missing images

Anax2i:
This is basically the follow-up of https://github.com/ppy/osu-wiki/pull/12966 (see the thread for more details)

- The temporary 2019, 2024 and current osu! lazer cookies have been added (all 1024x1024)
- Two categories have been created in the English translation ("History osu! stable" and "History osu! lazer")

## Self-check


wiki/Client/Interface/Cookie/en.md
```diff
@@ -12,17 +12,26 @@ tags:
 
 ![Window dialog of osu! crashing](img/Pippi_corruption.jpg)
 
-## History
+## History (osu! stable)
```
TicClick: two things:
1) it's `osu!(stable)` and `osu!(lazer)`
2) I think the following structure works better:

```md
## History

### osu!(stable)

...

### osu!(lazer)

...
```

TicClick (changes requested):

TicClick (approved):

PR #14861, opened 2026-06-17
Update `Medals/First medal unlocks`

AndiTafel:
Medals from April and June 2026 are missing in this list.

Got the data from [Osekai](https://inex.osekai.net/medals) and the medal icons from `https://assets.ppy.sh/medals/web/`.

The `First medals unlocked` leaderboard is cool, but a lot of medals are beatmap packs and these are just a matter of *who happens to be online when new medals drop and can NFDT the maps the fastest*. So I thought having a leaderboard for only first **Hush-Hush** unlocks would be neat as well, since these are not trivial to solve/obtain, so I added one.

I also updated the German translation.

## Self-check


Walavouchey:
@peppy mind cross-checking with the database

The-Last-Cookie (approved):
German looks good

peppy (approved):

PR #14869, opened 2026-06-20
[DE] Add `Beatmap ranking procedure/Ranking queue`

The-Last-Cookie:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #14920, opened 2026-07-05
Improve readability in `Beatmap ranking procedure/Ranking queue`

The-Last-Cookie:
`Ranking date` is ambigious and may be confusing.

Suggestion by @AndiTafel in https://github.com/ppy/osu-wiki/pull/14869/changes#r3445943074

## Self-check


peppy:
not sure this makes things better or not.

wiki/Beatmap_ranking_procedure/Ranking_queue/en.md
```diff
@@ -13,7 +13,7 @@ A timer of 7 days is set when a beatmap enters Qualified. It counts how many day
...
+Some beatmaps may be stalled from entering the ranking queue for a variety of reasons, usually due to remaining open problems or suggestions. A beatmap may only be qualified if these are resolved. If, on the other hand, issues are posted on a beatmap after it has already been qualified, these have to be addressed within a period of **48 hours**. Otherwise, the [Nomination Assessment Team](/wiki/People/Nomination_Assessment_Team) will [disqualify](/wiki/Beatmap_ranking_procedure#nomination-resets) the beatmap. After resolving all issues, the beatmap may be re-qualified, which can alter the timer. Beatmaps may also be disqualified earlier to apply changes from already addressed discussions.
```
Walavouchey: > A beatmap may only be qualified if these are resolved. If, on the other hand, issues are posted on a beatmap after it has already been qualified,

the topic at hand is that the beatmap is in qualified already, so this doesn't make sense. the article describes open problems stalling qualified beatmaps from automatically entering the ranked category

> since its original ranking date

this nuance is not reflected in the new wording? the "original ranking date" refers to the date that the qualified map would've entered ranked, were it not for open problems stalling the process

Walavouchey (changes requested):

wiki/Beatmap_ranking_procedure/Ranking_queue/en.md
AndiTafel: Your comment made me realise I completely misunderstood this paragraph. Sorry for the suggestion, I now see that it makes no sense.

However, I still feel like the current version of the article is confusing. On each map in Qualified it says something like

Image

i.e. each qualified map is already in the ranking queue, so

> Some beatmaps may be stalled from entering the ranking queue

doesn't make sense if this is supposed to be about qualified maps (which lead me to believe it's about pending maps).

How about this instead:
```suggestion
Some qualified beatmaps may be stalled from entering the ranked section for a variety of reasons, usually due to a lack of closure on open problems or suggestions. Any beatmap with issues that have not been addressed after a period of **48 hours** since the date it was supposed to be ranked will be [disqualified](/wiki/Beatmap_ranking_procedure#nomination-resets) by the [Nomination Assessment Team](/wiki/People/Nomination_Assessment_Team). After resolving all issues, the beatmap may be re-qualified, which can alter the timer. Beatmaps may also be disqualified earlier to apply changes from already addressed discussions.
```
Closer to the original, I hope I understood this now.

wiki/Beatmap_ranking_procedure/Ranking_queue/en.md
```diff
@@ -13,7 +13,7 @@ A timer of 7 days is set when a beatmap enters Qualified. It counts how many day
 
 Problem or suggestion stamps posted under a beatmap's discussion page will not stop the timer from counting down, but will stop it from being ranked after the timer reaches zero in 7 days. Once all problems and suggestions are resolved, the beatmap will be able to become ranked.
```
AndiTafel:
```suggestion
Problem or suggestion stamps posted on a beatmap's discussion page will not stop the timer from counting down, but will stop the beatmap from being ranked after the timer reaches zero in 7 days. Once all problems and suggestions are resolved, the beatmap will be able to become ranked.
```

wiki/Beatmap_ranking_procedure/Ranking_queue/en.md
Walavouchey: looks good except i'd change "whose issues" -> "with issues that" because "whose" would typically be used regarding people

AndiTafel: edited it

Walavouchey (approved):

PR #14925, opened 2026-07-06
Change graphic for CC redesign in `/wiki/People/Community_Contributors`

0x84f:
was reminded about this and tbh im not sure why we would keep the up scaled version when the proper one is used right below


Hiviexd:
i think the choice of the old badge is intentional given the "new" old one is a modern recreation of it? if you're looking to address the upscaling we can either use the original size (may be too small) or hope RBRat3 has the original files and may be able to provide a `@2x` export?

Walavouchey:
this doesn't make any sense and would misrepresent the evolution of the badge. if anything you'd switch to the non-upscaled version, not the new recreation

0x84f:
if you guys think its not worth changing im fine with it, just remembered about the upscaled version and at the moment it seemed like a good idea

PR #14927, opened 2026-07-07
Slightly rewrite `Gameplay/Choke`

AndiTafel:
"In osu! and osu!catch, the difference in total score between regular and reverse chokes may be drastic, as late misses cause a player to potentially lose more score due to the large score multiplier from the accumulated combo."

Doesn't make sense. Early misses also cause the player to miss out on the large score multiplier granted by a high/full combo.

## Self-check


Walavouchey (approved):
sure

PR #14968, opened 2026-07-16
Add missing download links to `Medals/Legacy beatmap packs`

### LINKED ISSUE #12449: Not all legacy beatmap packs have links
Only half the links in Legacy Beatmap Packs have links for them (Anime Pack and Rhythm Game Music Pack). Can we have pack links for The Internet and Video Game Music Packs?

https://github.com/ppy/osu-wiki/blob/master/wiki/Medals/Legacy_beatmap_packs/en.md

AndiTafel:
Closes https://github.com/ppy/osu-wiki/issues/12449.

I uploaded all the missing legacy packs to MediaFire. Downloading the existing ones from there was quite slow for me though, so I provided an additional download link for each file (Google Drive). I hope that's fine, if not just let me know where to upload them instead.

I also fixed an incorrect link and reuploaded Anime Pack 1 (the current archive contains `.zip` files rather than `.osz`).

## Self-check


TicClick:
I think this is sufficient. maybe Google Drive is not the best choice, but as long as there's at least one mirror for every link (Mediafire, for example -- by the way, why not mega.io?), we should be fine

AndiTafel:
> I think this is sufficient. maybe Google Drive is not the best choice, but as long as there's at least one mirror for every link (Mediafire, for example -- by the way, why not mega.io?), we should be fine

I just chose MediaFire cause it was already used for half the legacy packs and thought another mirror couldn't hurt, so I uploaded to Google Drive. But of course I can also upload them somewhere else if you like.

TicClick (approved):

Walavouchey:
hmm... maybe we could just upload these to osu!'s servers? (@peppy)

Issue #14970, opened 2026-07-18
User tag `streams/quads` Mismatch between translation


PaperMemo:
Some descriptions said "Patterns requiring consecutively tapping 2 note groups"; others said "4 note groups".

I assume that the "4 note groups" is the correct one instead of "2 note groups" because **quads means four in the Latin word**.

File list:
- 2 note groups
  - https://github.com/ppy/osu-wiki/blob/4d6ea28c27b587646969cbcff4a1adb2f6207d68/wiki/Beatmap/Beatmap_tags/en.md?plain=1#L66
  - https://github.com/ppy/osu-wiki/blob/4d6ea28c27b587646969cbcff4a1adb2f6207d68/wiki/Beatmap/Beatmap_tags/es.md?plain=1#L66

- 4 note groups (I assume this one is correct?)
  - https://github.com/ppy/osu-wiki/blob/4d6ea28c27b587646969cbcff4a1adb2f6207d68/wiki/Beatmap/Beatmap_tags/pl.md?plain=1#L70
  - https://github.com/ppy/osu-wiki/blob/4d6ea28c27b587646969cbcff4a1adb2f6207d68/wiki/Beatmap/Beatmap_tags/ru.md?plain=1#L66
  - https://github.com/ppy/osu-wiki/blob/4d6ea28c27b587646969cbcff4a1adb2f6207d68/wiki/Beatmap/Beatmap_tags/zh.md?plain=1#L66

P.S.: I'm sorry for the English, which may be a bit confusing to others. And I searched for similar issues and did not find any, so I created this issue.

PaperMemo:
@Walavouchey @TicClick Sorry for the mention. But since you're involved in the pull request #14971.

I think your pull requests are incorrect fixed the issue.
In the games, the description about `streams/quads` said: `Patterns requiring consecutively tapping 4 note groups`.

In-game screenshots:
<img width="478" height="114" alt="Image" src="https://github.com/user-attachments/assets/34ee97ce-a03e-4a87-ae26-579f6f6c82f2" />

AndiTafel:
You're right, it should be 2 for `streams/doubles` and 4 for `streams/quads`, see https://github.com/ppy/osu-wiki/pull/14972.

TicClick:
thanks for catching this

PR #14972, opened 2026-07-18
Fix incorrect number in `Beatmap/Beatmap tags` (again)

AndiTafel:
Sorry @Walavouchey, https://github.com/ppy/osu-wiki/pull/14971 was not correct, as pointed out in https://github.com/ppy/osu-wiki/issues/14970#issuecomment-5010388602, cause now `streams/doubles` and `streams/quads` have the exact same description.

DO_NOT_OUTDATE: Beatmap/Beatmap_tags/pl.md

## Self-check


Walavouchey (approved):
my bad

PR #14977, opened 2026-07-18
Fix typo in `Help_centre/Upgrading_to_lazer`

tsunderellaa:
DO_NOT_OUTDATE: *

## Self-check


Walavouchey (approved):

AndiTafel:
I guess the typo doesn't exist in translations, so you might want to add `DO_NOT_OUTDATE` here.

tsunderellaa:
> I guess the typo doesn't exist in translations, so you might want to add `DO_NOT_OUTDATE` here.

I'll keep that in mind. Just to be certain, is that something contributors typically do themselves or do you just leave that for the maintainers?

AndiTafel:
> Just to be certain, is that something contributors typically do themselves or do you just leave that for the maintainers?

You can do it yourself, although in most cases translations should be outdated, so it's not necessary. If you're unsure, leave it to the maintainers.

peppy (approved):

PR #15033, opened 2026-07-25
Fix and switch to new hatnote and notice style in `Beatmap`

TraceXR1:
## Self-check

Not a translations, so no fluent speaker review needed. Fixed new hatnotes and notices in Russian article (they were broken by the script, it seems) and switched to the new style in other languages' articles. No translation, just some ctrl+c ctrl+v work.


TraceXR1:
Oh, worth to note that hatnote and notice styles were applied as explained in ASC/Formatting

TraceXR1:
image

And it's how the Russian `Beatmap` article looks right now. And it definitely shouldn't look like this.

Walavouchey (approved):
ty

PR #15041, opened 2026-07-27
Fix Typo in KonKonKinakoN's Username in OWC 2023 Participants List

cesarila:
It doesn't look like usernames are localized in any way, so I made this change to all languages existing for this page. I don't trust myself to remember this correctly after 3 years, so I found some sources indicating that the proposed change is accurate.
- Japan 2023 OWC Roster Reveal (timestamped): https://youtu.be/yuc22UPuV-A?si=WIc6m2-HX4JglloX&t=47
- Japan 2023 quarterfinals vs Germany (timestamped to when KonKonKinakoN is playing a map): https://youtu.be/OczW9kKP6zg?si=DxyqJltw4ExWC1BH&t=1278
  - Note: While the intro screen showing both team rosters at around 30 seconds in the second video contains the typo I seek to correct, I assume this data comes from the wiki somehow. I am unfamiliar with the tournament client and how it functions. 

## Self-check


TicClick (approved):
ah, this is good, thank you for catching. reminds me that I should run the flag checker again sometime

PR #15060, opened 2026-08-01
Move lazer mod articles from `Game modifier` to `Game modifier (lazer)`

AndiTafel:
While working on https://github.com/ppy/osu-wiki/pull/14794, I found it quite annoying that the stable and lazer mods are mixed into the same directory. In my opinion it would make a lot more sense to separate them by moving all lazer mod articles to `Gameplay/Game modifier (lazer)`.

- [x] Update wikilinks (I'm sure I missed some, but CI will find those)

DO_NOT_OUTDATE: *

(shouldn't matter but just to be sure)

## Self-check


TicClick (commented):
I like the idea as well, but it needs to be checked by another pair of eyes
@Walavouchey @Nivalyx

Walavouchey:
i'm not really seeing the issue being solved (because the directory part isn't user-facing).. this just seems inconsistent with the direction we decided on in https://github.com/ppy/osu-wiki/issues/7139#issuecomment-1426416490 and puts ` (lazer)` into *two* parts of the breadcrumb/article path which is really weird

TicClick:
I understand the breadcrumbs argument, at the same time it becomes easier to manage the game mode articles specifically because there's a lot of them

regarding the plan, we would have `Game modifier` and `Game modifier (lazer)`, so it *kind of* falls under the plan still

TicClick:
maybe the reasoning behind my opinion is not completely sound, but I simply feel uneasy seeing that many articles within one directory.. hence the support

again, it's not fully rational (both your and my posts have points I can agree with), but that's what I would have done

(and the lazer prefix could be removed from the mod articles' names)

AndiTafel:
I agree, the paths becoming something like `Gameplay/Game_modifier_(lazer)/Hidden_(lazer)` is a bit weird, but

- for lazer mods that don't exist in stable it's no problem (`Gameplay/Game_modifier_(lazer)/Traceable` looks ok) and
- it doesn't matter that much because, as Walavouchey said, the directories/links aren't really visible to the user.

> I understand the breadcrumbs argument, at the same time it becomes easier to manage the game mode articles specifically because there's a lot of them

(I guess you mean game modifier articles.) This is exactly why I proposed the change (I want to make some more additions to the lazer mod articles later). It would be much more convenient to be able to tell at a glance which mods belong to stable/lazer when looking at the folders, instead of facing one `Game_modifier` directory with 87 subfolders while `Game_modifier_(lazer)` only contains a single article.

Another thing I just noticed is that this PR "solves" a small "issue" on the website, namely that when browsing through the lazer mod articles and using this bar to navigate back to the list of all lazer mods, it leads to the list of stable mods instead.

image

Walavouchey:
> Another thing I just noticed is that this PR "solves" a small "issue" on the website, namely that when browsing through the lazer mod articles and using this bar to navigate back to the list of all lazer mods, it leads to the list of stable mods instead.

very good point but at the same time i dread it eventually becomes `Gameplay (lazer)/Game modifier (lazer)/Hidden (lazer)`. i'd rather the lazer articles be beside the original ones at the leaves than in a whole different directory tree

the "many articles in one directory" thing is only an issue with the mod articles right? (and you can search/filter directory lists in any file manager.) i don't think this proposed structure makes sense elsewhere anyway, at least from the file management pov

also atm we have hatnotes linking up to the correct parent (and maybe the breadcrumbs could get a temporary fix on the osu!web side)

image

TicClick:
either way I don't have specifically strong arguments for my point this time, so if you want to uphold the current status, go ahead

AndiTafel:
> the "many articles in one directory" thing is only an issue with the mod articles right? (and you can search/filter directory lists in any file manager.)

Ok, but how? Filter for directories with the `_(lazer)` suffix? Doesn't work, because lazer mods that don't exist in stable don't have this.

Another thing is that there are other articles inside `Game_modifier` like `Mod_multiplier` (which only contains information about stable mods). If someone adds `Mod_multiplier_(lazer)` in the future, would this also be placed in `Game_modifier`? Ok I guess.

But what if someone adds an article about e.g. `Personal_presets`? Would that also be inside `Game_modifier` rather than `Game_modifier_(lazer)` (even though it has nothing to do with stable)?

Over time, more lazer-specific articles will be added, and this structure would mean that some of them will be in some nested directory with no way to easily find them/distinguish them from stable articles.

I just think it would be good to have some sort of separation between these articles, especially in an folder with that many subarticles. Though this is all I have to say, so if you're absolutely against it, it's ok to close the PR.

Walavouchey:
> Ok, but how? Filter for directories with the `_(lazer)` suffix? Doesn't work, because lazer mods that don't exist in stable don't have this.

if you want a list of lazer-only mods, it should rather be indicated inside the mod listing in some way probably. what i meant is that navigating around the file system and finding specific articles isn't very difficult

on the other hand, if i didn't remember that e.g. `Traceable` was lazer-only, i'd look in `Game_modifier` instead of `Game_modifier_(lazer)` and not find it. though of course, it can be found with the search function of the file manager. it's of very little importance in any case and i'm not understanding your use case when you say it's "quite annoying that the stable and lazer mods are mixed into the same directory"

> Another thing is that there are other articles inside `Game_modifier` like `Mod_multiplier` (which only contains information about stable mods). If someone adds `Mod_multiplier_(lazer)` in the future, would this also be placed in `Game_modifier`? Ok I guess.

ok so i find that `Game_modifier_(lazer)/Mod_multiplier_(lazer)` just looks incorrect/silly (why indicate lazer-ness in two places), but you don't and we may disagree there, but the way we have it now has a concrete advantage in allowing slow incremental migration without much complication

say you make a lazer version of `Client/Beatmap_editor/Song_setup`. you'll eventually need a `Client/Beatmap_editor_(lazer)`, and, well, maybe `Client_(lazer)` at that point too (kinda equivalent to `Game_modifier_(lazer)` as an index page)

if you want to do `Song_setup_(lazer)` first, it's still going under `Client/Beatmap_editor` (unless you do all the lazer parent articles at once), and if someone then creates `Beatmap_editor_(lazer)`, do we move `Song_setup_(lazer)` there like in this pr? and repeat that when `Client_(lazer)` gets created too?

after this pr, do we move `Game_modifier_(lazer)` to `Gameplay_(lazer)` when that gets created?

you either get a lot of work reorganising the directory tree all the time including a bunch of redirects (btw you're missing redirects in this pr - those are required for article moves), or you're forced to do more work at once instead of migrating articles in any order

> But what if someone adds an article about e.g. `Personal_presets`? Would that also be inside `Game_modifier` rather than `Game_modifier_(lazer)` (even though it has nothing to do with stable)?

the ` (lazer)` thing is disambiguation of the article title, not a category. it's not like it needs to be titled `Personal_presets_(lazer)`, so it doesn't need to be in a `Game_modifier_(lazer)` as a "lazer indicator" either. mod presets are a lazer-only concept, so there's no confusion

the indicator will be flipped around at some point and the "stable" articles will then get deprecated/stowed away somewhere eventually (and we can think how exactly at that point, but it will be all-at-once migrations anyhow)

does that make sense? i just see future headaches with this approach the more i think about it

AndiTafel:
> > But what if someone adds an article about e.g. `Personal_presets`? Would that also be inside `Game_modifier` rather than `Game_modifier_(lazer)` (even though it has nothing to do with stable)?
> 
> the ` (lazer)` thing is disambiguation of the article title, not a category. it's not like it needs to be titled `Personal_presets_(lazer)`, so it doesn't need to be in a `Game_modifier_(lazer)` as a "lazer indicator" either. mod presets are a lazer-only concept, so there's no confusion

My point was rather that someone searching for `Personal presets` after it's been created (without `(lazer)` suffix) would open `wiki/Gameplay`, see `Game modifier` and `Game modifier (lazer)` and then probably assume the article is inside `Game modifier (lazer)`, only to not find it there.

> the indicator will be flipped around at some point and the "stable" articles will then get deprecated/stowed away somewhere eventually (and we can think how exactly at that point, but it will be all-at-once migrations anyhow)
> 
> does that make sense? i just see future headaches with this approach the more i think about it

But yeah, makes sense.

> you either get a lot of work reorganising the directory tree all the time including a bunch of redirects (btw you're missing redirects in this pr - those are required for article moves), or you're forced to do more work at once instead of migrating articles in any order

Especially this, so let's leave it as it is. Thanks for taking the time here.

ApigeonAPE:
I have an opinion that maybe we can merge `mod(lazer)` into `mod` and add something like "availability"  at the beginning of the article.
For example:

---
# Hidden

| Client | Availability |
| :-- | :-- | 
| osu!lazer | Yes |
| osu!stable| Yes |

## osu!lazer
...

## osu!stable
...

---

I'm not sure if it kind of fixed the issue that "too  many articles in the same folder" or "cannot search lazer-only mod under `Game modifier` (cuz it's in `Game modifier (lazer)`)".

AndiTafel:
I think this would make everything even more convoluted, as for some mods there is a lot of information specific to one of the clients (like score multipliers).

> the indicator will be flipped around at some point and the "stable" articles will then get deprecated/stowed away somewhere eventually (and we can think how exactly at that point, but it will be all-at-once migrations anyhow)

Also, it would make this more complicated as soon as that time comes.

Walavouchey:
> I have an opinion that maybe we can merge `mod(lazer)` into `mod` and add something like "availability" at the beginning of the article.

i think that does make sense as a direction if it's just an index page, but maybe less so if the article(s) gets fleshed out with screenshots and other info

PR #15072, opened 2026-08-02
Update `Contest points`

pishifat:
from the last mbc


AndiTafel:
Can you maybe update German and Spanish too, so they won't be marked as outdated?

pishifat:
done!

TicClick (approved):

PR #15086, opened 2026-08-04
Update osu! spread rules with new thresholds

NoffyNoffeh:
Following the discussion and proposal from https://osu.ppy.sh/community/forums/topics/2213552?n=1 .

This will apply the proposed changes. This also moves spread to its own section as it is quite lengthy on its own, and needs to be very easily found.

- Additionally updates the simplified RC page to match changes made
- Splits "Spread" into its own section on each game mode page to match the changes made for osu!. On taiko additionally seperated "SV" into a category for consistent logic. No wording or rules outside of osu! spread have changed in wording or application.


wiki/Ranking_criteria/osu!/en.md
```diff
@@ -34,6 +29,19 @@ Overall rules and guidelines apply to every kind of osu! difficulty. Rhythm-rela
 - **Avoid using sound samples for sliderslide, sliderwhistle, and spinnerspin which do not naturally loop.** These hitsounds are continuous, meaning that their files play from start to end and loop as one continuous sound for the length of the object, so using sound files with a clear impact for them might lead to unwanted side effects.
 - **Avoid unjustified difficulty spikes.** Difficulty should be representative of the song's intensity.
 
+### Spread
```
Darius-Wattimena, line 32: Any reason to make this its own section now? Would it make more sense for me to keep it under the general section? Otherwise, it would be nice to have the other gamemodes also adjusted to be consistent?

Darius-Wattimena (changes requested):
We should not forget to update the simplified RC page. They currently also have a [spread section](https://osu.ppy.sh/wiki/en/Ranking_criteria/Simplified_ranking_criteria#spread) which still states the old values.

wiki/Ranking_criteria/osu!/en.md
Okorin: I think it being its separate heading is better because then they get sorted under that heading instead of being somewhere under general

NoffyNoffeh: I think it is beneficial as it is a specific enough topic that applies to all mapsets, while still belonging to the overall category. And as Oko said it makes for clearer sorting. I wouldn't want to update the other modes pages unless this is something they also need.

NoffyNoffeh:
> We should not forget to update the simplified RC page. They currently also have a [spread section](https://osu.ppy.sh/wiki/en/Ranking_criteria/Simplified_ranking_criteria#spread) which still states the old values.

Fixed the page in latest commit, combined the tables as osu! shares the same drain values as other modes now, besides osu!mania.

wiki/Ranking_criteria/osu!/en.md
Walavouchey: > I wouldn't want to update the other modes pages unless this is something they also need.

please do, there's no reason to leave them inconsistent is there <.<

Hiviexd: I think those should be updated too, can briefly loop in the other mode NATs for confirmation

Hiviexd (commented):
as commented, otherwise approved

wiki/Ranking_criteria/osu!/en.md
NoffyNoffeh: Updated with this section being added to the other pages. No other wording/rules have been changed on the other mode pages.

Hiviexd (approved):

peppy (approved):
seems fair enough 👍

Walavouchey (approved):

wiki/Ranking_criteria/osu!taiko/en.md
```diff
@@ -42,41 +42,56 @@ This set of **osu!taiko ranking criteria** lays out [rules and guidelines](/wiki
 
 General rules and guidelines apply to every kind of osu!taiko difficulty. Rhythm-related rules and guidelines apply to approximately 180 BPM beatmaps with 4/4 time signatures. If your song is drastically faster or slower, some variables might be different, as detailed in [Scaling BPM on the Ranking Criteria](/wiki/Ranking_criteria/Scaling_BPM).
 
-### Rules
+### Overall
```
ApigeonAPE, line 45: Noticed that we have taiko `Overall` part under `General` section but other mode `General` under `Overall`.
I'm wondering that if we can make these expressions consistent.

PR #15109, opened 2026-08-08
Add osu!dev development build information

gmangaming029:
I added osu!dev because why not

## Self-check


TicClick:
haven't seen this one, where does it come from? do you have to build the client locally?
also your change is ridden with typos, please fix them

AndiTafel:
You can get it like this: https://github.com/ppy/osu#building

But imo it's not really a *release stream*, since it isn't integrated in either of the clients.

Also, I've never seen the name `osu!dev` anywhere.

Hiviexd:
> But imo it's not really a release stream, since it isn't integrated in either of the clients.

> Also, I've never seen the name osu!dev anywhere.

I'm personally against this addition precisely for those reasons.

Walavouchey:
yeah it's not a release stream, and i don't think i've seen dev builds called "osu!dev" either

gmangaming029:
honestly this was never going to work

PR #15125, opened 2026-08-11
Update RC to allow repeats going through an uninherited timing point

osuWhite:
Following discussion from this thread: https://osu.ppy.sh/community/forums/topics/2203226?n=1

Simplified RC probably doesn't need an update.

### DIFF
--- a/wiki/Ranking_criteria/en.md
+++ b/wiki/Ranking_criteria/en.md
@@ -126,7 +126,7 @@ Each [osu!mania](/wiki/Game_mode/osu!mania) [key mode](osu!mania#common-terms) a
 - **Hit objects must be snapped within less than 2 ms of any timeline tick.** [AiMod](/wiki/Client/Beatmap_editor/AiMod) will report these issues, as well as rare false positives. False positives mainly occur on slider ends and reverses, and should be verified manually or with other tools. Objects in a musical section requiring unsupported beat snap divisors (e.g. 1/11) can either:
   - Remain unsnapped, as long as they align with the intended beat snap divisor.
   - Be snapped through a temporary change in [BPM](/wiki/Music_theory/Tempo).
-- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must have its tail snapped within the new timing section.**  This rule does not apply to slider repeats/reverses, as they are passive objects.
+- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must be snapped within the new timing section.** For osu! beatmaps, this does not apply to slider repeats/reverses, as they are passive objects. However, reverses must still be snapped for osu!catch beatmaps as they can affect droplet/drop generation.
   - To fix unsnapped spinners and osu!mania long notes, this can be achieved through dragging the end of the object in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the `.osu` file. 
 
 ## Audio
--- a/wiki/Ranking_criteria/en.md
+++ b/wiki/Ranking_criteria/en.md
@@ -126,7 +126,8 @@ Each [osu!mania](/wiki/Game_mode/osu!mania) [key mode](osu!mania#common-terms) a
 - **Hit objects must be snapped within less than 2 ms of any timeline tick.** [AiMod](/wiki/Client/Beatmap_editor/AiMod) will report these issues, as well as rare false positives. False positives mainly occur on slider ends and reverses, and should be verified manually or with other tools. Objects in a musical section requiring unsupported beat snap divisors (e.g. 1/11) can either:
   - Remain unsnapped, as long as they align with the intended beat snap divisor.
   - Be snapped through a temporary change in [BPM](/wiki/Music_theory/Tempo).
-- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must be snapped within the new timing section.** For osu! beatmaps, this does not apply to slider repeats/reverses, as they are passive objects. However, reverses must still be snapped for osu!catch beatmaps as they can affect droplet/drop generation.
+- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must be snapped within the new timing section.**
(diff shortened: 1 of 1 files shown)


wiki/Ranking_criteria/en.md
```diff
@@ -126,7 +126,8 @@ Each [osu!mania](/wiki/Game_mode/osu!mania) [key mode](osu!mania#common-terms) a
...
   - Remain unsnapped, as long as they align with the intended beat snap divisor.
   - Be snapped through a temporary change in [BPM](/wiki/Music_theory/Tempo).
-- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must have its tail snapped within the new timing section.** For spinners and osu!mania long notes, this can be achieved through dragging the end of the object in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the `.osu` file.
+- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must have its tail snapped within the new timing section.**  This rule does not apply to slider repeats/reverses, as they are passive objects.
```
Walavouchey:
```suggestion
- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must have its tail snapped within the new timing section.** This rule does not apply to slider repeats/reverses, as they are passive objects.
```

Walavouchey (commented):
:ok:

NoffyNoffeh:
Copying over a few notes from internal discussion for visibility and a pause on the PR:

All forum thread and nat discussion has been standard focused, but updating it in this way to general RC would also affect osu!catch. 

This will have to either be approved by catch NAT, or clarified as a standard specific exception.

osu!catch NAT are being looped in to see / figure this out

osuWhite:
I adjusted the wording to not allow the exemption to apply to catch per the recommendations of the catch NAT

wiki/Ranking_criteria/en.md
```diff
@@ -126,7 +126,8 @@ Each [osu!mania](/wiki/Game_mode/osu!mania) [key mode](osu!mania#common-terms) a
...
   - Remain unsnapped, as long as they align with the intended beat snap divisor.
   - Be snapped through a temporary change in [BPM](/wiki/Music_theory/Tempo).
-- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must have its tail snapped within the new timing section.** For spinners and osu!mania long notes, this can be achieved through dragging the end of the object in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the `.osu` file.
+- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must be snapped within the new timing section.** For osu! beatmaps, this does not apply to slider repeats/reverses, as they are passive objects. However, reverses must still be snapped for osu!catch beatmaps as they can affect droplet/drop generation.
```
NoffyNoffeh:
```suggestion
- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must be snapped within the new timing section.**
 - For osu! beatmaps, this does not apply to slider repeats/reverses, as they are passive objects. However, reverses must still be snapped for osu!catch beatmaps as they can affect droplet/drop generation.
```

NoffyNoffeh:
If mania's thing is a sub-bullet, this can be one as well for consistent layout. Otherwise confirmed with catch NAT now and content wise is good to go.

wiki/Ranking_criteria/en.md
```diff
@@ -126,7 +126,9 @@ Each [osu!mania](/wiki/Game_mode/osu!mania) [key mode](osu!mania#common-terms) a
...
   - Remain unsnapped, as long as they align with the intended beat snap divisor.
   - Be snapped through a temporary change in [BPM](/wiki/Music_theory/Tempo).
-- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must have its tail snapped within the new timing section.** For spinners and osu!mania long notes, this can be achieved through dragging the end of the object in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the `.osu` file.
+- **An object which is wrongly snapped due to passing through or ending slightly before a new [uninherited timing point](/wiki/Client/Beatmap_editor/Timing#uninherited-timing-point) must be snapped within the new timing section.**
+ - For osu! beatmaps, this does not apply to slider repeats/reverses, as they are passive objects. However, reverses must still be snapped for osu!catch beatmaps as they can affect droplet/drop generation.
```
Okorin: Maybe:
```md
- For osu!, this does not apply to slider repeats and reverses.
- For osu!catch, slider repeats and reverses have to be snapped as they affect droplet and drop generation.
```
I think making catch a separate point like mania is is better here because otherwise it's easy to miss that catch maps have another clarification. I left out the justification for osu! in the suggestion because i thought it doesn't need one, but that's optional. I think having repeats/reverses or droplets/drops is a bit vague because it can be and / or and i feel like just spelling it out without the `/` would be better?

Darius-Wattimena: Only drop matters here, so it can just be:
```
- For osu!catch, slider repeats and reverses have to be snapped as they affect drop generation.
```

wiki/Ranking_criteria/en.md
```diff
@@ -126,7 +126,9 @@ Each [osu!mania](/wiki/Game_mode/osu!mania) [key mode](osu!mania#common-terms) a
...
+  - To fix unsnapped spinners and osu!mania long notes, this can be achieved through dragging the end of the object in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the `.osu` file.
```
Okorin:
```suggestion
  - For osu!, this does not apply to slider repeats and reverses, as they are passive objects.
  - For osu!catch, slider repeats and reverses have to be snapped as they affect drop generation.
  - To fix unsnapped spinners and osu!mania long notes, this can be achieved through dragging the end of the object in the timeline. For sliders, this can be achieved through slider velocity manipulation or editing of the `.osu` file. 
```

Okorin (approved):

Okorin:
@peppy could you give this rule change a look over?

peppy (approved):
probably ok

PR #15147, opened 2026-08-17
[PL] Update `Mascots`

saltylucario7:
Basically translated the whole article from scratch, as the polish page was very heavily outdated, but I retained what I could from the old version if it didn't contain errors.

## Self-check


TicClick (approved):

PR #14210, opened 2026-01-16
Add Program_files_(lazer) page

Kraggle09:
This PR adds a lazer version of the Program files page. It contains all information in Client/Release_stream/Lazer/File_storage, and additional information about files stored in the osu folder. Closes #8649.

A couple of questions:

- Should I delete the en version of the old File_storage?
  - This would probably need a redirect to the new page if the user switches from the non-en version to the en version.
- Is there any information on what the client.realm.management folder does?

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #15196, opened 2026-08-22
Rewrite `Client/Release Stream/Lazer/File storage`

prplnorangesoda:
Goals for this were to be a bit more in-depth on what each file is used for. Inspired by all the people in #help that keep deleting their client.realm files. I think explicitly listing dangerous files/folders to delete is beneficial, so I included it.

Am a bit conflicted about the phrasing for the dangerous files, maybe alert-danger would be better? I wanted to avoid using `you` but it makes it feel a little odd. please lmk thoughts

## Self-check

### DIFF
--- a/wiki/Client/Program_files/de.md
+++ b/wiki/Client/Program_files/de.md
@@ -1,3 +1,8 @@
+---
+outdated_since: 3f39529f59e62379f36b04560cbe6eab1276e1db
+outdated_translation: true
+---
+
 # osu!-Programmdateien
 
 ::: alert-note
--- a/wiki/Client/Program_files/en.md
+++ b/wiki/Client/Program_files/en.md
@@ -1,6 +1,8 @@
 # osu! program files
 
 ::: alert-note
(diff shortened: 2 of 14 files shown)


Walavouchey:
you're not supposed to outdate because it's done automatically these days. the hash will be incorrect because of the squash merge

Walavouchey:
also duplicate of https://github.com/ppy/osu-wiki/pull/14210 which has a more correct direction

prplnorangesoda:
> you're not supposed to outdate because it's done automatically these days. the hash will be incorrect because of the squash merge

sorry, i did it as the check script told me to, my bad. will fix the commit tree if needed

> also duplicate of https://github.com/ppy/osu-wiki/pull/14210 which has a more correct direction

i saw this but it's been inactive for a few months. do you want me to adjust this PR or just close it outright?

i wrote this page this way because this is something i'd like to see on the osu! wiki, as someone who's had issues with storage recently.

PR #15200, opened 2026-08-23
Update information in `Help centre/Upgrading to lazer`

AndiTafel:
- Storyboard triggers were implemented in https://github.com/ppy/osu/pull/38387
- Flashlight in osu!taiko was adjusted in https://github.com/ppy/osu/pull/38374
- Lazer scores are displayed under `Best Performance` since https://github.com/ppy/osu-web/pull/11847

## Self-check

### DIFF
--- a/wiki/Help_centre/Upgrading_to_lazer/en.md
+++ b/wiki/Help_centre/Upgrading_to_lazer/en.md
@@ -59,7 +59,7 @@ The following is a comprehensive list of the **current state** of lazer in compa
 | Per-beatmap offset calibration | ![Partial][partial][^offset-calibration-stable] | ![Yes][true][^offset-calibration-lazer] |
 | osu! sliders "snake" while dragging | ![No][false] | ![Yes][true][^can-disable] |
 | osu! player-friendly "note lock" | ![No][false] | ![Yes][true][^note-lock] |
-| osu!mania and osu! timing-based note colouring | ![No][false] | ![Yes][true] |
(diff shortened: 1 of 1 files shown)


wiki/Help_centre/Upgrading_to_lazer/en.md
```diff
@@ -59,7 +59,7 @@ The following is a comprehensive list of the **current state** of lazer in compa
 | Per-beatmap offset calibration | ![Partial][partial][^offset-calibration-stable] | ![Yes][true][^offset-calibration-lazer] |
 | osu! sliders "snake" while dragging | ![No][false] | ![Yes][true][^can-disable] |
 | osu! player-friendly "note lock" | ![No][false] | ![Yes][true][^note-lock] |
-| osu!mania and osu! timing-based note colouring | ![No][false] | ![Yes][true] |
+| osu!, osu!catch and osu!mania timing-based note colouring | ![No][false] | ![Yes][true] |
```
AndiTafel, line 62: Not entirely sure about this. The setting `Timing-based note colouring` only exists in osu!mania, but osu! is listed here too, so I assume it refers to the Synesthesia mod?

I added osu!catch, because Synesthesia is available there too, but maybe it should only be osu!mania?

peppy (approved):

wiki/Help_centre/Upgrading_to_lazer/en.md
peppy: Probably fine to say this. We're still not sure if we'll combine the two in the future or not.

TicClick (approved):

PR #15199, opened 2026-08-23
Add jorj1357 and keyesgen to `Featured Artists with osu! accounts`

AndiTafel:
[yesandtheyes127](https://osu.ppy.sh/users/15943191) is already in the user group, [keyesgen](https://osu.ppy.sh/users/13278140) isn't.

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #14894, opened 2026-06-28
Update `People/Featured Artists`

AndiTafel:
Added missing FAs and updated some usernames. Also updated all non-outdated translations.

Edit: Maybe Adust Rain should be added? See [discord](https://discord.com/channels/188630481301012481/1094639474941636819/1520567606254371068).

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #15204, opened 2026-08-24
Update usernames in `People/Featured Artists`

AndiTafel:
Not super important, but it bugs me https://github.com/ppy/osu-wiki/pull/15199#issuecomment-5395967969

## Self-check


TicClick:
not sure this is the right direction, unless the artist(s) want it themselves or this is common practice — that I do not know

AndiTafel:
Idk, I also updated old usernames in https://github.com/ppy/osu-wiki/pull/14894 and it seemed to be okay.

(This only affects the left column (profile name), not the name on the FA listing.)

TicClick (approved):
yeah, I was looking for a signal like that, thank you

PR #15233, opened 2026-08-27
Mention recently added features for the wiki

The-Last-Cookie:
- styled alerts
- on-hover user card display

## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #13392, opened 2025-06-18
[ES] Update `History of osu!/osu! wiki`

UnknownOsu:
## Self-check

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)
- [x] *(translations only)* The changes are reviewed on GitHub [by a fluent speaker](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#review)

PR #13157, opened 2025-04-18
Clean up `History of osu! wiki`

Nivalyx:
Thought I'll tidy up the entire article as well after doing [#13150](http://github.com/ppy/osu-wiki/pull/13150) yesterday.

Please note that for the most part this is a cleanup/rewrite (not an update) and I didn't really add anything new to it. The formatting is based on the "[Mapping and modding timeline](http://osu.ppy.sh/wiki/en/History_of_osu!/Mapping_and_modding_timeline)" page.

- [x] The changes are tested against the [contribution checklist](https://osu.ppy.sh/wiki/osu!_wiki/Contribution_guide#self-check)

PR #15250, opened 2026-09-02
Correct word in `History of osu!/osu! wiki`

AndiTafel:
Didn't notice this in https://github.com/ppy/osu-wiki/pull/15233.

~~This PR also updates the `outdated_since` hash in `es.md`, because it seems to me like the translation was never updated to include the changes from https://github.com/ppy/osu-wiki/pull/13157 (it still uses tables). (I guess this is because https://github.com/ppy/osu-wiki/pull/13392 happened before https://github.com/ppy/osu-wiki/pull/13459.)~~

## Self-check


UnknownOsu:
Regarding what you mentioned about the use of tables in the `ES` article, the information in the article was updated, but the tables were kept.

It was mostly a stylistic decision, since there really wasn’t any issue with whether or not to use tables, as mentioned in [#13157](https://github.com/ppy/osu-wiki/pull/13157#issuecomment-2815762175), and the most important part of that pull request was updating the information; I decided to keep the tables but update the information

TicClick (approved):

AndiTafel:
> Regarding what you mentioned about the use of tables in the `ES` article, the information in the article was updated, but the tables were kept.

Oh okay, sorry, didn't know that was intentional.

PR #15270, opened 2026-09-06
Make article titles in `History of osu!/...` consistent

AndiTafel:
The article title is usually `History of osu! <year>` but for some reason in some years it's just `<year>`. Occasionally, there are also inconsistencies between English and translations.

image

vs.

image

The first one looks much better to me (and is used is the majority of these articles).

DO_NOT_OUTDATE: *

## Self-check


Joehuu (approved):
These titles are used for wiki search too, so makes sense. LGTM.

Screenshot 2026-09-05 at 6 53 14 PM

PR #15281, opened 2026-09-08
[EN] Add reading skill and rework tldr to Performance Points

stysan:
## Self-check

### DIFF
--- a/wiki/Performance_points/en.md
+++ b/wiki/Performance_points/en.md
@@ -10,7 +10,7 @@ It aims to shift the focus of skill progression from the amount of time played t
 
 Performance points are heavily based on calculated beatmap difficulty, which is determined by a unique algorithm constructed for each individual [game mode](/wiki/Game_mode).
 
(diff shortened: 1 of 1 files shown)


wiki/Performance_points/en.md
```diff
@@ -59,6 +59,12 @@ Scores with high accuracy values are considered by the algorithm to be highly im
 
 Sections or  of extremely high speed or difficulty [patterning](/wiki/Beatmap/Pattern) in a beatmap will significantly increase its considered strain values. E.g., beatmaps with more [streams](/wiki/Beatmap/Pattern/osu!/Stream) or fast jump waves will have high strain values, and thus increase the pp gain for that beatmap.
 
+### Reading
+
+*Reading* is a core value that considers how difficult it is to read patterns in a beatmap.
```
MChecaH, line 64: When writing a definition, it is often good practice to avoid using the same word that is being described. I'd avoid using *"read"* or any of its conjugations to define *"reading"*.

As such, I'd use other more basic synonyms so the term can be properly understood. Analyze, understand, react, or a combination of those (or any other).

As an example:
```suggestion
*Reading* is a core value that considers how difficult it is to understand and react to patterns in a beatmap.
```

PR #15334, opened 2026-09-17
Update `Client/Options`

saltylucario7:
When doing translation work, I noticed that some default values listed in the article are different than the defaults in the client, so changed them to be consistent with the client. Order of two renderer options was also different in the article so swapped them as well

## Self-check


wiki/Client/Options/en.md
```diff
@@ -285,7 +285,7 @@ This section is about input peripherals.
 | :-- | :-- | :-- | :-- |
 | `Sensitivity` | Adjust the sensitivity of the mouse cursor. If changed, it may automatically enable `Raw Input`. | Slider | `1x` |
 | `Raw input` | Read mouse/tablet positional values directly from the hardware, without any post-processing (this is to ignore mouse acceleration). | Checkbox | `Disabled` |
-| `Map absolute raw input to the osu! window` | Confine input devices with absolute positioning (e.g. pen tablets) to the game client window only. | Checkbox | `Disabled` |
+| `Map absolute raw input to the osu! window` | Confine input devices with absolute positioning (e.g. pen tablets) to the game client window only. | Checkbox | `Enabled` |
```
AndiTafel:
```suggestion
| `Map absolute raw input to the osu! window` | Confine input devices with absolute positioning (e.g. pen tablets) to the game client window only. | Checkbox | `Disabled` |
```
Image

For me `Disabled` is the default?

saltylucario7: ....huh?
image

saltylucario7: I'm gonna be honest, I don't understand how can we have different defaults, nor what to do in this situation. I double checked and my client is up-to date and i'm using stable release and all

AndiTafel: Yeah it's weird, I don't understand it either. To make sure this doesn't depend on other settings or anything like that, I just reinstalled stable and well.. for me it's still `Disabled` by default.

image

saltylucario7: I see, I'll try reinstalling tomorrow as well to see if it changes for me

saltylucario7: - after a clean reinstall this option was off by default for me as well. however, after restoring all of my settings i noticed it's default changed back to enabled. by testing every single setting I found out that the game changes its default upon enabling `Raw Input` and restarting the game
- also found out that `Compatibility mode` is enabled by default for me. says it forced it on for compatibility with NVIDIA Optimus (can't make sc in fullscreen mode for some reason to show it and once I switch to borderless it changes back to disabled by default)

anyhow, I believe these quirks still should be clarified in the article? can't say I feel qualified for that kind of job though, I did this only because it was a simple value change

Walavouchey: according to the source code, the default value is... whatever `Raw input` is currently set to

Walavouchey:
```suggestion
| `Map absolute raw input to the osu! window` | Confine input devices with absolute positioning (e.g. pen tablets) to the game client window only. | Checkbox | The same value as `Raw input` |
```

saltylucario7: applied, what about the compatibility mode? should that wait for another edit or is that not as important
edit: i guess that should be seperate update either way, i just found more stuff (missing `Generate conversion mappings` option that's present on cutting edge)

